PR #36, opened 2025-10-13
Fix debug_rollout_only incompatible with colocate in gpu allocation

fzyzcjy:


andyl98:
Round-3 batch C landed in f09225823 — all 11 adopted findings are now on the branch:

- **Active-generation drain**: `RolloutManager.dispose` cancels and drains in-flight `generate()` tasks inside the `try`, so a cancelled dispose still reaps the service owner in `finally` (new ordering test: generation-cancelled → close).
- **Prewarm refs**: every executor failure arm ray-cancels the prewarm calls before disposing provisional workers (a timed-out prewarm can no longer keep opening provider sandboxes); the manager constructor's failure path kills primer actors.
- **Driver drain bound**: 120s → 600s.
- **Submit bootstrap window** (the one previously-open item): `submit` writes an in-flight marker before the nohup; `sync-code` refuses while a marker is unresolved. The marker clears once the run's `ray_submission.json` lands (the job gate takes over) or after 15 minutes (launcher died pre-submit, reported as stale).
- Fail-closed probes (`miles_run` family scan, `stage-harbor --force` live-gate + move-aside instead of rm), shared lease policy-infra exception set, atomic preparer publication, loopback-only viewer with identity-checking `/healthz` (`view` never kills a mismatched server), honest docs for the retired lifecycle smoke.

Suites: 55 tooling/lifecycle tests + 183 harbor adapter tests green after the batch.

Issue #226, opened 2025-11-22
Contributing to miles


Ratish1:
Hello, I'm interested in miles and I want to work on it, is there anything in particular I can contribute to, I will give it my best. Thanks.

fzyzcjy:
cc @zhaochenyang20 

my personal thought is feel free to try whatever you like (to avoid conflicts w/ other people, maybe firstly briefly say what you want to do before getting started) :)

Ratish1:
Ok got it @fzyzcjy , I will mention what I want to do if I end up doing something, but I was mainly looking to see if there is anything in particular that the miles team would need a contribution for rn. Thanks.

fzyzcjy:
You are welcome and looking forward to your contributions :) We may provide a roadmap later as well

Ratish1:
@fzyzcjy , I see [here](https://github.com/radixark/miles/blob/5a47e45bd2f525db9ec4df6dc9f067ed570200ee/miles/utils/data.py#L20) that it says to #TODO so am I allowed to work on this, or is someone already working on it. Thanks.

fzyzcjy:
Yes I think that can be done (though not of high priority but is good to have). Btw, I personally prefer `polars` over `pandas`, and w/ polars ~~I guess this can be solved like `pl.scan_parquet` ("scan" is lazy while "read" is non-lazy)~~ hmm looks like that does not have iter_rows

EDIT: hmm I guess this may not be super useful, since the cpu/memory bottleneck is not at here - it seems Dataset.samples will occupy more memory compared w/ this. I will have a double check (discussion) tomorrow.

Ratish1:
Ok I will wait for your approval till then @fzyzcjy . Thanks for looking into it.

fzyzcjy:
After an internal discussion, the case is that, in SFT it can be e.g. 100B token and thus takes a lot of memory. Therefore, not only should we lazily read the file, but other memory-consuming things, such as Dataset.samples, also needs to be lazy to save memory.

Ratish1:
Thanks @fzyzcjy , I will start working on it then.

fzyzcjy:
there will be an example today.

Ratish1:
Ok got it I will be waiting for the example until then.

fzyzcjy:
Oops sorry, I mean there will be an example in https://github.com/radixark/miles/issues/238, not in this issue... please ignore my words above and feel free to start working on it!

Ratish1:
[...]

PR #246, opened 2025-11-24
feat: Implement lazy data loading for Dataset

Ratish1:
this pr improves the `Dataset` class in `miles/utils/data.py` to support lazy data loading, filtering, and indexed access for large datasets. This addresses memory consumption issues, in SFT workloads, by avoiding the in-memory materialization of entire datasets. @fzyzcjy #226 

- Modified `read_file` to use row-by-row iteration for JSONL and batched reading for Parquet.
- Refactored `Dataset` to build a lightweight index of valid sample locations on initialization, applying `max_length` filtering during this phase.
- Implemented `__getitem__` to read and process only the requested sample on demand.


fzyzcjy:
Too busy now, I will try to squeeze some time and have a review later :)

Ratish1:
No problem

Ratish1:
Hello @fzyzcjy, I wanted to follow up on this PR , but if you are busy its fine you can look at it later. Thanks.

fzyzcjy:
After an internal discussion, we think the most ideal code may be somehow different. Therefore, may be need to wait a little bit.

Ratish1:
Hello @fzyzcjy , I wanted to follow up on this. Thanks.

lhoestq:
Hi ! Quentin from HF Datasets here :)

I was checking if I could help with anything, in particular if you are interested in loading larger than RAM datasets. E.g. the `datasets` library loads many formats like Parquet and JSON, makes tokenization easy, and works with larger than RAM datasets since the data is loaded with memory mapped Arrow files on disk. It also has streaming features to load datasets bigger than disk

fzyzcjy:
@Ratish1 I am really too busy recently to check more details about this :/ @zhaochenyang20 do you want to find someone to review these?

@lhoestq looks great! Yes it would be great to handle huge datasets here, and at the same time keep code clean, b/c the framework supports SFT and can have big datasets.

Ratish1:
Np @fzyzcjy , maybe I can work with @lhoestq and use the HF `datasets` library here

zhaochenyang20:
@Ratish1 Have conflicts right now.

miles/utils/data.py
```diff
@@ -125,76 +75,283 @@ def __init__(
...
+
+        prompt = self._get_prompt_from_data(data)
+        sample = Sample(
+            prompt=prompt,
+            label=data.get(self.label_key) if self.label_key is not None else None,
+            metadata=data.get(self.metadata_key) or {},
+        )
+
+        # Basic cache management (simple LRU by always adding to end, but limited size)
+        if len(self._cache) >= 10000:  # Limit cache size
```
PopSoda2002: can you give this magic number a name?

PopSoda2002:
Hi @Ratish1 , are you working with `datasets` from HF? It that, that will be nice

Ratish1:
> Hi @Ratish1 , are you working with `datasets` from HF? It that, that will be nice

No I'm not. I will refactor this code with the `datasets` library instead now. Thanks.

miles/utils/data.py
Ratish1: I used HF `datasets` library now so this entire logic is not needed now.

requirements.txt
```diff
@@ -6,6 +6,8 @@ mcp[cli]
 memray  # needed for debugging (but is lightweight), we can put it to dev mode when using pyproject.toml
 omegaconf
 pillow
+polars
+pyarrow
```
PopSoda2002: I think these can be deleted?

miles/utils/data.py
```diff
@@ -125,52 +125,83 @@ def __init__(
...
-                    label=data[label_key] if label_key is not None else None,
-                    metadata=metadata,
-                )
+        if not os.path.exists(self.raw_file_path):
+            raise FileNotFoundError(f"Prompt dataset path '{self.raw_file_path}' does not exist.")
+
+        logger.info(f"Loading dataset from {self.raw_file_path} using Hugging Face datasets.")
+
+        # Determine file type and load using datasets library for memory-mapped access
+        file_type = "json" if self.raw_file_path.endswith(".jsonl") else "parquet"
```
PopSoda2002: should add an assert here I think

PopSoda2002 (commented):
Hi, thanks for the effort, can you please show some results or comparison?

miles/utils/data.py
Ratish1: Done

requirements.txt
Ratish1: Done

Ratish1:
> Hi, thanks for the effort, can you please show some results or comparison?

Hi, I haven't compared since I think `datasets` would be much faster for large datasets, but I ran a memory benchmark using a 500 MB dummy JSONL file (5 million rows) to verify the lazy loading of the current implementation. This is what I got 
```text
BENCHMARK RESULTS (Lazy Loading)
Initial Memory:   371.73 MiB
Peak Memory:      469.66 MiB
Memory Increment: 97.93 MiB

**Review by PopSoda2002 — APPROVED** [review]
It looks good for me but I think it needs a more careful look!

File: miles/utils/data.py [hunk 4]
```diff
@@ -125,52 +125,91 @@ def __init__(
...
     def shuffle(self, new_epoch_id):
         if self.epoch_id == new_epoch_id:
             return
 
-        random.seed(self.seed + new_epoch_id)
-        permutation = list(range(len(self.samples)))
-        random.shuffle(permutation)
-        self.samples = [self.origin_samples[i] for i in permutation]
+        logger.info(f"Shuffling dataset for epoch {new_epoch_id} with seed {self.seed + new_epoch_id}")
+        self.hf_dataset = self.hf_dataset.shuffle(seed=self.seed + new_epoch_id)
```
lhoestq: Here you may shuffle a dataset that has already been shuffled for the previous epoch no ?
You could have origin_hf_dataset and hf_dataset = origin_hf_dataset.shuffle(), as before

Ratish1: Yes thats correct. I changed it now, thank you for the review and spotting that out.

miles/utils/data.py
```diff
@@ -125,52 +125,93 @@ def __init__(
...
+            return not _should_skip_prompt(
+                prompt, self.tokenizer, self.processor, self.max_length, self.apply_chat_template_kwargs
+            )
+
+        original_size = len(self.hf_dataset)
+        self.hf_dataset = self.hf_dataset.filter(filter_func, num_proc=os.cpu_count())
+        new_size = len(self.hf_dataset)
+        logger.info(f"Filtered dataset from {original_size} to {new_size} samples.")
+
+        self.origin_hf_dataset = self.hf_dataset
```
zhaochenyang20: The __init__ method has become a dumping ground for logic and attributes. It violates our guideline: "Specific code path with multiple lines should be split into a function".

Please extract the dataset loading and filtering logic into private methods.

like:

```python

class Dataset:
    def __init__(self, ...):
        # 1. Store basic config
        self.tokenizer = tokenizer
        # ... (assign other attributes)

        # 2. Load and process dataset
        self.hf_dataset = self._load_and_filter_dataset(path, seed)

    def _load_and_filter_dataset(self, path, seed):
        # Encapsulate file type check, load_dataset, slice, and filter logic here
        pass
```

miles/utils/data.py
```diff
@@ -125,52 +125,93 @@ def __init__(
...
+
+        # Apply filtering using the existing helper functions
+        def filter_func(example):
+            prompt = _build_messages(example, self.prompt_key, self.multimodal_keys)
+            return not _should_skip_prompt(
+                prompt, self.tokenizer, self.processor, self.max_length, self.apply_chat_template_kwargs
+            )
+
+        original_size = len(self.hf_dataset)
+        self.hf_dataset = self.hf_dataset.filter(filter_func, num_proc=os.cpu_count())
```
zhaochenyang20: Do not use os.cpu_count() directly. In containerized environments (Kubernetes, Docker), this returns the host's total CPUs, not the container's quota. This will spawn too many processes, causing thrashing or OOM kills.

Use a safe default (e.g., 4 or 8) or allow it to be passed as an argument.

miles/utils/data.py
```diff
@@ -125,52 +125,93 @@ def __init__(
...
+        data = self.hf_dataset[idx]
+
+        # Process the data using existing logic
+        prompt = _build_messages(data, self.prompt_key, self.multimodal_keys)
+
+        metadata = data.get(self.metadata_key) or {}
+        if self.tool_key is not None and self.tool_key in data:
+            tools = data[self.tool_key]
+            if isinstance(tools, str):
+                tools = json.loads(tools)
```
zhaochenyang20, line 216: This is a comment. May not need to be done this time:

You are parsing JSON and building messages dynamically in every __getitem__ call. While this saves RAM, it adds significant CPU overhead during the training loop.

If the JSON parsing is heavy, we might need to use hf_dataset.map() during init to pre-process these fields into a more efficient format (Arrow-native), rather than parsing raw strings on the fly.

miles/utils/data.py
```diff
@@ -125,52 +125,93 @@ def __init__(
...
+
+        # Determine file type and load using datasets library for memory-mapped access
+        if self.raw_file_path.endswith(".jsonl"):
+            file_type = "json"
+        elif self.raw_file_path.endswith(".parquet"):
+            file_type = "parquet"
+        else:
+            raise ValueError(
+                f"Unsupported file format: {self.raw_file_path}. Supported formats are .jsonl and .parquet."
             )
```
zhaochenyang20:
```python
_FILE_TYPE_MAP = {
        ".jsonl": "json",
        ".parquet": "parquet",
    }

    def _get_file_type(self, path: str) -> str:
        _, ext = os.path.splitext(path)
        
        if ext not in self._FILE_TYPE_MAP:
            supported = list(self._FILE_TYPE_MAP.keys())
            raise ValueError(f"Unsupported file format: {ext}. Supported: {supported}")
            
        return self._FILE_TYPE_MAP[ext]
```

miles/utils/data.py
```diff
@@ -125,52 +125,93 @@ def __init__(
...
+
+        # Apply row slicing if specified
+        if self.row_slice:
+            num_rows = len(self.hf_dataset)
+            indices = range(num_rows)[self.row_slice]
+            self.hf_dataset = self.hf_dataset.select(indices)
+            logger.info(f"Applied slice {self.row_slice}, dataset size: {len(self.hf_dataset)}")
+
+        # Apply filtering using the existing helper functions
+        def filter_func(example):
```
zhaochenyang20: consider adding a tqdm progress bar or logging? Somehow I remember HuggingFace default has this. If so, no need to worrry.

zhaochenyang20: Adding a description for the filtering:

```python
self.hf_dataset = self.hf_dataset.filter(
    filter_func, 
    num_proc=..., 
    desc="Filtering invalid samples during init"  # Explicitly name the task
)
```

zhaochenyang20 (changes requested):

miles/utils/data.py
Ratish1: Done

miles/utils/data.py
Ratish1: Done

miles/utils/data.py
Ratish1: Done

miles/utils/data.py
Ratish1: I added the desc, and yes hugging face default has the progress bar. THanks.

miles/utils/data.py
Ratish1: Yes this is correct. But I have not done this change yet, since our primary goal for this PR was resolving the RAM spike during initialization, I believe this lazy approach is the safest first step. If we find that JSON parsing becomes a bottleneck for GPU throughput in future benchmarks, we can definitely change it to a .map() based pre-processing step to offload that work to the initialization phase.

docs/en/get_started/quick_start.md
```diff
@@ -580,3 +580,32 @@ miles has been deeply optimized for distributed training of large-scale Mixture
...
+    --num-rollout 1
+```
+
+### What to Observe
+
+1.  **Fast Startup**: Even with a large `.jsonl` file, the training process should start almost immediately. This confirms the data is being memory-mapped (Lazy Loading) instead of read entirely into RAM.
+2.  **Filtering Progress**: You will see a progress bar titled `Filtering invalid samples during init`. This confirms that the filtering logic is running and providing visual feedback.
+3.  **Configurable Parallelism**: By specifying `--num-proc 4`, we override the default value (8). You can verify that the system spawns exactly 4 worker processes for the data preparation phase.
+4.  **RAM Stability**: Monitor your system RAM during startup. It should remain stable because only the dataset indices are stored in memory, not the raw text.
+
```
zhaochenyang20: No need to add these verification commands in the quick start. Just add it to your PR's description, it's for developers only 😂

zhaochenyang20:
Reproduce/Verification commands:

```bash
docker pull radixark/miles:latest

docker run --rm --gpus all --ipc=host --shm-size=16g \
  --ulimit memlock=-1 --ulimit stack=67108864 \
  -it radixark/miles:latest /bin/bash

rm -rf /root/miles

git clone -b data-loading https://github.com/Ratish1/miles.git

cd miles

pip install -e .

hf download zai-org/GLM-Z1-9B-0414 --local-dir /data/ratish/GLM-Z1-9B-0414

hf download --repo-type dataset zhuzilin/dapo-math-17k --local-dir /data/ratish/dapo-math-17k

source scripts/models/glm4-9B.sh

PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py ${MODEL_ARGS[@]} --hf-checkpoint /data/ratish/GLM-Z1-9B-0414 --save /data/ratish/GLM-Z1-9B-0414_torch_dist

export CUDA_DEVICE_MAX_CONNECTIONS=1
export CUDA_VISIBLE_DEVICES=5,7

PYTHONPATH=/root/Megatron-LM python train.py \
    ${MODEL_ARGS[@]} \
    --hf-checkpoint /data/ratish/GLM-Z1-9B-0414 \
    --load /data/ratish/GLM-Z1-9B-0414_torch_dist \
    --prompt-data /data/ratish/dapo-math-17k/dapo-math-17k.jsonl \
    --input-key prompt \
    --label-key label \
    --apply-chat-template \
    --use-miles-router \
    --num-proc 4 \
    --rollout-batch-size 16 \
    --global-batch-size 16 \
    --n-samples-per-prompt 1 \
    --num-rollout 1 \
    --colocate \
    --sglang-mem-fraction-static 0.8
```

nanjiangwill:
@Ratish1 hi, i am wondering if lazy loading is still working during filtering `ds = ds.filter(partial(_filter_func, **filter_kwargs), num_proc=num_proc, desc="Filtering invalid samples")` (i.e. no peak memory consumption

zhaochenyang20:
> @Ratish1 hi, i am wondering if lazy loading is still working during filtering `ds = ds.filter(partial(_filter_func, **filter_kwargs), num_proc=num_proc, desc="Filtering invalid samples")` (i.e. no peak memory consumption

Hi, I've double-checked the underlying mechanism. ds.filter does indeed maintain the lazy loading behavior.

Under the hood, datasets processes the data in batches by reading them from the memory-mapped (mmap) storage into RAM. Once the filter function is executed, the Python objects are immediately released. It does not materialize the entire dataset in memory at once. Therefore, even during the filtering phase, memory consumption stays at a small constant level (determined by the batch_size and num_proc), avoiding any memory peaks that scale with the dataset size. This is exactly why integrating the datasets library is so beneficial for our use case.

zhaochenyang20:
https://github.com/radixark/miles/pull/372#issue-3771933179

Sorry for my bad. This is reverted. We need more unit tests for dataset consistency. 😭

PR #256, opened 2025-11-25
Remove outdated comments about start_rollout_id

Ratish1:
This PR fixes start_rollout_id which was not being correctly initialized when loading from a checkpoint. The fix reads the latest_checkpointed_iteration.txt file during argument validation and sets args.start_rollout_id accordingly, which was #TODO.

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -1314,7 +1314,6 @@ def miles_validate_args(args):
                 "please make sure it is a valid megatron checkpoint directory."
             )
 
-    # TODO: During loading, we need to set the start_rollout_id here.
     if (
         args.load is None
         or not os.path.exists(args.load)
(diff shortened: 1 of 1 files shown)


fzyzcjy:
iirc we do set the value during the next phase, and also I changed it here but was told it was incompatible w/ a fork. let me check internally.

fzyzcjy:
Thank you for the willing to contribute! However, after an internal discussion, I confirmed that this is not suggested to be changed :( (indeed I have changed it before as well but reverted)

However, feel free to make a pr to remove this todo line to more accurately reflect the state of code.

image

Ratish1:
Ok got it thanks for letting me know

Ratish1:
I reverted the changes and remove the todo @fzyzcjy.

Issue #259, opened 2025-11-26
fix: perform health check in background


Ratish1:
Hello, is this a TODO I could possibly work on, if so is there any particular requirements I need to satisy to make sure this is done properly. Thanks. https://github.com/radixark/miles/blob/9a5c10b8b3daf24b220fdb1b567718e3207f8259/miles/router/router.py#L67

fzyzcjy:
I double checked internally, and looks like just implement it naively (normally), no worries

Ratish1:
Ok will start working on it

PR #260, opened 2025-11-26
add background health check to miles native router

Ratish1:
This PR implements a background health check for the router component, resolving the TODO in miles/router/router.py. A background asyncio task is started on startup that periodically (default 30s) pings all registered workers. Workers that return a non-200 status or fail to connect are removed from the active pool. This ensures that traffic is not routed to unhealthy or dead workers.  #259 @fzyzcjy

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+.github/CODEOWNERS @fzyzcjy @Ying1123
+.github/workflows/ @yushengsu-thu
+/miles/ @fzyzcjy @yueming-yuan


Ratish1:
@fzyzcjy Does this look good or I need to change my approach?. Thanks.

Ratish1:
A gentle remainder about this @fzyzcjy if you were busy. Thanks

miles/router/router.py
```diff
@@ -63,9 +65,22 @@ def _setup_routes(self):
...
         self.app.api_route("/{path:path}", methods=["GET", "POST", "PUT", "DELETE"])(self.proxy)
 
-    async def health_check(self, request: Request):
-        # TODO: do health check in background
-        pass
+    async def _start_background_health_check(self):
+        asyncio.create_task(self._health_check_loop())
+
+    async def _health_check_loop(self):
+        interval = getattr(self.args, "rollout_health_check_interval", 30)
```
zhaochenyang20: We do not want to use `getattr` as suggested in https://github.com/sgl-project/sglang/pull/14836

miles/router/router.py
```diff
@@ -63,9 +65,22 @@ def _setup_routes(self):
...
+        while True:
+            await asyncio.sleep(interval)
+
+            for url in list(self.worker_urls.keys()):
+                try:
+                    response = await self.client.get(f"{url}/health", timeout=5.0)
+                    if response.status_code != 200:
+                        self.worker_urls.pop(url, None)
+                except Exception:
+                    self.worker_urls.pop(url, None)
```
zhaochenyang20: 1. Should we do an async query rather than query them one by one?
2. I think it's good to periodically log the health status, even if it's healthy.
3. If a specific worker's health status is wrong, should we raise a warning or consider restarting it?  cc @fzyzcjy

miles/router/router.py
Ratish1: Done

miles/router/router.py
Ratish1: 1. Yes I refactored it now to use async logic
2. I added logger warning and logger info messages.
3. Right now I have added logic for removing unhealthy workers instead of restarting them to avoid conflicts with the systems main process management. Lmk if you need more changes. Thanks

miles/router/router.py
```diff
@@ -63,9 +68,43 @@ def _setup_routes(self):
...
+                    logger.warning(f"[miles-router] Worker {url} health check failed: {e}")
+                return url, False
+
+            # Create a list of tasks for all workers to run in parallel
+            tasks = []
+            for url in urls:
+                tasks.append(check(url))
+
+            # Run all checks concurrently
+            results = await asyncio.gather(*tasks)
```
zhaochenyang20: While using `asyncio.gather` is the correct approach for concurrent health checks, defining the `async def check(url)` function inside the while True loop is an anti-pattern.

zhaochenyang20: This is a mock check:

```python
async def _check_worker_health(self, url):
    """Encapsulated health check logic for better maintainability."""
    try:
        response = await self.client.get(f"{url}/health", timeout=5.0)
        if response.status_code == 200:
            return url, True
        logger.warning(f"[miles-router] Worker {url} is unhealthy (Status: {response.status_code})")
    except Exception as e:
        logger.warning(f"[miles-router] Worker {url} health check failed: {e}")
    return url, False

async def _health_check_loop(self):
    interval = self.args.rollout_health_check_interval

    while True:
        await asyncio.sleep(interval)
        
        urls = list(self.worker_urls.keys())
        if not urls:
            continue

        # Concurrent execution using a generator expression
        results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))

        for url, is_healthy in results:
            if not is_healthy:
                logger.warning(f"[miles-router] Removing unhealthy worker from pool: {url}")
                self.worker_urls.pop(url, None)

        logger.info(f"[miles-router] Health check complete. {len(self.worker_urls)} workers active.")
```

Ratish1: Yes thats correct, I changed it now. thanks

miles/router/router.py
```diff
@@ -63,9 +68,39 @@ def _setup_routes(self):
...
+    async def _check_worker_health(self, url):
+        """Encapsulated health check logic for better maintainability."""
+        try:
+            response = await self.client.get(f"{url}/health", timeout=5.0)
+            if response.status_code == 200:
+                return url, True
+            logger.warning(f"[miles-router] Worker {url} is unhealthy (Status: {response.status_code})")
+        except Exception as e:
+            logger.warning(f"[miles-router] Worker {url} health check failed: {e}")
+        return url, False
```
zhaochenyang20, line 87: Thanks for your PR. I am just checking with Chengxing regarding wether we need a further refactor of miles router.

miles/router/router.py
```diff
@@ -63,9 +68,39 @@ def _setup_routes(self):
...
+            if not urls:
+                continue
+
+            # Concurrent execution using a generator expression
+            results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))
+
+            for url, is_healthy in results:
+                if not is_healthy:
+                    logger.warning(f"[miles-router] Removing unhealthy worker from pool: {url}")
+                    self.worker_urls.pop(url, None)
```
zhaochenyang20: The current logic is a bit too aggressive. If a single health check fails due to a transient network blip, the worker is permanently removed (pop) from the pool.

Unless the workers have an auto-re-registration mechanism, the router will "bleed" workers over time until the pool is empty.

1. Instead of removing the worker, mark it as "unhealthy" (e.g., maintain a separate set or status dict). Stop routing traffic to it, but continue health checks. If it recovers, mark it as healthy again.

2. Alternatively, implement a "strike system" (e.g., remove only after 3 consecutive failures).

miles/router/router.py
zhaochenyang20, line 87: This logs at INFO level every 30 seconds (default interval). In a production environment, this creates unnecessary log spam.

Suggestion: Change this to logger. debug, or only log at INFO level when the state actually changes (e.g., when a worker is removed or added).

miles/router/router.py
```diff
@@ -63,9 +68,39 @@ def _setup_routes(self):
...
+
+        while True:
+            await asyncio.sleep(interval)
+
+            urls = list(self.worker_urls.keys())
+            if not urls:
+                continue
+
+            # Concurrent execution using a generator expression
+            results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))
```
zhaochenyang20: I want to raise you the content of Fail Silently.

`_check_worker_health` catched the failure already, but the error happened in _health_check_loop is not caught.

When we spawn a background task using asyncio.create_task (fire-and-forget), if an unhandled exception occurs inside the task (e.g., an unpacking error, a logic bug, or a regression), the task terminates immediately.

The critical issue is that since the main application does not await this task, the exception is not raised immediately. Instead, it remains suppressed until the task object is garbage collected. This leads to a dangerous "Zombie State": The Router process and APIs remain active, and logs show no errors, but the health check logic has silently died in the background.

To ensure high availability in production, we need a "catch-all" block inside the while True loop to ensure the loop survives unexpected errors.

Try to modify it like:


```python
async def _health_check_loop(self):
        interval = getattr(self.args, "rollout_health_check_interval", 30)

        while True:
            try:
                await asyncio.sleep(interval)
                
                # ... existing logic ...
                # urls = list(self.worker_urls.keys())
                # ...
                
            except asyncio.CancelledError:
                # Allow the task to be cancelled gracefully on shutdown
                raise
            except Exception as e:
                # Catch-all: Log the full traceback and continue the loop
                # This prevents the thread from dying due to a single unexpected error
                logger.error(f"[miles-router] Unexpected error in health check loop: {e}", exc_info=True)
                await asyncio.sleep(5)  # Backoff to avoid busy-looping on persistent errors
```

miles/router/router.py
Ratish1: Done

miles/router/router.py
Ratish1: Thanks, thats a good catch. I changed it now.

miles/router/router.py
Ratish1: I went with the alternative approach you suggested. Instead of permanently popping workers from the pool, I introduced worker_failure_counts to track consecutive failures. A worker is now only marked as unhealthy after 3 consecutive strikes, and I updated the routing logic in _use_url to automatically exclude them.

docs/en/get_started/quick_start.md
```diff
@@ -580,3 +580,30 @@ miles has been deeply optimized for distributed training of large-scale Mixture
...
+    --rm-type deepscaler \
+    --use-miles-router \
+    --rollout-health-check-interval 10 \
+    --rollout-batch-size 16 \
+    --global-batch-size 16 \
+    --num-rollout 10 \
+    --colocate
+```
+
+The router monitors worker health silently. On the 3rd consecutive failure, it logs a `WARNING` and stops routing traffic to that worker. It logs an `INFO` message only when a worker recovers.
```
zhaochenyang20: No neeed to add this in the quick start.

miles/router/router.py
```diff
@@ -63,9 +69,60 @@ def _setup_routes(self):
...
+                    continue
+
+                # Concurrent execution using a generator expression
+                results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))
+
+                for url, is_healthy in results:
+                    prev_failures = self.worker_failure_counts.get(url, 0)
+
+                    if is_healthy:
+                        if prev_failures >= 3:
```
zhaochenyang20: `if xxx and xxx:`

zhaochenyang20: also, make 3 as an argument, default to 3.

miles/router/router.py
```diff
@@ -63,9 +69,60 @@ def _setup_routes(self):
...
+                        if new_failures == 3:
+                            logger.warning(
+                                f"[miles-router] Worker {url} failed 3 consecutive health checks. Stopping traffic."
+                            )
+
+                logger.debug("[miles-router] Health check complete.")
+
+            except asyncio.CancelledError:
+                # Normal shutdown
+                raise
```
zhaochenyang20, line 126: adding a log `warning` here.

miles/router/router.py
```diff
@@ -158,11 +216,16 @@ async def retrieve_from_text(self, request: Request):
         return result
 
     def _use_url(self):
-        """Select a worker URL using round-robin strategy"""
-        assert len(self.worker_urls) > 0, "No workers available"
+        """Select a worker URL using round-robin strategy, excluding unhealthy workers."""
+        # Filter for healthy workers (those with less than 3 consecutive failures)
```
zhaochenyang20: make 3 as a config.

zhaochenyang20 (changes requested):

miles/router/router.py
```diff
@@ -63,9 +69,60 @@ def _setup_routes(self):
...
+        try:
+            response = await self.client.get(f"{url}/health", timeout=5.0)
+            if response.status_code == 200:
+                return url, True
+            logger.debug(f"[miles-router] Worker {url} is unhealthy (Status: {response.status_code})")
+        except Exception as e:
+            logger.debug(f"[miles-router] Worker {url} health check failed: {e}")
+        return url, False
+
+    async def _health_check_loop(self):
```
zhaochenyang20, line 89: Good job. The update to _health_check_loop with the top-level try-except block correctly addresses the silent failure issue.

miles/router/router.py
```diff
@@ -63,9 +69,60 @@ def _setup_routes(self):
...
+                # Concurrent execution using a generator expression
+                results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))
+
+                for url, is_healthy in results:
+                    prev_failures = self.worker_failure_counts.get(url, 0)
+
+                    if is_healthy:
+                        if prev_failures >= 3:
+                            logger.info(f"[miles-router] Worker {url} has recovered and is now back in the pool.")
+                        self.worker_failure_counts[url] = 0
```
zhaochenyang20: The current logic allows a worker to auto-rejoin the pool (resetting failure count) once it becomes healthy. In distributed RL training, this is dangerous. A worker who has been disconnected for a while may hold stale model weights. Re-introducing it blindly will result in rollouts based on an old policy, causing severe off-policy issues and training instability. 

Please remove the auto-recovery logic. Once a worker is marked as Dead (>= 3 failures), it should remain quarantined permanently right now until the job restarts. Please add a TODO comment explaining that a mechanism to sync model versions is required before allowing reconnection.

miles/router/router.py
```diff
@@ -158,11 +216,16 @@ async def retrieve_from_text(self, request: Request):
         return result
 
     def _use_url(self):
-        """Select a worker URL using round-robin strategy"""
-        assert len(self.worker_urls) > 0, "No workers available"
```
zhaochenyang20, line 162: It takes me a while to understand the usage of `self.worker_urls` 🤣

The variable name self.worker_urls is misleading. It doesn't just store URLs; it maps URL -> Request Count (load state). The current name suggests a List or Set of URLs.

I suggest renaming it to self.worker_request_counts to clearly indicate that it tracks not just the worker identities but also their current load/state.

zhaochenyang20: At the same time, after the renaming, please follow this:

`_use_url` is a hot path function. And each time, we create a `healthy_urls` dict to find the lowest workload url. It incurs O(N) memory allocation, and makes routing slow.

I do not know whether you understand why we have a python router (miles-router) and a rust router (sgl-router) in Miles. Let me explain:

Routing is an extremely time - sensitive workload. We should reduce the time between "a URL is chosen" and "the request is sent to the url". Otherwise, we choose a url at timestamp `t`, when it had the minimal workload, but when the request is actually sent to the url, the workload has changed. This introduces the staleness of routing.

In this sense, please make sure your choosing algorithm is as quick as possible.

So, my suggestion is using this:

```python
def _use_url(self):
        """Select a worker URL using round-robin strategy."""
        
        # [Expert Design] Fast Path
        # In 99% of cases, all nodes are healthy.
        # If there are no dead nodes, run 'min' directly on the original dictionary
        # to avoid any memory allocation for new lists or dictionaries.
        if not self.dead_workers:
            # 'min' iterates directly over keys, resulting in zero extra memory allocation.
            url = min(self.worker_urls, key=self.worker_urls.get)
        else:
            # Slow Path - System is degraded
            # Only pay the cost of filtering when the system is in a degraded state.
            # Use a Generator Expression to further minimize memory allocation.
            valid_workers = (w for w in self.worker_urls if w not in self.dead_workers)
            try:
                url = min(valid_workers, key=self.worker_urls.get)
            except ValueError:
                # Generator is empty (all nodes are down)
                raise RuntimeError("No healthy workers available")

        self.worker_urls[url] += 1
        return url
```

I suggest using a "Fast Path" pattern:

Check if self.dead_workers is empty first. If so (the 99% case), run min directly on self.worker_urls for zero allocation.

Only when there are dead workers, use a generator expression (w for ...) to filter. This avoids creating a temporary list entirely.

zhaochenyang20 (changes requested):

miles/router/router.py
Ratish1: Thanks for the explanation, I renamed `self.worker_urls` and I implemented the fast path pattern as you suggested.

miles/router/router.py
Ratish1: Oh ok got it, that makes sense. I  removed the auto recovery logic and added the TODO comment aswell.

miles/router/router.py
Ratish1: done

miles/router/router.py
Ratish1: done

miles/router/router.py
Ratish1: done, I fixed the conditional logic. lmk if it is correct

zhaochenyang20:
@yitianlian approved this change

yueming-yuan:
lgtm~

yueming-yuan (approved):

zhaochenyang20:
@Ratish1 Will merge this PR after the code sync,.

PR #294, opened 2025-12-02
Feat: Add GitHub Actions workflow for auto linting

Lyken17:
currently miles repo requries user to run lint manually and locally before merging. This PR aims to ease the process -- users only need to tag the PR with `lint wanted` tag and the github action will automatically run the lint for us.


Lyken17:
example: https://github.com/Lyken17/miles/pull/1

fzyzcjy:
looks great! it seems we have https://github.com/radixark/miles/blob/main/.github/workflows/pre-commit.yml which contains lint as well. shall we somehow unify this new yml w/ that one

EDIT: oh I see, you are auto linting and may not want it for every commit. but it seems only admin can add labels, i.e. normal users cannot use this feature :/ if we want normal users to use this, maybe we can e.g. follow SGLang's bot, see https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests

Lyken17:
yup auto lint makes life easier. The action function is similar w/ sglang's one.

Lyken17:
btw @fzyzcjy we need to add a new tag (lint wanted) on repo to trigger the action

fzyzcjy:
(I updated my comment above a bit)

Lyken17:
@fzyzcjy  I have refactor the code to [slach-lint](https://github.com/Lyken17/miles/blob/945bc3a233ca07ff8871fb80c9cf21d05c0e4f90/.github/workflows/bot-slack-lint.yaml).  Now everyone can type `/tag-run-lint` and the bot will lint the codebase for you.

Lint is a feature that everyone needs and I feel it is not necessary to enable fine-grained permission like SGLang's `CI-run`

.github/workflows/bot-slash-lint.yaml
fzyzcjy, line 1: nit: slack->slash

fzyzcjy:
> Lint is a feature that everyone needs and I feel it is not necessary to enable fine-grained permission like SGLang's CI-run

Sure, that is pretty reasonable

fzyzcjy:
btw seems it may not be latest code, a bit confused seems there are code for labels and slash and also run-ci etc

image
image

.github/workflows/bot-slash-lint.yaml
Lyken17: fixed

Lyken17:
Have updated the codebase as suggested. Now we only have one action that triggers on `/run-lint` or `/tag-run-lint`

Btw tom could you add `lint wanted` label to miles repo? It is on the top right side

fzyzcjy:
> Btw tom could you add lint wanted label to miles repo? It is on the top right side

but as discussed before - it seems only maintainer can add that label, thus not very useful. and you have implemented hte `/run-lint` etc which may be useful I guess

EDIT: oh I see you temporarily added a label - may I know a bit why (is it for visibility - then shall we use a emoji to the comment instead, iirc many bots do that)

Lyken17:
updated  

example: https://github.com/Lyken17/miles/pull/9

Lyken17:
@fzyzcjy let's merge if no further comments?

.github/workflows/bot-slash-lint.yaml
```diff
@@ -0,0 +1,111 @@
+name: Slash Command Handler
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+permissions:
+  contents: write      # Required to push commits back to PR branch
+  pull-requests: write # Required to add labels and reactions
```
fzyzcjy: seems we do not need to add labels after the refactoring

.github/workflows/bot-slash-lint.yaml
```diff
@@ -0,0 +1,111 @@
...
+          repository: ${{ github.repository }}
+          ref: refs/pull/${{ github.event.issue.number }}/head    
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Test pre-commit hooks
```
fzyzcjy: nit: maybe change a bit of name (not "test")

Lyken17:
resolved

fzyzcjy:
Sorry I was too busy recently and missed this in my inbox. merged!

PR #299, opened 2025-12-03
Automatic HF and Megatron Checkpoint Transformation in Bash File

ZhichaoWang970201:
The original setting mannually transform hf checkpoints to megatron checkpoints. Then, the megatron checkpoints are utilized for training. Lastly, the megatron checkpoints will be transformed back to hf checkpoints. 

In this bash file, it can be done automatically. In addition, both the transformation to megatron or the transformation back to HF checkpoints can be skipped.


fzyzcjy:
some naive thoughts

1. the run-qwen3-4b.py (and other py files) does auto convert hf->megatron
2. we are migrating to Megatron Bridge, which supports loading hf ckpt directly
3. it is hard to directly run megatron->hf at the end of script, given training may not be finished (especially you specified --num-rollout 3000) and we may want intermediate ckpt

ZhichaoWang970201:
Got it.

PR #323, opened 2025-12-19
[fix] loss func return torch.tensor for new megatron compatibility

yueming-yuan:
return num_tokens as torch.tensor(1) rather than int(1)
Newer Megatron require receiving num_tokens as a torch tensor for later processing
location: `miles/backends/megatron_utils/loss.py`

### DIFF
--- a/miles/backends/megatron_utils/loss.py
+++ b/miles/backends/megatron_utils/loss.py
@@ -771,7 +771,7 @@ def loss_function(
 
     return (
         loss,
-        num_tokens if args.calculate_per_token_loss else 1,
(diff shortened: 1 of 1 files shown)


miles/backends/megatron_utils/loss.py
```diff
@@ -771,7 +771,7 @@ def loss_function(
 
     return (
         loss,
-        num_tokens if args.calculate_per_token_loss else 1,
+        num_tokens if args.calculate_per_token_loss else torch.tensor(1, dtype=torch.int, device=logits.device),
```
fzyzcjy, line 774: qq: does the `if` branch give tensor or not (seems your title says should give)

Issue #330, opened 2025-12-21
[Bug] deepscaler rm always return 0 when using non-think mode or instruct model


gongyisheng:
I'm running this [qwen3-30b-a3b example](https://github.com/radixark/miles/blob/main/scripts/run-qwen3-30B-A3B.sh), however after disable thinking mode, I found the reward is always 0.

config change I made: 
```
# added in rollout args
--apply-chat-template-kwargs '{"enable_thinking":false}'
```

result:
<img width="406" height="316" alt="Image" src="https://github.com/user-attachments/assets/b5005d13-1445-4f18-a706-884a8d140ffc" />

After debugging I found the problem is with deepscaler rm function: 
https://github.com/radixark/miles/blob/main/miles/rollout/rm_hub/deepscaler.py#L10
```
def get_deepscaler_rule_based_reward(response, label):
    if "</think>" in response:
        model_solution = response.split("</think>")[-1]
    elif "###Response" in response:
        model_solution = response.split("###Response")[1]
    else:
        return 0
```

This function always return 0 when both conditions are true:
- using a instruct model or disabled thinking
- use different ways to return response, no "###Response" detected in rollout, common in DAPO-17k dataset.

gongyisheng:
DAPO-17k prompt uses following output instruct:
```
Solve the following math problem step by step. The last line of your response should be of the form Answer: \\\\boxed{$Answer} where $Answer is the answer to the problem........Remember to put your answer on its own line after "Answer:".
```

And here's an example of rollout:
```
['[{\'content\': \'Solve the following math problem step by step. The last line of your response should be of the form Answer: \\\\boxed{$Answer} where $Answer is the answer to the problem.\\n\\nIf $A$ and $B$ are positive integers, find $A\\\\cdot B$ given that  \\\\[ \\\\begin{array}{c@{}c@{\\\\;}c@{}c@{}c@{}c}& & & A & B_{9}\\\\\\\\ &+& & & A_{9}\\\\\\\\ \\\\cline{2-5}& & & B & 0_{9}\\\\\\\\ \\\\end{array} \\\\] Express your answer in base 10 (you do not need to include the subscript).\\n\\nRemember to put your answer on its own line after "Answer:".\', \'role\': \'user\'}]

[...]

PR #332, opened 2025-12-21
[fix] fix deepscaler always return 0 when disable think

gongyisheng:
fix https://github.com/radixark/miles/issues/330


gongyisheng:
@Zhuohao-Li got you.

PR #334, opened 2025-12-21
feat: ray-less SFT

emergenz:


emergenz:
Sorry, was meant for our fork, but happy to re-open if this is interesting for upstream as well.

PR #338, opened 2025-12-23
Fix example rollout_temperature and top_k

gongyisheng:
discussed with @zhaochenyang20 , train and eval temperature / top_k is recommend to set to 1, avoid GRPO training policy mismatch and eval issues (degeneration)


zhaochenyang20 (approved):
Agreed by the team. Good to go.

Issue #315, opened 2025-12-15
[Feature Request] Improving examples scripts


BHZ-BER:
Right now, a lot of training examples are based on thinking models with default `rollout-max-response-len` and `eval-max-response-len`. This leads to a lot of truncations and a lot of examples ended up only learning to shorten the responses without really improving the performance.

We should consider switching to a better, more realistic examples in [quick_start](https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md). We can start from either Qwen 2.5 Instruct, or Qwen 3 4B base. And do RL with DAPO 17k. 

Similarly, we need the same changes for `scripts/run_qwen3_30b_a3b.py`, `scripts/run_glm45_355b_a32b.py` etc., and also provide better documentations there.

KennyYao2001:
would love to contribute!

gongyisheng:
I want to contribute on this as well

gongyisheng:
@BHZ-BER I notice that miles has two folders with run example scripts, `examples/` and `scripts/`, based on my understanding we need to update `scripts/` in this issue, is it correct?

fzyzcjy:
> based on my understanding we need to update scripts/ in this issue

Yes it is in scripts folder. I think the first step may be find such an example w/ (significantly) increasing accuracy.

BHZ-BER:
@gongyisheng @fzyzcjy Yes, and also it's best to include the best example in the [quick start doc here](https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md).

zhaochenyang20:
Hey Team, I already called out the community possibly on this. At the same time, I can be the example reviewer.

KennyYao2001:
> Hey Team, I already called out the community possibly on this. At the same time, I can be the example reviewer.

Thanks! Is there an existing public dashboard or doc that tracks the key metrics for these runs (e.g., AIME eval results, truncation rate, rollout/eval token stats)? If not, what metrics and reporting format would be most valuable?

zhaochenyang20:
Thanks, it really makes sense to me. @KennyYao2001 To be honest, I also do not know the answer. Could you possibly share your wandb link, and let's pick out of it.

KennyYao2001:
Sure! The wandb link is [here](https://wandb.ai/kenny_yao-carnegie-mellon-university/miles-dev/runs/b429qvir?nw=nwuserkenny_yao).

The args I was using for training qwen3-4B is the following：

[...]

PR #342, opened 2025-12-23
Update Qwen3-30B-A3B example

gongyisheng:
fix issue https://github.com/radixark/miles/issues/315

### DIFF
--- a/docs/en/examples/qwen3-30B-A3B.md
+++ b/docs/en/examples/qwen3-30B-A3B.md
@@ -27,6 +27,9 @@ cd /root/miles
 bash scripts/run-qwen3-30B-A3B.sh
 ```
 
+*Note* that we disabled think mode to in example script to speed up the experiment.  
+If you want to enable think, need to set `--rollout-max-response-len` to 32768
+
 ### Parameter Introduction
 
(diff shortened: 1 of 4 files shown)


docs/en/examples/qwen3-30B-A3B.md
```diff
@@ -27,6 +27,9 @@ cd /root/miles
 bash scripts/run-qwen3-30B-A3B.sh
 ```
 
+*Note* that we disabled think mode to in example script to speed up the experiment.  
+If you want to enable think, need to set `--rollout-max-response-len` to 32768
```
zhaochenyang20: I think disabled think mode is not only used to speed up. But we should describe the trade-off more clearly:

```
#  Note that: Enabling think mode requires high `rollout-max-response-len` (16k+) and 
# small batch sizes to avoid reward hacking, where the model learns to shorten 
# responses due to truncation, but does not improve on AIME scores. Disabling think 
# mode stabilizes training and VRAM but typically collapses performance on AIME.
```

zhaochenyang20 (changes requested):

docs/en/examples/qwen3-30B-A3B.md
gongyisheng: @zhaochenyang20 updated, please review.

PR #343, opened 2025-12-23
Update quick start with Qwen3-4B example

KennyYao2001:
## Motivation
Related issue: #315


zhaochenyang20 (commented):
Sounds reasonable. But I think we should follow my feedback in https://github.com/radixark/miles/pull/342/changes#r2644443037 to give users sense of tradeing off even in 4B model?

ChangyiYang:
Have rollout length 16k been tested? I'm worried that this length will cause OOM.

KennyYao2001:
> Have rollout length 16k been tested? I'm worried that this length will cause OOM.

Thanks for the question. Yes, the experiment with qwen3-4B using 16k rollout length have enough VRAM (~100GB) during rollout when using a single H200 GPU. Can adding this information to the quick_start as well to give a sense of how large the VRAM user need to have.

Single GPU:
image

ChangyiYang:
Yeah I encountered OOM when setting    --rollout-batch-size 128
   --n-samples-per-prompt 8. Great to have a ref value in the doc!

Issue #331, opened 2025-12-21
[Bug] quick start convert hf to torch dist command only use one gpu


gongyisheng:
I'm running Qwen3-30B-A3B DAPO example on 8x6000 instance, convert hf model to torch dist step, the command is copied from [quick start guide](https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md)
I found the script's speed is slow. It only uses one GPU to do conversion.

command: 
```
PYTHONPATH=/root/Megatron-LM nohup python tools/convert_hf_to_torch_dist.py \
    ${MODEL_ARGS[@]} \
    --hf-checkpoint /root/Qwen3-30B-A3B \
    --save /root/Qwen3-30B-A3B_torch_dist 2>&1 &
```

nvidia-smi screenshot: 
<img width="913" height="883" alt="Image" src="https://github.com/user-attachments/assets/fb0c86c0-8ee5-4ec5-b13e-f172587caf15" />

I tested the command on another 4xL40S instance, it raises CudaOOM error.
error: 
```
[rank0]: torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 20.00 MiB. GPU 0 has a total capacity of 44.53 GiB of which 3.25 MiB is free. Process 24355 has 44.52 GiB memory in use. Of the allocated memory 33.59 GiB is allocated by PyTorch, and 3.56 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables) when instantiating TEColumnParallelGroupedLinear when instantiating TEGroupedMLP when instantiating MoELayer
```

gongyisheng:
a potention fix: 

```
PYTHONPATH=/root/Megatron-LM torchrun --nproc_per_node=8 tools/convert_hf_to_torch_dist.py \
      ${MODEL_ARGS[@]} \
      --hf-checkpoint /root/Qwen3-30B-A3B \
      --save /root/Qwen3-30B-A3B_torch_dist 2>&1 &
```

confirmed on 4XL40S instance, time spent is 969s

<img width="900" height="745" alt="Image" src="https://github.com/user-attachments/assets/7d34207d-354c-4e06-abef-fb1e890d267b" />

NJX-njx:
[...]

PR #344, opened 2025-12-23
Add megatron conversion torchrun example in quick start

gongyisheng:
https://github.com/radixark/miles/issues/331


BHZ-BER (commented):
This looks great! Thanks! Can you add one additional sentence saying that this is an example conversion script with single node, 8 GPU?

gongyisheng:
@BHZ-BER Updated, thanks for your feedback!

Issue #314, opened 2025-12-15
[Feature Request] Improving logging processing logics


BHZ-BER:
Adding some feature requests for logging processing.

1. Add Max/Min/Std metrics for 0 average metrics like pg_loss, advantage. When we do GRPO with 1 step update per batch, because we're doing per sample mean, the pg_loss, advantage are usually 0 (or a very small number due to numerics). It's better to add max / mean / std metrics there to help people better understand the scale of advantage and loss for each token. 
2. Why do we have a different post-processing of log data for different training backend (FSDP & Megatron)? For example, we calculate sample mean of all metrics for megatron [here](https://github.com/radixark/miles/blob/520216497f9f0ba8833cbb7eeed5308bcef36878/miles/backends/megatron_utils/data.py#L332). Maybe a cleaner way is to have a unified class / pipeline for processing statistics of rollouts? Right after feedforward, we just calculate all metrics / statistics based on the shared pipeline? Is there any reason we don't do this? 
```
rollouts = await asyncio.gather()
rollouts = Postprocessing(rollouts)
```

aaaandychen:
Hi, I would love to contribute

Sophie8:
implemented my thoughts about how to unify the logging at the pr above, @Ratish1 could we work on this together> very happy to discuss further about the designs if we can cross-check and see how to consolidate, and learn from your thoughts, thanks!

Ratish1:
Sure @Sophie8 , it seems there is another PR for this now also.

Sophie8:
yes checked all and seems there are overlaps and just variants of postprocessor designs, as it's a new class with broad use cases later, people may have different opinions about the refactoring. @Ratish1 @aaaandychen let's have a sync and consolidate into one choice? left my contact info at the doc, thanks!

> Sure [@Sophie8](https://github.com/Sophie8) , it seems there is another PR for this now also.

Ratish1:
Hey @Sophie8 , you meant your wechat id?. I dont use wechat. Can you message me on slack for this? (my slack user is Ratish P). Thx

aaaandychen:
Hi @Sophie8  @Ratish1,glad to follow up on the Slack suggestion to hammer out the details. You can find me as Zhenyang Chen. We can summarize the final consensus back here.

Sophie8:
[...]

PR #348, opened 2025-12-24
feat: unify logging processing and add detailed metrics

Ratish1:
## Description
  This PR addresses two feature requests to improve the consistency of logging in miles #314 :

1. Detailed Metrics for Zero-Average Signals: For algorithms like GRPO, metrics such as pg_loss and advantages often have a near-zero mean. I have added global Max, Min, and Standard Deviation metrics to help researchers understand the actual scale of  these signals for each token.
3. Unified Processing Pipeline: I unified the post-processing logic for rollout  and training statistics across different backends (Megatron & FSDP). This is done via a new RolloutPostprocessor class that centralizes distribute reductions and log formatting.

 ## Key Changes
* `miles/utils/postprocessing.py`: Introduced a centralized pipeline for calculating global statistics (Mean, Max, Min, Std Dev) using numerically stable distributed reductions.
* Parallelism-Aware Stats: Implemented get_sample_means in both Megatron and FSDP backends. This ensures that metrics are correctly aggregated across Contex Parallelism (CP) ranks, reporting stats for full sequences rather than fragments.    
 * Backend Synchronization: Updated the training loops in both backends to call the unified pipeline immediately after feedforward, ensuring consistent logging regardless of the training backend.

### DIFF
--- a/miles/backends/fsdp_utils/actor.py
+++ b/miles/backends/fsdp_utils/actor.py
@@ -18,6 +18,7 @@ from miles.utils.data import get_minimum_num_micro_batch_size, process_rollout_d
 from miles.utils.distributed_utils import get_gloo_group
 from miles.utils.memory_utils import clear_memory, print_memory
 from miles.utils.metric_utils import compute_rollout_step
+from miles.utils.postprocessing import compute_detailed_stats, log_stats
 from miles.utils.ppo_utils import compute_approx_kl, compute_gspo_kl, compute_opsm_mask, compute_policy_loss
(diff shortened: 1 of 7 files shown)


yueming-yuan:
Thanks for your contribution! This is a great design, but we may consider a systematic refactoring to unify the training backend. We may need to consider this later when we finalize the overall design. Thank you again!

Ratish1:
Hello @yueming-yuan , has the training backend been unified?. If so, could I start fixing the conflicts on this PR to match the new unified interface?. Thanks

PR #383, opened 2026-01-02
update code

miles-code-angel:


yushengsu-thu:
This PR did not include CI/CD part syncing, right? 
miles: https://github.com/radixark/miles/tree/main/.github/workflows
slime: https://github.com/THUDM/slime/tree/main/.github/workflows
So, will I need to manually deal with `.github/workflows` after syncing?

PR #412, opened 2026-01-10
refactor [1/X]: unify training backends by general utils, tested Megatron & FSDP alignment

yueming-yuan:
### TODO list in this PR:

- [x]  Separate general logics from megatron utils and move to training_utils, remove megatron dependency there
- [x]  rewrite FSDP's logic, reuse general utils, remove duplicated implementations
- [x]  add unit tests to make sure Megatron & FSDP's grad norm match

### Description:

In this PR, the training backends are refactored by reusing general training utils (refactored from Megatron utils), rather than independent implementations. The FSDP part is very heavily modified.

Result: Now the grad norms of the same input, when CP is not enabled, have been aligned between Megatron & FSDP (dense text model tested).

Added CI to test Megatron & FSDP alignment. (use run-ci-precision or run-ci-fsdp to trigger)

For now, **FSDP CP is temporarily disabled** because non-zigzag style data sharding is not supported in general training utils. Will support it soon.

### Future refactor plan along this way:

- support non-zigzag CP slicing format in training_utils/cp_utils.py for FSDP CP & all-gather KV backend
- Fully remove megatron dependency from slime/utils (e.g. allow fallback to naive impls)
- support pip install slime without megatron
- more refactor: better abstraction...
- more unit tests
- support packed_seq/unpacked_seq & zigzag/non-zigzag data format on both side and make it cleaner, considering we may need to support more new models without comprehensive backend support

### DIFF
--- a/miles/backends/fsdp_utils/parallel.py
+++ b/miles/backends/fsdp_utils/parallel.py
@@ -21,9 +21,7 @@ def create_fsdp_parallel_state(args: Namespace) -> ParallelState:
     dp_rank = rank // cp_size
     cp_rank = rank % cp_size
 
-    mesh = init_device_mesh(
-        "cuda", mesh_shape=(world_size // cp_size, cp_size), mesh_dim_names=("dp", "cp")
-    )
+    mesh = init_device_mesh("cuda", mesh_shape=(world_size // cp_size, cp_size), mesh_dim_names=("dp", "cp"))
 
     logger.info(
         f"[Rank {rank}] Device mesh (2D): world_size={world_size}, "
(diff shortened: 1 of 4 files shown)


miles/backends/fsdp_utils/parallel.py
```diff
@@ -0,0 +1,56 @@
...
+from torch.distributed.device_mesh import init_device_mesh
+
+from miles.utils.distributed_utils import get_gloo_group
+
+from ..training_utils.parallel import ParallelState
+
+logger = logging.getLogger(__name__)
+
+
+class FSDPParallelState(ParallelState):
```
fzyzcjy: tiny nit: maybe favor creating vanilla object instead of subclassing

fzyzcjy:
discussed offline, not review in detail, but based on the facts that

* the impl is following what we discussed before
* megatron part is just code moving w/o changing any detail
* fsdp is rewritten but the old code is wrong
* CI passes

then I think the general direction looks somehow safe

miles/backends/fsdp_utils/parallel.py
yueming-yuan: (modified)

Issue #405, opened 2026-01-06
Weight update bitwise correctness checker


fzyzcjy:
Currently I have already supported some weight checking logic. It would be great if we make it a test, and/or check every time. I do not have time recently thus write this as an issue.

I guess there may be a few things to do:

* Try to enable those flags and see whether there are any issues (or it works well and only enable that flag), especially if we want to enable it by default for all cases
* May implement some logic like bf16 training fp8 rollout checker etc
* May support things like "compute checksum in miles weight, compute checksum in sglang weight (after updates), check whether they are identical"

Ratish1:
Hey @fzyzcjy , I'd like to work on this. Thanks.

PR #415, opened 2026-01-10
feat: Implement bitwise weight correctness checker for Miles-SGLang sync

Ratish1:
## Description
This PR implements a Bitwise Weight Correctness Checker to verify that model weights synced from the Miles training engine (Megatron-LM) to the SGLang inference engine are bit-for-bit identical.  #405 

## Key Changes

1. Miles Training Engine
- Arguments: Added --enable-weight-checker in miles/utils/arguments.py. Decoupled it from --check-weight-update-equal to allow independent bitwise verification.
- Updated miles/ray/rollout.py to accept and propagate a payload dictionary in the check_weights method.
- Updated miles/backends/sglang_utils/sglang_engine.py to include the payload in the weights_checker HTTP request.
- Modified UpdateWeightFromTensor and UpdateWeightFromDistributed to compute SHA256 hashes of the full parameters (after gathering shards into HF format) on Rank 0.
- Updated check_weights method to transmit these ground-truth hashes to SGLang engines using the compare_checksum action.

2. SGLang changes here sgl-project/sglang/pull/17009
- IOStruct: Updated CheckWeightsReqInput to accept optional checksums (list | dict)  and rank_offset (int).
- Scheduler/ModelRunner:  Propagated the checksum payload from the API endpoint to the WeightChecker.
- WeightChecker: Implemented a 3-stage verification protocol:
  - Stage A (Direct Match): Verifies replicated parameters (e.g., Norms, Biases) directly.
  - Stage B (Shard Reconstruction): Gathers local Tensor Parallel (TP) shards from all workers and reconstructs the full parameter by concatenating along Dim 0 (ColumnParallel) or Dim 1 (RowParallel).
  - Stage C (Dtype Alignment): Casts reconstructed tensors to bfloat16 to ensure bit-perfect representation matching the trainer's source format before hashing.

3. Tests
- Created tests/test_weight_update_correctness.py which runs Qwen2.5-3B on 2 GPUs (TP=2). It validates that the weight update handshake completes successfully and triggers no bitwise mismatches.


zhaochenyang20:
Great work so far. I strongly suggest that we shall first submit the patch of SGLang to SGLang directly. Then come back to use it in Miles. Could we do it as I described. Thanks!

Ratish1:
> Great work so far. I strongly suggest that we shall first submit the patch of SGLang to SGLang directly. Then come back to use it in Miles. Could we do it as I described. Thanks!

Yes, I will open a PR in SGlang today itself. Thanks.

fzyzcjy (commented):
tiny nit: would be great to improve code quality of the whole pr a little bit

Ratish1:
- I updated the Miles side of this PR to make the two verification paths explicit and independent. `--check-weight-update-equal` now stays on the original snapshot, reset, and compare flow, while the checksum-based path is owned by the explicit `--enable-weight-checksum-checker` flag. I kept `--enable-weight-checker` as a backward-compatible alias, but renamed the primary flag because this path is specifically doing checksum verification.
- I centralized checksum generation so the tensor update path and the distributed update path use the same hashing logic, which removes duplicated checksum code and keeps the two sync implementations aligned.
- I also fixed the Megatron argument normalization path so existing Qwen model scripts that pass `--norm-epsilon` validate cleanly against HF config checks that expect `layernorm_epsilon`.
- On the test side, I added focused e2e coverage for the legacy equality checker, the normal checksum path, the `bf16 training -> fp8 rollout` checksum path, and a non-colocated distributed checksum smoke test.
- The main design point on the Miles side is that the checksum contract follows the actual rollout payload that Miles produces and sends to SGLang after conversion. In the normal rollout case that means the normal converted rollout payload, and in the mixed-precision case that means the FP8 rollout payload plus its scale tensors, not the raw BF16 trainer tensor. This keeps the checker aligned with what is actually being synced into the rollout engine.
- When testing locally, the checksum checker now passes both before rollout and again after a real weight update on the normal BF16 path and on the BF16-train / FP8-rollout path.


bf16 train, fp8 rollout:
image

normal checksum test:
image

weight update equal check:
image

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -274,6 +279,28 @@ def _send_hf_params(self, hf_named_tensors) -> tuple[list[ObjectRef], Any]:
 
         return all_refs, long_lived_tensors
 
+    def check_weights(self, action: str) -> None:
+        if self.args.enable_weight_checksum_checker and self._last_checksums:
+            if dist.get_rank() == 0:
+                target_engines = list(self.rollout_engines)
+                if self.use_distribute:
+                    target_engines.extend(self.distributed_rollout_engines)
```
fzyzcjy: dup 1a

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -274,6 +279,28 @@ def _send_hf_params(self, hf_named_tensors) -> tuple[list[ObjectRef], Any]:
...
+                            action="compare_checksum", payload={"checksums": self._last_checksums}
+                        )
+                        for engine in target_engines
+                    ]
+                )
+        else:
+            if dist.get_rank() == 0:
+                target_engines = list(self.rollout_engines)
+                if self.use_distribute:
+                    target_engines.extend(self.distributed_rollout_engines)
```
fzyzcjy: dup 1b

miles/backends/megatron_utils/update_weight/update_weight_from_distributed.py
```diff
@@ -140,6 +142,22 @@ def update_weights(self) -> None:
...
+                    [
+                        engine.check_weights.remote(
+                            action="compare_checksum", payload={"checksums": self._last_checksums}
+                        )
+                        for engine in self.rollout_engines
+                    ]
+                )
+        else:
+            if dist.get_rank() == 0:
+                ray.get([engine.check_weights.remote(action=action) for engine in self.rollout_engines])
```
fzyzcjy: dup 2a

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -274,6 +279,28 @@ def _send_hf_params(self, hf_named_tensors) -> tuple[list[ObjectRef], Any]:
...
+                ray.get([engine.check_weights.remote(action=action) for engine in target_engines])
```
fzyzcjy: dup 2b

fzyzcjy (commented):
quick question: wondering whether these are a little bit of code dup that may be somehow unified to make it easier to maintain

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
Ratish1: done

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
Ratish1: done

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
Ratish1: done

miles/backends/megatron_utils/update_weight/update_weight_from_distributed.py
Ratish1: done

fzyzcjy:
wondering whether the things we compare are able to be bitwise equal, will have a check later

Ratish1:
> wondering whether the things we compare are able to be bitwise equal, will have a check later

thanks

PR #426, opened 2026-01-13
[squashed] Support VLM Multi-turn Training with Megatron, Support INT4 training, bug fix, etc.

zhaochenyang20:
**Thanks to miles's clean design, multi-turn RL aligns with first principles: with a custom rollout function, any training backend (e.g. Megatron/FSDP) can use it.**

Thanks so much for Xiaole Guo, Nan Jiang and Jin Pan's contribution.


.gitmodules
yueming-yuan, line 1: This should not be changed.. May need to skip `.gitsubmodule` in sync script

zhaochenyang20: thanks for poninting this up.

Issue #423, opened 2026-01-13
[Refactor & fix] Request: refactor and fix swe-agent example


yueming-yuan:
Refactor: 

The current swe-agent example (examples/experimental/swe-agent) is based on `nemo-gym` and `nv-mini-swe-agent`. This was a temporary solution with internal considerations. However, this solution is not clean when provided as an example for miles. We may not need to use `nemo-gym` for server setup, but directly use `mini-swe-agent` as a lightweight external module for simplicity, which is more like other examples.

Fix:

_(If you choose to refactor, you may not need to follow the current design and fix it in this way. But just to show the problem for reference)_

Also, the current `nv-mini-swe-agent` does not apply chat templates to sglang engine. It (`sglang_model.py`) currently uses `/generate`, but should be changed to OpenAI APIs instead. Also, the processing in `generate_with_swe_agent.py` needs to be modified along with this change.

Request:

Overall, I suggest doing a full refactor of this example. If anyone has time, feel free to look at it. Also, I'm happy to chat at anytime to discuss or help.

Ratish1:
Hello @yueming-yuan , I'd like to work on this. Thanks.

JunzheJoe:
Hi! I’m a student in the CodePath AI301 program, and I’d like to work on this issue.

I noticed #442 attempted this refactor but is currently still open and blocked by merge conflicts / a failing fast test. I’m going to start by reading the current implementation and the prior PR, then put together a focused approach based on what still applies to current main.

Any pointers or preferences on scope would be appreciated! Thanks!

PR #442, opened 2026-01-14
[Refactor]: SWE-agent integration and chat template fix

Ratish1:
**Description**

This PR refactors the experimental swe-agent example to make environment setup significantly simpler and fix model reasoning collapse / infinite loops. #423 

**Key Changes**

- Removed the nemo-gym dependency. The generate function now instantiates DockerEnvironment and DefaultAgent directly from minisweagent, eliminating the need for an external HTTP server and
  reducing setup complexity.
- Integrated asyncio.to_thread in the rollout process. This allows Miles to run multiple agent tasks in parallel across all allocated GPUs while maintaining Ray cluster responsiveness and heartbeats.
- Chat Template Fix: Updated the mini-swe-agent model logic to use /v1/chat/completions instead of /generate. Models now correctly apply their internal chat templates


Ratish1:
Hey @yueming-yuan , I wanted to follow up on this in case you missed it. Thanks.

PR #485, opened 2026-01-17
Improve fault tolerance for router session retrieval

fzyzcjy:


miles/rollout/base_types.py
```diff
@@ -1,22 +1,103 @@
...
+from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
+from miles.rollout.data_source import DataSource
 from miles.utils.types import Sample
 
+if TYPE_CHECKING:
+    from miles.rollout.modular_rollout.orchestration_common import GenerateState
+
+
+@dataclass(frozen=True)
```
Skylion007, line 15: Nit, frozen usually should also imply slots=True as long as Python 3.10 or newer

PR #492, opened 2026-01-18
Enable experimental rollout flag for CI tests

fzyzcjy:


yueming-yuan (approved):
(approved, designs are clear, and tests are widely covered.)

fzyzcjy:
hmm it would be great to not merge in one pr, which introduces a 6000 line of code change, and the logical history is lost, making it a bit hard to debug and change part of it.

image

PR #498, opened 2026-01-20
[VLM Multi-turn] delete fsdp

gxlvera:
This PR delete fsdp-related content for geo3k vlm multi-turn example, and only keeps Megatron-LM as training backend.


examples/geo3k_vlm_multi_turn/run_geo3k_vlm_multi_turn.py
```diff
@@ -168,4 +159,4 @@ def execute():
 
 if __name__ == "__main__":
     prepare()
-    execute()
+    execute()
```
zhaochenyang20: Why change this?

PR #502, opened 2026-01-22
[feat] Support OAI TITO v1

guapisolo:
This PR support several things (sorry I found it hard to split it into seperate PR.. Will do better in later PR) 

Under /chat/completions endpoint with miles router enable:
- [x] Support token consistency between train & inferenc by adding trajectory manager middleware to store history samples.
- [x] Support token in for SGLang. Miles can enable token input to SGLang by setting `miles_router_enable_token_input_for_chat_completions`. (text model only). 
- [x] Refine token out in SGLang engine, including the prompt part token out and token id (`convert_ids_to_tokens` sometimes has bug when the output token text include white space).
- [x] Refine agentic tool call impl. 
- [x] Refine some code logic in router to make it decouple from radix tree. 
- [x] Doc for refactored multi-turn & agentic 
- [ ] Cross turn token consistency is not supported. which means, in each turn we retokenize the OAI messages in the trajectory.
Why? Chat template is not append-only. e.g. Different model cut think part in very different timing and scenario. It makes cross turn token inherit tightly coupled with the chat template, and very hard to find a general solution.、

TODO:
- [ ] API wrapper.

Thanks @ChangyiYang and @swordfaith for discussing details 

Test details:
Performance comparison
image
image
There is definitely some overhead, but generally acceptable.

Accuracy comparison:
The change of raw reward aligns very well.
image

### DIFF
--- a/tests/fast/utils/test_utils/test_mock_sglang_server.py
+++ b/tests/fast/utils/test_utils/test_mock_sglang_server.py
@@ -23,13 +23,9 @@ def expected_logprobs(tokenizer, text: str) -> list[dict]:
     ]
 
 
-def expected_input_logprobs(tokenizer, messages: list[dict], tools: list[dict] | None) -> list[dict]:
+def expected_input_token_ids(tokenizer, messages: list[dict], tools: list[dict] | None) -> list[int]:
     prompt_str = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True, tools=tools)
-    prompt_ids = tokenizer.encode(prompt_str, add_special_tokens=False)
-    return [
-        {"token": tokenizer.convert_ids_to_tokens(tid), "token_id": tid, "logprob": -i / 128}
-        for i, tid in enumerate(prompt_ids)
-    ]
+    return tokenizer.encode(prompt_str, add_special_tokens=False)
 
 
 @pytest.fixture(scope="module")
@@ -266,7 +262,7 @@ class TestChatCompletionsEndpoint:
                     "index": 0,
                     "message": {"role": "assistant", "content": "\\boxed{6}", "tool_calls": None},
                     "logprobs": {"content": expected_logprobs(mock_server.tokenizer, "\\boxed{6}")},
-                    "input_logprobs": {"content": expected_input_logprobs(mock_server.tokenizer, messages, None)},
+                    "input_token_ids": expected_input_token_ids(mock_server.tokenizer, messages, None),
                     "finish_reason": "stop",
                 }
             ],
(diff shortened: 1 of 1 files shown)


guapisolo:
@yueming-yuan Do you have some time to review this PR? I think it's almost ready. Thanks!

docker/patch/v0.5.7/sglang.patch
```diff
@@ -74,6 +74,213 @@ index 0478526ef..cfb1aa669 100644
...
+         Literal[
+diff --git a/python/sglang/srt/entrypoints/openai/serving_chat.py b/python/sglang/srt/entrypoints/openai/serving_chat.py
+index cb0c084a3..02babd90c 100644
+--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
++++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
+@@ -146,6 +146,62 @@ class OpenAIServingChat(OpenAIServingBase):
+ 
+         return None
+ 
++    def _hack_convert_input_ids_to_messages(
```
maocheng23: NIT, a better name would be helpful, this one is a little misleading

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -17,16 +17,21 @@ class OpenAIEndpointTracer:
...
+        session_id = response["session_id"]
         return OpenAIEndpointTracer(router_url=router_url, session_id=session_id)
 
     async def collect_records(self) -> list[SessionRecord]:
-        response = await post(f"{self.router_url}/sessions/{self.session_id}", {}, action="get")
+        try:
+            response = await post(f"{self.router_url}/sessions/{self.session_id}", {}, action="get")
+        except Exception as e:
+            logger.warning(f"Failed to get session {self.session_id} records: {e}")
+            return []
```
maocheng23: Would return [] be sufficient or the error here would trigger some further problems?

miles/router/session/sessions.py
```diff
@@ -0,0 +1,92 @@
...
+        response = json.loads(result["response_body"])
+
+        choice = response.get("choices", [{}])[0]
+        # messages = request_body["messages"] + [choice["message"]]
+
+        assert "logprobs" in choice and "content" in choice["logprobs"], "logprobs must be in choice"
+        logprobs_content = choice["logprobs"]["content"]
+
+        for item in logprobs_content:
+            assert "token_id" in item, "token_id must be in item"
```
maocheng23: In R71 and R75, use raise instead of assert may be more robust.

maocheng23:
Added some tiny comments, overall good work! And can you please attach your test results to this PR? Thanks

PR #505, opened 2026-01-22
feat: support context parallel for FSDP backend

Beichen-Ma:
# Description
This PR implements context parallel support for FSDP backend with all-gather communication type. Motivated by #412, support non-zigzag CP slicing format for FSDP CP & all-gather KV backend. 

# Tests
- `tests/test_qwen3_0.6B_megatron_fsdp_align.py` passed when CP enabled.
- Test numerical correctness by compare:
  - CP=1 and CP=2 produce identical gradient norms
  - Megatron CP=2 and FSDP CP=2 produce identical gradient norms
Screenshot 2026-01-21 at 10 29 17 PM

# Next steps
- [ ] rollout data logging for fsdp cp>1
- [ ] global padding for cp>1
- [ ] all_gather_with_cp for non-zigzag

### DIFF
--- a/miles/backends/training_utils/cp_utils.py
+++ b/miles/backends/training_utils/cp_utils.py
@@ -285,126 +285,45 @@ def get_logits_and_tokens_offset_with_cp(
 
 
 def get_sum_of_sample_mean(
+    specs: list[CPSliceSpec],
+    loss_masks: list[torch.Tensor],
     total_lengths: list[int],
     response_lengths: list[int],
-    loss_masks: list[torch.Tensor],
-    parallel_state: ParallelState,
     calculate_per_token_loss: bool = False,
-    qkv_format: str = "thd",
-    max_seq_lens: list[int] | None = None,
-    chunk_size: int | None = None,
 ) -> Callable[[torch.Tensor], torch.Tensor]:
     """
-    Calculate correct sample mean for CP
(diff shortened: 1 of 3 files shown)


maocheng23:
Hi Beichen,

Great work — this is a very important feature and we believe it will bring significant benefits to miles. 

For further validation, there are a few details we’d like to clarify:
The plot shows two fsdp-cp2 runs, but their metrics differ. Could you help explain why that is?

Additionally, we have a couple of follow-up requests:
1. Could we also run this test on other models, for example Qwen3-4B?
2. Would it be possible to add unit tests for this PR?

Thanks! 

cc @yueming-yuan

Beichen-Ma:
Thanks for the review and the kind words. The two `fsdp-cp` runs come from two different test scripts (test_fsdp_cp1_vs_cp2_align.py and test_fsdp_megatron_cp_align.py) that both log to the same wandb project - they use different rollout and serve different test purposes (fsdp-cp1 vs. fsdp-cp2 alignment, and megatron vs. fsdp alignment). I can include those tests in following commits.

I'm also refactoring the the cp slicing abstraction to better encapsulate the logic (unify the zigzag and non-zigzag slicing). I'll push that refactoring along with unit tests and the fixes for the limitations listed in the "Next Steps" section as part of this PR. Thanks again!

PR #511, opened 2026-01-24
rather professional readme document

zhaochenyang20:


README.md
```diff
@@ -1,212 +1,154 @@
...
 ---
 
-## Quick Start
-
-> **Note:** Miles is under active development. Commands and examples may evolve; please check the repo for the latest instructions.
+## 🚀 What is Miles?
 
-For a comprehensive quick start guide covering environment setup, data preparation, training startup, and key code analysis, please refer to:
-- [Quick Start Guide](./docs/en/get_started/quick_start.md)
+**Miles** is a high-performance, enterprise-ready reinforcement learning (RL) framework specifically optimized for **Large-Scale MoE (Mixture-of-Experts) models**. Built as a powerful fork of **[slime](https://github.com/THUDM/slime)**, Miles bridges the gap between research-grade RL and production-grade reliability by integrating **SGLang** for high-throughput rollout and **Megatron-LM** for scalable training.
```
yueming-yuan: maybe change **Large-Scale MoE (Mixture-of-Experts) models** to just **Large-Scale model training** or something else; because we are not only optimizing for MoE

README.md
```diff
@@ -1,212 +1,159 @@
...
-For more detailed usage, please refer to the documentation and example configs in the repo as they become available.
- 
+*   **Unified FP8 Pipeline**: The first framework to implement end-to-end FP8 sampling and training. By unifying precision across rollout and training, Miles eliminates the quantization-induced discrepancy that causes RL collapse in large MoE models.
+*   **Rollout Routing Replay (R3)**: Records expert routing decisions during SGLang inference and replays them during training to ensure bit-wise expert alignment.
+*   **INT4 QAT Support**: Recommendation for 1TB+ models to enable single-machine (e.g., H200) deployment by significantly reducing memory footprint.
 
+### 🛡️ Eliminating Train-Inference Mismatch
 
-## Recent Updates
+*   **Bit-Wise Identical Truly On-Policy**: System-level solution achieving deterministic forward/backward passes through kernel-level optimization (FlashAttention-3, DeepGEMM).
```
BHZ-BER: I think Ying also mentioned a few times that truly on-policy is a bit confusing. Shall we change the name a bit? Say "Bit-wise identical training and inference alignment"?

README.md
```diff
@@ -1,212 +1,159 @@
...
+Miles supports a wide range of state-of-the-art architectures, with a special emphasis on **DeepSeek, Qwen, Llama** and mainstream models.
 
-This makes Miles suitable for **high-stakes experiments** where repeatability, auditability, and production debugging matter.
+| Family | Supported Models |
+| :--- | :--- |
+| **DeepSeek** | **R1, V3, V3.2** |
+| **Qwen** | **Qwen 2, 2.5, 3** |
+| **Llama** | **Llama 3, 3.1, 3.3, 4** |
+| **Gemma** | **Gemma 2, 3, 3N** |
+| **Others** | **Mistral, Mixtral, Phi, gpt-oss and any model supported by SGLang and Megatron** |
```
BHZ-BER, line 69: Can we be a bit more explicit and comprehensive? Say at least we know GLM 4.5, 4.6, 4.7 are supported. What about Minimax M2 / M2.1?

README.md
```diff
@@ -1,212 +1,159 @@
...
+```bash
+python train.py \
+    --train-backend fsdp \
+    --advantage-estimator grpo \
+    --model-name qwen3-30b-a3b \
+    --hf-checkpoint /path/to/qwen3-30b-a3b-hf \
+    --true-on-policy-mode \
+    --rollout-batch-size 512 \
+    --n-samples-per-prompt 8
+```
```
BHZ-BER, line 106: For here is it possible to use a megatron based backend rather than fsdp? Given that fsdp is still a bit experimental and may have some memory issues.

Also at some point we shall say we support both FSDP2 and Megatron training backend, but the FSDP backend is still under development and experimental?

README.md
BHZ-BER: "any model supported by SGLang and Megatron" maybe we shall also add FSDP here? But also we need to mention that FSDP is still experimental. It may be unstable etc.

PR #513, opened 2026-01-24
CI: Add data packing test

guapisolo:
Some people have concern about miles's data packing logic will reduce accuracy. So add this CI.


yushengsu-thu (approved):

yushengsu-thu (commented):
Add this test into
https://github.com/radixark/miles/blob/main/.github/workflows/pr-test.yml.j2
and create label - `ci-unit-test`

PR #517, opened 2026-01-25
docs: add Miles server arguments

Ratish1:
This PR adds complete docs for Miles server arguments.

### DIFF
--- a/docs/en/advanced/miles_server_args.md
+++ b/docs/en/advanced/miles_server_args.md
@@ -325,7 +325,7 @@ Arguments for configuring reward signals and post-processing.
 | `--reward-key` | JSON key to extract the numerical reward from a returned dictionary if reward model returns a dict instead of a value. | `None` | Type: str | Miles Native |
 | `--eval-reward-key` | Evaluation variant for `--reward-key`. | `None` | Type: str | Miles Native |
 | `--custom-rm-path` | Path to a custom Python reward function. [Ref](../get_started/customization.md#3-reward-model---custom-rm-path) | `None` | Type: str | Miles Native |
-| `--group-rm` | Defer reward computation to process the entire group of samples (per-prompt) at once. Essential for comparative/ranking reward models and improves throughput. **Training only**. | `False` | bool flag (set to enable) | Miles Native |
+| `--group-rm` | Defer reward computation to process the entire group of samples (per-prompt) at once. Essential for comparative/ranking reward models and improves throughput. **Not supported in eval**. | `False` | bool flag (set to enable) | Miles Native |
 | `--custom-reward-post-process-path` | Path to a custom reward post-processor. [Ref](../get_started/customization.md#12-reward-post-processing---custom-reward-post-process-path) | `None` | Type: str | Miles Native |
 | `--custom-convert-samples-to-train-data-path` | Path to a custom data format converter. [Ref](../get_started/customization.md#13-samples-to-train-data-conversion---custom-convert-samples-to-train-data-path) | `None` | Type: str | Miles Native |
 


miles/utils/arguments.py
```diff
@@ -1022,7 +1022,7 @@ def add_wandb_arguments(parser):
                 default=None,
                 help=(
                     "Log statistics of the category of reward, such as why the reward function considers it as failed. "
-                    "Specify the key in the reward dict using this argument.",
+                    "Specify the key in the reward dict using this argument."
```
Ratish1, line 1039: This was needed so that the command `python3 train.py --help` would work or else it would return an error

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,372 @@
...
+
+---
+
+## Algorithm and RL Arguments
+
+Arguments for reinforcement learning algorithms, loss calculation, and off-policy correction.
+
+| Argument | Description | Default | Options |
+| :--- | :--- | :--- | :--- |
+| `--advantage-estimator` | Selection of the reinforcement learning algorithm. | `grpo` | `grpo`, `gspo`, `ppo`, `reinforce_plus_plus`, `on_policy_distillation` |
```
zijiexia:
```suggestion
| `--advantage-estimator` | Selection of the reinforcement learning algorithm. | `grpo` | `grpo`, `gspo`, `ppo`, `reinforce_plus_plus`, `reinforce_plus_plus_baseline`, `on_policy_distillation` |
```

zijiexia: refer to: https://github.com/radixark/miles/blob/55858437b778eda000c8d41b339b8ddfee4d2759/miles/utils/arguments.py#L787

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,372 @@
...
+
+---
+
+## Reward Model Arguments
+
+Arguments for configuring reward signals and post-processing.
+
+| Argument | Description | Default | Options |
+| :--- | :--- | :--- | :--- |
+| `--rm-type` | Built-in reward model selection. | `None` | `math`, `deepscaler`, `gpqa`, `ifbench`, `remote_rm` |
```
zijiexia:
```suggestion
| `--rm-type` | Built-in reward model selection. | `None` | `remote_rm`, `deepscaler`, `dapo`, `math`, `f1`, `gpqa`, `ifbench`, `random` |
```

zijiexia: refer to: https://github.com/radixark/miles/blob/55858437b778eda000c8d41b339b8ddfee4d2759/miles/rollout/rm_hub/__init__.py#L45

zijiexia:
Hi @Ratish1 does it makes sense to add the arguments for checkpointings? Also, I think there're some arguments not covered in this PR, e.g., FSDP: --deterministic-mode. It might make sense to provide a comprehensive overview of all arguments

Ratish1:
> Hi @Ratish1 does it makes sense to add the arguments for checkpointings? Also, I think there're some arguments not covered in this PR, e.g., FSDP: --deterministic-mode. It might make sense to provide a comprehensive overview of all arguments

Hey @zijiexia , I have added even more server arguments, could you let me know if it looks good now?. Thanks

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,457 @@
...
+| `--load` | Path to the training checkpoint to resume from. | `None` | Type: str |
+| `--save` | Directory where training checkpoints will be saved. | `None` | Type: str |
+| `--save-interval` | Interval (in rollout steps) for saving checkpoints. | `None` | Type: int |
+| `--async-save` | Enable background checkpoint saving. | `False` | bool flag |
+| `--save-hf` | Automatically save a HuggingFace-formatted model at intervals. | `None` | Type: str |
+| `--no-save-optim` | Skip saving optimizer states to save disk space. | `False` | bool flag |
+| `--ref-load` | Path to the reference model checkpoint. | `None` | Type: str |
+| `--ref-ckpt-step` | Specific step to load from the reference checkpoint. | `None` | Type: int |
+| `--critic-load` | Path to the critic model checkpoint. | `None` | Type: str |
+| `--critic-save` | Path to save the critic model checkpoint. | `None` | Type: str |
```
zijiexia: I'm not sure if this is the right place to put checkpointing arguments, maybe make sense to have a separate section for it, following the same idea as the job scripts.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,457 @@
...
+
+## FSDP Specific Arguments
+
+Arguments applicable when using `--train-backend fsdp`. **Note: The FSDP backend is still under development and experimental.**
+
+| Argument | Description | Default | Options |
+| :--- | :--- | :--- | :--- |
+| `--optimizer` | Optimizer type for FSDP. | `adam` | `adam`, `sgd` |
+| `--fsdp-cpu-offload` | Offload parameters and gradients to CPU. | `False` | bool flag (set to enable) |
+| `--fsdp-state-dict-cpu-offload` | Offload full state dict to CPU during collection. | `True` | bool flag (set to enable) |
```
zijiexia: Can you double check if the default is True? check: https://github.com/radixark/miles/blob/a8c8687ca7f1e3cf98aa6a63c3c3c0e95a79b6b1/miles/backends/fsdp_utils/arguments.py#L79

docs/en/advanced/miles_server_args.md
Ratish1: done

docs/en/advanced/miles_server_args.md
Ratish1: done

docs/en/advanced/miles_server_args.md
Ratish1: Yup I thought that too, I created a new section for it. thanks

docs/en/advanced/miles_server_args.md
Ratish1: Yes good catch, it is actually false. while the dataclass says its true, the parser argument actually set it false.

miles/utils/arguments.py
Ratish1: this is merged through another PR now.

zijiexia:
Hi @Ratish1 , I've made some changes to the docs, see this PR: https://github.com/Ratish1/miles/pull/1, please let me know what you think. Thanks!

zhaochenyang20:
1. Total GPUs per node on the machine. Specify if using fewer than 8 GPUs per node in colocate mode.

This do not affect disaggregated mode?

zhaochenyang20:
These expressions are clear. I just feel quite strange.

image

What if users turn on these params when using disaggregate? In other words, I think we should not have these parameters. Just have `--collocate` is enough.

zhaochenyang20:
Disable weights backuper to save host memory. By default, this feature is enabled.

Please explain in one or two lines what is weights-backuper and it's trade off.

zhaochenyang20:
image

explain in what case we shall keep the old actor, and the trade off to keeep it.

zhaochenyang20:
Path to the Huggingface checkpoint used to initialize SGLang and provide the tokenizer. It must have the same architecture as the model being trained. It doesn't necessarily need to contain the most up-to-date parameters.

looks weird to me. I think we can only keep this: `Path to the Huggingface checkpoint used to initialize SGLang and provide the tokenizer.`

zhaochenyang20:
image


explain what's the relationship between these three parameters.

zhaochenyang20:
Skip special tokens in the response. Useful when the response is used as a prompt for the next rollout.

is this needed in multi-turn RL. If so, please stress this

zhaochenyang20:
> It may be hard to pass special tokens in command line, in that case --rollout-stop-token-ids can be used.

what does this mean?

zhaochenyang20:
image

explain the relationship

zhaochenyang20:
> If both --num-epoch and --num-rollout are set, --num-epoch will be ignored.

`--num-epoch` will overwrite `--num-rollout` if both are set.

zhaochenyang20:
image

Put prompt data before --disable-rollout-global-dataset

zhaochenyang20:
> Disable the global dataset for rollout. If set, the rollout will use the --prompt-data as the prompt dataset, and the prompts for rollout will be sampled from the dataset. If not set, you need to manage the data by yourself.

What do you mean by manage the data by yourself? Please make it rather clearer.

zhaochenyang20:
image

I think this `If you want to use a custom template, you can set --apply-chat-template to true` is redundant. Put these descriptions in `--apply-chat-template` and explain how to use customized chat template

zhaochenyang20:
image
Put the first two arguements later with its related arguements.

zhaochenyang20:
image

What is gbs here? And, please state that settng `--num-steps-per-rollout` to `n` means that each batch of rollout data should update the policy model `n` times.

And, should the default value be `1` but not None?

zhaochenyang20:
Did you guys checked the consistency with https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md?

If not, please check it 😂

zhaochenyang20:
image

I think "max tokens per gpu should be around max_response_len // cp_size instead of max_response_len"

max tokens per gpu should strongly related with sequence length (promt + response). Not only response length?

zhaochenyang20:
image

I am quite confused with this parameters. I think for a given response, we should calculate all the tokens' log probs. The current explanation let me feel that some tokens out of the `--log-probs-max-tokens-per-gpu` will not calculate the log probs.
 This seems to be batch size parameter for megatron that how many log probs are calcualted each batch?

zhaochenyang20:
I quite don't understand:

image

why this is related with verl? Make clearer explanation and trade off.

zhaochenyang20:
image

1. explain the first one more detailedly.
2. are you sure it's for per GPU? Micro batch size per GPU? the parame name does not has per GPU.

zhaochenyang20:
two more suggestions:

Based on our discussion, here are the documentation optimization requirements for the `miles_server_args.md` file, focused on classification and mapping:

1. Clear Argument Attribution and Categorization

* **Source Labeling:** Each argument must be explicitly labeled with its origin to distinguish between **Miles native** arguments, **Megatron** passthroughs, and **SGLang/SGLang Model Gateway** arguments.

2. Explicit Mapping of Passthrough Relationships

* **Passthrough Logic:** Clearly define the relationship between Miles and its underlying frameworks.
* **Naming Conventions:** Document the existing prefix rules to help users identify the target backend:
* `--sglang-*`: Arguments passed directly to SGLang.
* `--router-*`: Arguments directed to the SGLang Model Gateway/Router.
* **No Prefix**: Default arguments corresponding to Megatron-LM.

Ratish1:
> image
> What is gbs here? And, please state that settng `--num-steps-per-rollout` to `n` means that each batch of rollout data should update the policy model `n` times.
> 
> And, should the default value be `1` but not None?

Hey @zhaochenyang20 , Currently in the codebase it seems to be None. is that intended behaviour?, Should I update it in this PR itself?

Ratish1:
> image
> 1. explain the first one more detailedly.
> 2. are you sure it's for per GPU? Micro batch size per GPU? the parame name does not has per GPU.

Yes micro batch size is per GPU

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--max-tokens-per-gpu` | The maximum number of tokens (Prompt + Response combined) per GPU for dynamic batch size. This parameter defines the total sequence length budget for packing samples into micro-batches during training. Note that when enabling context parallel (CP), the effective capacity is shared, so the value should be approximately `(Total_Sequence_Length) // cp_size`. | `None` | Type: int | Miles Native |
+| `--log-probs-max-tokens-per-gpu` | The maximum number of tokens per GPU for calculating log probs. This is used to calculate the log probs of the responses during rollout, and should be set to a larger value than `max_tokens_per_gpu` if you want better performance. | `None` | Type: int | Miles Native |
+| `--balance-data` | Balance the number of tokens between data parallel ranks with `karmarkar_karp` for verl. Note that this may allocate the different response of the same prompt into different training steps. | `False` | Type: bool | Megatron-LM |
```
zijiexia:
```suggestion
| `--balance-data` | Repartition each rollout batch so each data-parallel rank gets a similar total token count via Karmarkar-Karp method. It may be beneficial for training speed but changes per-rank sample grouping and adds a small CPU scheduling overhead. | `False` | Type: bool | Miles Native |
```

zijiexia: Hi @Ratish1 , can you also change the help in arguments.py accordingly?

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--adam-beta1` | Beta1 for Adam optimizer. | `0.9` | Type: float | Megatron-LM |
+| `--adam-beta2` | Beta2 for Adam optimizer. | `0.95` | Type: float | Megatron-LM |
+| `--adam-eps` | Epsilon for Adam optimizer. | `1e-8` | Type: float | Megatron-LM |
+| `--true-on-policy-mode` | Strictly align SGLang's log probs and training engine's log probs to bit-wise equal. This parameter is only used for FSDP right now. [Ref](https://github.com/zhaochenyang20/Awesome-ML-SYS-Tutorial/blob/main/rlhf/slime/mismatch/blog-en.md#truly-on-policy-training) | `False` | bool flag (set to enable) | Miles Native |
+| `--train-env-vars` | Extra environment variables for training process, e.g., PyTorch memory management ones. | `{}` | Type: JSON / Dict | Miles Native |
+| `--train-memory-margin-bytes` | Reserved memory margin for training in bytes. Defaults to 1GB. | `1073741824` | Type: int | Miles Native |
+| `--disable-weights-backuper` | Disables the system that backups model weights (Actor, Ref, Old Actor) to CPU RAM. Disabling saves significant host memory but prevents weight-swapping features like KL-divergence. | `False` | bool flag (set to disable) | Miles Native |
```
zijiexia:
```suggestion
| `--disable-weights-backuper` | Applies to `megatron` training backend only. Disables the system that backups model weights (Actor, Ref, Old Actor) to CPU RAM. Disabling saves significant host memory but prevents features that rely on weight-swapping, such as computing KL-divergence against a reference model. **Note**: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper. | `False` | bool flag (set to disable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+
+---
+
+## Data Arguments
+
+Arguments for dataset configuration, prompt mapping, and training batch sizes.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--prompt-data` | Path to the prompt dataset (JSONL format) and each line should contains `--input-key` and `--label-key` which will be used as the prompt and the label respectively. If you want to use a custom template, you can set `--apply-chat-template` to true | `None` | Type: str | Miles Native |
```
zijiexia:
```suggestion
| `--prompt-data` | Path to the prompt dataset (JSONL format) and each line should contains `--input-key` and `--label-key` which will be used as the prompt and the label respectively. | `None` | Type: str | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--disable-rollout-global-dataset` | Disable the global dataset for rollout. If set, the rollout will use the `--prompt-data` as the prompt dataset, and the prompts for rollout will be sampled from the dataset. If not set, you need to manage the data by your self. | `False` | bool flag (set to disable) | Miles Native |
```
zijiexia:
```suggestion
| `--disable-rollout-global-dataset` | Disable the global dataset for rollout. By default, Miles loads `--prompt-data` into a global dataset and samples from it for rollout. Setting this flag turns off this behavior, Use this flag only when providing a custom `--rollout-function-path` (and usually a custom `--data-source-path`) that handles data loading independently. | `False` | bool flag (set to disable) | Miles Native |
```

zijiexia:
> image
> I am quite confused with this parameters. I think for a given response, we should calculate all the tokens' log probs. The current explanation let me feel that some tokens out of the `--log-probs-max-tokens-per-gpu` will not calculate the log probs. This seems to be batch size parameter for megatron that how many log probs are calcualted each batch?

I feel like this argument is deprecated, I couldn't find anywhere it was referred.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--train-backend` | The backend for training. Highly suggest Megatron for numerical stability and efficiency. | `"megatron"` | `megatron`, `fsdp` | Miles Native |
+| `--qkv-format` | The QKV layout. | `"thd"` | `thd`, `bshd` | Miles Native |
+| `--optimizer` | Optimizer type. | `adam` | `adam`, `sgd` | Megatron-LM |
+| `--lr` | Learning rate for the Actor. | `1e-6` | Type: float | Megatron-LM |
+| `--lr-warmup-init` | Initial learning rate for warmup. | `0.0` | Type: float | Megatron-LM |
+| `--min-lr` | Minimum learning rate after decay. | `0.0` | Type: float | Megatron-LM |
+| `--lr-decay-style` | Learning rate decay style. | `constant`(FSDP), `linear`(Megatron) | Type: str | Megatron-LM |
```
zijiexia: Hi @Ratish1 , I think most of the arguments I put outside the Megatron/FSDP sections should be able to source back to both backend, that's why I marked both FSDP and Megatron defaults here. Could you help me double check? Thanks!

zijiexia (changes requested):

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--num-epoch` | Number of epochs for the training. If set, `num_rollout` is calculated as `(num_epoch * dataset_size) // rollout_batch_size`. **Note:** This argument takes precedence and will overwrite `--num-rollout` if both are specified. | `None` | Type: int | Miles Native |
+| `--rollout-batch-size` | Number of prompts per rollout batch. The total data returned should be `rollout_batch_size` * `n_samples_per_prompt`. | Required | Type: int | Miles Native |
+| `--n-samples-per-prompt` | Number of responses to generate for each prompt. | `1` | Type: int | Miles Native |
+| `--num-steps-per-rollout` | The number of training steps (optimizer updates) to perform using the data collected in a single rollout round. Setting this to **n** means the policy model will be updated **n** times using the same batch of rollout data. This parameter is used to automatically calculate the **Global Batch Size** using the formula: `(rollout_batch_size * n_samples_per_prompt) // num_steps_per_rollout`. **Constraint:** The system ensures that `(rollout-batch-size * n-samples-per-prompt) = (global-batch-size * num-steps-per-rollout)`. | `None` | Type: int | Miles Native |
```
zijiexia: imho, I think we should phrase it this way.

```suggestion
| `--num-steps-per-rollout` | The number of training steps to perform using the data collected in a single rollout round. Setting this to `n` means the policy model will be updated `n` times using the same batch of rollout data. This parameter will **override** the global batch size using the formula: `(rollout_batch_size * n_samples_per_prompt) // num_steps_per_rollout`. The system ensures that `(rollout-batch-size * n-samples-per-prompt) = (global-batch-size * num-steps-per-rollout)`. If this value is not provided, you have to set `--global-batch-size` explicitly. | `None` | Type: int | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--custom-model-provider-path` | Path to a custom function that replaces the default model provider. [Detail](../get_started/customization.md#20-model-provider---custom-model-provider-path) | `None` | Type: str | Miles Native |
+| `--recompute-loss-function` | Enable recomputing the loss function to save memory during training. | `False` | bool flag (set to enable) | Miles Native |
+| `--log-probs-chunk-size` | Specifies the chunk size for logprobs computation to reduce peak memory usage. Processing logits in smaller batches, it prevents CUDA OOM errors during long-context prefilling or re-computation. Set to `-1` to disable chunking. [Ref](https://github.com/sgl-project/sglang/pull/6318) | `-1` | Type: int | Miles Native |
+| `--keep-old-actor` | Maintains a "Model Queue" (Actor, Rollout Actor, Old Actor) to ensure importance sampling ratios are calculated against the exact policy version that generated the data. Essential for asynchronous RL or high-throughput scenarios where training and inference versions may diverge. **Trade-off:** Consumes additional Host Memory (~1x model size in System RAM). | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia: I think the description here is not accurate. If the `update_weights_interval == 1`, it will also save a copy of the `rollout_actor`. I suggest the following changes:
```suggestion
| `--keep-old-actor` | Maintains a "Model Queue" (Actor, Rollout Actor, Old Actor) to ensure importance sampling ratios are calculated against the exact policy version that generated the data. Essential for asynchronous RL where training and inference are decoupled, preventing mathematical incorrectness due to model staleness. It consumes additional Host Memory (extra ~1x model size for `update_weights_interval > 1` or 2x for `update_weights_interval == 1`) depending on update interval. | Miles Native |
```

zhaochenyang20:
https://github.com/radixark/miles/pull/517#issuecomment-3826106876

~~This is still quite weird to me. I am confirming with other teams.~~

This is related to B300's VRAM and host mem. Let's make it deep in the docs @zijiexia

zhaochenyang20:
Also, we should update the parameter help string in the code repo. I am thinking how to do a high level sync/auto-generation with miles arguments.py and this document. But, let's finish the document first.

Ratish1:
> Also, we should update the parameter help string in the code repo. I am thinking how to do a high level sync/auto-generation with miles arguments.py and this document. But, let's finish the document first.

Ok got it, thanks

zhaochenyang20:
Deploy training and rollout on the same GPUs. Enables --offload-train and --offload-rollout. Tip: When enabled, it is highly recommended to set --sglang-mem-fraction-static to 0.8 to ensure Megatron has enough memory to initialize before weights are offloaded to CPU.

This should be dive deep and explain specific cases on B300. Wether colocate always turn on two offload?

Ratish1:
> Wether colocate always turn on two offload?

Yes colocate always turn on two offload. And I updated the command specific for gpus. Lmk if it looks good. thanks

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+
+---
+
+## Evaluation Arguments
+
+Arguments for configuring the evaluation process during training.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--eval-interval` | Interval (in rollout steps) between evaluations. | `None` | Type: int | Megatron-LM |
```
zijiexia:
```suggestion
| `--eval-interval` | Interval (in rollout steps) between evaluations. | `None` | Type: int | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+
+---
+
+## Checkpointing and Resuming
+
+Arguments for saving and loading model states.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--load` | Path to the training model checkpoint to load. | `None` | Type: str | Megatron-LM |
```
zijiexia: This argument, though is original from Megatron, but it has been reset by miles and support FSDP backend too, maybe we should change it to `Mile Native`?

```suggestion
| `--load` | Path to the training model checkpoint to load. | `None` | Type: str | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--save` | Path to save checkpoints. | `None` | Type: str | Megatron-LM |
```
zijiexia: same

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--save-interval` | Interval (in rollout steps) to save checkpoints. Requires `--save` to be set. | `None` | Type: int | Megatron-LM |
```
zijiexia: same

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,498 @@
...
+| `--async-save` | Enable asynchronous checkpoint saving (Megatron backend only). | `False` | bool flag (set to enable) | Megatron-LM |
+| `--save-hf` | Path to save the model in HuggingFace format when using Megatron backend. The model will be saved to `save_hf.format(rollout_id)`. | `None` | Type: str | Miles Native |
+| `--no-save-optim` | If set, optimizer state is not saved with checkpoints to reduce size, but prevents resumption of training. | `False` | bool flag (set to enable) | Miles Native |
+| `--ref-load` | Path to the reference model checkpoint. Used as initial checkpoint if `--load` is not set. | `None` | Type: str | Miles Native |
+| `--ref-ckpt-step` | The checkpoint step for reference model. | `None` | Type: int | Miles Native |
+| `--critic-load` | Checkpoint to load for the critic model. | value of `--load` | Type: str | Miles Native |
+| `--critic-save` | Path to save the critic model. | `None` | Type: str | Miles Native |
+| `--start-rollout-id` | The starting rollout step, if not set, will try to load the step from `--load` when doing continue training, otherwise will be set to `0`, meaning training from start. | `None` | Type: int | Miles Native |
```
zijiexia:
```suggestion
| `--start-rollout-id` | The starting rollout step. If not set, it is inferred from the --load checkpoint when resuming training; otherwise, if training is not continuous, miles will start training from scratch | `None` | Type: int | Miles Native |
```

Ratish1: done

docs/en/advanced/miles_server_args.md
Ratish1: discussed offline. kept it as megatron-lm and mentioned that miles resets the default value.

docs/en/advanced/miles_server_args.md
Ratish1: same

docs/en/advanced/miles_server_args.md
Ratish1: same

docs/en/advanced/miles_server_args.md
Ratish1: same

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--critic-lr` | Learning rate for the Critic. Defaults to `--lr`. | `None` | Type: float | Miles Native |
+| `--critic-lr-warmup-iters` | Number of iterations for Critic learning rate linear warmup. | `0` | Type: int | Miles Native |
+| `--num-critic-only-steps` | Number of initial steps dedicated to training only the Critic. | `0` | Type: int | Miles Native |
+| `--eps-clip` | PPO clip range. | `0.2` | Type: float | Miles Native |
+| `--eps-clip-high` | PPO clip upper range (defaults to `--eps-clip` if not set). | `None` | Type: float | Miles Native |
+| `--eps-clip-c` | Lower bound for [Dual-clip PPO](https://arxiv.org/pdf/1912.09729). | `None` | Type: float | Miles Native |
+| `--value-clip` | Clip range for value loss. | `0.2` | Type: float | Miles Native |
+| `--kl-coef` | KL penalty coefficient for reward shaping. This is applied to the reward signal before advantage calculation. | `0.00` | Type: float | Miles Native |
```
zijiexia:
```suggestion
| `--kl-coef` | KL penalty coefficient for reward shaping. This is applied to the reward signal before advantage calculation for PPO and REINFORCE-style estimator. | `0.00` | Type: float | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--use-kl-loss` | Enable KL loss term in the final objective (as in GRPO). | `False` | bool flag (set to enable) | Miles Native |
+| `--kl-loss-coef` | Weight of the KL loss term in the final objective. | `0.0` | Type: float | Miles Native |
+| `--kl-loss-type` | Selection of the KL loss implementation. | `k1` | `k1`, `k2`, `k3`, `low_var_kl` | Miles Native |
```
zijiexia:
```suggestion
| `--kl-loss-type` | Selection of the KL loss implementation. See [Approximating KL Divergence](http://joschu.net/blog/kl-approx.html) for more details. | `k1` | `k1`, `k2`, `k3`, `low_var_kl` | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--use-unbiased-kl` | Enable unbiased KL estimation. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--use-unbiased-kl` | Apply Importance Sampling (IS) correction to the KL estimator. Reduces bias from distribution shift. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--entropy-coef` | Coefficient for entropy regularization. | `0.0` | Type: float | Miles Native |
```
zijiexia:
```suggestion
| `--entropy-coef` | Coefficient for entropy regularization term. Penalizes low entropy to encourage exploration and prevent premature convergence. | `0.0` | Type: float | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--gamma` | PPO GAE gamma. | `1.0` | Type: float | Miles Native |
```
zijiexia:
```suggestion
| `--gamma` | Discount factor for future rewards. Used in PPO (GAE) and REINFORCE++. | `1.0` | Type: float | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--lambd` | PPO GAE lambda. | `1.0` | Type: float | Miles Native |
+| `--normalize-advantages` | Normalize advantages within each batch. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--normalize-advantages` | Performs distributed masked whitening of advantages. Normalization statistics are computed globally across the Data-Parallel group, ignoring padding tokens. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--tis-clip` | Clipping threshold C for importance sampling ratios to control variance. | `2.0` | Type: float | Miles Native |
+| `--tis-clip-low` | Lower bound clipping threshold C for importance sampling ratios to control variance. | `0.0` | Type: float | Miles Native |
+| `--custom-tis-function-path` | Path to a custom TIS or MIS function. See [customization](../get_started/customization.md#10-custom-tisrs-function---custom-tis-function-path) for more details. | `None` | Type: str | Miles Native |
+| `--custom-pg-loss-reducer-function-path` | Custom reducer function for policy gradient loss. See [customization](../get_started/customization.md#11-custom-pg-loss-reducer---custom-pg-loss-reducer-function-path) for more details. | `None` | Type: str | Miles Native |
+| `--use-routing-replay` | Enable [Routing Replay](https://arxiv.org/abs/2507.18071). | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-routing-replay` | Enable R3: [Rollout Routing Replay](https://arxiv.org/pdf/2510.11370). | `False` | bool flag (set to enable) | Miles Native |
+| `--use-opsm` | Enable Off-Policy Sequence Masking (OPSM). | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--use-opsm` | Enable Off-Policy Sequence Masking (OPSM). Filters sequences that have **BOTH** negative advantages (bad results) AND high KL divergence (stale data). This stabilizes training by preventing updates from unreliable, highly off-policy samples. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--opsm-delta` | The threshold for Off-Policy Sequence Masking (OPSM). | `1e-4` | Type: float | Miles Native |
+| `--get-mismatch-metrics` | Whether to calculate the mismatch metrics. It will **only return mismatch metrics** but not change the loss in any way. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--get-mismatch-metrics` | Calculate mismatch metrics. If it is set, you need to provide a custom TIS function via `--custom-tis-function-path`. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--ref-update-interval` | Interval (in rollout steps) to update ref model from actor. If `None`, ref model is not updated. | `None` | Type: int | Miles Native |
+| `--reset-optimizer-states` | Resets the optimizer state after each rollout round. This clears the optimization history, which can improve stability or satisfy specific experimental requirements. | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-grpo-std-normalization` | Disable standard deviation normalization for GRPO. From [Dr.GRPO](https://arxiv.org/pdf/2503.20783) | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-rewards-normalization` | Disable rewards normalization. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-entropy` | Enable entropy calculation when calculating the logprobs from actor and reference model. This is useful for doing special loss mask. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-logprobs` | Use rollout logprobs for importance sampling ratios, use the logprobs from the actor model if not set. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--use-rollout-logprobs` | Use rollout logprobs for importance sampling ratios, use the logprobs from the actor model if not set. If `--get-mismatch-metrics` is set, the log probs will be recomputed by training engine, one more forward pass will be applied. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--use-opsm` | Enable Off-Policy Sequence Masking (OPSM). | `False` | bool flag (set to enable) | Miles Native |
+| `--opsm-delta` | The threshold for Off-Policy Sequence Masking (OPSM). | `1e-4` | Type: float | Miles Native |
+| `--get-mismatch-metrics` | Whether to calculate the mismatch metrics. It will **only return mismatch metrics** but not change the loss in any way. | `False` | bool flag (set to enable) | Miles Native |
+| `--ref-update-interval` | Interval (in rollout steps) to update ref model from actor. If `None`, ref model is not updated. | `None` | Type: int | Miles Native |
+| `--reset-optimizer-states` | Resets the optimizer state after each rollout round. This clears the optimization history, which can improve stability or satisfy specific experimental requirements. | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-grpo-std-normalization` | Disable standard deviation normalization for GRPO. From [Dr.GRPO](https://arxiv.org/pdf/2503.20783) | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-rewards-normalization` | Disable rewards normalization. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--disable-rewards-normalization` | Disable the default group-wise reward normalization for GRPO, GSPO, and REINFORCE++. This effectively skips the baseline subtraction step. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| :--- | :--- | :--- | :--- | :--- |
+| `--use-wandb` | Enable WandB logging. | `False` | bool flag (set to enable) | Miles Native |
+| `--wandb-mode` | WandB operating mode. Overrides `WANDB_MODE`. | `None` | `online`, `offline`, `disabled` | Miles Native |
+| `--wandb-project` | WandB project name. | `None` | Type: str | Megatron-LM (Reset by Miles) |
+| `--wandb-group` | WandB group name. | `None` | Type: str | Miles Native |
+| `--wandb-team` | WandB team name. | `None` | Type: str | Miles Native |
+| `--wandb-host` | WandB host address. | `None` | Type: str | Miles Native |
+| `--wandb-key` | WandB API key. | `None` | Type: str | Miles Native |
+| `--wandb-run-id` | Specific WandB run ID to resume. | `None` | Type: str | Miles Native |
+| `--wandb-dir` | Directory to store WandB logs. Default is ./wandb in current directory. | `None` | Type: str | Miles Native |
```
zijiexia:
```suggestion
| `--wandb-dir` | Directory to store WandB logs. Default is `./wandb` in current directory. | `None` | Type: str | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+
+---
+
+## Miles Router
+
+Arguments for the specialized Miles text-based router.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--use-miles-router` | Use text-based routing instead of token-based routing. | `False` | bool fla (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--use-miles-router` | Use text-based routing instead of token-based routing. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--get-mismatch-metrics` | Whether to calculate the mismatch metrics. It will **only return mismatch metrics** but not change the loss in any way. | `False` | bool flag (set to enable) | Miles Native |
+| `--ref-update-interval` | Interval (in rollout steps) to update ref model from actor. If `None`, ref model is not updated. | `None` | Type: int | Miles Native |
+| `--reset-optimizer-states` | Resets the optimizer state after each rollout round. This clears the optimization history, which can improve stability or satisfy specific experimental requirements. | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-grpo-std-normalization` | Disable standard deviation normalization for GRPO. From [Dr.GRPO](https://arxiv.org/pdf/2503.20783) | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-rewards-normalization` | Disable rewards normalization. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-entropy` | Enable entropy calculation when calculating the logprobs from actor and reference model. This is useful for doing special loss mask. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--use-rollout-entropy` | Enable entropy calculation when calculating the logprobs from actor and reference model. This is useful for implementing custom entropy-based loss masking. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+Arguments for configuring reward signals and post-processing.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--rm-type` | Built-in reward model selection. | `None` | `remote_rm`, `deepscaler`, `dapo`, `math`, `f1`, `gpqa`, `ifbench`, `random` | Miles Native |
+| `--rm-url` | URL for the reward model service (used with `--rm-type remote_rm`). | `None` | Type: str | Miles Native |
+| `--reward-key` | JSON key to extract the numerical reward from a returned dictionary if reward model return a dict instead of a value. | `None` | Type: str | Miles Native |
+| `--eval-reward-key` | Evaluation variant for `--reward-key`. | `None` | Type: str | Miles Native |
+| `--custom-rm-path` | Path to a custom Python reward function. See [customization](../get_started/customization.md#3-reward-model---custom-rm-path) for more details. | `None` | Type: str | Miles Native |
+| `--group-rm` | Compute rewards for an entire group of samples at once. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--group-rm` | Defer reward computation to process the entire group of samples (per-prompt) at once. Essential for comparative/ranking reward models and improves throughput. **Training only**. | `False` | bool flag (set to enable) | Miles Native |
```

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+
+## Rollout Buffer Management
+
+Arguments for managing the rollout data buffer.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--rollout-buffer-url` | URL for the rollout buffer service. | `None` | Type: str | Miles Native |
+| `--fetch-trajectory-retry-times` | Number of times to retry fetching trajectory, -1 means unlimited retry. | `-1` | Type: int | Miles Native |
+| `--min-batch-collection-ratio` | Minimum batch collection ratio before proceeding. | `1.0` | Type: float | Miles Native |
```
zijiexia, line 342: I couldn't find anywhere this arg is used, can you double check @Ratish1 ?

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+---
+
+## Rollout Buffer Management
+
+Arguments for managing the rollout data buffer.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--rollout-buffer-url` | URL for the rollout buffer service. | `None` | Type: str | Miles Native |
+| `--fetch-trajectory-retry-times` | Number of times to retry fetching trajectory, -1 means unlimited retry. | `-1` | Type: int | Miles Native |
```
zijiexia, line 341: I couldn't find anywhere this arg is used

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+
+---
+
+## Rollout Buffer Management
+
+Arguments for managing the rollout data buffer.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--rollout-buffer-url` | URL for the rollout buffer service. | `None` | Type: str | Miles Native |
```
zijiexia, line 340: I couldn't find anywhere this arg is used

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,501 @@
...
+| `--num-gpus-per-node` | Total GPUs per node on the physical machine. This informs the Ray scheduler of the hardware capacity. In **Colocate mode**, it is required if the machine has fewer than 8 GPUs to calculate correct VRAM offsets. In **Disaggregated mode**, it ensures SGLang engines are distributed correctly across nodes without exceeding per-node GPU limits. | `8` | Type: int | Miles Native |
+| `--colocate` | Deploy training and rollout on the same GPUs. This mode automatically enables `--offload-train` and `--offload-rollout` to facilitate weight-swapping between the training actor and inference engine. **Memory Tip:** When colocating, it is highly recommended to set `--sglang-mem-fraction-static` to **0.8** (especially on **NVIDIA Blackwell B200/B300** GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200, values up to 0.75 are safer for long-running jobs to prevent potential OOMs. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia: Is this statement about GB200 still valid? @zhaochenyang20

Ratish1:
Hey @zijiexia , those args are present here?
https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207
https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1214
https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207

zhaochenyang20:
> Deploy training and rollout on the same GPUs. This mode automatically enables --offload-train and --offload-rollout to facilitate weight-swapping between the training actor and inference engine. Memory Tip: When colocating, it is highly recommended to set --sglang-mem-fraction-static to 0.8 (especially on NVIDIA Blackwell B200/B300 GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200, values up to 0.75 are safer for long-running jobs to prevent potential OOMs.

Are you sure these numbers are right? And, I want to have a comment as I put it in TODO in the docs.

Note that most use cases do not need to consider offload parameters, including `--offload-rollout, --no-offload-rollout, --offload-training, --no-offload-training`. They are used only on AMD GPUs and will eventually be removed. [Please help me check this].

zijiexia:
> Hey @zijiexia , those args are present here?
> 
> https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207
> 
> 
> https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1214
> 
> 
> https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207

Hi @Ratish1 , these arguments are added here but never referred in the elsewhere of the miles codebase. Can you double check?

zhaochenyang20:
Note that most use cases do not need to consider offload parameters, including --offload-rollout, --no-offload-rollout, --offload-training, --no-offload-training. They are used only on AMD GPUs and will eventually be removed. [Please help me check this].

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,499 @@
...
+| `--num-gpus-per-node` | Total GPUs per node on the physical machine. This informs the Ray scheduler of the hardware capacity. In **Colocate mode**, it is required that the machine has fewer than 8 GPUs to calculate correct VRAM offsets. In **Disaggregated mode**, it ensures SGLang engines are distributed correctly across nodes without exceeding per-node GPU limits. | `8` | Type: int | Miles Native |
+| `--colocate` | Deploy training and rollout on the same GPUs. This mode automatically enables `--offload-train` and `--offload-rollout` to facilitate weight-swapping between the training actor and inference engine. **Memory Tip:** When colocating, it is highly recommended to set `--sglang-mem-fraction-static` to **0.8** (especially on **NVIDIA Blackwell B200/B300** GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200, values up to 0.75 are safer for long-running jobs to prevent potential OOMs. | `False` | bool flag (set to enable) | Miles Native |
```
zijiexia:
```suggestion
| `--colocate` | Deploy training and rollout on the same GPUs. This mode automatically enables `--offload-train` and `--offload-rollout` to facilitate weight-swapping between the training actor and inference engine. **Note:** The offload parameters are currently only used for AMD GPUs and will be removed soon. **Memory Tip:** When colocating, it is highly recommended to set `--sglang-mem-fraction-static` to **0.8** (especially on **NVIDIA Blackwell B200/B300** GPUs). On GB200/GB300, values up to 0.75 are safer for long-running jobs to prevent potential OOMs. | `False` | bool flag (set to enable) | Miles Native |
```

zijiexia (changes requested):

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,499 @@
...
+| `--prefill-num-servers` | Number of dedicated prefill servers for PD disaggregation. | `None` | Type: int | Miles Native |
+| `--distributed-backend` | Backend for distributed communication. | `nccl` | `nccl`, `gloo` | Megatron-LM (Reset by Miles) |
+| `--distributed-timeout-minutes` | Timeout for distributed operations in minutes. | `10` | Type: int | Megatron-LM (Reset by Miles) |
+
+Note that most use cases do not need to consider offload parameters, including `--offload-rollout, --no-offload-rollout, --offload-training, --no-offload-training`. They are used only on AMD GPUs and will eventually be removed. [Please help me check this].
```
zijiexia: Based on the discussion in group chat, this statement is correct.

Ratish1:
> > Deploy training and rollout on the same GPUs. This mode automatically enables --offload-train and --offload-rollout to facilitate weight-swapping between the training actor and inference engine. Memory Tip: When colocating, it is highly recommended to set --sglang-mem-fraction-static to 0.8 (especially on NVIDIA Blackwell B200/B300 GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200, values up to 0.75 are safer for long-running jobs to prevent potential OOMs.
> 
> Are you sure these numbers are right? And, I want to have a comment as I put it in TODO in the docs.
> 
> Note that most use cases do not need to consider offload parameters, including `--offload-rollout, --no-offload-rollout, --offload-training, --no-offload-training`. They are used only on AMD GPUs and will eventually be removed. [Please help me check this].

Hey @zhaochenyang20 , I did those numbers based on the numbers used for the training scripts present in the codebase.

> > Hey @zijiexia , those args are present here?
> > https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207
> > 
> > https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1214
> > 
> > https://github.com/radixark/miles/blob/0ee8847d11208dcdd900584999e957537c764f37/miles/utils/arguments.py#L1207
> 
> Hi @Ratish1 , these arguments are added here but never referred in the elsewhere of the miles codebase. Can you double check?

Hey @zijiexia , yeah those commands are not used in the codebase, should I remove them from arguments.py or from our docs only?

Ratish1:
> > Deploy training and rollout on the same GPUs. This mode automatically enables --offload-train and --offload-rollout to facilitate weight-swapping between the training actor and inference engine. Memory Tip: When colocating, it is highly recommended to set --sglang-mem-fraction-static to 0.8 (especially on NVIDIA Blackwell B200/B300 GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200, values up to 0.75 are safer for long-running jobs to prevent potential OOMs.
> 
> Are you sure these numbers are right? And, I want to have a comment as I put it in TODO in the docs.
> 
> Note that most use cases do not need to consider offload parameters, including `--offload-rollout, --no-offload-rollout, --offload-training, --no-offload-training`. They are used only on AMD GPUs and will eventually be removed. [Please help me check this].

Hey @zhaochenyang20 , I did those numbers based on the training scripts values present and I added the TODO to the command.

zhaochenyang20:
image

The rendering for this parameter is broken. Please preview it locally. 😂

zhaochenyang20:
image
 
Put this above:

image

I've done this

zhaochenyang20:
image

The definition of `--over-sampling-batch-size` is quite unclear. Please refer to https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md#dynamic-sampling to change this parameters definition. And, make a reference link in the parameter docs.

zhaochenyang20:
image

following my comments here:

https://github.com/radixark/miles/pull/517#issuecomment-3863129507

As for partial rollouts, explain it more in detail and add ra eference link to it:

https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md#partial-rollout

zhaochenyang20:
image

I find that on github, none of these links can lead me to jump to that part.

zhaochenyang20:
change all the:

`See customization for more details.`
 
 to

`[Ref](xxxx)`

zhaochenyang20:
After this is merged:

https://github.com/radixark/miles/pull/538#pullrequestreview-3765717485

Link R3 docs to it:

image

have the paper and document link.

zhaochenyang20:
After this is merged:

https://github.com/radixark/miles/pull/538#pullrequestreview-3765717485

Link docs of miles router to it:

image

Ratish1:
> I find that on github, none of these links can lead me to jump to that part.

I think it is a github problem when viewing it like that. Click view file and then the link clicking works for me fine
image

zhaochenyang20:
R3 docs is merged, we can work on this:

https://github.com/radixark/miles/pull/517#issuecomment-3863190839

And, please rebase with main.

zhaochenyang20 (approved):
Great work! @Ratish1 @zijiexia

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+Arguments for configuring reward signals and post-processing.
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--rm-type` | Built-in reward model selection. | `None` | `remote_rm`, `deepscaler`, `dapo`, `math`, `f1`, `gpqa`, `ifbench`, `random` | Miles Native |
+| `--rm-url` | URL for the reward model service (used with `--rm-type remote_rm`). | `None` | Type: str | Miles Native |
+| `--reward-key` | JSON key to extract the numerical reward from a returned dictionary if reward model returns a dict instead of a value. | `None` | Type: str | Miles Native |
+| `--eval-reward-key` | Evaluation variant for `--reward-key`. | `None` | Type: str | Miles Native |
+| `--custom-rm-path` | Path to a custom Python reward function. [Ref](../get_started/customization.md#3-reward-model---custom-rm-path) | `None` | Type: str | Miles Native |
+| `--group-rm` | Defer reward computation to process the entire group of samples (per-prompt) at once. Essential for comparative/ranking reward models and improves throughput. **Training only**. | `False` | bool flag (set to enable) | Miles Native |
```
guapisolo: What's the meaning of **training only** here

guapisolo:
Thanks for writing the docs. I will carefully review this.

docs/en/advanced/miles_server_args.md
zijiexia: I think I mentioned it because `--group-rm` is explicitly disabled in evaluation.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--disable-rewards-normalization` | Disable the default group-wise reward normalization for GRPO, GSPO, and REINFORCE++. This effectively skips the baseline subtraction step. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-entropy` | Enable entropy calculation when calculating the logprobs from actor and reference model. This is useful for implementing custom entropy-based loss masking. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-logprobs` | Use rollout logprobs for importance sampling ratios, use the logprobs from the actor model if not set. If `--get-mismatch-metrics` is set, the log probs will be recomputed by training engine, one more forward pass will be applied. | `False` | bool flag (set to enable) | Miles Native |
+| `--calculate-per-token-loss` | Calculate loss on a per-token basis. | `False` | bool flag (set to enable) | Megatron-LM (Reset by Miles) |
+| `--seed` | Random seed for the training process. | `1234` | Type: int | Megatron-LM (Reset by Miles) |
```
guapisolo: This `--seed` also affect sglang server's `--random-seed` param.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--use-tis` | Enable Token-level Importance Sampling (TIS) from this [blog](https://fengyao.notion.site/off-policy-rl#279721e3f6c48092bbe2fcfe0e9c6b33). | `False` | bool (set to enable) | Miles Native |
+| `--tis-clip` | Clipping threshold C for importance sampling ratios to control variance. | `2.0` | Type: float | Miles Native |
+| `--tis-clip-low` | Lower bound clipping threshold C for importance sampling ratios to control variance. | `0.0` | Type: float | Miles Native |
+| `--custom-tis-function-path` | Path to a custom TIS or MIS function. [Ref](../get_started/customization.md#10-custom-tisrs-function---custom-tis-function-path) | `None` | Type: str | Miles Native |
+| `--custom-pg-loss-reducer-function-path` | Custom reducer function for policy gradient loss. [Ref](../get_started/customization.md#11-custom-pg-loss-reducer---custom-pg-loss-reducer-function-path) | `None` | Type: str | Miles Native |
+| `--use-routing-replay` | Enable [Routing Replay](https://arxiv.org/abs/2507.18071). | `False` | bool flag (set to enable) | Miles Native |
```
guapisolo: Enable r2 for MoE: record expert routing decisions during forward and replay them during backward. Will be automatically set to True when enable `--use-rollout-routing-replay`

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--log-probs-max-tokens-per-gpu` | The maximum number of tokens per GPU for calculating log probs. This is used to calculate the log probs of the responses during rollout, and should be set to a larger value than `max_tokens_per_gpu` if you want better performance. | `None` | Type: int | Miles Native |
+| `--balance-data` | Repartition each rollout batch so each data-parallel rank gets a similar total token count via the Karmarkar-Karp method. It may be beneficial for training speed, but changes per-rank sample grouping and adds a small CPU scheduling overhead. | `False` | bool flag (set to enable) | Miles Native |
+| `--data-pad-size-multiplier` | Multiplier used to calculate the sequence padding boundary. Miles rounds sequence lengths up to a multiple of `tensor_parallel_size * data_pad_size_multiplier`. This optimization ensures that matrix dimensions are aligned with NVIDIA Tensor Core requirements, maximizing throughput and reducing VRAM fragmentation. | `128` | Type: int | Miles Native |
+| `--micro-batch-size` | Micro batch size per GPU. Ignored when `--use-dynamic-batch-size` is enabled. | `1` | Type: int | Megatron-LM (Reset by Miles) |
+
```
guapisolo, line 170: Could you add one more parameter `--seq-length`? It's a very confusing param in Megatron but not effective in miles at all. Ref https://github.com/radixark/miles/issues/574#issuecomment-3873918902

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--use-dynamic-batch-size` | Dynamically packs variable-length samples into micro-batches to maximize GPU utilization, ensuring the total token count per batch does not exceed `--max-tokens-per-gpu`. For example, with a 300-token limit, samples of lengths 100, 200, and 300 would be packed into two batches: `[100, 200]` and `[300]`. **Note:** Miles ensures that enabling this optimization does not affect the mathematical correctness of per-sample or per-token loss calculation. It is **strongly recommended** to enable this for maximum efficiency. | `False` | bool flag (set to enable) | Miles Native |
```
guapisolo: This can only be enabled when `--qkv-format` is `thd`, not work for `bshd`

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+Note that most use cases do not need to consider offload parameters, including `--offload-rollout, --no-offload-rollout, --offload-train, --no-offload-train`. They are used only on AMD GPUs and will eventually be removed.
+
+## Training Backend
+
+Arguments for configuring the training engine (Megatron or FSDP).
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--train-backend` | The backend for training. Highly suggest Megatron for numerical stability and efficiency. | `"megatron"` | `megatron`, `fsdp` | Miles Native |
+| `--qkv-format` | The QKV layout. | `"thd"` | `thd`, `bshd` | Miles Native |
```
guapisolo: Write more about this param. New models may not support `thd` but `bshd` only.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--log-probs-max-tokens-per-gpu` | The maximum number of tokens per GPU for calculating log probs. This is used to calculate the log probs of the responses during rollout, and should be set to a larger value than `max_tokens_per_gpu` if you want better performance. | `None` | Type: int | Miles Native |
+| `--balance-data` | Repartition each rollout batch so each data-parallel rank gets a similar total token count via the Karmarkar-Karp method. It may be beneficial for training speed, but changes per-rank sample grouping and adds a small CPU scheduling overhead. | `False` | bool flag (set to enable) | Miles Native |
+| `--data-pad-size-multiplier` | Multiplier used to calculate the sequence padding boundary. Miles rounds sequence lengths up to a multiple of `tensor_parallel_size * data_pad_size_multiplier`. This optimization ensures that matrix dimensions are aligned with NVIDIA Tensor Core requirements, maximizing throughput and reducing VRAM fragmentation. | `128` | Type: int | Miles Native |
+| `--micro-batch-size` | Micro batch size per GPU. Ignored when `--use-dynamic-batch-size` is enabled. | `1` | Type: int | Megatron-LM (Reset by Miles) |
```
guapisolo: Not work for `--qkv-format=thd`

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+Most SGLang server arguments can be passed through by adding the `--sglang-` prefix (some are intentionally skipped, e.g. `model_path`, `tp_size`, `port`, `nnodes`, `node_rank`). For a full list, refer to the [SGLang Server Arguments documentation](https://docs.sglang.io/advanced_features/server_arguments.html).
+
+Commonly used arguments:
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--sglang-mem-fraction-static` | Fraction of GPU memory to reserve for SGLang KV cache. | `0.9` | Type: float | SGLang |
+| `--sglang-server-concurrency` | Maximum number of concurrent requests. | `512` | Type: int | SGLang |
+| `--sglang-router-ip` | IP address of the SGLang router. | `None` | Type: str | SGLang Gateway |
+| `--sglang-router-port` | Port of the SGLang router. | `None` | Type: int | SGLang Gateway |
```
guapisolo: And miles router.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+
+## SGLang Backend Arguments
+
+Most SGLang server arguments can be passed through by adding the `--sglang-` prefix (some are intentionally skipped, e.g. `model_path`, `tp_size`, `port`, `nnodes`, `node_rank`). For a full list, refer to the [SGLang Server Arguments documentation](https://docs.sglang.io/advanced_features/server_arguments.html).
+
+Commonly used arguments:
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--sglang-mem-fraction-static` | Fraction of GPU memory to reserve for SGLang KV cache. | `0.9` | Type: float | SGLang |
```
guapisolo, line 370: Why 0.9 here ? It's too large. 0.7~0.8 is good.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--wandb-run-name` | Specific run name for WandB (FSDP backend). | `None` | Type: str | Miles Native |
+
+---
+
+## Debug and Profiling
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--check-weight-update-equal` | Verify that weight updates are equal across ranks. | `False` | bool flag (set to enable) | Miles Native |
+| `--save-debug-rollout-data` | Path to save rollout data for offline analysis. | `None` | Type: str | Miles Native |
```
guapisolo: Add `--save-debug-rollout-data`, `--load-debug-rollout-data`, `--debug-rollout-only`, `--debug-train-only` refers to `debug.md`

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--disable-grpo-std-normalization` | Disable standard deviation normalization for GRPO. From [Dr.GRPO](https://arxiv.org/pdf/2503.20783) | `False` | bool flag (set to enable) | Miles Native |
+| `--disable-rewards-normalization` | Disable the default group-wise reward normalization for GRPO, GSPO, and REINFORCE++. This effectively skips the baseline subtraction step. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-entropy` | Enable entropy calculation when calculating the logprobs from actor and reference model. This is useful for implementing custom entropy-based loss masking. | `False` | bool flag (set to enable) | Miles Native |
+| `--use-rollout-logprobs` | Use rollout logprobs for importance sampling ratios, use the logprobs from the actor model if not set. If `--get-mismatch-metrics` is set, the log probs will be recomputed by training engine, one more forward pass will be applied. | `False` | bool flag (set to enable) | Miles Native |
```
guapisolo: Plz check code logics in `loss.py`. Maybe it's should be "use rollout logprobs as the old policy logprobs for importance sampling ratios in GRPO/GSPO" ?

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--rollout-batch-size` | Number of prompts per rollout batch. The total data returned should be `rollout_batch_size` * `n_samples_per_prompt`. | Required | Type: int | Miles Native |
+| `--n-samples-per-prompt` | Number of responses to generate for each prompt, e.g., the group size of GRPO. | `1` | Type: int | Miles Native |
+| `--global-batch-size` | Total samples per optimizer step. Automatically calculated or **overridden** if `num_steps_per_rollout` is set. | `None` | Type: int | Megatron-LM (Reset by Miles) |
+| `--num-steps-per-rollout` | The number of training steps to perform using the data collected in a single rollout round. Setting this to `n` means the policy model will be updated `n` times using the same batch of rollout data. Miles ensures that `(rollout-batch-size * n-samples-per-prompt) = (global-batch-size * num-steps-per-rollout)`. If this value is not provided, you have to set `--global-batch-size` explicitly. If both are provided, `--num-steps-per-rollout`  will **override** the global batch size with `num_steps_per_rollout = (rollout_batch_size * n_samples_per_prompt) // num_steps_per_rollout`. | `None` | Type: int | Miles Native |
```
guapisolo, line 163: Could you check what will happend if multi samples are returned in the generate function (maybe the refactored one)?

miles/utils/arguments.py
```diff
@@ -124,7 +124,7 @@ def add_train_arguments(parser):
                 type=str,
                 choices=["thd", "bshd"],
                 default="thd",
-                help="The qkv layout for Megatron backend.",
+                help="The qkv layout.",
```
guapisolo, line 127: Why change here? More details needed

miles/utils/arguments.py
```diff
@@ -513,9 +516,7 @@ def add_data_arguments(parser):
                 help=(
                     "The path to the prompt data. "
                     "Currently we only support jsonl format, and each line should contains --input-key and --label-key, "
-                    "which will be used as the prompt and the label respectively. "
-                    "If you want to use a custom template, you can set --apply-chat-template to true, in that case, "
-                    "the input should be the same structure as an openai message, e.g. [{'role': 'user', 'content': 'blabla'}]. "
+                    "which will be used as the prompt and the label respectively."
```
guapisolo, line 519: Why change here?

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--log-probs-max-tokens-per-gpu` | The maximum number of tokens per GPU for calculating log probs. This is used to calculate the log probs of the responses during rollout, and should be set to a larger value than `max_tokens_per_gpu` if you want better performance. | `None` | Type: int | Miles Native |
+| `--balance-data` | Repartition each rollout batch so each data-parallel rank gets a similar total token count via the Karmarkar-Karp method. It may be beneficial for training speed, but changes per-rank sample grouping and adds a small CPU scheduling overhead. | `False` | bool flag (set to enable) | Miles Native |
+| `--data-pad-size-multiplier` | Multiplier used to calculate the sequence padding boundary. Miles rounds sequence lengths up to a multiple of `tensor_parallel_size * data_pad_size_multiplier`. This optimization ensures that matrix dimensions are aligned with NVIDIA Tensor Core requirements, maximizing throughput and reducing VRAM fragmentation. | `128` | Type: int | Miles Native |
```
guapisolo: Add a notice: Better not change this. <128 may trigger accuracy loss under thd with TP>=4.

guapisolo:
This line-level comment is hard to use :( —if a comment spans multiple lines, the line to change should be the last one.

docs/en/advanced/miles_server_args.md
yueming-yuan: you mean `--micro-batch-size`? this also works for thd; both dynamic/specified micro batch size work for thd but only specified one works for bshd

miles/utils/arguments.py
zijiexia: I think I changed here because the same parameter is also applies to FSDP backend. I wonder if this could make it less confusing.

docs/en/advanced/miles_server_args.md
zijiexia: I think the default value for sglang mem fraction static is 0.9.

miles/utils/arguments.py
yueming-yuan: Ah, yes, this was a historical issue, at the beginning I only supported for megatron, but after refactor, as general utils, it should also work for fsdp... (though I did not test fsdp+bshd)

docs/en/advanced/miles_server_args.md
yueming-yuan: may say sth like: If or not to pack all sequences with variant lengths into tokens dimension. By default use `thd` format because it will faster than `bshd` by saving padding overhead. However, for new models with novel attention architectures (e.g. sparse, attention sink), `thd` format may lack training backend support. Use `bshd` to train those models.

docs/en/advanced/miles_server_args.md
```diff
@@ -0,0 +1,491 @@
...
+| `--override-lr-scheduler` | Override the loaded LR scheduler state. | `False` | bool flag (set to enable) | Miles Native |
+| `--wandb-run-name` | Specific run name for WandB (FSDP backend). | `None` | Type: str | Miles Native |
+
+---
+
+## Debug and Profiling
+
+| Argument | Description | Default | Options | Source |
+| :--- | :--- | :--- | :--- | :--- |
+| `--check-weight-update-equal` | Verify that weight updates are equal across ranks. | `False` | bool flag (set to enable) | Miles Native |
```
yueming-yuan:
```suggestion
| `--check-weight-update-equal` | Use SGLang's weight checker to check and ensure that the loaded weight from HF checkpoint and received from Megatron are bit-wise equal. | `False` | bool flag (set to enable) | Miles Native |
```

(suggest to be more specific on this)

Ratish1:
> This line-level comment is hard to use :( —if a comment spans multiple lines, the line to change should be the last one.

Correct me if i'm wrong, but I think you can drag with the + sign specifically like this. I'm not sure if this already existed since I dont reveiw PR's often. lmk about it.
image

Ratish1:
Hey @guapisolo @yueming-yuan , thank you so much for the reviews. I have addressed all of them, let me know if you need more changes. 

Also @guapisolo , thanks for the note about “multi-sample” returns from the  generate function. I clarified this under --custom-generate-function-path: in the refactored interface a custom generate can return list[Sample], but the default rollout/training pipelines expect each prompt group to be a flat list[Sample] of length --n-samples-per-prompt (they assert len(group) == n_samples_per_prompt). so if users return multi-samples per generate call, they’ll need a compatible rollout pipeline that handles that structure. Lmk if this sounds good. Thanks again

docs/en/advanced/miles_server_args.md
guapisolo: Okay That's fine..

docs/en/advanced/miles_server_args.md
guapisolo: Sorry, Yueming is right. Ignore my words.

docs/en/advanced/miles_server_args.md
guapisolo: Just say the parameter is not work in eval. "Training only" is confusing.

guapisolo (approved):
LGTM. Only the description of `group-rm` needs change.

zhaochenyang20:
after fixing Jiajun's comment:

https://github.com/radixark/miles/pull/517#pullrequestreview-3781107286

I will merge this PR.

@Ratish1 @zijiexia then, we can further work on this. https://github.com/radixark/miles/issues/578

Great job so far! with over 130+ conversations 😂

PR #524, opened 2026-01-28
[fix] mbridge incorrectly handle all weight precision to bf16

yueming-yuan:


fzyzcjy (approved):
nit: shall we add a UT

nit: wondering whether there is a better patching method...

PR #525, opened 2026-01-28
Add debug weight sync feature

zianglih:
@humansand

This PR conducts per-tensor bit-exact check for first weight sync.

Example:
```
python scripts/run_qwen3_30b_a3b.py --no-enable-eval --hardware GB200 --num-gpus-per-node 8 --rollout-fp8 --extra-args "--debug-first-weight-sync /root/models/debug-weight-sync "
```


guapisolo:
Thanks for contribution. I am looking into this PR.

PR #526, opened 2026-01-28
Remove apply-chat-template argument from SFT_ARGS

rainj-me:
Removed the '--apply-chat-template' argument from SFT_ARGS.

- With the --apply-chat-template, the samples are applied prompt templates which lead to message["role"] failure since the message is str instead of object.
- Support Qwen3 VL 32B model sft


hhzguo (approved):
looks great

zhaochenyang20:
@gxlvera helped to check this PR is right. Thanks!

rainj-me:
@zhaochenyang20, any followup on this?

PR #531, opened 2026-01-29
fix: missing comma in runtime env JSON for qwen3-235B-A22B

harvenstar:
Summary

Fix missing comma after NCCL_NVLS_ENABLE in RUNTIME_ENV_JSON (line 164 of scripts/run-qwen3-235B-A22B.sh), which causes ray job submit to fail JSON parsing before training starts.


fzyzcjy (approved):

fzyzcjy:
looks safe

Issue #530, opened 2026-01-29
Reported Failure of Qwen3 14B


zhaochenyang20:
I highly think that Qwen3 235B should work, since we do a bunch of experiments on it. But, Qwen3 14B (I think it's a dense model), is also reported to not work.

https://huggingface.co/Qwen/Qwen3-14B

My suggestion is that, could try to install latest Miles docker and debug with it. May just need to use Megatron + SGLang.

Beichen-Ma:
I can help with investigation

WWWjiahui:
I can try to reproduce this with the latest Miles docker and testing Megatron + SGLang. I’ll report back if I find anything useful.

Warlord437:
Hi would love to solve the issue. I just discovered while applying through a linkedin post.

PR #532, opened 2026-01-30
Qwen3-14B Test Script

Beichen-Ma:
As https://github.com/radixark/miles/issues/530, try to reproduce with scripts/run-qwen3-14B.sh. The script succeeds with no failure.


Beichen-Ma:
Hi @zhaochenyang20, I put the script here for reference.

PR #535, opened 2026-01-31
Add Qwen3-14B B300 training script

harvenstar:
## Summary
- Add training script for Qwen3-14B with batch size 300 configuration

## Test plan
- [ ] Verify script runs correctly on target cluster


harvenstar:
[train-b300.log](https://github.com/user-attachments/files/24989845/train-b300.log)

PR #536, opened 2026-01-31
Add an nvfp4 PTQ convert script

zianglih:
@humansand


zianglih:
Move the full implementation to #546 .

PR #563, opened 2026-02-05
[AMD] Bump Megatron to 3714d81d and SGLang v0.5.7

zyzshishui:
TODO: 
1. Offload is supported now but need to add `--sglang-disable-custom-all-reduce`, will solve this in the future.
2. Further guarantee correctness.
3. (Exists from previous version) ppo_kl sometimes goes to non-0 value with mini-batch off, seems have a precision issue.

### DIFF
--- a/docker/Dockerfile.rocm_MI350-5
+++ b/docker/Dockerfile.rocm_MI350-5
@@ -114,7 +114,7 @@ RUN pip install "ray[data,train,tune,serve]==2.47.1"
 #########################################
 ###6. Install torch_memory_saver#########
 #########################################
-RUN pip install torch_memory_saver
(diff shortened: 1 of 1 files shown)


yushengsu-thu (approved):
This only modifies the AMD Dockerfile and patch, and will not break the codebase. It’s good to merge after testing on the AMD side.

zyzshishui:
> This only modifies the AMD Dockerfile and patch, and will not break the codebase. It’s good to merge after testing on the AMD side.

No other problem on my side (except those I already claimed in description). @gramesh-amd Gowtham do you have other issue?

gramesh-amd:
I tested some MLA model training and it works fine. Didnt find any issues

PR #575, opened 2026-02-09
Refactor and fix rollout routing replay (R3)

yueming-yuan:
## Summary
- Support rollout replay for `qkv_format=bshd` (non-packing)
- Replace env-var-driven `RoutingReplay` with a cleaner `BaseReplayManager` / `RoutingReplayManager` pattern
- Remove `os.environ` usage for replay stage control in favor of direct manager state
- Generalize `fill_routing_replay` into `_fill_replay_data` with extracted layer registration logic in `replay_utils`
- Delete old `routing_replay.py`, replaced by `replay_base.py`

## Files changed
- `miles/utils/replay_base.py` — New: replay framework (`Replay`, `BaseReplayManager`, `RoutingReplayManager`)
- `miles/utils/routing_replay.py` — Deleted: old implementation
- `miles/backends/megatron_utils/replay_utils.py` — New: MoE layer replay data distribution
- `miles/backends/megatron_utils/actor.py` — Support `bshd` format; use `all_replay_managers` instead of `RoutingReplay` + `os.environ`
- `miles/backends/megatron_utils/model.py` — Use `all_replay_managers` instead of `os.environ`
- `miles/ray/actor_group.py` — Remove old `ENABLE_ROUTING_REPLAY` env var
- `miles/rollout/generate_utils/generate_endpoint_utils.py` — Generalize `_get_rollout_topk_from_response`

## CI changed
- re-apply `dev/megatron.patch` each time before run CI.

### DIFF
--- a/miles/utils/replay_base.py
+++ b/miles/utils/replay_base.py
@@ -182,9 +182,7 @@ class BaseReplayManager:
             )
 
         if mismatch_count > mismatch_threshold:
-            raise AssertionError(
-                f"R3 mismatch tokens ({mismatch_count}) > threshold ({mismatch_threshold:.0f})"
-            )
+            raise AssertionError(f"R3 mismatch tokens ({mismatch_count}) > threshold ({mismatch_threshold:.0f})")
 
 
 class RoutingReplayManager(BaseReplayManager):


fzyzcjy (commented):
TODO(tom): I am still not very satisfied w/ my refactor b/c the time was urgent when I did it, but anyway this is not a blocker and feel free to merge it, and I will work on it later when having time

miles/utils/replay_base.py
fzyzcjy, line 1: nit: btw when I named this file I originally thought there will be several files like routing_replay and indexer_replay, but later it seems no need for heavy weight files, thus should this be renamed

EDIT: @yueming-yuan already pinged me offline me further modifying this pr, I will do it as soon as having time

guapisolo:
Megatron patch need some small fix, like: https://github.com/guapisolo/Megatron-LM/commit/0a21682109610b9bfd6aa4da1f216130c5490e49

miles/utils/replay_base.py
```diff
@@ -0,0 +1,194 @@
...
+
+                self.check_replay_result(old_topk_fn, scores, topk, top_indices, **kwargs)
+
+                return get_probs_and_top_indices(top_indices, return_probs)
+
+            elif stage == "replay_backward":
+                replay_top_indices = replay.pop_backward()
+
+                shape_sanity_check(replay_top_indices, scores, topk)
+                top_indices = replay_top_indices[..., :topk].view(scores.shape[:-1] + (topk,))
```
guapisolo: Sorry. I am a little bit confused by this line. I think the `replay_top_indices` has a shape `(num_tokens, topk)` then why do we need to change the view?

miles/utils/replay_base.py
```diff
@@ -0,0 +1,194 @@
...
+        replay = self.create_replay()
+        setattr(module, attr_name, replay)
+        manager = self
+
+        def pre_forward_hook(*args, **kwargs):
+            manager.set_current(replay)
+
+        module.register_forward_pre_hook(pre_forward_hook)
+
+    def check_replay_result(self, old_topk_fn, scores, topk, top_indices, **kwargs):
```
guapisolo: I runned `test_moonlight_16B_A3B_r3.py` with env var `MILES_CHECK_REPLAY_RESULT=1`, but the replay check failed frequently, which says `token xxx failed replay check, len(orig_set & replay_set)=2 len(replay_set)=6`. The default `thresh_check_replay_result=0.5` maybe too strict?

guapisolo: Could we make the thresh an env var? So the r3 CI can hack this value.

miles/utils/replay_base.py
yueming-yuan: oh it's related to another feature, for generalization changed to this one. maybe I can change back

miles/utils/replay_base.py
yueming-yuan: > Could we make the thresh an env var? So the r3 CI can hack this value.

Yes plan to add this in CI

.github/workflows/pr-test.yml.j2
```diff
@@ -171,6 +171,7 @@ jobs:
         shell: bash
         run: |
           cd /sgl-workspace/sglang && git fetch origin sglang-miles && git checkout FETCH_HEAD && git log --oneline -1 && pip install -e python --no-deps --break-system-packages
+          cd /root/Megatron-LM && git reset --hard HEAD && git log --oneline -1 && git apply $GITHUB_WORKSPACE/docker/patch/dev/megatron.patch && pip install -e . --no-deps --break-system-packages
```
yushengsu-thu, line 174: Discuss with Tom to see whose PR will be merged first. If Tom's merge first. Might need to change here?~

Issue #493, opened 2026-01-18
Miles Rollout Enhancement and Refactor


fzyzcjy:
Main goals trying to achieve
* For users: Try to allow users to easily understand quickly, directly use, customize, or deeply hack the rollout component
	* For quickstart users: The code may be understood more easily, after the refactor targeting simplicity and modularity. For example, to understand how to write an agent or multi-turn, the main code is ~50 lines.
	* For users wanting to train with their agents/tasks: Directly use the demos, or lightweight customize using the provided util functions
		* Besides, the tracing-based approach for blackbox agents also hope to train agents more easily by avoiding modifying blackbox, i.e. no need to extract in/out messages from agent
	* For users requiring deep customization: The provided utility functions, designed to be pluggable utils instead of a big framework, can either be picked partially or served as a reference to write new code. The test mocks and existing tests may also be used and referenced.
		* It may also be more possible for framework to add new features without requiring users to manually change their code, by abstracting functions like Sample filling logic with SGLang calls
* For maintainers: Try to enhance code quality and a testing safety net, and to more easily accommodate for potential future changes

Details
* Add 5 general examples: single-turn, multi-turn demo of general tool calling (variants: one-sample-per-trajectory, multi-sample-per-trajectory), agentic demo of general tool calling (same two variants)
* Add integration testing infra and many concrete tests: Tools e.g. mock SGLang and mock tool calling.  Testing e.g. various corner cases, various flags, for generate component and rollout component; comparison tests, etc.
* Refactor the rollout logic (`sglang_rollout.py`) (details are merged to above)
* The code still have full compatibility, i.e. existing users should not need to modify a single line of code when upgrading the framework

Code structure (tentative)
* `rollout/generate_hub/*`: The 5 examples
* `rollout/generate_utils/*`: Utility functions for generation
* `rollout/inference_rollout/*`: The rollout function minus the generate function
* `tests/fast/*`: Integration and unit tests

Steps
- [ ] Design and implement the main enhancement and refactor code, see above for details (Tom) (almost done, waiting for CI)
- [ ] Implement simple multi-turn and agent examples based on the new code, expect the code to be very short (@PopSoda2002 @GuanxingLu)
- [ ] Wait for @guapisolo's token-id-consistent OpenAI endpoint handling
- (Only at this point can we test end-to-end thoroughly, thus surely expect to have bugs before this step)
- [ ] Thorough end-to-end testing
[...]

PR #577, opened 2026-02-10
Add minimal multi turn examples based on the new rollout functions (#493)

GuanxingLu:


examples/retool_v2/run_multi_turn.sh
fzyzcjy, line 1: nit: maybe we can use python script, and maybe run_agentic.sh and run_multi_turn.sh can become one single file w/ a simple bool flag and a short if-else

fzyzcjy (commented):
the general direction LGTM, i.e. almost zero lines of code change, except for copy-pasting the sandbox code

PR #579, opened 2026-02-10
[low-precision] add asymmetric INT4 quantizer & INT4 converter; add INT4 rollout CI

yueming-yuan:
asymmetric quantization
`convert_hf_to_int4.py` and `convert_hf_to_int4_direct.py` utils

updated CI with INT4 test


scripts/run_qwen3_30b_a3b.py
yueming-yuan, line 1: Note: will do a further refactor to unify and clean up all low-precision stuff, across all the model scripts, make a unified "rollout-precision" and "train-precision" selector.

This is a intermediate version, but has been functional.

yushengsu-thu: ok

PR #581, opened 2026-02-11
[CI] Add env var enables all test using 8 gpus, and add abs tolerance to log probs diff check.

guapisolo:
Speed up CI in current CI design.

Known bugs:
- Moonlight test sometimes OOM.
- For fsdp true on policy, both test and code have bug.


guapisolo:
@yushengsu-thu  Plz review, thx

yushengsu-thu (approved):

PR #587, opened 2026-02-12
[CI] fix true on policy CI

yueming-yuan:


fzyzcjy (approved):
LGTM after discuss offline

PR #588, opened 2026-02-12
[AMD] Remove Deprecated Args: no-offload, no-gradient-accumulation-fusion

zyzshishui:
1. Remove `no-offload` since torch memory saver has now been fixed. https://github.com/fzyzcjy/torch_memory_saver/pull/69
2. Remove `no-gradient-accumulation-fusion` as now supported
3. Add `--sglang-disable-custom-all-reduce`, otherwise cause deadlock with offload enabled


guapisolo (approved):
LGTM.  Could you add description?

PR #589, opened 2026-02-12
[fix] Fix true on policy in new FSDP backend

yueming-yuan:
1. Fix true on policy (bug from refactor): add back true on policy's log probs compute path
2. Fix CI: change from 2 GPUs -> 4 GPUs to aovid OOM

### DIFF
--- a/miles/backends/training_utils/loss.py
+++ b/miles/backends/training_utils/loss.py
@@ -9,7 +9,6 @@ from miles.utils.distributed_utils import distributed_masked_whiten
 from miles.utils.misc import load_function
 from miles.utils.ppo_utils import (
     calculate_log_probs_and_entropy,
-    calculate_log_probs_and_entropy_for_true_on_policy,
     compute_approx_kl,
     compute_gspo_kl,
     compute_opsm_mask,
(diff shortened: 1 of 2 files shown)


miles/utils/ppo_utils.py
```diff
@@ -681,3 +681,36 @@ def calculate_log_probs_and_entropy(logits, tokens, tp_group, with_entropy: bool
             entropy = logits.new_zeros((0,))
 
     return log_prob, entropy
+
+
+def calculate_log_probs_and_entropy_for_true_on_policy(
```
fzyzcjy: hmm wondering whether we can be a bit more unified w/ existing code (which may or may not be possible)

miles/backends/training_utils/loss.py
```diff
@@ -155,13 +156,20 @@ def get_log_probs_and_entropy(
...
+                with_entropy=with_entropy,
+            )
+        else:
+            log_prob, entropy = calculate_log_probs_and_entropy(
+                logits_chunk,
+                tokens_chunk,
+                parallel_state.tp_group,
+                with_entropy=with_entropy,
+                chunk_size=args.log_probs_chunk_size,
+            )
```
fzyzcjy: what about

```suggestion
        log_prob, entropy = calculate_log_probs_and_entropy(
            logits_chunk,
            tokens_chunk,
            parallel_state.tp_group,
            with_entropy=with_entropy,
            chunk_size=args.log_probs_chunk_size,
        )
```

and `if true on policy` inside calculate_log_probs_and_entropy, since it looks a bit more like impl details

yushengsu-thu (approved):
Sync off-line already

PR #591, opened 2026-02-12
[feat]support rollout data transfer with mooncake

zxpdemonio:
- Add Mooncake backend for rollout data transfer
- Use zero-copy put_from/get_into for better performance
- Auto cleanup data after retrieval to free storage


stmatengss:
@zhaochenyang20 Hi, could you help trigger the CI?

miles/utils/data.py
```diff
@@ -269,9 +258,9 @@ def get_minimum_num_micro_batch_size(total_lengths, max_tokens_per_gpu):
     return len(batches)
 
 
-def process_rollout_data(args, rollout_data_ref, dp_rank, dp_size):
+def process_rollout_data(args, rollout_data_ref, dp_rank, dp_size, transfer_backend):
     assert len(rollout_data_ref) == dp_size
-    rollout_data = ray.get(rollout_data_ref[dp_rank].inner)
+    rollout_data = transfer_backend.get(rollout_data_ref[dp_rank])
```
stmatengss:
```
if transfer_backend not None:
    rollout_data = transfer_backend.get(rollout_data_ref[dp_rank])
else:
    rollout_data = ray.get(rollout_data_ref[dp_rank].inner)

**Comment by zxpdemonio** [comment]
performence:
# Single Node

| Data Size | Backend  | Put (ms)        | Get (ms)        |
|-----------|----------|-----------------|-----------------|
| 100MB     | Mooncake | 225.6 ± 15.5    | 333.4 ± 35.0    |
|           | Ray      | 222.6 ± 35.4    | **292.7 ± 18.8** |
|           | Disk     | **187.9 ± 1.4** | 341.2 ± 19.8    |
| 500MB     | Mooncake | 1046.8 ± 106.8  | 1647.3 ± 186.7  |
|           | Ray      | 1097.3 ± 176.4  | **1513.3 ± 120.0** |
|           | Disk     | **974.0 ± 55.6** | 1742.2 ± 96.3   |
| 1GB       | Mooncake | 2098.2 ± 137.3  | 3040.8 ± 277.6  |
|           | Ray      | 2481.5 ± 286.1  | **2977.3 ± 347.3** |
|           | Disk     | **2196.7 ± 131.4** | 3384.5 ± 321.5 |

# Double Node (RDMA)

| Data Size | Backend  | Put (ms)      | Get (ms)      |
|-----------|----------|---------------|---------------|
| 100MB     | Ray      | 887 ± 208     | 375 ± 44      |
|           | Mooncake | **740 ± 70**  | **329 ± 32**  |
| 500MB     | Ray      | 4316 ± 469    | 1873 ± 25     |
|           | Mooncake | **3894 ± 304**| **1654 ± 43** |
| 1GB       | Ray      | 8772 ± 655    | 3794 ± 283    |
|           | Mooncake | **8036 ± 660**| **3148 ± 20** |
| 5GB       | Ray      | 52830 ± 8817  | 19435 ± 1662  |
|           | Mooncake | **39739 ± 201** | **16085 ± 186** |

File: scripts/run-qwen3-0.6b.sh [hunk 2]
```diff
@@ -0,0 +1,157 @@
+#!/bin/bash
```
XucSh, line 1: maybe this file need to be deleted

fzyzcjy (commented):
naive thought: briefly glanced and wondering whether we need to add much complexities, or can it be (nearly) as simple as a ray object.get/put

btw a bit more context: ray RDT may also become a backend (see https://github.com/THUDM/slime/pull/2088 where zilin integrated it into slime), i.e. support vanilla ray, ray rdt, mooncake store backends. thus would be great to keep things simple when possible

zxpdemonio:
> naive thought: briefly glanced and wondering whether we need to add much complexities, or can it be (nearly) as simple as a ray object.get/put
> 
> btw a bit more context: ray RDT may also become a backend (see [THUDM/slime#2088](https://github.com/THUDM/slime/pull/2088) where zilin integrated it into slime), i.e. support vanilla ray, ray rdt, mooncake store backends. thus would be great to keep things simple when possible

@fzyzcjy   Thanks, I agree with the concern.

  The `mooncake_dataproto` / remote batch pieces were added with a longer-term direction in mind: supporting remote batch reads and future async RL training patterns, where the trainer may not always materialize the whole rollout batch eagerly and we may eventually want field-level reads, remote tensor batches, delayed materialization, or other richer data-plane semantics.

  That said, for the current PR and the immediate rollout-to-train transfer bottleneck, I agree this is more complexity than we need. We can keep the first integration much closer to Ray `put/get` and treat Mooncake only as a pluggable transport backend.
  
  I will simplify this PR to keep only the thin transfer abstraction:
  
  ```python
  handle = transfer_backend.put(rollout_data)
  rollout_data = transfer_backend.get(handle)
  ```
  Concretely, I plan to:
  - keep Ray object store as the default path;
  - keep Mooncake as an optional rollout data transfer backend;
  - preserve the existing rollout data structure and control flow;
  - keep the abstraction open so Ray RDT/NIXL or other backends can fit later;
  - remove/postpone the DataProto, RemoteBatch, field-level materialization, and hard-pin pieces from this PR.

fzyzcjy:
That looks reasonable, yes we may firstly add a minimal version to miles to keep complexity low, and then add fancier features when they are to be needed.

miles/backends/experimental/fsdp_utils/actor.py
```diff
@@ -437,9 +437,18 @@ def train(
         with inverse_timer("train_wait"), timer("train"):
             rollout_data = get_rollout_data(self.args, rollout_data_ref, witness_info=None)
             if self.args.debug_rollout_only:
+                if getattr(self.args, "transfer_backend", "ray") == "mooncake":
+                    from miles.utils.data_transfer import release_mooncake_rollout_data
```
fzyzcjy: nit: make all imports global, while make imports to mooncake optional, e.g. `try: import mooncake; except ImportError: pass`

train.py
```diff
@@ -110,6 +110,11 @@ async def save(rollout_id):
         else:
             await actor_model.train(rollout_id, rollout_data_ref)
 
+        if getattr(args, "transfer_backend", "ray") == "mooncake":
+            from miles.utils.data_transfer import cleanup_mooncake_rollout_refs
+
+            cleanup_mooncake_rollout_refs(args, rollout_data_ref)
```
fzyzcjy: wondering whether we should cleanup at here or at trainer side

miles/ray/rollout/train_data_conversion.py
```diff
@@ -124,9 +153,34 @@ def _post_process_rewards(args, samples: list[Sample] | list[list[Sample]], cust
 def split_train_data_by_dp(args, data, dp_size):
     """Split the train data by data parallel size."""
     rollout_data_list = split_train_data_by_dp_raw(args, data, dp_size=dp_size)
+    if getattr(args, "transfer_backend", "ray") == "mooncake":
+        from miles.utils.data_transfer import put_mooncake_rollout_data
+
+        return [
+            put_mooncake_rollout_data(
```
fzyzcjy: btw another way is to make a unified key value object abstraction

fzyzcjy (commented):
random optional & incomplete list of nits; I may cleanup by myself when refactoring this part as well

miles/backends/experimental/fsdp_utils/actor.py
stmatengss: Fixed in 74406ff2. Moved to global optional import with try/except ImportError. See zxpdemonio/miles#2.

train.py
stmatengss: By design: cleanup is done at the train.py side to keep the trainer-side code agnostic to the transfer backend. The trainer receives already-deserialized rollout_data and does not need to know about mooncake references.

miles/backends/experimental/fsdp_utils/actor.py
stmatengss: Fixed

train.py
stmatengss: In my mind, we should clean up immediately to simplify the workflow.

miles/backends/experimental/fsdp_utils/actor.py
zxpdemonio: Thanks. I pushed 68bdf54 to finish this cleanup across the rollout-transfer touch points, not only this FSDP call site.

The PR-added `miles.utils.data_transfer` helper imports are now module-level imports, and the real Mooncake dependency remains optional inside `miles/utils/data_transfer.py` (plus an optional `FieldSchema` import in `train_data_conversion.py`). This avoids function-local imports while keeping Miles importable without Mooncake installed.

train.py
zxpdemonio: I kept the remote-object cleanup at the train-loop side and pushed 68bdf54 to make the failure path safer.

The train loop owns the rollout refs and knows when all consumers (actor and optional critic) have finished. The trainer actor only owns the deserialized local result and releases local Mooncake-backed buffers after use. In the new code, `cleanup_mooncake_rollout_refs` runs in `finally`, but only after the critic task is awaited, so we do not remove the remote rollout object while another consumer may still need it.

stmatengss:
@fzyzcjy PTAL. Thanks~

fzyzcjy (commented):
quick questions about mooncake apis

* it would be great to support num replicas > 1 for the objects to have better fault tolerance (I guess maybe need to call the mooncake apis w/ some args etc)
* it seems we are calling a legacy mooncake api, shall we call a non legacy one

zxpdemonio:
> quick questions about mooncake apis
> 
> * it would be great to support num replicas > 1 for the objects to have better fault tolerance (I guess maybe need to call the mooncake apis w/ some args etc)
> * it seems we are calling a legacy mooncake api, shall we call a non legacy one

@fzyzcjy Thanks for the questions. I think there are two separate API decisions here.

For replica support, Mooncake’s lower-level Store APIs already model this as a per-write option via ReplicateConfig. So one possible Miles-side design is to add a rollout replica-count argument, e.g. rollout_num_replicas, and pass that through to the Mooncake put call as the write-time replica config. This keeps the user-facing config simple while matching Mooncake’s existing per-object replication model.

For the legacy_dict API name, I agree it is confusing. In our use case the payload is still a normal flat dict, not a DataProto object, and “legacy” can make it sound like we are calling an old Mooncake API. One possible cleanup is to expose a unified put/get API and make the payload format an argument, for example format="dict". Then Miles can call put(..., format="dict") / get(..., format="dict"), and the DataProto-vs-dict distinction stays explicit without leaking into method names or implying this is a legacy API.

If this sounds reasonable, I can first update the Mooncake-side structured API to support the write-time replica config and the unified put/get format option, then update Miles to pass rollout_num_replicas through.

fzyzcjy:
yes both looks like a reasonable vanilla approach (and btw as long as "legacy" does not mean "mooncake is going to throw away" then no need to be worry about it, though it is best to avoid the name)

fzyzcjy:
(FYI I will do some refactors)

zxpdemonio:
> A note on how `mooncake_master` is started, after checking this PR end to end:
> 
> **This PR never starts `mooncake_master` anywhere — the master is assumed to be external infrastructure.** Evidence:
> 
> * All 7 commits in this PR contain no `mooncake_master` launch logic (no subprocess spawn, no startup code); the new `MooncakeBundleTransfer` code only calls `store.setup(...)` as a client connecting to an already-running master.
> * The master address comes purely from configuration: `--mooncake-store-init-kwargs '{"master_server_address": "ip:port", ...}'`, or the `MOONCAKE_MASTER` environment variable read by the mooncake library itself.
> * A full code search of the source fork for `mooncake_master` returns 0 hits, and the PR description/comments never mention starting one.
> 
> Accordingly, on this branch the framework still does not manage the master lifecycle. The CI test starts `mooncake_master` itself as an environment-preparation step (with a `TODO` noting that ideally miles itself, e.g. the RolloutManager, would launch the master when none is configured).

@fzyzcjy Thanks for checking this end to end. I think the right contract here is that Miles should treat the Mooncake control plane as external infrastructure, and this PR should stay focused on the client-side rollout object-store backend.

One nuance is that the configured `master_server_address` is effectively a Mooncake master entry, not necessarily a single direct `ip:port`. In direct mode it can point to one `mooncake_master`, while in HA mode Mooncake can use entries such as `etcd://...`, `redis://...`, or `k8s://...`; the Mooncake client then resolves the current leader internally.

Based on that, my preference for this PR is:

- Miles should not auto-launch `mooncake_master` in the rollout transport path.
- Miles should pass through the Mooncake master entry from config/env so both direct mode and HA mode work.
- The CI test can start a direct-mode `mooncake_master` as test infrastructure, but that should not imply Miles owns the Mooncake control-plane lifecycle.
- We should clarify the TODO/comments/docs so this contract is explicit.

This also matches the pattern used by SGLang's Mooncake integration: the framework loads Mooncake config and calls `store.setup(...)`, while the master/HA backend lifecycle is managed by deployment or test infrastructure.

Does this sound like the direction you prefer? If yes, I can update the comments/docs accordingly and keep the runtime code focused on the client-side object-store integration. cc @stmatengss

fzyzcjy:
oh that comment was sent by agent and I deleted it after a minute. I guess it would be great for users to launch miles in one command and enjoy mooncake (just like how they enjoy ray object store), but I may do the (default) launching later during my refactor, and no worries about it.

fzyzcjy:
did a vanilla refactor and passed a bit of test. agent report:

Ran both e2e short tests for this branch (at 66eaba9b5) on 4x H200 with the `radixark/miles:dev` (CUDA 13) image, `MILES_TEST_FEW_GPU=1`.

**Mooncake setup.** The released `mooncake-transfer-engine` 0.3.11.post1 is too old for this branch: it lacks the `structured_object_store` bundle API, and the `global_segment_size=0` "pure client" setup semantics only landed on mooncake main on 2026-06-02. I used the v0.3.12 wheel that Mooncake CI built today from `release/v0.3.12` (c7ae97fd, CI run 30010874753, artifact `mooncake-wheel-cu130-ubuntu-py312`), plus `structured_object_store.py` from mooncake main (a1ca71f4) because even the 0.3.12 wheel does not yet package the new structured object store API:

```bash
pip install --force-reinstall --no-deps mooncake_transfer_engine_cuda13-0.3.12-cp312-cp312-manylinux_2_35_x86_64.whl
cp Mooncake/mooncake-wheel/mooncake/structured_object_store.py \
   /usr/local/lib/python3.12/dist-packages/mooncake/structured_object_store.py
```

**1. Mooncake object store e2e** (the test starts `mooncake_master` itself):

```bash
cd miles && PYTHONPATH=$PWD MILES_TEST_FEW_GPU=1 python tests/e2e/short/test_qwen2.5_0.5B_gsm8k_async_short.py
```

Result: `Job 'raysubmit_8KZnENwXYbEBQxr4' succeeded` — full run green: 3 rollouts with mooncake put/get/remove across rollout manager, train actors, and driver, plus weight updates, eval, and fault-tolerance health checks.

**2. Non-mooncake regression** (default `--object-store-backend ray`):

```bash
cd miles && PYTHONPATH=$PWD MILES_TEST_FEW_GPU=1 python tests/e2e/short/test_qwen2.5_0.5B_gsm8k_short.py
```

Result: `Job 'raysubmit_pbywk6EB6VxE22Qv' succeeded`.

guapisolo (approved):
LGTM. Need to handle mooncake install issue.

fzyzcjy (approved):

guapisolo:
Let's wait for ci and then we can merge it.

PR #573, opened 2026-02-08
feature: docker build and release workflow

dougyster:
Created build.py based on justfile to get the correct tags and build args for each build type and added a dry-run flag to preview args before building. 

Added dockerfile path into the workflow to specify which dockerfile to build.

Added auto image building:
- Push to main touching Dockerfile.dev → builds dev immediately
- Every 12h cron → polls sglang (sglang-miles branch) and Megatron-LM (main branch) for new commits, only builds dev if either changed                                                                                                   
- Manual dispatch → builds user-selected variant with user-selected Dockerfile
                                                                                                                                                                                                                                           
For automatically built images:                           
  - Tags: radixark/miles:dev (rolling) + radixark/miles:dev-{YYYYMMDDHHMM} (preserved)
  - Overrides MEGATRON_COMMIT=main (latest, not pinned)
  - Prunes Docker Hub to keep only the 20 most recent dev-* tags

PR #593, opened 2026-02-13
[CI] Reorg test file and fix moonlight oom

guapisolo:
TODO: Replace moonlight-16B-A3B test by GLM-4.7-flash after new sglang docker bump and  https://github.com/radixark/miles/pull/566 merge. It's a 30b model with mla and mtp.

Upd: new dev docker has been bumped. wip https://github.com/radixark/miles/pull/566

### DIFF
--- a/tests/e2e/megatron/test_moonlight_16B_A3B.py
+++ b/tests/e2e/megatron/test_moonlight_16B_A3B.py
@@ -80,7 +80,9 @@ def execute():
     )
 
     sglang_args = (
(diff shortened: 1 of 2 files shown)


yushengsu-thu (approved):
It looks good, but I have some advice.

Comments on deleting CI under image tags:

For the CI under image tags, I planned to run all tests under those images to validate the dev Docker, even though this may leave duplicate CI in different tags.

If we adopt the PR above:

We may need to update/add tags in the Docker release workflow (it seems this has not been added yet - https://github.com/radixark/miles/pull/573).
I think it’s better to add/update these tags every time we release a new Docker image.

Issue #578, opened 2026-02-10
Dynamic sync parameter docs with `--help` and arguments docs string


zhaochenyang20:
Thanks to @Ratish1 and @zijiexia 

https://github.com/radixark/miles/pull/517

This doc is so far so good and after fixing Jiajun's review @guapisolo , we can merge it by @fzyzcjy 

My further plan is to make a dynamic sync with the `argument.py`, especially with the `python3 train.py --help`.

We can have a script like `python sync_param_docs.py`, which dynamically scans all the params and syncs them in `docs/en/advanced/miles_server_args.md`. I do not have a clear plan how to do this. But, the first thing I want to do is that:

1. Update our current  params' descriptions in `python3 train.py`;
2. Later, all the changes in `python3 train.py` should be synced to the docs by `python sync_param_docs.py`.

We shall first make a plan on how to do this. After we agree on the plan, let's make the code. Welcome to use AI to help on this issue, but vibe coding with caution. Do not be blind to vibed coded codes.

Ratish1:
Hey @zhaochenyang20 , here is the plan below we could follow:
  1. Scope and source of truth
  - Source of truth is Miles-native args explicitly defined in `miles/utils/arguments.py` (plus approved scoped sections).
  - We will not try to fully mirror all `python train.py --help` output.
  - We will not auto-document all SGLang passthrough flags; only supported ones in Miles docs.
 
  2. Script-first (before the CI)
  - Add `sync_param_docs.py` with:
    - `--write`: sync docs from supported argument definitions
    - `--check`: validation-only; fails when either side drifts (args -> docs or docs -> args).
  - Validation will also fail if docs include scoped args not present in native args definitions.
 
  3. “Bi-directional” behavior
  - Implement bi-directional consistency checks, not bi-directional file mutation:
    - args -> docs sync (--write)
    - docs -> args and args -> docs validation (--check)
  - We will not auto-edit `arguments.py` from docs.
 
  4. CI (after script is stable)
  - Add a CPU-only CI job running `python sync_param_docs.py --check`.

Diagram:

```mermaid
flowchart TD
    A["miles/utils/arguments.py<br/>(explicit Miles-native args)"] --> B["Argument Reader<br/>- read parser details<br/>- standardize defaults/types/choices/help text"]
    B --> C["Scope Filter<br/>- include supported sections only<br/>- skip non-native/unsupported args"]

    C --> D["Write Path"]
    C --> E["Check Path"]

[...]

PR #594, opened 2026-02-13
[docs]: sync args with doc

Ratish1:
This is part 1 of #578 . Manual sync of current doc's description of parameters to the argument.py.


miles/utils/arguments.py
```diff
@@ -38,49 +38,56 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
...
             )
             parser.add_argument(
                 "--colocate",
                 action="store_true",
                 default=False,
                 help=(
-                    "Whether to colocate the inference engines and the actor. "
-                    "Turning this on will also set --offload to true."
+                    "Deploy training and rollout on the same GPUs. This mode automatically enables `--offload-train` and `--offload-rollout` to facilitate weight-swapping between the training actor and inference engine. **Note:** The offload parameters are currently only used for AMD GPUs and will be removed soon. **Memory Tip:** When colocating, it is highly recommended to set `--sglang-mem-fraction-static` to **0.8** (especially on **NVIDIA Blackwell B200/B300** GPUs). This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. On GB200/GB300, values up to 0.75 are safer for long-running jobs to prevent potential OOMs. #TODO: Verify optimal fraction for Blackwell in production"
                 ),
```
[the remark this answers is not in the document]
zijiexia: We should move TODOs out of the help text

zijiexia:
Hi @Ratish1 , I think for the sake of better code visualization, we should not put the helper text as a single line, I think a better way to do is instead of using a single string, stacking a bunch of substrings. e.g., 
use
```
help=(
    "Applies to `megatron` training backend only. "
    "Disables the system that backups model weights (Actor, Ref, Old Actor) to CPU RAM. "
    "Disabling saves significant host memory but prevents features that rely on weight-swapping, such as computing KL-divergence against a reference model. "
    "Note: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper."
),
```
instead of
```
help="Applies to `megatron` training backend only. Disables the system that backs up model weights (Actor, Ref, Old Actor) to CPU RAM. Disabling saves significant host memory but prevents features that rely on weight-swapping, such as computing the KL-divergence against a reference model. **Note**: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper."
```
I believe there're a lot of linting tools can help you automate this. Please let me know what you think. Thanks.

Also, for those single lined help text, why we need bracket around the text? I think we should remove those to make the coding style aligned.

miles/utils/arguments.py
```diff
@@ -190,22 +205,14 @@ def add_rollout_arguments(parser):
...
                 type=str,
                 default=None,
-                help=(
-                    "The huggingface checkpoint of the trained model. "
-                    "This is used to initialize sglang and also provide the tokenizer. "
-                    "Note that, we will always update the parameters in sglang with that of megatron before training, "
-                    "so you only need to provide a huggingface checkpoint that has the same architecture as the model you want to train. "
-                    "It doesn't necessary need to contain the most up-to-date parameters."
-                ),
+                help=("Path to the HuggingFace checkpoint used to initialize SGLang and provide the tokenizer."),
```
zijiexia: e.g.,
```suggestion
                help="Path to the HuggingFace checkpoint used to initialize SGLang and provide the tokenizer.",
```

miles/utils/arguments.py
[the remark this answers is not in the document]
Ratish1: Yes, I thought of this too, but then the script would just keep failing since docs and arguments are out of sync.

cc: @zhaochenyang20

Ratish1:
> Hi @Ratish1 , I think for the sake of better code visualization, we should not put the helper text as a single line, I think a better way to do is instead of using a single string, stacking a bunch of substrings. e.g., use
> 
> ```
> help=(
>     "Applies to `megatron` training backend only. "
>     "Disables the system that backups model weights (Actor, Ref, Old Actor) to CPU RAM. "
>     "Disabling saves significant host memory but prevents features that rely on weight-swapping, such as computing KL-divergence against a reference model. "
>     "Note: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper."
> ),
> ```
> 
> instead of
> 
> ```
> help="Applies to `megatron` training backend only. Disables the system that backs up model weights (Actor, Ref, Old Actor) to CPU RAM. Disabling saves significant host memory but prevents features that rely on weight-swapping, such as computing the KL-divergence against a reference model. **Note**: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper."
> ```
> 
> I believe there're a lot of linting tools can help you automate this. Please let me know what you think. Thanks.
> 
> Also, for those single lined help text, why we need bracket around the text? I think we should remove those to make the coding style aligned.

Yes I will follow this, thanks @zijiexia . I assumed the pre commit would fix this but it didnt. I will look into it and fix it.

miles/utils/arguments.py
```diff
@@ -1068,149 +1278,154 @@ def add_debug_arguments(parser):
                 "--save-debug-rollout-data",
                 type=str,
                 default=None,
-                help=(
-                    "Save the rollout data to this path for debugging. "
-                    "The file will be saved to `save_debug_rollout_data.format(rollout_id)`."
-                ),
+                help=("Path to save rollout data for offline analysis. " "[Ref](../developer_guide/debug.md)"),
```
zijiexia: Here we split on periods even though it's a short string

miles/utils/arguments.py
```diff
@@ -1068,149 +1278,154 @@ def add_debug_arguments(parser):
...
                 type=str,
                 choices=["torch", "memray"],
                 default="torch",
+                help="Selection of the memory recording backend.",
+            )
+            parser.add_argument(
+                "--check-weight-update-equal",
+                action="store_true",
+                help=(
+                    "Use SGLang's weight checker to check and ensure that the loaded weight from HF checkpoint and received from Megatron "
```
zijiexia: Here we split on random place in the sentence. I was wondering what is the logic behind the splitting? Will this works with your sync script?

miles/utils/arguments.py
```diff
@@ -967,10 +1181,8 @@ def add_router_arguments(parser):
                 action="store_true",
                 default=False,
                 help=(
-                    "This is an experimental feature, and only supports for text model."
-                    "Whether to enable token input for chat completions. If set, we will calculate "
-                    "the input_ids for the prompt part inside miles and add it to the request body."
-                    "This is reserved for cross turn token in under OAI format."
+                    "This is an experimental feature, and only supports for text model.Whether to enable token input for chat completions. "
+                    "If set, we will calculate the input_ids for the prompt part inside miles and add it to the request body.This is reserved for cross turn token in under OAI format."
```
zijiexia: This arg is not in the miles_server_args, should we add it to the docs?

zijiexia: Lets revert this change as this arg won't be sync to the docs anyway

miles/utils/arguments.py
```diff
@@ -979,64 +1191,60 @@ def add_router_arguments(parser):
...
             )
             parser.add_argument(
                 "--log-reward-category",
                 type=str,
                 default=None,
                 help=(
-                    "Log statistics of the category of reward, such as why the reward function considers it as failed. "
-                    "Specify the key in the reward dict using this argument.",
+                    "Log reward-category statistics (e.g., why the reward function marked a failure). "
+                    "Use this argument to specify the key in the reward dict.",
```
zijiexia:
```suggestion
                    "Use this argument to specify the key in the reward dict."
```

miles/utils/arguments.py
```diff
@@ -38,76 +38,102 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
...
                 action="store_true",
                 default=False,
                 help=(
-                    "Whether to colocate the inference engines and the actor. "
-                    "Turning this on will also set --offload to true."
+                    "Deploy training and rollout on the same GPUs. "
+                    "This mode automatically enables `--offload-train` and `--offload-rollout` to facilitate weight-swapping between the training actor and inference engine. "
+                    "**Note:** The offload parameters are currently only used for AMD GPUs and will be removed soon. "
+                    "**Memory Tip:** When colocating, it is highly recommended to set `--sglang-mem-fraction-static` to **0.8** (especially on **NVIDIA Blackwell B200/B300** GPUs). "
+                    "This leaves sufficient VRAM (~20%) for Megatron to initialize its structures before the first weight offload to CPU occurs. "
```
zijiexia:
```suggestion
                    "This leaves sufficient VRAM (~20%%) for Megatron to initialize its structures before the first weight offload to CPU occurs. "
```

miles/utils/arguments.py
```diff
@@ -38,76 +38,102 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
...
-                help=(
-                    "Whether to offload the rollout generator to CPU during training. "
-                    "This will always be true when --colocate is set."
-                ),
+                help="Whether to offload the rollout generator to CPU during training. This will always be true when --colocate is set.",
             )
 
-            reset_arg(parser, "--distributed-backend", type=str, default="nccl")
-            reset_arg(parser, "--distributed-timeout-minutes", type=int, default=10)
+            reset_arg(
```
zijiexia: For these args updated by `reset_arg`, does it make sense to add the help text? The help will be ignored anyway.

zijiexia (changes requested):
Can you also run `python3 train.py --help` to see if it works as expected?

Ratish1:
> Can you also run `python3 train.py --help` to see if it works as expected?

Yes it works now. Thanks @zijiexia

Ratish1:
Hey @yueming-yuan , could you take a look at this if you are not too busy?. It is a pretty straightforward PR. Thanks.

PR #597, opened 2026-02-13
[AMD] Unify run-qwen3-4B.sh to support both AMD and NVIDIA GPUs

lizamd:
Auto-detect GPU vendor (/dev/kfd or torch.version.hip for AMD, nvidia-smi for NVIDIA) and conditionally apply platform-specific settings:
- AMD: HIP_VISIBLE_DEVICES, RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES, --no-gradient-accumulation-fusion, --no-offload-train/rollout
- NVIDIA: NVLink detection, NCCL_NVLS_ENABLE
- Both: dynamic Megatron-LM path detection, configurable MODEL_DIR/DATA_DIR

This eliminates the need for a separate run-qwen3-4B-amd.sh script.

### DIFF
--- a/scripts/run-qwen3-4B.sh
+++ b/scripts/run-qwen3-4B.sh
@@ -36,9 +36,13 @@ if [ "$GPU_VENDOR" = "amd" ]; then
     HAS_NVLINK=0
 else
     NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
-    if [ "$NVLINK_COUNT" -gt 0 ]; then HAS_NVLINK=1; else HAS_NVLINK=0; fi
+    if [ "$NVLINK_COUNT" -gt 0 ]; then
+        HAS_NVLINK=1
+    else
+        HAS_NVLINK=0
+    fi
     echo "HAS_NVLINK: $HAS_NVLINK (detected $NVLINK_COUNT NVLink references)"
-    NUM_GPUS=8
+    NUM_GPUS=$(nvidia-smi -L 2>/dev/null | wc -l)
 fi
 
 # will prevent ray from buffering stdout/stderr
@@ -140,15 +144,6 @@ MISC_ARGS=(
    --attention-backend flash
 )
 
-# ==================== Platform-Specific Args ====================
-PLATFORM_TRAIN_ARGS=()
-if [ "$GPU_VENDOR" = "amd" ]; then
-    # Apex not available on ROCm
-    MISC_ARGS+=(--no-gradient-accumulation-fusion)
-    # Disable offloading (torch_memory_saver may not support ROCm; MI300X has 192GB HBM)
-    PLATFORM_TRAIN_ARGS+=(--no-offload-train --no-offload-rollout)
-fi
-
 # ==================== Launch Ray ====================
 export MASTER_ADDR=${MASTER_ADDR:-"127.0.0.1"}
 ray start --head --node-ip-address ${MASTER_ADDR} --num-gpus ${NUM_GPUS} --disable-usage-stats --dashboard-host=0.0.0.0 --dashboard-port=8265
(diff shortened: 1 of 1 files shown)


scripts/run-qwen3-4B.sh
```diff
@@ -10,33 +10,54 @@ sleep 3
...
+
+# ==================== Configurable Paths ====================
+MODEL_DIR="${MODEL_DIR:-/root}"
+DATA_DIR="${DATA_DIR:-/root}"
+export MODEL_DIR DATA_DIR
+
+# ==================== Platform-Specific Setup ====================
+if [ "$GPU_VENDOR" = "amd" ]; then
+    export RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES=${RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES:-"1"}
+    export HIP_VISIBLE_DEVICES=${HIP_VISIBLE_DEVICES:-"0,1,2,3,4,5,6,7"}
```
yushengsu-thu, line 34: @lizamd, if my memory is right, `CUDA_VISIBLE_DEVICES` also works on the AMD side (.... if torch version and the ray version are new enough). Could we use the same ENV name in case the script is too long?

scripts/run-qwen3-4B.sh
```diff
@@ -10,33 +10,54 @@ sleep 3
...
+echo "Detected GPU vendor: ${GPU_VENDOR}"
+
+# ==================== Configurable Paths ====================
+MODEL_DIR="${MODEL_DIR:-/root}"
+DATA_DIR="${DATA_DIR:-/root}"
+export MODEL_DIR DATA_DIR
+
+# ==================== Platform-Specific Setup ====================
+if [ "$GPU_VENDOR" = "amd" ]; then
+    export RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES=${RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES:-"1"}
```
yushengsu-thu, line 33: @lizamd 
Here, we might not need `RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES` and `HIP_VISIBLE_DEVICES`; using `CUDA_VISIBLE_DEVICES` might be enough. We fixed this issue in Ray last year, and `RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES` is just an experimental env var. It should be removed by Ray. You can confirm these env variables with @vickytsang Vicky Tsang within AMD. 

cc. @zyzshishui @gramesh-amd 
Hello, if the Ray version in the AMD Docker image is too old, we can upgrade it to the same version used in the RadixArk NVIDIA Docker image.

scripts/run-qwen3-4B.sh
```diff
@@ -119,14 +140,26 @@ MISC_ARGS=(
    --attention-backend flash
 )
 
-# launch the master node of ray in container
+# ==================== Platform-Specific Args ====================
+PLATFORM_TRAIN_ARGS=()
+if [ "$GPU_VENDOR" = "amd" ]; then
+    # Apex not available on ROCm
+    MISC_ARGS+=(--no-gradient-accumulation-fusion)
```
yushengsu-thu: We do not need this: MISC_ARGS+=(--no-gradient-accumulation-fusion)
The megatron and related stuff within the AMD docker img is already supported. 

cc. @zyzshishui to confirm this.

scripts/run-qwen3-4B.sh
```diff
@@ -119,14 +140,26 @@ MISC_ARGS=(
...
+    # Disable offloading (torch_memory_saver may not support ROCm; MI300X has 192GB HBM)
+    PLATFORM_TRAIN_ARGS+=(--no-offload-train --no-offload-rollout)
```
yushengsu-thu: We do not need this: `PLATFORM_TRAIN_ARGS+=(--no-offload-train --no-offload-rollout)` 
The torch_memory_saver has already resolved this issue,e and the AMD docker img is already supported. 

cc. @zyzshishui to confirm this.

scripts/run-qwen3-4B.sh
zyzshishui: yes, no need

scripts/run-qwen3-4B.sh
zyzshishui: correct, can be removed

scripts/run-qwen3-4B.sh
JessicaJiang-123: From my side, it might be better to keep `HIP_VISIBLE_DEVICES` for AMD compatibility.

PR #601, opened 2026-02-14
Make large model training scripts one-click

fzyzcjy:


fzyzcjy:
definitely, it is in my long long todo list

PR #619, opened 2026-02-19
[fix] bypass r3 for mtp layer.

guapisolo:
Related megatron commit: https://github.com/radixark/Megatron-LM/pull/10

R3 cannot be applied to mtp layer, since mtp will not forward verify bonus tokens during rollouts. So bypass r3 replay for mtp head.

This impl is a little bit hard code, but maybe the best impl in current design. Mainly because we cannot get the module name before the layer is fully registered, so the Replay class cannot record anything about the layer name. Also, the original MoE routing function does not included module info, so we cannot directly skip MoE mtp replay during runtime... So add the element `is_mtp` as instead.

### DIFF
--- a/miles/backends/megatron_utils/model_provider.py
+++ b/miles/backends/megatron_utils/model_provider.py
@@ -192,14 +192,15 @@ def get_model_provider_func(
                 mtp_kwargs["vp_stage"] = vp_stage
 
             # hard code here to skip r3 registration for mtp layers
-            if args.use_rollout_routing_replay:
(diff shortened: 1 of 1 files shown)


yueming-yuan:
could you also create a patch-equivalent PR in megatron fork `miles-main` branch? thanks! https://github.com/radixark/Megatron-LM/tree/miles-main

guapisolo:
> could you also create a patch-equivalent PR in megatron fork `miles-main` branch? thanks! https://github.com/radixark/Megatron-LM/tree/miles-main

done

PR #638, opened 2026-02-25
[feat] GB300 (sm103a / cu13 arm64) support

kaixih:
Three changes for running Miles on GB300 / Blackwell hardware:

1. docker/Dockerfile.dev — cu13 build improvements:
   - Upgrade transformer_engine to 2.12.0 for cu13 (was 2.10.0). TE 2.12 adds THD layout + head_dim=256 support in the unfused attention backend, required for Qwen3-Next-80B-A3B on sm103a (cuDNN and FA2/FA3 don't support this config).
   - Apply docker/patch/cu13/patch_fla_blackwell.py at build time to fix three Triton 3.5.1 / fla 0.4.0 bugs on Blackwell (see patch docstring for details).
   - Set TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas in /root/.bashrc since Triton's bundled ptxas does not support sm_103a.

2. docker/patch/cu13/patch_fla_blackwell.py — new file: Patches installed fla 0.4.0 + Triton 3.5.1 packages for Blackwell (sm103a):
   - Bug 1: fla wy_fast.py backward kernel: Triton 3.5.1 lowers `+= tl.dot()` to a three-operand SSA form that TritonGPUHoistTMEMAlloc rejects with a dominance violation. Fix: safe_dot() wraps the result in inline ASM to prevent the fusion (fla PR #687, Triton issue #8695).
   - Bug 2/3: gluon_ir not compiled in the arm64+cu130 Triton build. Affects any @triton.autotune kernel on a cache miss. Fix: try/except guards in triton/runtime/jit.py and triton/experimental/gluon/language/_semantic.py.

~~3. update_weight_from_distributed.py / update_weight_from_tensor.py: Call post_process_weights() unconditionally instead of only for compressed-tensors / mxfp8. The function is already a no-op when no module implements restore_weights_before_loading / process_weights_after_loading, so this is safe for all existing models. Required for BF16 MoE models where the SGLang flashinfer_trtllm backend reshapes w13_weight after load and needs the restore→load→re-apply cycle on update_weights.~~
(**3 is removed from this PR, it will be added back when fixing the trtllm_moe backend. Currently we use triton backend.**)

For repro and more details: https://gist.github.com/kaixih/7ca3ff5ca41e7e78b6cf6c6bc56d4dc1

### DIFF
--- a/.github/workflows/pr-test.yml.j2
+++ b/.github/workflows/pr-test.yml.j2
@@ -27,6 +27,9 @@
     {'test_file': 'e2e/short/test_qwen2.5_0.5B_gsm8k_async_short.py', 'num_gpus': 8},
     {'test_file': 'e2e/short/test_qwen2.5_0.5B_gsm8k_short.py', 'num_gpus': 8},
     {'test_file': 'e2e/short/test_qwen3_0.6B_fsdp_colocated_2xGPU.py', 'num_gpus': 8},
+    {'test_file': 'e2e/sglang_config/test_sglang_config.py', 'num_gpus': 8},
+    {'test_file': 'e2e/sglang_config/test_sglang_config_mixed_offload.py', 'num_gpus': 8},
+    {'test_file': 'e2e/sglang_config/test_sglang_config_mixed_offload_ft.py', 'num_gpus': 8},
 ] %>
 
 <% set precision_tests = [
@@ -60,6 +63,7 @@
       'test_executor': 'pytest',
       'tests': [
         {'test_file': 'fast', 'num_gpus': 0},
+        {'test_file': 'utils/test_sglang_config.py', 'num_gpus': 0},
       ],
     },
     'unit-test': {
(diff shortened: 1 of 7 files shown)


yueming-yuan:
Could you please briefly explain the scale of this fix, e.g., which kind of model will fail without this PR (is it only for Qwen3-Next-80B-A3B?) and what will happen without the FLA/triton patch? thx!

kaixih:
yes, the TE upgrade would be a must for gb300 workloads that uses attention. the triton patch is mainly for the linear attention. It would be qwen3next and the new qwen3.5.

kaixih:
Any updates @yushengsu-thu ?

yushengsu-thu:
@kaixih 
Could you 
1. create a docker file name as `Dockerfile_GB300` 
2. use as [Dockerfile](https://github.com/radixark/miles/blob/main/docker/Dockerfile) the template and add/modify the GB300 parts. 
After finish above, we will merge this PR - `Dockerfile_GB300`.

kaixih:
> @kaixih Could you
> 
> 1. create a docker file name as `Dockerfile_GB300`
> 2. use as [Dockerfile](https://github.com/radixark/miles/blob/main/docker/Dockerfile) the template and add/modify the GB300 parts.
>    After finish above, we will merge this PR - `Dockerfile_GB300`.

Done. PTAL. @yushengsu-thu

PR #645, opened 2026-02-26
[fix] fix R3 padding handling

yueming-yuan:
The previous version used incorrect padding handling to skip sequence w/ paddings, rather than the padding tokens, which causes R3 to not be fully functional in training. Fix by using a token padding mask and fill with the range routing indices value.

Thanks @zianglih for pointing out the issue

### DIFF
--- a/miles/utils/replay_base.py
+++ b/miles/utils/replay_base.py
@@ -91,12 +91,13 @@ class BaseReplayManager:
 
             if self.enable_check_replay_result:
                 self.check_replay_result(old_topk_fn, scores, topk, top_indices, *args, **kwargs)
-            
+
             padding_mask = top_indices == -1
             if padding_mask.any():
-                top_indices[padding_mask] = torch.arange(
-                    padding_mask.sum(), device=top_indices.device, dtype=top_indices.dtype
-                ) % scores.shape[1]
+                top_indices[padding_mask] = (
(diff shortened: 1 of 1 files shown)


fzyzcjy (approved):

miles/utils/replay_base.py
```diff
@@ -93,11 +92,15 @@ def _get_replay_result(top_indices, scores, topk, *args, **kwargs):
             if self.enable_check_replay_result:
                 self.check_replay_result(old_topk_fn, scores, topk, top_indices, *args, **kwargs)
 
+            padding_mask = top_indices == -1
+            if padding_mask.any():
+                top_indices[padding_mask] = (
+                    torch.arange(padding_mask.sum(), device=top_indices.device, dtype=top_indices.dtype)
+                    % scores.shape[1]
+                )
```
fzyzcjy, line 100: qq shall we return -1 instead of randomly picked value

yueming-yuan: the top_indices of padding tokens will be used in moe calculation (though meaningless), so may need to fill with some true expert index to avoid ima (and try to make it balance)

fzyzcjy: I see, then LGTM. nit: shall we assert such scenario never happens: for a specific token, some topk are valid and some are -1. (I think it never happens for today's model, thus is of low priority)

PR #648, opened 2026-02-27
Fix the missing cu_seqlens in GDN

nono-Sang:
`cu_seqlens` is necessary for varlen sequence


guapisolo:
I'm already testing this PR with my qwen-next mtp support. If there is no problem, we will merge it.

nono-Sang:
> I'm already testing this PR with my qwen-next mtp support. If there is no problem, we will merge it.

@guapisolo No problem, thanks for the review.

PR #649, opened 2026-02-28
[CI] improve ckpt save/load check with hash checks & add moe model test

guapisolo:
as the title.


.github/workflows/pr-test.yml
```diff
@@ -1045,11 +1047,54 @@ jobs:
           rm -rf /tmp/ray/* 2>/dev/null || true
           sleep 3
 
+      - name: Resolve dependency refs
```
guapisolo, line 1071: I dont know why this is changed..

yueming-yuan: it's auto-generated, don't need to worry

guapisolo:
@yushengsu-thu Could you review this ? Thanks

PR #655, opened 2026-03-02
[model] Support qwen3 next mtp training

guapisolo:


scripts/models/qwen3-next-80B-A3B.sh
```diff
@@ -25,7 +25,6 @@ MODEL_ARGS=(
    --num-layers 48
    --hidden-size 2048
    --ffn-hidden-size 5120
-   --use-gated-attention
```
guapisolo, line 28: remove this deprecated param, new megatron has delete it。

PR #668, opened 2026-03-03
CI: Test short suite

dougyster:
Trigger PR for `short` test suite. Add label `run-ci-short` to run tests.


dougyster:

PR #708, opened 2026-03-06
fix: validate --use-miles-router is required for --use-rollout-routing-replay

### LINKED ISSUE #407: Rollout Routing Replay (R3) error: rollout_routed_experts is required in rollout_data
### Environment
- **Base script**: `scripts/run-qwen3-30B-A3B.sh`
- **Docker image**: `radixark/miles:dev-20260105`
- **Model**: Qwen3-30B-A3B (MoE)

### Description

We are running RL training based on  
`scripts/run-qwen3-30B-A3B.sh` and made a minimal configuration change to enable rollout routing replay (R3) for Qwen3-30B-A3B.

Specifically, we added the following flags:

```yaml
use_routing_replay: true
use_rollout_routing_replay: true
```
No other MoE- or routing-related flags (e.g. `enable_return_routed_experts`, `DeepEP`, or custom dispatchers) were enabled.

### Observed behavior

Training crashes with the following error:

```bash
ValueError: rollout_routed_experts is required in rollout_data when use_rollout_routing_replay is set
```

This suggests that when `use_rollout_routing_replay=true`, miles expects a `rollout_routed_experts` field to be present in `rollout_data`, but this field is not being produced by the current rollout / SGLang integration in this setup.

**Additional observation**: when making direct requests to SGLang (v0.5.7) with `enable_return_routed_experts=True` and `return_routed_experts=True`, SGLang correctly returns routed expert information. However, when SGLang is used through Miles rollout, the routed expert data is not propagated into `rollout_data`.

NJX-njx:
## Related Issue
Fixes #407

## Problem
When users enable R3 routing replay with \use_rollout_routing_replay=true\ without setting \--use-miles-router\, the SGLang engine is launched externally without \enable_return_routed_experts=True\. This causes:
1. SGLang silently doesn't return \
outed_experts\ data in responses
2. \sample.rollout_routed_experts\ remains \None\
3. Training crashes with a confusing error about missing \
ollout_routed_experts\ in \
ollout_data\

## Fix
Added an early validation check in argument parsing that raises a clear \ValueError\ when \--use-rollout-routing-replay\ is set without \--use-miles-router\, explaining the dependency.

## Changes
- [miles/utils/arguments.py](miles/utils/arguments.py): Added validation check with descriptive error message

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -1802,6 +1802,13 @@ def miles_validate_args(args):
 
     if args.use_rollout_routing_replay:
         args.use_routing_replay = True
+        if not args.use_miles_router:
+            raise ValueError(
+                "--use-rollout-routing-replay requires --use-miles-router to be set. "
+                "Without the Miles router, the SGLang engine will not be launched with "
+                "enable_return_routed_experts=True and rollout_routed_experts data "
+                "will not be available."
+            )
 
     if args.custom_config_path:
         with open(args.custom_config_path) as f:


behroozazarkhalili:
Two findings from running this on top of current main (2ed189c0d):

1. It fails 4 existing tests: tests/fast/utils/test_arguments.py::TestSessionServerPauseGenerationMode::test_retract_r3_warning parses --use-session-server --use-rollout-routing-replay --pause-generation-mode retract without --use-miles-router, so the new ValueError fires (4 failed, 184 passed in test_arguments.py + test_megatron_cli_flags.py). Those cases would need updating or the check would need to allow the session-server path.

2. The mechanism in the description ("only the Miles router path launches SGLang with enable_return_routed_experts=True") does not match the code: miles/backends/sglang_utils/sglang_engine.py L133-134 sets enable_return_routed_experts from args.use_rollout_routing_replay alone, and it already did at the commit #407 was filed against (2cdc1f7189, same two lines). Whatever made --use-miles-router fix #407 for zianglih sits downstream of engine launch (how rollout_routed_experts reaches the sample), not in the launch flags. Worth pinning that path before adding the hard requirement.

cc @fzyzcjy @guapisolo

PR #710, opened 2026-03-06
[rollout] (2/N) Add chat template utils to select fixed templates automatically

guapisolo:
prerequisite: #709 

Many chat template is not append-only after the last user turn, they may not add special tokens correctly, especially for old qwen3 models. This PR add an interface to fix this issue.

For reviewer: This PR not included full logic of why a specific chat template need a fix.. it will release in the next PR. mostly the `miles/utils/chat_templates/fix_chat_template.py` and `--chat-template-path == autofix` design need review.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -20,6 +20,8 @@
 
 
 
+
+
 name: PR Test
 
 on:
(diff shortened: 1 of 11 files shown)


miles/utils/chat_templates/qwen3_fixed.jinja
yueming-yuan, line 1: minor: should we somehow improve the file structures, e.g. put all jinja files into one directory  that only contain .jinja, separate with python script, for better readability? given the number of models will increase later

guapisolo: Good idea. Done.

PR #711, opened 2026-03-06
[rollout, CI] (3/N) Add one-click recipe and tests to verify whether a chat template is append-only after last user message

guapisolo:
- Add a recipe and doc to tell whether a model's chat template is append only after last user message.
- Reorg chat template utils structure
- Fix qwen3.5 chat template with intermediate system message. Originally qwen3.5 banned intermediate system message, we should change it to role = user in text, like tool response goes.


miles/utils/test_utils/mock_trajectories.py
Shi-Dong, line 1: Seems that all mock trajectories end with tool call.  Does `chat_template_verify.py` also work when the final turn is user?  E.g.
```text
sys, user, assistant, tool, system_retry, assistant, tool, user
```

guapisolo: On one hand, this case should fail the append-only check, because the user input itself can already break certain chat templates append-only.

On the other hand, this case would not appear in real training data. The tool responses & user message after the assistant message, would always have a loss mask of 0, so they are useless for training. Also, under the current TITO design, we trace the trajectory only after the model has finished inference and then use that trace for training. As a result, any captured trajectory would necessarily end with an assistant message (tool call or normal end.)

PR #715, opened 2026-03-08
[AMD] Support AMD True on-policy Using Triton Backend

JessicaJiang-123:
## Motivation

Support true on-policy RL for text models on AMD GPU with triton attention backend.

For bit-wise identical log-probs between sglang rollout (decode) and FSDP training (extend), the inference and training must execute exactly the same numerical operations.

## Modifications

- **SGLang Triton Attention Bridge**: Add a lightweight adapter (`sglang_attn_bridge`) that patches only the core HF attention computation with SGLang's Triton kernel (see sgl-project/sglang#20143). Uses per-request index pointers (qo_indptr/kv_indptr) for batch-invariant, numerically deterministic computation.

- **Triton Attention Bridge Integration**: Actor and ref model construction now use `_build_model_with_attn_bridge()` with a two-stage attention switching strategy. Triton kernel is used for forward log-prob computation, and eager attention is used for backward pass.

- **AMD Triton Runner Script**: Add `run_simple_amd_triton.py` configured with `--sglang-attention-backend triton` and `--attn-implementation triton`, replacing NVIDIA-specific env vars with AMD-compatible options.

- **Conditional MOE Patching**: `_enable_true_on_policy_optimizations()` now detects whether the model is a Mixture-of-Experts model via num_experts config attribute, and only applies MOE-specific patches accordingly, preventing unnecessary patching of dense models.

## Accuracy Tests

Verified using [`run_simple_amd_triton.py`](https://github.com/JessicaJiang-123/miles/blob/amd-top-final-version/examples/true_on_policy/run_simple_amd_triton.py) from the [miles](https://github.com/radixark/miles) RL training framework, on Qwen3-0.6B (single AMD GPU, triton attention backend, `rl_on_policy_target=fsdp`):

train_rollout_logprob_abs_diff: 0.0
mismatch_k3_kl: 0.0

top-wandb-fv-new

## Collaborators
@XinyuJiangCMU
@JessicaJiang-123


guapisolo:
Impressive!

yushengsu-thu (changes requested):
@JessicaJiang-123 
1. Is it possible to move `hf_sglang_triton_patch.py` to the [patch directory under Docker](https://github.com/radixark/miles/tree/main/docker/amd_patch)
 and avoid modifying the original Miles codebase?

2. In `actor.py`, if your code needs to follow a different path from NV, could you use is_hip()? However, I still recommend use my first suggestion. I strongly believe that hardware-specific issues or dependencies should be resolved in the Docker image instead of the Miles codebase; otherwise, it will make the whole codebase harder to maintain.

XinyuJiangCMU:
> @JessicaJiang-123
> 
> 1. Is it possible to move `hf_sglang_triton_patch.py` to the [patch directory under Docker](https://github.com/radixark/miles/tree/main/docker/amd_patch)
>    and avoid modifying the original Miles codebase?
> 2. In `actor.py`, if your code needs to follow a different path from NV, could you use is_hip()? However, I still recommend use my first suggestion. I strongly believe that hardware-specific issues or dependencies should be resolved in the Docker image instead of the Miles codebase; otherwise, it will make the whole codebase harder to maintain.

Thanks, that makes sense. We opened a new PR following your suggestion:
https://github.com/radixark/miles/pull/1047
The AMD-specific SGLang Triton bridge is now moved under the Docker AMD patch directory, and the actor.py changes are applied only through the Docker patch instead of directly modifying the original Miles codebase.
I think we can close this PR and continue the review in #1047.

PR #719, opened 2026-03-09
[feat] CPU mem: release and realloc gradient buffers when offload train engine

yueming-yuan:
ci-megatron-pr: #17

### core changes
Megatron: https://github.com/radixark/Megatron-LM/pull/17
torch memory saver: https://github.com/fzyzcjy/torch_memory_saver/pull/72

### result
Green is the newest
compare_all

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -1750,9 +1750,6 @@ def miles_validate_args(args):
         args.offload_rollout = True
     del args.offload
 
-    if args.offload_train:
-        args.disable_grad_buffers_cpu_backup = True
-
     if args.debug_rollout_only:
         if args.colocate and (not args.rollout_num_gpus):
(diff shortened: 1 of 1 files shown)


miles/utils/arguments.py
```diff
@@ -167,6 +167,17 @@ def add_train_arguments(parser):
                     "Note: do not set `--ref-load` and `--keep-old-actor` if disable weights backuper."
                 ),
             )
+            parser.add_argument(
```
yueming-yuan: Do people think we need to keep this argument? Is there any senario that we need to keep cpu copy of gradient and distributed optimizer's parameter gather buffer? if no i will delete

fzyzcjy: I personally do not have an example to need this flag in my mind

yueming-yuan: okay

PR #729, opened 2026-03-12
[feat] (2/n) fault tolerance and RolloutManager async

yushengsu-thu:
Make rolloutmanager become async version and add APIs for FT:
Generated_image

and impl:
start_cell/stop_cell/etc..
to support: https://github.com/radixark/miles/pull/754


miles/ray/placement_group.py
```diff
@@ -170,8 +170,11 @@ def create_rollout_manager(args, pg):
     rollout_manager = RolloutManager.options(
         num_cpus=1,
         num_gpus=0,
+        max_concurrency=10,
```
fzyzcjy: maybe not use max_concurrency>1 since we are async actor (?)

miles/ray/rollout.py
```diff
@@ -43,19 +48,426 @@
...
+        async with self._lock:
+            if self._status == JobStatus.STOPPED:
+                return
+            self._status = JobStatus.PENDING
+            await self._async_stop()
+            self._status = JobStatus.STOPPED
+
+    async def _async_start(self) -> int:
+        group = self.server_group
+        handles, _ = group.start_engines(target_indices=self.engine_indices)
```
fzyzcjy: I may personally mv the logic from ServerGroup.start_engines into here (and simplify it)

miles/ray/rollout.py
```diff
@@ -80,22 +492,47 @@ def __init__(self, args, pg):
         logger.info(f"import {self.args.eval_function_path} as eval_generate_rollout function.")
 
         if self.args.debug_train_only:
-            self.all_rollout_engines = []
+            self.servers: dict[str, RolloutServer] = {}
+            self._init_handles: list = []
```
fzyzcjy: hmm if it is like this (have to await), then what about
split `__init__` into `__init__` + `async def init`, and put all heavy async work into the latter

miles/ray/rollout.py
```diff
@@ -167,68 +636,153 @@ def eval(self, rollout_id):
...
 
-    def save(self, rollout_id):
-        self.data_source.save(rollout_id)
+    async def save(self, rollout_id):
+        await asyncio.to_thread(self.data_source.save, rollout_id)
 
-    def load(self, rollout_id=None):
-        self.data_source.load(rollout_id)
+    async def load(self, rollout_id=None):
+        await asyncio.to_thread(self.data_source.load, rollout_id)
```
fzyzcjy: to be honest I am a bit worried, this makes DataSource have to understand multithreading. it may be better to let it just block the async thread and we do not care.
(similar for other changes)

yushengsu-thu: fix

miles/ray/placement_group.py
yushengsu-thu: done

miles/ray/rollout.py
yushengsu-thu: ok

PR #741, opened 2026-03-16
[1/N][Weight_update] Use `BucketedWeightGatherMixin` to isolate the weight all_gather logic

JensenFire:
Refactor the original `UpdateWeightFromDistributed`.
- Involve `DistBucketedWeightUpdateMixin` (old `UpdateWeightFromRemote`), to isolate the logic of `weight_allgather` , `convert_to_hf` and `update_weight`.
- involve `collect_named_tensors_for_weight_transfer` in `common.py`, identifying whether a tensor is in non-experts/experts clustering.
- others. Like move the function `post_process_weights` to `common.py` since it's also used int `UpdateWeightFromTensor`

Currently we could extract the logic of updating weights in `update_weight_from_distributed` as:
- `_pause_and_prepare_engines`: pause rollout engines, flush caches and run pre-process.
- `_gather_and_update_non_expert_weights`
- `_gather_and_update_expert_weights`
- `_finalize_and_resume_engines`: run post-process, resume rollout generation.

# Tests

```python tests/e2e/megatron/test_quick_start_glm4_9B.py``` , without `--colocate` and `num_gpus = 8` on h100

### CODE REVIEW

**Comment by JensenFire** [comment]
cc @Risc-lt @JD-ETH

**Comment by JD-ETH** [comment]
we will introduce a new `UpdateWeightPointToPoint` logic that implements the point to point weight update via RDMA transfer and extends the bucketed update Mixin interface.

File: miles/backends/megatron_utils/update_weight/update_weight_from_distributed.py [hunk 1]
```diff
@@ -335,20 +168,4 @@ def update_weights_from_distributed(
...
-    ray.get(
-        [
-            engine.post_process_weights.remote(
-                restore_weights_before_load=restore_weights_before_load,
-                post_process_quantization=post_process_quantization,
-            )
-            for engine in rollout_engines
-        ]
-    )
+#
```
yueming-yuan: nit: format

JensenFire:
cc @yueming-yuan for the next step

PR #713, opened 2026-03-07
[rollout] (4/N) [to be break down] Support OAI TITO with cross token inherit after last user turn

guapisolo:
ci-sglang-pr: #20066

Need this SGLang PR as prerequisite.
Be broken down into 4 parts. #744 #745 #746 #747

SGLang prerequisite: https://github.com/sgl-project/sglang/pull/20066
tested, and rebased on swe agent v2: https://github.com/radixark/miles/pull/726
Also has some modification to the agent framework.

PR #746, opened 2026-03-17
[rollout] (4.4/N) Add and fix e2e tests for session server and pretokenized TITO 

guapisolo:
ci-sglang-pr: #20066

- Add e2e tests to test TITO refinement logics, including glm-4.7-flash and qwen3-4b, with tool call error append and retry.
- Change to `sglang` folder instead of `sglang_patch`
Diff from #784  https://github.com/guapisolo/miles/compare/tito/2b-rollout-tito-integration...tito/3-e2e-tests 

This PR is the 4th part of #713 

A pr description generated by cc:
## Summary

Add end-to-end and integration tests for the TITO pretokenized session pipeline. The centerpiece is a **logprob equivalence test** that proves the incremental TITO tokenization path produces bit-identical model states to a full re-tokenization from scratch.

### What changed

**New: `test_tito_logprob_equivalence.py`** — E2E logprob + R3 equivalence verification (1 GPU)
- After the agent completes a multi-turn tool-call conversation through the session server, sends the full `accumulated_token_ids` to `/generate` with `max_new_tokens=0` for a single prefill pass
- Compares per-turn `input_token_logprobs` (prefill) against `output_token_logprobs` (decode) — token IDs must match exactly, logprob values within `1e-8` tolerance
- When `ENABLE_R3=1`, also compares `routed_experts` arrays (MoE expert routing decisions) between session decode and re-prefill
- Uses `--sglang-enable-deterministic-inference` + `temperature=0` for bit-reproducible results
- Verified: **Qwen3-4B** (diff=0.0) and **GLM-4.7-Flash** (diff=0.0, including R3)

**New: `test_session_server_tool_call.py`** — E2E session server with real inference (1 GPU)
- Starts the full miles pipeline via `execute_train --debug-rollout-only` with session server enabled
- Custom agent (`session_tool_agent.py`) runs multi-turn weather tool calls, randomly exercises both retry strategies (rollback + tool-message retry)
- Token-level correctness guaranteed server-side by `update_pretokenized_state` prefix check — if TITO produces wrong tokens, the request fails with HTTP 500

**New: `test_session_pretokenized_e2e.py`** — Fast integration test (no GPU)
- Mock SGLang server + MilesRouter + SessionServer wired together with predefined trajectories
- Parametrized across 6 working model/template configs (Qwen3-fixed, Qwen3-Thinking-fixed, GLM-5-native, Qwen3.5-native, Qwen3-Next-Instruct-native, Qwen3-Next-Thinking-fixed)
- Negative tests prove 3 broken native templates (Qwen3, Qwen3-Thinking, Qwen3-Next-Thinking) violate the prefix invariant

**New: `tests/e2e/sglang/utils/`** — shared test utilities
- `session_tool_agent.py` — custom agent for multi-turn tool-call conversations with random rollback/tool-retry strategies; retry appends a `tool`-role message with `tool_call_id: "invalid_tool_call"`
- `logprob_verify_generate.py` — custom generate function that inlines the production `agentic_tool_call.generate` flow and inserts a re-prefill verification step between record collection and sample finalization (avoids modifying production code; tradeoff is lockstep maintenance)
- `sglang_server.py` — helper to start/manage a real SGLang server subprocess

**Moved: `test_chat_input_ids_equivalence.py`** — relocated from `tests/e2e/sglang_patch/` to `tests/e2e/sglang/`
- Enhanced to handle tokenizer output format variations (dict vs. raw list)
- Added null-value guards for logprob handling
- Refined output token ID verification to encode chat text directly instead of relying on logprob token IDs

**Fix: `token_seq_comparator.py`** — remove `.strip()` from text comparison
- `_compare_single_segment` now compares decoded text exactly, without stripping leading/trailing whitespace
- Assistant text differences are already ignored at a higher level, so remaining (non-assistant) segments must match exactly — `.strip()` was masking real mismatches

**CI: added sglang e2e tests to matrix** (`.github/workflows/pr-test.yml.j2`)
- `test_chat_input_ids_equivalence` (1 GPU)
- `test_session_server_tool_call` × {qwen3, glm47} (1 GPU each)
- `test_tito_logprob_equivalence` × {qwen3, glm47} (1 GPU each)

### test_tito_logprob_equivalence — "Does TITO produce identical model states?"

Wraps the normal agentic flow with a verification step. After the agent finishes, the custom generate function (`logprob_verify_generate.py`) does:

1. **Session-level check**: assert `tito_session_mismatch == []` and `accumulated_token_ids` is non-empty
2. **Re-prefill**: send the full `accumulated_token_ids` to `/generate` with `logprob_start_len=first_prompt_len` and `max_new_tokens=0` — this runs a single forward pass over the entire conversation
3. **Per-turn comparison**: walk each session record, greedily match output token IDs against `accumulated_token_ids`, handle trailing token trimming (GLM's ambiguous boundary tokens), then compare:
   - Token IDs: exact match (fatal if different — TITO tokenization bug)
   - Logprobs: `abs_tol=1e-8` (with deterministic inference, observed diff=0.0)
   - Routed experts (R3): `np.testing.assert_array_equal` on matched slices
4. **Statistics**: log mean/max/p99 diff across all tokens

**Why this proves correctness**: if the prefill logprobs over the full `accumulated_token_ids` are identical to the per-turn decode logprobs, the KV cache states must be identical — meaning TITO's incremental token construction produces the exact same sequence as tokenizing the entire conversation from scratch.

### test_session_server_tool_call — "Does the full pipeline work?"

Starts the real miles training pipeline in `--debug-rollout-only` mode. A custom agent performs a multi-turn weather lookup with random retry strategies:
- 50% chance of **rollback** (discard failed assistant, trigger `_detect_and_rollback` on the server)
- 50% chance of **tool-message retry** (keep failed assistant, append a `tool`-role message with `tool_call_id: "invalid_tool_call"`)
- If no prior assistant checkpoint exists (`total_tool_calls == 0`), tool-message retry is used unconditionally since there is nothing to roll back to

Server-side prefix check in `update_pretokenized_state` is the implicit assertion — if TITO produces wrong tokens, the request fails with HTTP 500.

### test_session_pretokenized_e2e — "Does pretokenized injection work across templates?"

Uses mock SGLang server (no GPU) with predefined trajectories. For each model/template config:
- **Positive tests** (6 configs × 2 trajectory types): verify `input_ids == prompt_token_ids` on turn 2+, verify token growth across turns, verify session records
- **Negative tests** (3 broken configs): verify that native (unfixed) templates cause `input_ids != full_retokenization`, proving the prefix invariant is actually violated

### Not covered
- R3 with Qwen3 (not an MoE model)
- Multi-GPU tensor parallel configurations
- Rollback + R3 interaction (would need MoE model + agent that triggers rollback)

### DIFF
--- a/tests/e2e/sglang/utils/session_tool_agent.py
+++ b/tests/e2e/sglang/utils/session_tool_agent.py
@@ -199,6 +199,8 @@ async def run_agent(base_url, prompt, request_kwargs, metadata, **kwargs):
                     )
                 else:
                     messages.append(assistant_msg)
+                    # Add invalid tool call id field here. Otherwise template like kimi-k2 which explicitly
+                    # insert tool call id might failed. This won't affect most of chat templates.
                     messages.append(
                         {"role": "tool", "content": RETRY_TOOL_MESSAGE, "tool_call_id": "invalid_tool_call"}
                     )


miles/utils/chat_template_utils/token_seq_comparator.py
```diff
@@ -221,8 +221,8 @@ def _compare_single_segment(
                 )
             return None
 
-        exp_text = self._decode(exp.token_ids).strip()
-        act_text = self._decode(act.token_ids).strip()
+        exp_text = self._decode(exp.token_ids)
```
guapisolo, line 226: After ignoring assistant text diff, there is no need to keep the strip operator, as other text should be exact match.

tests/e2e/sglang/utils/session_tool_agent.py
```diff
@@ -0,0 +1,226 @@
...
+                    logger.info(
+                        "Turn %d: discarded failed assistant, " "will trigger session rollback (%d/%d)",
+                        turn,
+                        consecutive_retries,
+                        MAX_RETRIES,
+                    )
+                else:
+                    messages.append(assistant_msg)
+                    messages.append(
+                        {"role": "tool", "content": RETRY_TOOL_MESSAGE, "tool_call_id": "invalid_tool_call"}
```
guapisolo, line 205: Add invalid tool call id field here. Otherwise template like kimi-k2 will failed

tests/e2e/sglang/utils/logprob_verify_generate.py
```diff
@@ -0,0 +1,411 @@
+"""Custom generate function: agentic flow + re-prefill logprob verification.
+
+Design rationale
+~~~~~~~~~~~~~~~~
+This file needs to run the *exact same* pipeline as the production
```
maocheng23, line 5: This one could be a little tricky, so we can't import the code from the production?

guapisolo: > This one could be a little tricky, so we can't import the code from the production?

It's hard to directly import the code. This mainly because the CI need intermediate variables `records` and `session_metadata` to do verifications.

PR #747, opened 2026-03-17
[rollout] (4.5/N) Update swe-agent-v2 example scripts using current TITO design. 

guapisolo:
ci-sglang-pr: #20066

- Update params, suitable for session server.
- Fix script bug, should not include `DEPRECATED_MEGATRON_COMPATIBLE`
- add py scripts for swe agent v2

This PR is the 5th part of #713

### DIFF
--- a/scripts/models/glm4.7-flash.sh
+++ b/scripts/models/glm4.7-flash.sh
@@ -38,6 +38,7 @@ MODEL_ARGS=(
     --position-embedding-type rope
     --no-position-embedding
     --normalization RMSNorm
+    --norm-epsilon 1e-5
     --qk-layernorm
     --multi-latent-attention
     --q-lora-rank 768


scripts/models/glm4.7-flash.sh
```diff
@@ -38,6 +38,7 @@ MODEL_ARGS=(
     --position-embedding-type rope
     --no-position-embedding
     --normalization RMSNorm
+    --norm-epsilon 1e-5
```
maocheng23, line 41: Do we need this @Shi-Dong

guapisolo: > Do we need this @Shi-Dong

All other scripts have this variable... But without this param the glm4.7 it's still runnable.

guapisolo: > Do we need this @Shi-Dong

`argument.py` will automatically read `norm_epsilon` from hf config if we don't set it actively, and if set, it will check whether the input `norm_epsilon` in model script is equal to the value in hf config.

PR #750, opened 2026-03-18
Add environment report collection for training runs

fzyzcjy:
## Summary

Add per-node environment report collection for training runs. When `MILES_ENV_REPORT` env var is set (JSON string from external launcher), each actor collects and prints environment info during init.

### What it does

- **New `miles/utils/env_report.py`**: collects per-node environment info via single `pip inspect` call (JSON output):
  - Editable pip packages with locations
  - Git commit hash + dirty status for each editable package repo
  - Full pip package list
  - Launcher config (parsed from `MILES_ENV_REPORT`)
- **Actor init integration**: `TrainRayActor.init()` and `SGLangEngine.init()` call env report when `args.env_report` is non-empty
- **wandb config**: driver process writes `launcher_env_report` to wandb config
- **`--env-report` CLI arg** in `arguments.py`: reads from `MILES_ENV_REPORT` env var
- **`command_utils.py`**: forwards `MILES_ENV_REPORT` from host env into Ray runtime env vars

### Output format

Single-line JSON with parseable prefix, one line per actor:
```
ENV_REPORT_JSON={"role":"training","rank":0,"launcher_env_report":{...},"editable_packages":[...],"git_repos":[...],"full_pip_list":[...]}
```

### Design decisions

- **Opt-in**: only triggers when `MILES_ENV_REPORT` / `--env-report` is non-empty
- **Per-node**: every actor (training + rollout) independently collects — detects node-to-node environment inconsistencies
- **Graceful degradation**: all subprocess calls have timeouts and try/except — never crashes training
- **Single command**: replaced N+1 pip calls with one `pip inspect` (JSON output)

## Test plan
- [x] 17 unit tests (mock-based) — all pass
- [x] 3 integration tests (real editable package + git repo, clean/staged-dirty/unstaged-dirty) — pass on ac-h200-n04, skipped in CI (read-only env)
- [x] Pre-commit pass
- [ ] CI fast tests

### DIFF
--- a/miles/utils/env_report.py
+++ b/miles/utils/env_report.py
@@ -92,6 +92,7 @@ def _collect_pip_info() -> tuple[list[EditablePackageInfo], list[dict[str, str]]
     Returns (editable_packages, full_pip_list).
     """
     try:
+        # TODO: remove this workaround and still make Megatron detected
         env = {k: v for k, v in os.environ.items() if k != "PYTHONPATH"}
         result = subprocess.run(
             ["pip", "inspect"],


fzyzcjy:
failure seems unrelated

image

scripts/run_qwen3_4b.py
Shi-Dong, line 1: Curious: How are changes in this file related to the rest of this PR?

miles/utils/wandb_utils.py
```diff
@@ -88,6 +88,12 @@ def _compute_config_for_logging(args):
     ]
     output["env_vars"] = {k: v for k, v in os.environ.items() if k in whitelist_env_vars}
 
+    if env_report_raw := args.env_report:
+        from miles.utils.env_report import decode_env_report
```
Shi-Dong: Nit: import at the beginning?

miles/utils/env_report.py
```diff
@@ -0,0 +1,178 @@
...
+class GitRepoInfo:
+    package_name: str
+    location: str
+    commit: str
+    dirty: bool
+    diff_stat: str
+
+
+@dataclass
+class NodeEnvReport:
```
Shi-Dong, line 29: I think the current design is totally OK, but I'm worried that `pip inspect` would return a long list of installed packages, and manually comparing the list of packages across processes would be rather cumbersome.

I wonder—is it feasible to "freeze" the serialization of instances of this class (like using `sort_keys` in `json.dumps` or something like that), so that we can directly compare strings?

scripts/run_qwen3_4b.py
fzyzcjy: unrelated, I already splitted it

miles/utils/env_report.py
fzyzcjy: sorting LGTM, let me add it

miles/utils/wandb_utils.py
fzyzcjy: Done — moved to top-level import.

miles/utils/env_report.py
fzyzcjy: Added `sort_keys=True` to `json.dumps` in `_print_report`, so env reports from different processes can be directly compared as strings.

miles/utils/external_utils/command_utils.py
```diff
@@ -184,7 +184,7 @@ def execute_train(
         exec_command(
             f"export no_proxy=127.0.0.1 && export PYTHONBUFFERED=16 && "
             f"{cmd_megatron_model_source}"
-            f'ray job submit --address="http://127.0.0.1:8265" '
+            f"""ray job submit {'--address="http://127.0.0.1:8265" ' if 'RAY_ADDRESS' not in os.environ else ''}"""
```
maocheng23: This one could be conflict with new PR? Let's rebase onto the new one

miles/utils/env_report.py
```diff
@@ -0,0 +1,178 @@
...
+    return report
+
+
+def _collect_pip_info() -> tuple[list[EditablePackageInfo], list[dict[str, str]]]:
+    """Collect all pip info in a single `pip inspect` call.
+
+    Returns (editable_packages, full_pip_list).
+    """
+    try:
+        env = {k: v for k, v in os.environ.items() if k != "PYTHONPATH"}
```
maocheng23, line 96: QQ: will this strips the Megatron env?

Consider [command_utils.py:149]

miles/utils/env_report.py
```diff
@@ -0,0 +1,178 @@
...
+            env=env,
+        )
+        if result.returncode != 0:
+            logger.warning("pip inspect failed: %s", result.stderr)
+            return [], []
+
+        data = json.loads(result.stdout)
+        installed: list[dict[str, Any]] = data.get("installed", [])
+
+        full_pip_list = [_parse_pip_entry(pkg) for pkg in installed]
```
maocheng23, line 111: nit: Shall we excldue the `editable_packages` from `full_pip_list`?

miles/utils/external_utils/command_utils.py
fzyzcjy: sure

miles/utils/env_report.py
fzyzcjy: I guess adding it do not harm, thus keep it there

miles/utils/env_report.py
fzyzcjy: that's what is done deliberately by agents, o/w it says cannot find megatron as editable pkg. the phenomenon may be real while this work around may be suboptimal. maybe let me put a todo here.

maocheng23: I see, with a todo label would be good for us now.

PR #756, opened 2026-03-19
switch model to actor

maocheng23:
Always restore GPU model to actor before leaving init. Required for:
        1. offload_train: sleep() offloads self.model; wake_up restores it — must be actor.
        2. disagg: first update_weights() reads from self.model to push to rollout engine — must be actor.


yueming-yuan (approved):

guapisolo (approved):
LGTM. verified on some examples.

PR #774, opened 2026-03-20
Add routing replay for megatron runner

fzyzcjy:


miles/backends/megatron_utils/model.py
```diff
@@ -498,6 +498,8 @@ def forward_step(data_iterator: DataIterator, model: GPTModel, return_schedule_p
 
     dumper_phase_util.finalize(model)
```
Zhichenzzz: why adding two finalize here?

Zhichenzzz (approved):
LGTM! See the comments

miles/backends/megatron_utils/model.py
fzyzcjy: should be merge issue

PR #744, opened 2026-03-17
[rollout] (4.1/N) Add TITOTokenizer and TokenSeqComparator and upd template utils to align with SGLang

guapisolo:
ci-sglang-pr: #20066

- Add TITOTokenizer, which is a customization interface for agentic trajectory management.
- Add TokenSeqComparator, which is used for comparing whether a token sequence follow the behavior of chat template.
- Modify `template.py` to make sure it follow SGLang's behavior.

This PR is the 1st part of #713

PR #783, opened 2026-03-21
[rollout] (4.2/N) Support single user turn trajectory manager, add trim util

guapisolo:
ci-sglang-pr: #20066

- Impl `SingleUserTurnTrajectoryManager` to support better trajectory management.
- Make all token info returned from SGLang in `meta_info`

NOTICE: **CI fails will be fixed in the next PR #784 to reduce the size of this PR.**

Diff from #744 link: https://github.com/guapisolo/miles/compare/tito/1-chat-template-utils...tito/2a-session-trajectory

This PR is the 2nd part of #713.
Also, need this SGLang PR as prerequisite https://github.com/sgl-project/sglang/pull/20066.

A summary generated by cc:
## Summary

Replace `NaiveTrajectoryManager` (record-only, no cross-turn token reuse) with `SingleUserTurnTrajectoryManager` that implements **TITO pretokenized prefix caching** across agentic tool-call turns within a single user conversation.

### What changed

**New: `SingleUserTurnTrajectoryManager`** (`single_user_turn_trajectory.py`)
- Manages per-session state: message history, pretokenized `token_ids` checkpoints, and session records
- `try_prepare_pretokenized()`: on each turn, merges stored token IDs with new messages via `TITOTokenizer.merge_tokens()` and returns `input_ids` for SGLang — skipping redundant re-tokenization of the entire conversation
- `update_pretokenized_state()`: after each completion, appends a new checkpoint (`prompt_token_ids + completion_token_ids`) and validates that stored tokens are a prefix of the new ones (detects SGLang/tokenizer divergence)
- **Rollback support** (`_detect_and_rollback`): when an agent retries from an earlier point (e.g. re-running a tool call with different arguments), detects the divergence point, truncates state back to the last matching assistant checkpoint, and continues from there
- **Invariant enforcement**: rejects user messages after the first assistant (single-user-turn constraint), rejects non-append-only message modifications, raises on token prefix mismatch

**Updated: session HTTP routes** (`sessions.py`)
- Swapped `NaiveTrajectoryManager` → `SingleUserTurnTrajectoryManager`
- Proxy endpoint now calls `try_prepare_pretokenized()` before forwarding to SGLang, injecting `input_ids` when available
- After response, calls `update_pretokenized_state()` to accumulate token checkpoint
- GET session response now includes `metadata` with `accumulated_token_ids`, `tito_session_mismatch`, and `max_trim_tokens` for observability
- Hardcodes `logprobs=True`, `return_prompt_token_ids=True`, `return_meta_info=True`, `no_stop_trim=False` to prevent agent-side overrides from breaking token accumulation invariants
- Parses `meta_info.output_token_logprobs` instead of `logprobs.content[].token_id`

**Other changes**
- `SessionRecord` → added `metadata: dict` field to `GetSessionResponse`
- `Sample.strip_last_output_tokens()` — utility to remove trailing output tokens and all associated per-token info (tokens, log_probs, loss_mask, routed_experts)
- Deleted `naive_trajectory.py`

### Deleted
- `NaiveTrajectoryManager` / `naive_trajectory.py` — fully replaced

## Tests

Unit tests for `SingleUserTurnTrajectoryManager` in `test_single_user_turn_trajectory.py` — validates session state management and pretokenized token accumulation in isolation (no HTTP server, no real tokenizer). Uses a mock `TITOTokenizer` that passes through token IDs unchanged.

### Session CRUD (6 tests)
- Create, get, delete, append record
- Not-found edge cases (get returns `None`, delete returns `None`, append returns `None`)

### Multi-turn pretokenized trajectory (11 tests)
- **Happy path**: 2-turn and 3-turn agentic loops (`user → assistant(tool_call) → tool → … → final`), verifying `messages` and `token_ids` accumulate correctly across turns
- **First turn**: returns `None` (no prior tokens to reuse, falls back to normal tokenization)
- **Message structure variants**: no system message, multiple system messages at start, mid-trajectory retry system message (e.g. "Please try using the tools"), retry system message followed by full completion

### Invariant enforcement (5 tests)
- Token prefix mismatch between stored state and new prompt+completion → `ValueError`
- Non-append-only message modification (stored prefix altered) → `ValueError`
- User message after first assistant (violates single-user-turn constraint) → `ValueError`
- Unknown session ID → `ValueError` with diagnostic info

### Rollback (8 tests)
- Rollback to 1st assistant checkpoint (after 2 completions, agent sends different tool result)
- Rollback to 2nd assistant checkpoint (after 3 completions, diverges at 3rd tool)
- Rollback preserves exact prefix token IDs at the target checkpoint
- Rollback + continue: complete a full new trajectory from the rolled-back state
- Fewer request messages than stored → triggers rollback
- Normal append-only flow does NOT trigger rollback (no false positives)
- No assistant in matched prefix → `ValueError` (can't determine safe rollback point)
- Session `records` truncated in sync with trajectory checkpoints on rollback

### Integration tests (`test_sessions.py`)
- Updated to use `SingleUserTurnTrajectoryManager` and new `meta_info`-based response format
- Existing HTTP route tests (create/get/delete/proxy) continue to pass

### Not covered
- `compute_session_mismatch()` — requires real tokenizer + comparator; deferred to integration/e2e
- `get_session_token_ids()` — trivial delegation, indirectly exercised
- `Sample.strip_last_output_tokens()` — no unit test in this PR


miles/utils/chat_template_utils/template.py
```diff
@@ -97,19 +95,150 @@ def apply_chat_template_from_str(
...
+def message_matches(stored: dict[str, Any], new: dict[str, Any]) -> bool:
+    """Compare only the fields that affect chat-template tokenization.
+
+    External client libraries (e.g. litellm) may inject extra keys like
+    ``provider_specific_fields`` into messages.  These have no effect on
+    the Jinja2 chat template output, so we only compare the keys that
+    templates actually read: role, content, reasoning_content, tool_calls.
+    """
+    for key in _TEMPLATE_RELEVANT_KEYS:
+        if _normalize_value(stored.get(key)) != _normalize_value(new.get(key)):
```
maocheng23, line 159: For my understanding, here we use bit-wsie comparison between the key values w/o striping space and line breakers right?

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+class SingleUserTurnTrajectory(BaseModel):
+    """State for a single-user-turn trajectory.
+
+    Tracks the full message history and accumulated token IDs for one session.
+    The message sequence is: [system?, user, assistant, tool, assistant, tool, …].
+    """
+
+    messages: list[dict[str, Any]] = Field(default_factory=list)
```
maocheng23, line 26: nit: put them into a __init__()

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+    records: list[SessionRecord] = Field(default_factory=list)
```
maocheng23, line 27: Do we have weight version in this SessionRecord?

miles/router/session/sessions.py
```diff
@@ -71,12 +105,40 @@ async def chat_completions(request: Request, session_id: str):
 
         choice = response.get("choices", [{}])[0]
 
-        if "logprobs" not in choice or "content" not in choice["logprobs"]:
-            raise RuntimeError("logprobs must be in choice")
-        logprobs_content = choice["logprobs"]["content"]
-        for item in logprobs_content:
-            if "token_id" not in item:
-                raise RuntimeError("token_id must be in choice's logprobs content item")
+        if "meta_info" not in choice or "output_token_logprobs" not in choice.get("meta_info", {}):
```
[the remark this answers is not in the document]
maocheng23: nit here

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+        completion_token_ids: list[int],
+    ) -> None:
+        """Store raw token IDs after a successful response (no stripping).
+
+        Appends a new checkpoint to ``trajectory_token_ids``.  Validates that
+        the previously stored token_ids are a prefix of the new
+        ``prompt_token_ids + completion_token_ids``, confirming SGLang
+        actually reused our pretokenized input.
+        """
+        with self._lock:
```
maocheng23, line 269: Will we meet situation that two parts read into the same session?

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+        session_id: str,
+        request_messages: list[dict[str, Any]],
+        tools: list[dict[str, Any]] | None = None,
+    ) -> dict[str, Any] | None:
+        """Compute a merged prompt via ``TITOTokenizer.merge_tokens`` and
+        return it as ``input_ids`` for SGLang.
+
+        Returns ``None`` on the first turn (no stored token_ids yet).
+        """
+        with self._lock:
```
maocheng23, line 211: Same as below

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+                    + "\n".join(previews)
+                    + "\nThis usually means a stale agent environment from a previous "
+                    "training run is still sending requests after the router restarted. "
+                    "Ensure all agent containers are fully stopped before restarting training."
+                )
+
+            if not session.token_ids:
+                return None
+
+            self._assert_no_user_after_assistant(request_messages)
```
maocheng23, line 232: nit, add a comments for row 224-226

miles/router/session/single_user_turn_trajectory.py
```diff
@@ -0,0 +1,301 @@
...
+
+        Appends a new checkpoint to ``trajectory_token_ids``.  Validates that
+        the previously stored token_ids are a prefix of the new
+        ``prompt_token_ids + completion_token_ids``, confirming SGLang
+        actually reused our pretokenized input.
+        """
+        with self._lock:
+            session = self.sessions.get(session_id)
+            if session is None:
+                return
```
maocheng23: try_prepare_pretokenized raises a detailed ValueError for the same case.

miles/router/session/single_user_turn_trajectory.py
guapisolo: > Do we have weight version in this SessionRecord?

Yes, the `ChatCompletionResponse` has `meta_info`, which includes the weight version field.

miles/utils/chat_template_utils/template.py
guapisolo: > For my understanding, here we use bit-wsie comparison between the key values w/o striping space and line breakers right?

Yes. message will not strip any space or line breakers.

miles/router/session/single_user_turn_trajectory.py
guapisolo: Sorry, I think `BaseModel` do not require an `init` to do initialization 🤔 Do you think I should remove `BaseModel` to other impl ?

miles/router/session/single_user_turn_trajectory.py
guapisolo: We assume the agent’s execution flow is serial: the next inference request is only sent after the previous one returns a result. This holds for most of agent frameworks. In other words, for the same session ID, there should not be concurrent requests hitting the session router.

We can add a safeguard in later PR to prevent concurrent requests.

miles/router/session/single_user_turn_trajectory.py
guapisolo: Given a better impl

yueming-yuan (approved):

PR #784, opened 2026-03-21
[rollout] (4.3/N) Integrate TITO main logic into generations and fix uts.

guapisolo:
ci-sglang-pr: #20066

Diff from #783  link: https://github.com/guapisolo/miles/compare/tito/2a-session-trajectory...tito/2b-rollout-tito-integration
The run-ci-sglang issue will be handled in the next PR.

This PR is the 3rd part of #713.
Also, need this SGLang PR as prerequisite https://github.com/sgl-project/sglang/pull/20066.

A pr description generated by cc:
## Summary

tito/2a introduced `SingleUserTurnTrajectoryManager` and the session route changes that enable pretokenized prefix reuse across agentic turns — the session server now tracks `accumulated_token_ids`, detects token mismatches, and exposes session metadata via `GetSessionResponse`. However, the rollout pipeline didn't consume any of this: `compute_samples_from_openai_records` treated each turn's output tokens as-is, and stop tokens emitted by the model (e.g. `<|observation|>`) that also serve as the next turn's prompt delimiter were double-counted in the training sequence.

This PR closes the loop by wiring TITO session outputs into sample construction, training metrics, and CI validation.

### What changed

**Trailing-token trimming in sample construction** (`openai_endpoint_utils.py`)
- `compute_samples_from_openai_records` now accepts `accumulated_token_ids` and `max_trim_tokens` from session metadata
- Walks the accumulated sequence with a cursor, aligning each turn's `output_token_ids` against it; trailing tokens that don't match (consumed by the next turn's chat template) are trimmed via `Sample.strip_last_output_tokens()`
- Last turn is never trimmed; non-final turns assert `trim_count <= max_trim_tokens` to catch unexpected divergence
- After all records, asserts cursor == len(accumulated) to verify full sequence coverage

**Session metadata propagation** (`openai_endpoint_utils.py`, `agentic_tool_call.py`)
- `collect_records()` returns `(records, metadata)` instead of just `records`
- `agentic_tool_call.py` passes `accumulated_token_ids` and `max_trim_tokens` into sample construction, and attaches session metadata (`tito_session_mismatch`, etc.) to the final sample

**Data source migration** (`openai_endpoint_utils.py`)
- Token IDs and logprobs now read from `meta_info.output_token_logprobs` as `(logprob, token_id)` tuples, matching the format the session server already returns — replaces `logprobs.content[].token_id`/`.logprob`
- `build_chat_request_kwargs` no longer injects `logprobs=True` / `return_prompt_token_ids=True` — these are now hardcoded in the session route to prevent agent-side overrides from breaking token accumulation invariants

**TITO mismatch metrics** (`ray/rollout.py`)
- `compute_metrics_from_samples` reports `tito_session_mismatch_rate` broken down by type (`special_token_count`, `special_token_type`, `non_assistant_text`, `assistant_text`)
- In CI (`args.ci_test`), hard-asserts the first three are zero — `assistant_text` mismatch is non-critical since pretokenized prefix reuse can produce benign tokenization differences

**Other**
- `--tito-model` argument added to control TITO tokenizer type
- Agentic variant now returns `rollout_routed_experts` (was `pytest.skip`)
- Port-availability check before launching router process
- `_get_rollout_topk_from_response` → `get_rollout_topk_from_response` (used by both paths now)

### TITO trailing-token trimming (6 tests, `TestTITOTrailingTokenTrim`)

Validates the cursor-based trimming algorithm using concrete token ID sequences. See the test class docstring for a full worked example with three agentic retry turns.

- **3-turn trim**: non-final turns each have 1 trailing stop token trimmed, final turn keeps all tokens
- **No-trim**: output tokens fully match accumulated — `trim_count=0` for all turns
- **Single turn**: last turn never trims, even with `accumulated_token_ids` provided
- **Feature off**: `accumulated_token_ids=None` → no trimming performed (backward compat)
- **Over-trim assertion**: trailing tokens exceed `max_trim_tokens` → `AssertionError`
- **Cursor completeness**: accumulated sequence not fully consumed → `AssertionError`

### Routed experts multi-turn (`TestRoutedExpertsMultiTurn`)
- Removed `pytest.skip` for agentic variant — now fully asserts logprobs and `rollout_routed_experts` shape/values
- Agentic requests verified: `logprobs`, `return_prompt_token_ids`, `return_meta_info`, `no_stop_trim`, `return_routed_experts` all present

### Agent metadata (`TestAgentMetadata`)
- New: `test_tito_session_mismatch_present_in_metadata` — verifies `tito_session_mismatch` is a list of dicts with schema `{type, segment_index, expected_text, actual_text, detail}`

### Test infrastructure updates
- `MockSGLangServer` returns `meta_info.output_token_logprobs` in real SGLang format; supports `input_ids` passthrough for pretokenized requests
- `_make_record` helper uses `(logprob, token_id)` tuple format
- `generation_fixtures` passes `chat_template_path`, `tito_model`, `use_rollout_routing_replay` through to router
- `_strip_pretokenized()` helper strips session-injected `input_ids` from requests for deterministic comparison
- Existing `TestComputeSamplesFromRecords`, `TestMultiTurnPrefixChain`, `TestThinkingTokenPrefixBreak` adapted for new `args` parameter

### Not covered
- `compute_metrics_from_samples` TITO metric aggregation and CI assertion logic — no direct unit test; exercised via e2e
- Port-availability check in `_start_router` — runtime guard only

### DIFF
--- a/tests/fast/router/test_sessions.py
+++ b/tests/fast/router/test_sessions.py
@@ -81,7 +81,7 @@ class TestSessionRoutes:
     def test_get_session_not_found(self, router_env):
         response = requests.get(f"{router_env.url}/sessions/nonexistent", timeout=5.0)
         assert response.status_code == 404
-        assert response.json()["error"] == "session not found"
+        assert response.json()["error"] == "session not found: session_id=nonexistent"
 
     def test_delete_session(self, router_env):
         session_id = requests.post(f"{router_env.url}/sessions", timeout=5.0).json()["session_id"]
@@ -95,7 +95,7 @@ class TestSessionRoutes:
     def test_delete_session_not_found(self, router_env):
         response = requests.delete(f"{router_env.url}/sessions/nonexistent", timeout=5.0)
         assert response.status_code == 404
-        assert response.json()["error"] == "session not found"
+        assert response.json()["error"] == "session not found: session_id=nonexistent"
 
 
 class TestSessionProxy:


miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -26,47 +27,124 @@ async def create(args: Namespace):
...
         assert (
             request_input_ids == prompt_token_ids
         ), "for prompt part, input_ids return by sglang should match with the request input_ids"
+
     sample.tokens = prompt_token_ids + output_token_ids
     sample.rollout_log_probs = output_log_probs
     sample.response = tokenizer.decode(output_token_ids)
     sample.response_length = len(output_token_ids)
     sample.loss_mask = [1] * len(output_token_ids)
+    sample.rollout_routed_experts = get_rollout_topk_from_response(args, choice, sample, "routed_experts")
```
maocheng23, line 144: So by default, we will apply R3?

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -26,47 +27,124 @@ async def create(args: Namespace):
...
 
-        return records or []
-
-
-def compute_samples_from_openai_records(input_sample: Sample, records: list[SessionRecord], tokenizer) -> list[Sample]:
-    return [_compute_sample_from_openai_record(input_sample, record, tokenizer) for record in records]
-
-
-def _compute_sample_from_openai_record(input_sample: Sample, record: SessionRecord, tokenizer) -> Sample:
+        return records or [], metadata
```
maocheng23: nit: return (records or []), metadata

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -26,47 +27,124 @@ async def create(args: Namespace):
...
+    samples = []
+    cursor = 0
+
+    for i, record in enumerate(records):
+        is_last = i == len(records) - 1
+        prompt_ids = record.response["choices"][0]["prompt_token_ids"]
+        output_ids = [t[1] for t in record.response["choices"][0]["meta_info"]["output_token_logprobs"]]
+
+        trim_count = 0
+        if accumulated_token_ids is not None:
```
maocheng23, line 78: For my understanding, here `accumulated_token_ids` is based on the inference of each turn, and here we what to make sure the merged and rollout is the same for safety. And ideally, we can directly use the last turn?

miles/rollout/generate_utils/openai_endpoint_utils.py
guapisolo: The `get_rollout_topk_from_response` function will automatically detect whether the key field exists in the `meta_info`. if `routed_experts` does not exist, `rollout_routed_experts` will become `None`. No need to worry about this.

miles/rollout/generate_utils/openai_endpoint_utils.py
guapisolo: Yes. you're correct.

miles/rollout/generate_utils/tool_call_utils.py
```diff
@@ -82,8 +82,12 @@ def _tokenize_postfix_messages(
     messages_without = base_messages
     messages_with = base_messages + postfix_messages
 
-    tokens_with = tokenizer.apply_chat_template(messages_with, tokenize=True, add_generation_prompt=True)
-    tokens_without = tokenizer.apply_chat_template(messages_without, tokenize=True, add_generation_prompt=False)
+    tokens_with = tokenizer.apply_chat_template(
+        messages_with, tokenize=True, add_generation_prompt=True, return_dict=False
```
guapisolo, line 86: transformers 5.0.0 change default `return_dict` from False to True, which will return `BatchDecoding` instead of `list[int]`

guapisolo: Lets migrate all apply_chat_template to `template.py` in later PRs.

maocheng23: Approved to unblock.

yushengsu-thu (approved):
sync-offline

PR #785, opened 2026-03-21
[swe] Add truncation fix for miles agentic tool call

maocheng23:
As title, add length limit for agentic tool call

### DIFF
--- a/miles/rollout/generate_hub/agentic_tool_call.py
+++ b/miles/rollout/generate_hub/agentic_tool_call.py
@@ -83,7 +83,7 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
     if not samples:
         logger.warning("All samples truncated (prompt already exceeds max_seq_len)")
         sample = deepcopy(input.sample)
-        sample.status = Sample.Status.TRUNCATED
+        sample.status = Sample.Status.ABORTED
         return GenerateFnOutput(samples=sample)
 
     if not input.args.generate_multi_samples:


examples/experimental/swe-agent-v2/server.py
```diff
@@ -85,6 +87,8 @@ def _extract_exit_status(result) -> str:
         exc_type = getattr(exc, "exception_type", "")
         if exc_type in _TIMEOUT_EXCEPTIONS:
             return "LimitsExceeded"
+        if exc_type in _OUTPUT_LIMIT_EXCEPTIONS:
+            return "OutputLimitExceeded"
```
Shi-Dong: It's a bit weird to have both `LimitsExceeded` and `OutputLimitExceeded`. I'd propose changing `LimitsExceeded` to `TimeLimitExceeded`, and `OutputLimitExceeded` to `SequenceLengthLimitExceeded`.

More generally I think passing literal strings around isn't really the best practice? I'm OK with it as-is, though, since everything in under `experimental`.

miles/rollout/generate_hub/agentic_tool_call.py
```diff
@@ -50,11 +51,17 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
         custom_agent_function is not None
     ), f"Custom agent function {input.args.custom_agent_function_path} not found"
 
+    max_total_response_tokens = getattr(input.args, "max_total_response_tokens", None)
+
+    metadata = input.sample.metadata
+    if max_total_response_tokens is not None:
+        metadata = {**metadata, "max_total_response_tokens": max_total_response_tokens}
```
Shi-Dong: Super nit: I think Python style guide would recommend `metadata |= {"max_total_response_tokens": max_total_response_tokens}` since this modifies the dictionary `metadata` in-place, instead of creating a new dictionary instance and assigning it to `metadata`.

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -78,3 +78,44 @@ def _compute_sample_from_openai_record(input_sample: Sample, record: SessionReco
...
+    for sample in samples:
+        total = len(sample.tokens)
+        if total <= max_total_response_tokens:
+            result.append(sample)
+            continue
+
+        overshoot = total - max_total_response_tokens
+        allowed_output = sample.response_length - overshoot
+        if allowed_output <= 0:
+            break
```
Shi-Dong, line 102: Here if I understand correctly, if the prompt part of a sample already exceeds the length limit, then the entire sample is dropped.  I think this is quite dangerous since if `max_total_response_tokens` is so small such that the prompt of the first sample crosses the limit, then the returned list will be empty, which crashes `merge_samples` in `sample_utils.py`.

I'm not sure whether we can do the following: if the prompt in a sample has 500 tokens, and the response in the same sample has 200 tokens, with token limit 400, then we clip the sample, keeping just the first 400 tokens in the prompt with no response.

Also I'd recommend changing the name `max_total_response_tokens` into `max_total_sample_tokens` or something similar, since we are counting both prompt and response tokens.

miles/rollout/generate_hub/agentic_tool_call.py
maocheng23: Here I would prefer to do copy and paste as it's safer.

miles/rollout/generate_utils/openai_endpoint_utils.py
maocheng23: There are two parts:
1. Here we actually collect all samples from a single trajectory, and then merge later, so we will still have some data (unless the initial prompt is super long)
2. update the name to `max_seq_len`

examples/experimental/swe-agent-v2/server.py
maocheng23: Updated

miles/rollout/generate_utils/openai_endpoint_utils.py
Shi-Dong: I'm still worried that if `max_seq_len` is set to a small value (e.g. in a debugging run), the initial prompt would exceed the length limit, crashing the run.  I'd recommend keeping the truncated version of the initial prompt in this case (the mask would be all-zero, but at least this won't crash the run).

It's hard to predict how long the initial prompt would be: in swe-bench some prompts are a few thousand tokens long.

maocheng23: Processed in `miles/rollout/generate_hub/agentic_tool_call.py` row 83-87, and the `input.sample.tokens` should be [] by default so it should not include a long token ids

miles/rollout/generate_hub/agentic_tool_call.py
```diff
@@ -70,6 +77,15 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
     for s in samples:
         s.metadata.update(agent_metadata or {})
 
+    if max_seq_len is not None:
+        samples = truncate_samples_by_total_tokens(samples, max_seq_len, input.state.tokenizer)
+
+    if not samples:
+        logger.warning("All samples truncated (prompt already exceeds max_seq_len)")
+        sample = deepcopy(input.sample)
+        sample.status = Sample.Status.ABORTED
```
maocheng23, line 86: @Shi-Dong @yueming-yuan Now we aborted the cases where intial prompt>max_seq_len

PR #801, opened 2026-03-24
[Kimi k2.5] [Part 3] data processors

yueming-yuan:


miles/backends/training_utils/mm_data.py
```diff
@@ -0,0 +1,183 @@
...
+    rollout_data["loss_masks"] = expanded_loss_masks
+    rollout_data["total_lengths"] = expanded_total_lengths
+    rollout_data["response_lengths"] = expanded_response_lengths
+
+    metadata_changed = (expanded_total_lengths != old_total_lengths) or (
+        expanded_response_lengths != old_response_lengths
+    )
+    if metadata_changed:
+        cp_size = parallel_state.cp_size
+        if cp_size > 1 and qkv_format == "thd":
```
Zhichenzzz, line 156: In `mm_data.py`, `_expand_image_tokens_for_sample` has a branch handling image tokens in the response (if pos >= prompt_len), but the log_prob fix-up only runs when CP>1. Does the K2.5 rollout response ever contain media_token_id?

PR #803, opened 2026-03-25
[fix, CI] fixed megatron-bridge mode; add bridge mode CI on MoE model

yueming-yuan:
Fix 1: added missed arguments in `model_provider.py`
Fix 2: https://github.com/yushengsu-thu/Megatron-Bridge/pull/3


yushengsu-thu (approved):
sync off-line already

PR #804, opened 2026-03-25
Add pin_rollout_manager_to_head for rollout manager

maocheng23:


miles/ray/placement_group.py
```diff
@@ -166,11 +166,26 @@ def create_training_models(args, pgs, rollout_manager):
     return actor_model, critic_model
 
 
+def _get_head_node_id() -> str:
+    from ray.util.state import list_nodes
```
Shi-Dong: nit: import at the beginning?

fzyzcjy: btw I find seems claude always not want to import at beginning...

PR #810, opened 2026-03-26
[rollout] (5/N) Decouple session server out of miles router

guapisolo:
ci-sglang-pr: #21644

- Decouple session server out of miles router.
Diff from #746 https://github.com/guapisolo/miles/compare/guapisolo:tito/3-e2e-tests...radixark:tito/5-session-server

(Some small ut will be fixed tonight. just need review main changes.)

A description generated by cc:
## Summary

Decouple the session server from Miles Router into a standalone process and add `--use-session-server` flag for explicit lifecycle control.

### What changed

**Standalone `SessionServer`** (`miles/rollout/session/session_server.py`)
- Lightweight FastAPI server that manages TITO sessions and proxies inference to any backend (SGLang Rust Router, Miles Router, or direct worker)
- Session logic (`setup_session_routes`) shared between `MilesRouter` and `SessionServer`; removed `SessionBackend` Protocol in favor of duck-typed `do_proxy`/`build_proxy_response`
- Started as daemon process via `--use-session-server`, auto-allocates port in 5000-6000 range

**Move session module** `miles/router/session/` → `miles/rollout/session/`

**New CLI flags** (`miles/backends/sglang_utils/arguments.py`)
- `--use-session-server` / `--session-server-ip` / `--session-server-port`
- `agentic_tool_call.generate` asserts ip/port are set with actionable error message

**E2E tests** (all 1 GPU, added to CI matrix)
- `test_session_server_tool_call` — full pipeline with real inference, multi-turn tool calls with random rollback/retry strategies, prefix check as implicit assertion
- `test_tito_logprob_equivalence` — re-prefill verification proving TITO produces bit-identical KV cache states (token ID exact match, logprob diff=0.0 on Qwen3-4B and GLM-4.7-Flash, R3 routed_experts match when `ENABLE_R3=1`)

**Fast integration test** (`test_session_pretokenized_e2e.py`, no GPU)
- Mock SGLang + MilesRouter + SessionServer, parametrized across 6 model/template configs
- Negative tests verify 3 broken native templates violate prefix invariant

**Bugfix**
- `get_rollout_topk_from_response` — guard missing `num_layers` in debug-rollout-only mode
- All example scripts and E2E tests updated to pass `--use-session-server`

### DIFF
--- a/miles/rollout/session/session_server.py
+++ b/miles/rollout/session/session_server.py
@@ -10,6 +10,7 @@ import json
 import logging
 
 import httpx
+import setproctitle
 import uvicorn
 from fastapi import FastAPI, Request
 from fastapi.responses import JSONResponse
@@ -81,6 +82,9 @@ class SessionServer:
 
 def run_session_server(args, backend_url: str):
     """Entry point to start the standalone session server as a subprocess."""
+    # Visible to `pkill -9 miles`; without this the daemon inherits "python".
+    setproctitle.setproctitle("miles-session-server")
+
     server = SessionServer(args, backend_url)
     logger.info(
         "[session-server] Starting on %s:%s, proxying to %s",
--- a/miles/router/router.py
+++ b/miles/router/router.py
@@ -4,6 +4,7 @@ import json
 import logging
 
 import httpx
+import setproctitle
 import uvicorn
 from fastapi import FastAPI, Request
 from fastapi.responses import JSONResponse
@@ -18,6 +19,9 @@ def run_router(args):
     """
     Run the Miles router with the specified configuration.
     """
+    # Visible to `pkill -9 miles`; without this the daemon inherits "python".
+    setproctitle.setproctitle("miles-router")
+
     # Initialize the router with tokenizer and lazy worker initialization
     miles_router = MilesRouter(args, verbose=False)
 


miles/ray/rollout.py
```diff
@@ -1083,6 +1084,37 @@ def _resolve_sglang_config(args) -> SglangConfig:
 # ---------------------------------------------------------------------------
 
 
+def _start_session_server(args):
+    """Start a standalone session server when ``--use-session-server`` is set.
+
+    The session server runs as a separate process with its own port and proxies
+    inference requests directly to SGLang worker engines.  It is always started
+    as a standalone process regardless of whether ``--use-miles-router`` is active.
```
guapisolo, line 1099: In this PR, we still use miles router instead of sglang router since it need further fix.

miles/ray/rollout.py
```diff
@@ -1083,6 +1084,37 @@ def _resolve_sglang_config(args) -> SglangConfig:
...
+
+    router_url = f"http://{args.sglang_router_ip}:{args.sglang_router_port}"
+
+    from miles.rollout.session.session_server import run_session_server
+
+    process = multiprocessing.Process(target=run_session_server, args=(args, router_url))
+    process.daemon = True
+    process.start()
+    time.sleep(3)
+    assert process.is_alive(), "Session server process died on startup"
```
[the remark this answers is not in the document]
maocheng23: And alive is not enough, you will need to check the port is actually listening. 

```
for _ in range(30):
  time.sleep(0.5)
  if not process.is_alive():
      raise RuntimeError("Session server process died on startup")
  try:
      requests.get(f"http://{args.session_server_ip}:{args.session_server_port}/docs", timeout=1)
      break
  except Exception:
      continue
else:
  raise RuntimeError("Session server did not become ready within timeout")
```

miles/rollout/session/session_server.py
```diff
@@ -0,0 +1,84 @@
...
+        self.backend_url = backend_url
+        self.app = FastAPI()
+
+        timeout = getattr(args, "miles_router_timeout", None)
+        self.client = httpx.AsyncClient(
+            limits=httpx.Limits(max_connections=1024),
+            timeout=httpx.Timeout(timeout),
+        )
+
+        setup_session_routes(self.app, self, args)
```
maocheng23, line 41: QQ: would this server be auto closed when shutdown miles?

miles/rollout/session/session_server.py
```diff
@@ -0,0 +1,84 @@
...
+
+class SessionServer:
+    """Lightweight FastAPI server that manages sessions and proxies inference
+    requests through the inference router (sglang or miles)."""
+
+    def __init__(self, args, backend_url: str):
+        self.backend_url = backend_url
+        self.app = FastAPI()
+
+        timeout = getattr(args, "miles_router_timeout", None)
```
maocheng23: Are we sure we want None for default?

tests/fast/router/test_session_pretokenized_e2e.py
```diff
@@ -75,6 +74,11 @@ class ModelTemplateConfig:
 }
 
 BROKEN_CONFIGS: dict[str, ModelTemplateConfig] = {
+    # TODO: GLM-5 is here because the mock server lacks a reasoning parser, so
```
guapisolo, line 77: glm-5 has bug here, will fix in later PR.

In #746 it did not generate the bug, since #746 is in my forked repo, the secret HF_TOKEN is not injected to the container. So the CI https://github.com/radixark/miles/actions/runs/23634201581/job/68839874631?pr=746 sliently skipped the glm5-native test and passed

But current PR #810 is in a branch in miles root repo with HF_TOKEN. And the bug was exposed. It needs some modifications to mock_sglang_server and will be fixed in later PRs.

miles/rollout/session/session_server.py
guapisolo: > Are we sure we want None for default?

I didn't actively set default `timeout` since there's no logic to handle a real request failed or just no response temporarily.. Do you think add a number, like 600.0 (10min) is ok?

miles/rollout/session/session_server.py
guapisolo: > QQ: would this server be auto closed when shutdown miles?

From cc: Yes — on normal exit, both the miles-router and session-server are daemon processes, so Python auto-kills them when the parent exits.     

On abnormal exit (SIGKILL, OOM, crash), they become orphaned. The existing cleanup sequence (pkill -9 miles) now catches them because we set recognizable process titles via setproctitle (miles-router / miles-session-server).

miles/ray/rollout.py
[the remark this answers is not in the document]
guapisolo: good point, fixed by adding `wait_for_server_ready()` function

miles/ray/rollout.py
```diff
@@ -1083,6 +1090,43 @@ def _resolve_sglang_config(args) -> SglangConfig:
...
+
+    The session server runs as a separate process with its own port and proxies
+    inference requests directly to SGLang worker engines.  It is always started
+    as a standalone process regardless of whether ``--use-miles-router`` is active.
+    """
+    if not getattr(args, "use_session_server", False):
+        return
+
+    hf_checkpoint = getattr(args, "hf_checkpoint", None)
+    if not hf_checkpoint:
```
guapisolo, line 1105: This reserved for `--chat-template-path autofix`. Can be removed later.

PR #817, opened 2026-03-29
[rollout] (5.1/N) Refactor session impl for better lock control

guapisolo:
Remove old run.py impl.


miles/rollout/session/sessions.py
```diff
@@ -0,0 +1,191 @@
...
+            meta_info = choice.get("meta_info")
+            if not isinstance(meta_info, dict) or "output_token_logprobs" not in meta_info:
+                raise UpstreamResponseError(
+                    "meta_info and output_token_logprobs must be in choice (requires logprobs=True)"
+                )
+            assistant_message = choice.get("message", {})
+            # SGLang may return content=None when the assistant only emits tool_calls or
+            # reasoning tokens. Normalize to empty string to keep session tracking robust.
+            if assistant_message.get("content") is None:
+                assistant_message["content"] = ""
```
maocheng23: Why do we give the slient processing instead of the raising error as previously

miles/rollout/session/single_user_turn_trajectory.py
```diff
@@ -13,30 +12,125 @@
...
-    records: list[SessionRecord] = Field(default_factory=list)
-    trajectory_token_ids: list[list[int]] = Field(default_factory=list)
+    lock: asyncio.Lock = field(default_factory=asyncio.Lock, repr=False, compare=False)
+    closing: bool = field(default=False, repr=False, compare=False)
+    messages: list[dict[str, Any]] = field(default_factory=list)
+    records: list[SessionRecord] = field(default_factory=list)
+    trajectory_token_ids: list[list[int]] = field(default_factory=list)
     num_assistant: int = 0
 
-    @computed_field  # type: ignore[prop-decorator]
```
maocheng23, line 31: Why remove this?

guapisolo: > Why remove this?

From cursor: 
`@computed_field` is a Pydantic-specific decorator that makes a `@property` visible in `.model_dump()` serialization. This PR changed `SingleUserTurnTrajectory` from Pydantic `BaseModel` to `@dataclass` (to embed `asyncio.Lock` as a field for per-session lock control — `asyncio.Lock` is not serializable by Pydantic), so `@computed_field` was removed along with it. The `@property` itself is kept and works the same way.

miles/rollout/session/sessions.py
guapisolo: > Why do we give the slient processing instead of the raising error as previously

I revert this change, e2e swe agent v2 tests work well. Previously I modified this because of some reason in old refactor version (which is not very functionable).

PR #828, opened 2026-03-30
Rename dp groups with intra_ prefix

fzyzcjy:


Zhichenzzz:
**miles/backends/fsdp_utils/parallel.py:55** `parallel_state.dp_mesh = mesh["dp"]`

do we need to change this `dp_mesh` into `intra_dp_mesh`?

fzyzcjy:
Good point, given fsdp not support the indep dp mode, I temporarily keep it unchanged to minimize fsdp code change

Zhichenzzz:
nice, LGTM!

PR #833, opened 2026-03-30
Remove src_rank and use default behavior in gather_object

fzyzcjy:
## What

Remove `src_rank` field from `GroupInfo`. Replace with `dist.get_global_rank(gloo_group, 0)` at the call site.

## Why src_rank existed

`dist.gather_object(obj, gather_list, dst=..., group=...)` requires `dst` to be a **global rank** (not group-local rank). When gathering to group-rank-0, `dst` must be the global rank that corresponds to group-rank-0.

`src_rank` stored this value: the global rank of rank-0 within the dp_cp group. It was set via `mpu.get_data_parallel_src_rank(with_context_parallel=True)` (Megatron) or `dp_rank // world_size` (FSDP, which was always 0).

## Why it can be removed

`dist.get_global_rank(group, group_rank)` translates a group-local rank to a global rank at runtime. So instead of storing `src_rank` at construction time, we compute `dist.get_global_rank(gloo_group, 0)` at the call site. This is equivalent:

```python
# Before:
dst=parallel_state.intra_dp_cp.src_rank

# After:
dst=dist.get_global_rank(parallel_state.intra_dp_cp.gloo_group, 0)
```

Both resolve to the same global rank — the rank-0 member of the gloo group.

### DIFF
--- a/miles/backends/training_utils/log_utils.py
+++ b/miles/backends/training_utils/log_utils.py
@@ -168,7 +168,14 @@ def log_rollout_data(
             ):
                 # When R3 (rollout routing replay) is enabled, ref model does not use R3
                 # so log_probs and ref_log_probs may diverge; use a relaxed tolerance.
-                abs_tol = 1e-5 if args.use_rollout_routing_replay else 1e-9
+                # When --sglang-config deploys multiple models, the heavier offload/onload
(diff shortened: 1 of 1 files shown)


fzyzcjy:
image

PR #614, opened 2026-02-17
[feat] Expand mxfp8 quantization to support fine-grained layer precision config

zianglih:
@humansand

Expand mxfp8 quantization utils to support fine-grained layer precision config:
- Megatron `--num-layers-at-start-in-bf16`, `--num-layers-at-end-in-bf16`
- `--extra-high-precision-layers-hf`, `--extra-high-precision-layers-megatron` used by weight conversion
    - Intended to be used with Megatron `--te-precision-config-file`
- Ability to convert 128x128 blockscaled fp8 checkpoint to mxfp8 checkpoint in `tools/convert_hf_to_mxfp8.py`
- Use FlashInfer mxfp8 quantizer for faster weight sync and more explicit rounding, fall back to Triton if unavailable

This cannot be merged untill SGLang v0.6.10 bump with the following PRs;

- https://github.com/sgl-project/sglang/pull/18742
- https://github.com/sgl-project/sglang/pull/19537
- https://github.com/sgl-project/sglang/pull/20214
- https://github.com/sgl-project/sglang/pull/21280

PR #838, opened 2026-03-31
Support jdopensource/JoyAI-LLM-Flash

zianglih:
@humansand
Small-scale DeepSeek V3 model useful for 1 node experiment.


scripts/run_qwen3_30b_a3b.py
zianglih, line 1: No functional changes were made to this file.

yueming-yuan:
can we make a separate script from qwen3-30b for this model? and it's best if we can just run with "python scripts/run_joyai_llm_flash.py" without parsing extra args

zianglih:
Added a dedicated script. Since it contains certain mxfp8 features, hold untill sglang bump + https://github.com/radixark/miles/pull/614 .

zianglih:
Current reward issue from the JoyAI e2e: the rollout sample text printed by Miles is coherent and includes the JoyAI prompt-side delimiter:

```text
...<|Assistant|><|end_of_thought|>...Answer: \boxed{8}<|end▁of▁sentence|>
```

But the reward path calls `get_deepscaler_rule_based_reward(sample.response, label)`, so by the time we enter `deepscaler.py`, `response` starts after `<|end_of_thought|>` and no longer contains any of the delimiters that deepscaler currently uses (`</think>`, `<|end_of_thought|>`, `###Response`). That is why a correct sample was logged with `reward: 0` before the local fix.

A fix purely inside `deepscaler.py` is only possible if we relax its fallback behavior, e.g. treat the whole string as the model solution when no delimiter is present. That is what `reward-fix` tries. It would make this JoyAI case work when only `sample.response` is passed in, but it also changes deepscaler semantics for every caller: previously delimiter-less text returned `0`; afterward any delimiter-less text with a boxed answer can receive credit.

The narrower local fix I am testing is to keep `deepscaler.py` strict, and only for `rm_type == "deepscaler"` pass `sample.prompt + sample.response` when `sample.prompt` is a string. That lets deepscaler see the existing JoyAI delimiter without changing its parser contract. It is technically a caller-side change, but it is scoped to the deepscaler branch of `async_rm`, and the regression test covers the exact prompt-delimiter case.

Validation so far on the current devbox after syncing the `sglang-miles` fork:
- `tests/fast/rollout/rm_hub/test_deepscaler.py tests/fast/rollout/rm_hub/test_rm_hub.py`: 34 passed.
- The first JoyAI e2e rollout sample is not garbled and the same 2400 sample now logs `reward: 1` for `Answer: \boxed{8}`.

I am letting the full e2e continue to see whether anything remains after reward correctness, especially R3/replay consistency.

zianglih:
Update on the reward design and validation:

I switched this away from the earlier deepscaler/caller-side prompt concatenation approach. Following the reviewer suggestion in THUDM/slime#1170, this JoyAI math workload now uses `--rm-type math` instead of `--rm-type deepscaler`, since we only need answer correctness here and do not need DeepScaler format-loss enforcement. That keeps both `get_deepscaler_rule_based_reward` and the upstream async reward caller semantics unchanged.

Pushed commit: `75d67c875` (`Use math reward for JoyAI MXFP8 run`).

What changed:
- `scripts/run_joy_ai_llm_flash.py` and `tests/e2e/megatron/test_joyai_llm_flash_mxfp8.py` now use `--rm-type math`.
- Dropped the temporary `--no-save-optim --no-save-rng` extra args; the final checkpoint save completed successfully, though it took about 170s.
- Kept the e2e args explicit in the test file, including `--num-rollout 2`, 4 rollout GPUs + 4 train GPUs, `--sglang-fp8-gemm-backend flashinfer_cutlass`, and the HF/Megatron high-precision layer lists.

Validation:
- `pre-commit run --all-files`: passed locally before syncing.
- Remote fast reward tests in the final e2e run: `33 passed, 6 warnings in 0.03s`.
- SGLang fork validation used the `sglang-miles` checkout at `/sgl-workspace/sglang/python`; `TestFlashinferTrtllmRoutedMxfp8MixedBF16.test_gsm8k` passed with `gsm8k_score=0.93` using `flashinfer_cutlass` + `flashinfer_trtllm_routed`.
- Direct smoke on the locally converted JoyAI MXFP8 model generated coherent math samples without garbled text.
- Final end-to-end run `joyai_issue8_e2e_math_rm_20260527_001848` passed: Ray job `raysubmit_MRA4T9gt9GB3rMQb` succeeded. Rollout samples were readable and reward-bearing; the first rollout sample answered `\\boxed{8}` with reward `1`, rollout metrics showed `raw_reward=0.8359375`, and the second rollout showed `raw_reward=0.859375`. Logs were pulled under `logs/joyai_issue8_e2e_math_rm_20260527_001848/`.

zianglih:
Follow-up after removing checkpoint saves from the JoyAI e2e test.

Pushed commit: `00915ce2d` (`Skip checkpoint saving in JoyAI e2e`).

What changed:
- The JoyAI e2e test no longer passes `--load`, `--save`, `--save-interval`, or `--save-retain-interval` for training checkpoints.
- The HF->MXFP8 conversion `--save-dir` path is unchanged.

Validation:
- Local `python3 -m py_compile tests/e2e/megatron/test_joyai_llm_flash_mxfp8.py`: passed.
- Local `pre-commit run --all-files`: passed.
- Remote e2e rerun `joyai_e2e_no_ckpt_20260527_005138` passed: Ray job `raysubmit_UtLz6Di3BxrNZZWz` succeeded.
- The run went from train end directly to weight update; no `Timer save_model` path ran.
- Post-run checkpoint scan printed `__CHECKPOINTS_AFTER_TEST__` with no checkpoint directories under `/root/shared_data`. Logs are pulled under `logs/joyai_e2e_no_ckpt_20260527_005138/`.

zianglih:
Split the `--fp8-param-gather` work into a stacked PR here: https://github.com/zianglih/miles/pull/2

Design summary:
- keep the JoyAI MXFP8 high-precision controls from this PR (`--extra-high-precision-layers-hf`, `--extra-high-precision-layers-megatron`, and the TE precision config);
- use a direct MXFP8 rollout weight-sync payload for FP8 primary weights, carrying raw quantized weight bytes plus `weight_scale_inv`, while falling back to dense transfer for first/last BF16 and explicit high-precision layers;
- avoid `--optimizer-cpu-offload` with MXFP8 `--fp8-param-gather`; use `--offload-optimizer-states` with the MCore offload/reload/release lifecycle instead;
- work around the train/eval transition issue tracked in NVIDIA-NeMo/Megatron-Bridge#3801 by suspending DDP forward pre-hooks during ref/old-logprob eval and restoring them for train.

Validation passed on `joyai-llm-e2e4`: 1 rollout GPU + 7 train GPUs, PP7, `--num-rollout 2`, `flashinfer_cutlass`, run `joyai_fp8_param_gather_pp7_1r7t_train1k_try25_execute_20260527_112331`. The run completed two rollout/train cycles and final direct weight sync with Ray job success.

yueming-yuan (approved):

PR #843, opened 2026-03-31
[rollout] Rebuild dockerfile using sgl router

guapisolo:
ci-sglang-pr: #21571

After we do modifications to SGLang router, we need to rebuild the SGLang router in docker.


yueming-yuan:
how long time will this build take?

guapisolo:
> how long time will this build take?

Let check on meeting.

PR #857, opened 2026-04-02
[session] Limit roll back to one step

guapisolo:
As @maocheng23 previous suggestion, the default behavior should not allow agent roll back trajectory by more than 1 assistant step. One step is just retry one assistant generation, this is common across harness when assistant generate invalid tool calls. But roll back more than one step might be some error.

### DIFF
--- a/miles/rollout/session/single_user_turn_trajectory.py
+++ b/miles/rollout/session/single_user_turn_trajectory.py
@@ -12,6 +12,11 @@ from miles.utils.chat_template_utils.tito_tokenizer import TITOTokenizer
 logger = logging.getLogger(__name__)
 
 
+# TODO: hardcoded to 1 for now; if multi-step rollback is actually needed,
+#  raise this limit or make it configurable and remove the restriction.
+MAX_ASSISTANT_ROLLBACK_STEPS = 1
+
+
 def _assert_no_user_after_assistant(messages: list[dict[str, Any]]) -> None:
     """Assert no user message appears after the first assistant message."""
     seen_assistant = False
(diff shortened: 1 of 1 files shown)


miles/rollout/session/single_user_turn_trajectory.py
```diff
@@ -137,29 +140,15 @@ def _try_detect_and_rollback_to_assistant_checkpoint(
...
-              num_assistant        = 1
+        Only a single-step rollback is allowed (controlled by
+        ``MAX_ASSISTANT_ROLLBACK_STEPS``).  Discarding exactly one assistant
+        message means the agent is retrying from the latest checkpoint —
+        the request shares the stored prefix up to that assistant and then
+        continues with whatever the agent chooses (same or different tool
+        result, additional messages, etc.).  Any request that would need to
+        discard more than one assistant (i.e. jump back across multiple
+        turns) is rejected with ``MessageValidationError`` and no state is
+        modified.
```
Shi-Dong, line 158: Can you keep the example in the original docstring?  It makes things much easier to understand.

miles/rollout/session/single_user_turn_trajectory.py
```diff
@@ -32,12 +32,15 @@ class SingleUserTurnTrajectory:
...
-    in which case the session is rolled back to the last matching assistant
-    checkpoint and re-extended from there.
+    in which case the session is rolled back at most one assistant step.
 
     Concurrency contract: all mutating methods must be called under ``self.lock``.
     """
 
+    # TODO: hardcoded to 1 for now; if multi-step rollback is actually needed,
+    #  raise this limit or make it configurable and remove the restriction.
+    MAX_ASSISTANT_ROLLBACK_STEPS = 1
```
Shi-Dong: Nit: I would make this a script-level global variable (instead of a class attribute), and set class attribute `max_assistant_rollback_steps = MAX_ASSISTANT_ROLLBACK_STEPS` here.

maocheng23: Let's set it a global var, with default=1?

miles/rollout/session/single_user_turn_trajectory.py
guapisolo: ok. I will do that.

miles/rollout/session/single_user_turn_trajectory.py
guapisolo: I intended to hardcode this since I consider the roll back operation > 1 step is abnormal (there might be some error....).

guapisolo: ok. I will changed it to a global variable.

PR #859, opened 2026-04-02
[session] Rename SingleUserTurnTrajectory to LinearTrajectory

guapisolo:
Modify the class name to make it align with definition, only one trajectory as training sample.

In the future, we will support something like `TreeTrajectory` to support multi training sample collection.


Shi-Dong (approved):
LGTM

PR #864, opened 2026-04-02
[Do not merge] nvfp4 rollout with bf16 Megatron training

Zhichenzzz:
deepseek v3 with nvfp4 rollout and bf16 training


zianglih:
Hi, on the SGLang side you may find https://github.com/sgl-project/sglang/pull/18012 useful for activation per-tensor fp32 scales.

Zhichenzzz:
@zianglih Thank you for your suggestions! I will look into that.

PR #865, opened 2026-04-02
Fix/agentic error hendle

maocheng23:
Fix run.py


maocheng23:
cc @Shi-Dong

guapisolo (approved):
LGTM. But I am not quite familiar with prometheus. As long as it looks good on your side, we're good.

PR #866, opened 2026-04-03
Add heartbeat and id to session server

maocheng23:
## Summary
- Add `/health` endpoint to session server that returns `session_server_instance_id`
- Propagate `session_server_id` (host:port) and `session_server_instance_id` (uuid) through agentic_tool_call metadata to the agent server
- `OpenAIEndpointTracer.create()` fetches instance_id from `/health` on startup
- Generate uuid-based `session_server_instance_id` in `_start_session_server`
- Forward `session_server_id` and `session_server_instance_id` in swe_agent_function to agent server requests

## Test plan
- [x] Unit test: `test_create_fetches_session_server_instance_id` — verifies tracer fetches instance_id from health endpoint
- [x] Unit test: `test_health_reports_stable_instance_id` — verifies health endpoint returns consistent uuid
- [x] Unit test: `test_session_server_identity_forwarded_to_agent_metadata` — verifies id propagation through generate pipeline

### DIFF
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -36,7 +36,7 @@ class OpenAIEndpointTracer:
         session_url = f"http://{session_ip}:{session_port}"
         session_server_instance_id = None
         try:
-            health = await post(f"{session_url}/health", {}, action="get", timeout=_SESSION_REQUEST_TIMEOUT)
+            health = await post(f"{session_url}/health", {}, action="get")
             if isinstance(health, dict):
                 session_server_instance_id = health.get("session_server_instance_id")
                 if session_server_instance_id is not None:


examples/experimental/swe-agent-v2/swe_agent_function.py
```diff
@@ -60,6 +60,17 @@ async def run(
...
         request["max_seq_len"] = int(max_seq_len)
 
+    session_server_id = metadata.get("session_server_id")
+    if session_server_id is not None:
+        if external_host:
+            port = urlsplit(f"http://{session_server_id}").port
+            session_server_id = f"{external_host}:{port}"
+        request["session_server_id"] = session_server_id
+
+    session_server_instance_id = metadata.get("session_server_instance_id")
```
Shi-Dong, line 70: I'd suggest adding a comment here to explain what `session_server_id` and `session_server_instance_id` are for.

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -33,9 +34,22 @@ async def create(args: Namespace):
...
         session_url = f"http://{session_ip}:{session_port}"
+        session_server_instance_id = None
+        try:
+            health = await post(f"{session_url}/health", {}, action="get")
+            if isinstance(health, dict):
+                session_server_instance_id = health.get("session_server_instance_id")
+                if session_server_instance_id is not None:
+                    args.session_server_instance_id = session_server_instance_id
+        except Exception as e:
+            logger.warning("Failed to get session server health from %s: %s", session_url, e)
```
Shi-Dong, line 45: It feels like this warning is only logged when the post times out (the following `if` section is unlikely to throw an exception).  I wonder whether it'd be more helpful to also log the warning when the session doesn't have a `session_server_instance_id`.

tests/fast/rollout/generate_hub/test_multi_turn.py
```diff
@@ -667,6 +668,29 @@ def test_agent_returns_none_metadata_unchanged(self, variant, generation_env):
             assert s.metadata.get("instance_id") == "test-123"
             assert "reward" not in s.metadata
 
+    def test_session_server_identity_forwarded_to_agent_metadata(self, variant, generation_env):
+        from miles.utils.test_utils import mock_tools
```
Shi-Dong, line 672: Nit: the `import` here is rather weird because some functions in `mock_tools` are already imported at the beginning.

Shi-Dong (approved):
LGTM with minor comments.

miles/ray/rollout.py
Shi-Dong, line 1130: Super nit: can we move this import to the beginning as a drive-by?

I think the style guide mandates that all imports be placed at the beginning.  Personally it'd be OK for me make an exception if the import is heavy-weight, but it doesn't seem so in this case :/

PR #868, opened 2026-04-03
Fix session request hanging with asyncio timeout

maocheng23:
## Summary
- Add 120s `asyncio.wait_for` timeout to `collect_records()` (GET) and session deletion (DELETE) in `OpenAIEndpointTracer`
- Prevents indefinite hangs when the session server has stale HTTP keepalive connections — instead of blocking forever, the tracer logs an error and returns empty records
- Timeout applies to both the GET (collect) and DELETE (cleanup) calls independently

## Test plan
- [ ] Run agentic rollout end-to-end; verify sessions are collected normally when server is healthy
- [ ] Kill/restart session server mid-rollout; confirm the tracer logs timeout errors and returns gracefully instead of hanging

### DIFF
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -14,7 +14,7 @@ from miles.utils.types import Sample
 
 logger = logging.getLogger(__name__)
 
-SESSION_REQUEST_TIMEOUT = 120
+_SESSION_REQUEST_TIMEOUT = 120
 
 
 class OpenAIEndpointTracer:
@@ -38,17 +38,24 @@ class OpenAIEndpointTracer:
         return OpenAIEndpointTracer(router_url=session_url, session_id=session_id)
 
     async def collect_records(self) -> tuple[list[SessionRecord], dict]:
-        url = f"{self.router_url}/sessions/{self.session_id}"
         try:
             response = await asyncio.wait_for(
-                post(url, {}, action="get"),
-                timeout=SESSION_REQUEST_TIMEOUT,
+                post(self.base_url, {}, action="get"),
+                timeout=_SESSION_REQUEST_TIMEOUT,
             )
         except asyncio.TimeoutError:
             logger.error(
-                f"Timed out waiting for session {self.session_id} records after {SESSION_REQUEST_TIMEOUT}s "
+                f"Timed out waiting for session {self.session_id} records after {_SESSION_REQUEST_TIMEOUT}s "
                 f"(likely stale HTTP keepalive connection). Returning empty records."
             )
+            # Still attempt to clean up the session.
+            try:
+                await asyncio.wait_for(
+                    post(self.base_url, {}, action="delete"),
+                    timeout=_SESSION_REQUEST_TIMEOUT,
+                )
+            except Exception:
+                logger.warning(f"Failed to delete session {self.session_id} after timeout")
             return [], {}
         except Exception as e:
             logger.warning(f"Failed to get session {self.session_id} records: {e}")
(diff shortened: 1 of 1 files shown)


miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -35,8 +38,18 @@ async def create(args: Namespace):
         return OpenAIEndpointTracer(router_url=session_url, session_id=session_id)
 
     async def collect_records(self) -> tuple[list[SessionRecord], dict]:
+        url = f"{self.router_url}/sessions/{self.session_id}"
         try:
-            response = await post(f"{self.router_url}/sessions/{self.session_id}", {}, action="get")
+            response = await asyncio.wait_for(
+                post(url, {}, action="get"),
+                timeout=SESSION_REQUEST_TIMEOUT,
+            )
```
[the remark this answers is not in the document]
fzyzcjy: +1

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -13,6 +14,8 @@
 
 logger = logging.getLogger(__name__)
 
+SESSION_REQUEST_TIMEOUT = 120
```
fzyzcjy: _SESSION_REQUEST_TIMEOUT

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -35,8 +38,18 @@ async def create(args: Namespace):
...
+            response = await asyncio.wait_for(
+                post(url, {}, action="get"),
+                timeout=SESSION_REQUEST_TIMEOUT,
+            )
+        except asyncio.TimeoutError:
+            logger.error(
+                f"Timed out waiting for session {self.session_id} records after {SESSION_REQUEST_TIMEOUT}s "
+                f"(likely stale HTTP keepalive connection). Returning empty records."
+            )
+            return [], {}
```
fzyzcjy, line 59: qq: why we return empty instead of raise (explicit fail)

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -45,7 +58,12 @@ async def collect_records(self) -> tuple[list[SessionRecord], dict]:
         metadata = response.metadata
 
         try:
-            await post(f"{self.router_url}/sessions/{self.session_id}", {}, action="delete")
+            await asyncio.wait_for(
+                post(url, {}, action="delete"),
+                timeout=SESSION_REQUEST_TIMEOUT,
+            )
+        except asyncio.TimeoutError:
+            logger.warning(f"Timed out deleting session {self.session_id} after {SESSION_REQUEST_TIMEOUT}s")
```
fzyzcjy: does `except Exception` already handles it

miles/rollout/generate_utils/openai_endpoint_utils.py
maocheng23: Raising here would propagate through asyncio.gather in generate_and_rm_group and abort the entire sample group, not just this one sample. Returning empty isolates the failure — the caller in agentic_tool_call.generate already handles not records by marking the single sample as ABORTED. Added a comment explaining this + also attempting DELETE cleanup before returning.

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -13,6 +14,8 @@
 
 logger = logging.getLogger(__name__)
 
+_SESSION_REQUEST_TIMEOUT = 120
```
guapisolo, line 17: I am thinking shall we add a `timeout` argument in `post()` of `miles/utils/http_utils.py` ? Since it's very common to set `timeout` for communication between service components.

guapisolo: Then we have no need to write this every time. It will be handled in the post function.

```
await asyncio.wait_for(
    post(self.base_url, {}, action="delete"),
    timeout=_SESSION_REQUEST_TIMEOUT,
) 
```

guapisolo: Although the wrapped `post()` function can only be used in the miles main process which launch http_client globally in `init_http_client()`

maocheng23: Good idea — the underlying issue is that `httpx.Timeout(None)` on the shared client means a single stale TCP connection can hang forever. Adding a `timeout` param to `post()` would be cleaner, but it touches all callers and needs to work for both local and Ray-distributed paths. Keeping the localized `asyncio.wait_for` for now since it's explicit about why the timeout is needed here (session GET/DELETE over potentially stale keepalive connections). Happy to do the `post()` refactor as a follow-up.

guapisolo: You can still control the per-request timeout.. Doesn't matter I think we can still go with this now. It's still clean

PR #870, opened 2026-04-03
Split session lock: release during proxy to unblock DELETE

maocheng23:
## Summary
- Refactor `chat_completions` lock into three phases: Phase 1 (prepare, lock held) → Phase 2 (proxy to SGLang, **no lock**) → Phase 3 (update state, lock held)
- Previously the lock was held for the entire duration of `do_proxy`, blocking DELETE and causing agent disconnection to leave sessions stuck
- Add stale-update guard: captures `expected_num_assistant` before proxy, checks it after to detect concurrent modifications
- Add closing check after proxy: if session was marked closing during the proxy call, skip state update gracefully
- Fix `_inflight_chat` counter to use `try/finally` so it always decrements (was only on the happy path before)

## Test plan
- [ ] Run multi-session agentic rollout; verify sessions complete and records are collected correctly
- [ ] Disconnect an agent mid-request; confirm DELETE proceeds without waiting for the proxy call to finish
- [ ] Verify no state corruption when concurrent chat_completions race on the same session

### DIFF
--- a/tests/fast/router/test_session_race_conditions.py
+++ b/tests/fast/router/test_session_race_conditions.py
@@ -1,10 +1,17 @@
 """E2E session stress tests.
 
-Contract under test (with session.lock / session.closing):
-- Requests within the same session are serialized via session.lock.
+Contract under test (with split-lock / session.closing):
+- Phase 1 (prepare) and Phase 3 (state update) hold session.lock briefly;
+  Phase 2 (proxy to SGLang) does NOT hold the lock.
+- Concurrent same-session requests can overlap at the backend (Phase 2),
+  but state updates (Phase 3) are serialized; stale-update guard
+  (expected_num_assistant check) ensures only one concurrent writer wins.
 - Different sessions can run in parallel (no global lock).
 - Per-session clients can run turn-by-turn without idle gaps while global load stays parallel.
-- Delete marks session.closing=True, waits for session.lock, then removes.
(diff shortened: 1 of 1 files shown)


miles/rollout/session/sessions.py
```diff
@@ -83,62 +108,58 @@ async def delete_session(session_id: str):
...
-            if pretokenized is not None:
-                request_body["input_ids"] = pretokenized["input_ids"]
-                logger.debug(
-                    "Using pretokenized input_ids: %d tokens",
-                    len(pretokenized["input_ids"]),
+            # --- Phase 1: prepare request (lock held briefly) ---
+            async with session.lock:
+                if session.closing:
+                    raise SessionNotFoundError(f"session not found: session_id={session_id}")
+
```
guapisolo: Keep the original code comments here?

miles/rollout/session/sessions.py
```diff
@@ -83,62 +108,58 @@ async def delete_session(session_id: str):
...
+                        "Using pretokenized input_ids: %d tokens",
+                        len(pretokenized["input_ids"]),
+                    )
+
+                body = json.dumps(request_body).encode()
+                expected_num_assistant = session.num_assistant
+            # --- lock released here ---
+
+            # --- Phase 2: proxy to SGLang (NO lock held) ---
             result = await backend.do_proxy(request, "v1/chat/completions", body=body)
```
guapisolo: nit: Also keep comments here

Issue #366, opened 2025-12-29
MTP not working with Qwen3 Next


nutriarch:
```

2025-12-28 15:13:50

b200-202…4kot98i7
(MegatronTrainRayActor pid=40438) [2025-12-28 23:13:50] rerun_state_machine.py:238 - RerunStateMachine initialized in mode RerunMode.DISABLED
2025-12-28 15:13:51

b200-202…4kot98i7
Traceback (most recent call last):
2025-12-28 15:13:51

b200-202…4kot98i7
  File "/tmp/ray/session_2025-12-28_22-37-40_093785_577/runtime_resources/working_dir_files/_ray_pkg_3dd0fddcec28ee00/train.py", line 106, in <module>
2025-12-28 15:13:51

b200-202…4kot98i7
    train(args)
2025-12-28 15:13:51

b200-202…4kot98i7
  File "/tmp/ray/session_2025-12-28_22-37-40_093785_577/runtime_resources/working_dir_files/_ray_pkg_3dd0fddcec28ee00/train.py", line 79, in train
2025-12-28 15:13:51

b200-202…4kot98i7
    ray.get(actor_model.async_train(rollout_id, rollout_data_ref))
2025-12-28 15:13:51

b200-202…4kot98i7
  File "/usr/local/lib/python3.12/dist-packages/ray/_private/auto_init_hook.py", line 22, in auto_init_wrapper
2025-12-28 15:13:51

b200-202…4kot98i7
    return fn(*args, **kwargs)
2025-12-28 15:13:51

b200-202…4kot98i7
           ^^^^^^^^^^^^^^^^^^^
2025-12-28 15:13:51

b200-202…4kot98i7
  File "/usr/local/lib/python3.12/dist-packages/ray/_private/client_mode_hook.py", line 104, in wrapper
2025-12-28 15:13:51

b200-202…4kot98i7
    return func(*args, **kwargs)
2025-12-28 15:13:51

b200-202…4kot98i7
           ^^^^^^^^^^^^^^^^^^^^^
2025-12-28 15:13:51

b200-202…4kot98i7
  File "/usr/local/lib/python3.12/dist-packages/ray/_private/worker.py", line 2961, in get
2025-12-28 15:13:51

b200-202…4kot98i7
    values, debugger_breakpoint = worker.get_objects(
2025-12-28 15:13:51

b200-202…4kot98i7
                                  ^^^^^^^^^^^^^^^^^^^
2025-12-28 15:13:51

[...]

PR #876, opened 2026-04-03
Fix Qwen3 Next MTP bridge block-spec wiring and add regression coverage

taivu1998:
## Summary

Addresses radixark/miles#366.

This PR fixes the missing Qwen3 Next bridge-side MTP block-spec wiring and adds dedicated regression coverage for Qwen3 Next MTP bridge/raw-export behavior.

## Root Cause

Qwen3 Next already had partial MTP support in Miles:

- MTP config was exposed through `num_nextn_predict_layers`
- direct MTP tensors were mapped in `miles_plugins/mbridge/qwen3_next.py`
- raw Megatron-to-HF export already handled Qwen3 Next MTP tensor names

However, unlike the existing Qwen3.5 and MiMo bridge implementations, the Qwen3 Next bridge did not inject `mtp_block_spec` from `_get_gptmodel_args()`. That left Qwen3 Next without bridge-side MTP model-construction parity.

The repo also had no dedicated Qwen3 Next MTP bridge regression tests, which made this gap easy to miss.

## Changes

- add `get_gpt_mtp_block_spec` import to `miles_plugins/mbridge/qwen3_next.py`
- cache the active transformer layer spec in `Qwen3NextBridge._get_transformer_layer_spec()`
- add `Qwen3NextBridge._get_gptmodel_args()` so bridge mode injects `mtp_block_spec` when `num_nextn_predict_layers` is present
- tighten the `mtp_num_layers` config check to use `getattr(..., None)` instead of a bare `hasattr(...)`
- add `tests/test_qwen3_next_mtp_bridge_mapping.py` covering:
  - direct MTP tensor name mappings
  - transformer-layer attention remapping under `mtp.layers.*`
  - transformer-layer dense/MoE MLP remapping
  - bridge-side `mtp_block_spec` propagation
  - current `eh_proj` passthrough behavior
  - raw export consistency for direct and transformer-layer MTP tensors

## Validation

Ran locally in the isolated worktree:

- `python3.12 -m py_compile miles_plugins/mbridge/qwen3_next.py tests/test_qwen3_next_mtp_bridge_mapping.py`
- `git diff --check`
- direct Python 3.12 semantic validation script covering:
  - Qwen3 Next bridge-side `mtp_block_spec` injection
  - Qwen3 Next MTP direct / attention / MLP remapping
  - raw-export reuse of decoder mapping for MTP tensors

## Notes

- This PR is intentionally narrow. It fixes bridge-construction parity and adds regression coverage; it does not change Qwen3 Next `eh_proj` layout semantics or add model-specific numerical tuning.
- I could not run the full pytest path in this local environment because the repo requires Python `>=3.10`, while the locally available Python with `torch` installed is older; the available Python 3.12 environment here does not have `torch` installed.

### DIFF
--- a/miles_plugins/mbridge/qwen3_next.py
+++ b/miles_plugins/mbridge/qwen3_next.py
@@ -1,4 +1,6 @@
 import torch
+from megatron.core.models.gpt.gpt_layer_specs import get_gpt_mtp_block_spec
+
 from mbridge.core import register_model
 from mbridge.models import Qwen2MoEBridge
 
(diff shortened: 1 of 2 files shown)


taivu1998:
Hi @miles-code-angel, could you help review this PR? Thanks!

behroozazarkhalili:
Verified on top of current main (2ed189c0d); 1104 commits behind but merges with no conflicts.

- tests/test_qwen3_next_mtp_bridge_mapping.py on the merge: 9 passed. With miles_plugins/mbridge/qwen3_next.py reverted to main, 1 of the 9 fails, so the regression test pins the wiring change.
- ruff/black/isort at the pre-commit pins are clean on both files.

One blocking detail: the test sits at tests/ root, and CI only discovers tests/fast, tests/fast-gpu, tests/e2e and tests/ci (tests/ci/file_run.py L17, ci_register.py L271), so as placed it never runs on a PR. Moving it under tests/fast/plugins/ (no register call needed there) makes it a real guard.

cc @yueming-yuan @Zhichenzzz

PR #885, opened 2026-04-04
feat: real cp support with relayout fix for qwen3.5 train/rollout mismatch

### LINKED ISSUE #878: [Feature] Support real CP for Qwen 3.5
Right now, when using CP with Qwen 3.5, the GDN layer does not truly support CP but instead duplicates the computation. For example, with CP8, the computation will be duplicated 8 times.

This feature is available upstream in flash-linear-attention: https://github.com/fla-org/flash-linear-attention/pull/691

Example implementation: https://github.com/PrimeIntellect-ai/prime-rl/pull/2080

Zhichenzzz:
## Summary

This PR addresses #878 , adding real Context Parallelism (CP) support and fixing a train/rollout mismatch bug that was overlooked in [PrimeIntellect prime-rl#2080](https://github.com/PrimeIntellect-ai/prime-rl/pull/2080).

## Problem

When using hybrid linear attention with CP, there is a silent but critical layout mismatch between training and rollout: `flash-linear-attention` uses a **zig-zag layout** for CP, while Megatron uses a **packed layout**. The approach taken in prime-rl #2080 did not account for this incompatibility, resulting in undetected logprob divergence during training.

## Solution

We identified this root cause and implemented proper **shard and reshard** operations to correctly bridge the zig-zag ↔ packed layout between `flash-linear-attention` and Megatron under CP, ensuring full consistency between train and rollout.

## Results

After 10 training steps, logprob diff across configurations:

| Config | logprob_diff |
|---|---|
| CP=1, EP=4 | 0.0121 |
| CP=2, EP=4 | 0.0113 |
| CP=2, EP=8 | 0.0105 |

With the shard/reshard fix applied, logprob diff is further reduced to **~0.001**. Thanks @guapisolo for the great help!

## References
- Closes #878
- Related (flawed approach): [PrimeIntellect/prime-rl#2080](https://github.com/PrimeIntellect-ai/prime-rl/pull/2080)


miles/utils/arguments.py
```diff
@@ -2088,6 +2088,7 @@ def equal(x, y):
         if "rope_theta" in hf_config.rope_parameters:
             hf_config.rope_theta = hf_config.rope_parameters["rope_theta"]
 
+    is_moe = hasattr(hf_config, "num_experts") or hasattr(hf_config, "moe_intermediate_size")
```
Zhichenzzz: @guapisolo I added this line to ensure the qwen 3.5 hf_config intermediate_size=5632 not equal with ffn_hidden_size

guapisolo: Let me double check.

guapisolo: I think we can remove this line? This line is not related to your motivation.

Zhichenzzz: Curious that around few weeks before when I integrated qwen3.5, there was no assert errors here. (what about you I believe you also tested it before) is it from transformer version update or sth else?

miles/utils/arguments.py
```diff
@@ -2102,6 +2103,8 @@ def equal(x, y):
         ("rope_theta", "rotary_base", equal),
     ]:
         if hasattr(hf_config, hf_config_name):
+            if is_moe and hf_config_name == "intermediate_size":
```
guapisolo: On my side this line might be redundant? https://huggingface.co/Qwen/Qwen3.5-35B-A3B/blob/main/config.json . There is no `intermediate_size=5632` in qwen3.5

Zhichenzzz: oh i found it... we should update into` pip install transformers==5.2.0`.

guapisolo: I add a walkaround to avoid this under transformers 4.57.1

guapisolo (approved):
LGTM. On my side, Qwen-3.5-35b-a3b, cp=2 logprob diff 0.013
Need code owner review.

scripts/run-qwen3.5-35B-A3B-mtp-cp2-ep8.sh
yueming-yuan, line 1: nit: can we use `.py` style script instead?

miles/backends/training_utils/cp_utils.py
```diff
@@ -336,3 +345,52 @@ def slice_log_prob_with_cp(
...
+    if cu_seqlens is None or cu_seqlens.numel() < 2:
+        raise ValueError(f"Hybrid CP requires valid cu_seqlens (at least 2 elements) but got {cu_seqlens}")
+    return _fla_build_cp_context(
+        cu_seqlens=cu_seqlens.to(device=device, dtype=torch.int32),
+        group=cp_group,
+        conv1d_kernel_size=module.conv_kernel_size,
+    )
+
+
+def detect_and_setup_hybrid_cp(
```
yueming-yuan: nit: for this function, I would recommend to put at somewhere inside `miles_plugin` given that it's coupled with the implementation there

miles/backends/megatron_utils/actor.py
```diff
@@ -110,6 +110,17 @@ def init(
             args, role
         )
 
+        from megatron.core import mpu
```
yueming-yuan: nit: we usually don't want to import mpu inside these files.. maybe use `parallel_state` instead?

yueming-yuan (approved):

PR #893, opened 2026-04-05
Fix: truncate rollout_routed_experts in _truncate_sample_output

### LINKED ISSUE #861: `_truncate_sample_output` does not truncate `rollout_routed_experts`
## Bug

`_truncate_sample_output()` in [`openai_endpoint_utils.py`](https://github.com/radixark/miles/blob/cc758db65423ba8507e49a32a5691150ed79260a/miles/rollout/generate_utils/openai_endpoint_utils.py#L196-L208) truncates `tokens`, `rollout_log_probs`, and `loss_mask`, but does not truncate `rollout_routed_experts`.

After truncation, `len(sample.tokens) - 1 != rollout_routed_experts.shape[0]`, which causes `Sample.validate()` to fail with an assertion error.

## When it triggers

`agentic_tool_call.generate()` calls `truncate_samples_by_total_tokens(samples, max_seq_len, tokenizer)`, which calls `_truncate_sample_output()` when a sample's total token count exceeds `max_seq_len`. This is a realistic scenario in long multi-turn agentic sessions with `--use-rollout-routing-replay` enabled.

## Comparison

`Sample.strip_last_output_tokens()` in [`types.py`](https://github.com/radixark/miles/blob/cc758db65423ba8507e49a32a5691150ed79260a/miles/utils/types.py#L170-L185) handles this correctly:

```python
if self.rollout_routed_experts is not None:
    self.rollout_routed_experts = self.rollout_routed_experts[:-n]
```

## Suggested fix

Add the corresponding truncation in `_truncate_sample_output`:

```python
if sample.rollout_routed_experts is not None:
    prompt_experts = prompt_len - 1
    sample.rollout_routed_experts = sample.rollout_routed_experts[:prompt_experts + keep_tokens]
```

(`rollout_routed_experts` has shape `(len(tokens) - 1, num_layers, moe_router_topk)`, so the slicing accounts for the offset.)

DavidBellamy:
## Summary
Add missing `rollout_routed_experts` truncation in `_truncate_sample_output()`, fixing `Sample.validate()` assertion failures when agentic sessions exceed `max_seq_len` with `--use-rollout-routing-replay` enabled.

Closes #861.

## Context
`_truncate_sample_output()` truncates `tokens`, `rollout_log_probs`, and `loss_mask`, but not `rollout_routed_experts`. After truncation, `len(tokens) - 1 != rollout_routed_experts.shape[0]`, which fails validation. `Sample.strip_last_output_tokens()` in `types.py` already handles this correctly; this change adds the same logic to the truncation path.

## Changes
- `miles/rollout/generate_utils/openai_endpoint_utils.py`: Truncate `rollout_routed_experts` to `prompt_len - 1 + keep_tokens` (matching the `(len(tokens) - 1, ...)` shape convention)

### DIFF
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -228,4 +228,8 @@ def _truncate_sample_output(sample: Sample, keep_tokens: int, tokenizer) -> None
         sample.rollout_log_probs = sample.rollout_log_probs[:keep_tokens]
     if sample.loss_mask is not None:
         sample.loss_mask = sample.loss_mask[:keep_tokens]
+    if sample.rollout_routed_experts is not None:
+        # rollout_routed_experts has shape (len(tokens) - 1, ...), so slice to
+        # match the new total token count minus one.
+        sample.rollout_routed_experts = sample.rollout_routed_experts[:prompt_len - 1 + keep_tokens]
     sample.status = Sample.Status.TRUNCATED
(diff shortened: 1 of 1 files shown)


guapisolo:
Could you pick up the suggestion in https://github.com/radixark/miles/issues/861. Thanks.

PR #896, opened 2026-04-05
Lora update weight refactory

gongyisheng:
current problem:
when using LoRA finetuning, colocate mode, we always have `enable_weights_cpu_backup` to be True.
The base weights are moved between CPU and GPU when onload/offload sglang engine, it's slow for big models (>32B)

Fix it by sync both base weight and lora weight when using colocate mode, cuda ipc is faster than cpu/gpu pcie transmission. For disaggregate mode, the behavior is not changed.

sglang side fix: 
- \[main\]: https://github.com/sgl-project/sglang/pull/22169
- \[sglang-miles\]: https://github.com/sgl-project/sglang/pull/22846


gongyisheng:
@yushengsu-thu this PR is ready for review.

gongyisheng:
observed that LoRA RL train time and update_weight time is larger than full parameter RL, need to further tune the performance. 
image

gongyisheng:
after refactory, the update_weight time for qwen2.5-3B lora RL (colocate mode) reduced from ~5s to ~0.3s, same as full param RL (colocate mode)
disaggragate mode is not implemented, but should also benefit from the change.
image
train step time reduced from 30s to 21s, a X1.5 speed up
image

yushengsu-thu (changes requested):
need to pass the rest of 2 ci

yushengsu-thu (approved):
off-line sync each line of codes

PR #909, opened 2026-04-06
Explicitly error when wanting multiple updatable servers

fzyzcjy:


miles/ray/rollout/rollout_manager.py
```diff
@@ -217,10 +217,12 @@ def recover_updatable_engines(self) -> None:
...
-        for srv in self.servers.values():
-            if srv.update_weights:
-                return srv
-        return None
+        updatable = [srv for srv in self.servers.values() if srv.update_weights]
+        assert len(updatable) <= 1, (
+            f"Multiple servers have update_weights=True: {[srv.model_name for srv in updatable]}. "
+            f"Only one updatable server is supported."
+        )
+        return updatable[0] if updatable else None
```
Shi-Dong: Style-wise I prefer the following
```python
match updatable:
    case []:
        return None
    case [x]:
        return x
    case _:
        raise ValueError("blablabla")
```

Shi-Dong (approved):

miles/ray/rollout/rollout_manager.py
fzyzcjy: yes that would be more FP, let me update it later

PR #942, opened 2026-04-07
Allow engine recovery to be executed in parallel with generation and wait for aliveness

fzyzcjy:


miles/ray/rollout/rollout_server.py
```diff
@@ -209,3 +209,12 @@ async def onload(self, tags: list[str] | None = None):
 
     async def check_weights(self, action: str):
         return await asyncio.gather(*[g.check_weights(action=action) for g in self.server_groups])
+
+    async def wait_all_engines_alive(self, timeout: float = 600):
```
Shi-Dong, line 217: I wonder whether it is a good idea to hardcode the timeout limit to 600 seconds.  Perhaps add a todo here to flag it?

PR #949, opened 2026-04-07
fix: handle non-tool appended messages in TITO incremental tokenization

guapisolo:
## Summary

This PR fixes TITO incremental tokenization for non-assistant appends by replacing the old single dummy-diff approach with role-aware segmentation.

Previously, append tokenization could become unstable when appending `user`/`system` messages (and mixed `tool` + non-tool sequences), because boundary tokens were inferred from a single synthetic context.  
Now we tokenize appended content segment-by-segment using role-specific synthetic prefixes.

## What Changed

- Refactored `tokenize_additional_non_assistant` in `tito_tokenizer.py` to a role-segmented pipeline.
- Added segment splitting rules:
  - contiguous `tool` messages are grouped and tokenized together. use `[dummy_system, dummy_assistant]` in additional tokenzer without `dummy_user` to avoid any cut think issue across models.
  - each `user` and `system` message is tokenized as a singleton segment, and use `[dummy_system]` in additional tokenzer
- Keep Qwen-3.5 template ban `system` message behavior, and revert previous modifications.
- Modify Qwen-3.5 chat template logic to skip no user message check.

### Updated unit tests (`test_tito_tokenizer.py`)
- Add coverage for `user` message.
- Add coverage for intermediate system check, which can detect templates that ban intermediate system like qwen3.5
- `test_tito_tokenizer_model_matrix.py` run many additional tokenize checks across different models.

### New model-matrix tests (`test_tito_tokenizer_model_matrix.py`)
- Added cross-model matrix cases and existing failure explanation.
- Assertions focus on using `TokenSeqComparator` to check mismatches except assistant-text diff.

### DIFF
--- a/miles/utils/chat_template_utils/tito_tokenizer.py
+++ b/miles/utils/chat_template_utils/tito_tokenizer.py
@@ -5,11 +5,17 @@
 token sequence, then merges them with the pretokenized prefix — handling
 model-specific boundary tokens at the junction.
 
-The default implementation uses a dummy-message diff: it tokenizes a
-synthetic ``[dummy_user, dummy_assistant]`` base with and without the
-appended messages, then takes the suffix difference as the incremental IDs.
-Model-specific subclasses override ``merge_tokens`` to handle boundary
-quirks at the junction.
+The default implementation incrementally tokenizes appended non-assistant turns
+with role-specific synthetic prefixes:
+
+- contiguous ``tool`` runs use ``[dummy_system, dummy_assistant]``
+- each ``user`` message uses ``[dummy_system, dummy_user]``
+- each ``system`` message uses ``[dummy_system]``
+
+The appended suffix is processed left-to-right, then the generation prompt for
+the next assistant turn is appended once at the end.  Model-specific
+subclasses only override ``merge_tokens`` for boundary quirks at the prefix
+junction.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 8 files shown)


miles/utils/chat_template_utils/tito_tokenizer.py
```diff
@@ -23,25 +23,30 @@
...
-        "content": "",
-        "reasoning_content": " ",
-        "tool_calls": [
+def _build_dummy_assistant(appended_messages: list[dict[str, Any]]) -> dict[str, Any]:
+    """Build a dummy assistant message for the dummy-prefix diff.
+
+    All ``tool`` messages in *appended_messages* (not just leading contiguous
+    ones) get matching ``tool_calls``.  If there are no tool messages the dummy
+    assistant has no ``tool_calls`` — so the template renders the correct
+    turn-transition tokens (e.g. ``<|user|>`` instead of ``<|observation|>``).
```
maocheng23: For my sanity check, now our TITO would only support GLM style? Can we still keep it flexible somehow?

If it would be too hard to support Qwen3 chat template, I think it's still good here

maocheng23: And we don't need `reasoning_content` any more?

guapisolo: > For my sanity check, now our TITO would only support GLM style? Can we still keep it flexible somehow?
> 
> If it would be too hard to support Qwen3 chat template, I think it's still good here

This walkaround for glm 4.7 break qwen impl. Need fix to qwen3 chat template. Unblock TITO dev.

miles/utils/chat_template_utils/tito_tokenizer.py
```diff
@@ -5,11 +5,17 @@
...
 
-The default implementation uses a dummy-message diff: it tokenizes a
-synthetic ``[dummy_user, dummy_assistant]`` base with and without the
-appended messages, then takes the suffix difference as the incremental IDs.
-Model-specific subclasses override ``merge_tokens`` to handle boundary
-quirks at the junction.
+The default implementation incrementally tokenizes appended non-assistant turns
+with role-specific synthetic prefixes:
+
+- contiguous ``tool`` runs use ``[dummy_system, dummy_assistant]``
```
guapisolo, line 11: Remove user in tool to avoid boundary issue around think.

jybsuper (approved):
LGTM

maocheng23 (approved):

PR #953, opened 2026-04-07
feat: add session record disk logging (RecordLogger)

guapisolo:
## Summary
- Add `RecordLogger` class that writes `SessionRecord` objects as one-JSON-per-line to per-session files under `--session-record-log-dir`
- Records are flushed immediately so partial sessions survive crashes
- Integrate into `SessionRegistry`: records are written on each chat completion and file handles are closed when sessions are deleted
- Add `--session-record-log-dir` CLI argument (disabled when not set)

## Test plan
- [ ] Verify session records are written to disk when `--session-record-log-dir` is set
- [ ] Verify no logging occurs when the flag is not set
- [ ] Verify file handles are properly closed on session deletion

### DIFF
--- a/miles/rollout/session/record_logger.py
+++ b/miles/rollout/session/record_logger.py
@@ -1,53 +1,105 @@
 """Disk-based JSONL logger for session records.
 
 Each session gets its own file: ``<log_dir>/<session_id>.jsonl``.
+A dedicated background thread handles serialization and disk I/O so that
+callers (typically async request handlers) never block on file operations.
 Records are flushed immediately so that partial sessions are preserved
 if the process crashes.
 """
 
 import json
 import logging
+import threading
 from pathlib import Path
+from queue import SimpleQueue
 from typing import IO
 
 from miles.rollout.session.session_types import SessionRecord
 
 logger = logging.getLogger(__name__)
 
+# Sentinel object to signal the writer thread to shut down.
+_SHUTDOWN = object()
+
 
 class RecordLogger:
-    """Writes ``SessionRecord`` objects as one-JSON-per-line to per-session files."""
+    """Writes ``SessionRecord`` objects as one-JSON-per-line to per-session files.
+
+    All disk I/O runs on a single background daemon thread, making
+    :meth:`log_record` and :meth:`close_session` non-blocking.
+    """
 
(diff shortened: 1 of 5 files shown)


miles/rollout/session/record_logger.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def log_record(self, session_id: str, record: SessionRecord) -> None:
+        """Enqueue a record to be written. Returns immediately."""
+        self._queue.put(("write", session_id, record))
+
+    def close_session(self, session_id: str) -> None:
+        """Enqueue a session-close event (flushes and closes the file handle)."""
+        self._queue.put(("close", session_id, None))
+
+    def close_all(self) -> None:
```
maocheng23, line 41: `close_all` here is not called other than test?

Do you want to add an atexit with close all?

e.g. 
```
import atexit

def __init__(self, log_dir: str):
    ...
    self._thread.start()
    atexit.register(self.close_all)
```

miles/rollout/session/sessions.py
```diff
@@ -209,6 +209,7 @@ async def chat_completions(request: Request, session_id: str):
                     response=response,
                 )
                 session.append_record(record)
+                registry.log_record(session_id, record)
```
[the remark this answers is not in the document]
maocheng23: Here I think we dont need asyncio, the record logging should be very light-weighted? Not sure

miles/rollout/session/record_logger.py
```diff
@@ -0,0 +1,105 @@
+"""Disk-based JSONL logger for session records.
+
+Each session gets its own file: ``<log_dir>/<session_id>.jsonl``.
+A dedicated background thread handles serialization and disk I/O so that
+callers (typically async request handlers) never block on file operations.
+Records are flushed immediately so that partial sessions are preserved
+if the process crashes.
+"""
```
maocheng23, line 6: Can you refactor this file? I think the logic is good but a little hard to follow.

maocheng23 (changes requested):

miles/rollout/session/record_logger.py
guapisolo: Okay. Let me check in more detail

PR #971, opened 2026-04-10
chore: remove deterministic configs from lora example

gongyisheng:
Following configs are used for deterministic inference, set these env configs will have performance impact, not needed for general lora training:
- CUBLAS_WORKSPACE_CONFIG
- NVTE_ALLOW_NONDETERMINISTIC_ALGO
- NCCL_ALGO: "Ring"


gongyisheng:
@yushengsu-thu The configs are inside a dumped json, I cannot comment them out. Remove them instead.

yushengsu-thu (commented):
LFTM, but we only need to ensure the functional part works, and regarding training stability, we can defer it to the model training phase.

PR #976, opened 2026-04-13
Finalize all weight transfer P2P examples 

JD-ETH:
Final Fix for Enabling Kimi and GLM5. 

- directly allocate replica model on CPU 
- force skip `post_weight_load`, which is a GPU-dependent requantization process for DeepSeek model with MLA.  It does not affect result because the quantized tensors are not part of `model_parameters` and needs to be requantized on the rollout side after weight transfer completes. 

All validation successful. 

image


docs/en/advanced/p2p-weight-transfer.md
```diff
@@ -26,3 +26,86 @@ P2P mode addresses this by having each training rank transfer only the specific
...
+
+## Profiling Results
+
+All profiling is run on H100-80GB clusters.
+Timing measures `update_weights_implementation` (the actual weight transfer), averaged over
+steady-state steps 3–12 (steps 0–2 are warmup).
+Consolidated timer logs are stored in
+`/mnt/vast/checkpoints/jiadongguo/rdma/logs/profile-for-release/<model>/{broadcast,p2p}/`.
+
+| Model Family | Model Name | Total Param | sglang Model Class | Train Config | Inference Config | NCCL (ms) | RDMA (ms) | Delta |
```
JensenFire, line 110: it would be better to highlight whether the models are of MoE archs, where the p2p mode shows its potential. 



Shall we also point out that why p2p mode performs not that good in some case, where the training/rollout scenarios are the single node case with `pp_size = 1`, meaning that the nccl broadcast will be relatively fast and the replica loading in p2p node becomes something non-negligible.

JensenFire: and the key word `WS` in the column `Inference Config` is kinda confusing. I think it's SGLangs'  Tp?

JD-ETH: good tip, thanks!

examples/p2p_weight_transfer/run.py
```diff
@@ -0,0 +1,1151 @@
...
+    if not is_single_node:
+        print(f"  Node rank  : {node_rank}, Head: {head_ip}")
+    print("=" * 60)
+    print()
+
+    # --- Cleanup stale processes (exclude self) ---
+    my_pid = os.getpid()
+    my_ppid = os.getppid()
+    # Kill stale python/ray but spare this process tree
+    pkill_python = f"pgrep -x python3 python | grep -v -w {my_pid} | grep -v -w {my_ppid} | xargs -r kill -9 || true"
```
Zhichenzzz: `pgrep -x python3 python` -> `pgrep -x 'python|python3'`

examples/p2p_weight_transfer/run.py
```diff
@@ -0,0 +1,1151 @@
...
+    # --- Submit Ray job (head node only, or single-node) ---
+    if is_single_node or node_rank == 0:
+        import shlex
+
+        args_str = " ".join(shlex.quote(a) for a in args)
+        run_cmd(
+            f"{model_args_source} && "
+            f"ray job submit --address='http://127.0.0.1:8265' "
+            f"--runtime-env-json='{runtime_env_json}' "
+            f"-- python3 train.py "
```
Zhichenzzz: `f"-- python3 \"{MILES_ROOT}/train.py\" "`

.gitignore
```diff
@@ -193,3 +193,9 @@ _examples_synced/
 .env
 .DS_Store
 .claude/
+
+# Slurm job scripts (local, not tracked)
+scripts/slurm/
+
+# Profiling devlog (local, not tracked)
+devlog-profiling.md
```
Zhichenzzz: Could you handle this ignore locally instead for the whole repo?

Zhichenzzz (changes requested):

JD-ETH:
addressed, thank you!

Zhichenzzz:
LGTM! Thank you @JD-ETH for the great job!

Zhichenzzz (approved):

yueming-yuan (approved):

PR #986, opened 2026-04-15
GLM47 full cmd for async and sync reasoning

maocheng23:


yushengsu-thu (approved):
LGTM but might need to
1. pre-commit
and it's better to
2. run `run-ci-sglang` to ensure this change

PR #988, opened 2026-04-16
[lora] support disaggregate model lora training

gongyisheng:
This PR supports disaggregate and distributed model lora training.

sglang side PR: https://github.com/sgl-project/sglang/pull/27268


gongyisheng:
to check with @yushengsu-thu: 
- [update_weight_from_tensor] do we need to raise error when base weight is empty? currently it does not raise error but will cause degeneration problem.
- [update_weight_from_tensor] do we need to raise error when lora weight is chunked? currently load_lora_adapter_from_tensor does not support chunked requests, so it will cause error on sglang side, better to raise error + prompt user to raise update_weight buffer size.

gongyisheng:
compared colocate vs disaggregate lora: 
eval:
image
rollout:
image

gongyisheng:
distributed (broadcast) training done
eval: 
image
image
train:
image

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/p2p.py
yueming-yuan, line 1: if lora is not supported in p2p now, should we just `if use_lora: assert not p2p` in `argument.py`, and no need to init lora and define lora functions? (also there seems not to have strong motivation to support lora+p2p given lora transfer is lightweight)

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
```diff
@@ -210,19 +304,34 @@ def update_weights(self) -> None:
...
-            pbar = tqdm(desc=f"[{self._group_name}] Update weights", total=0) if self._is_source else None
-
-            self._gather_and_update_non_expert_weights(self._update_weight_implementation, pbar)
-            dist.barrier(group=get_gloo_group())
-            self._gather_and_update_expert_weights(self._update_weight_implementation, pbar)
-            dist.barrier(group=get_gloo_group())
+            # Base weight sync model:
+            #   full-param RL: base weights change every step -> always sync.
+            #   LoRA RL: base is frozen -> only sync once, on the first iteration.
+            if not (getattr(self, "is_lora", False) and self._lora_base_synced):
```
yueming-yuan: nit: we would prefer directly use `self.is_lora` instead of getattr for unconditional args

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/p2p.py
gongyisheng: @yueming-yuan updated, agree. p2p lora support looks not necessary for now

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
gongyisheng: @yueming-yuan updated.

yushengsu-thu (commented):
@gongyisheng 
Need to check these (potential issue):

 - p2p full-parameter training might crash with AttributeError (reproduced)

  - mixin.py:321 — if not (self.is_lora and self._lora_base_synced): reads self.is_lora unconditionally
  (miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py)
  - Root cause in p2p.py:48–64 — UpdateWeightP2P.__init__ only does assert not is_lora; it never sets
  self.is_lora and never calls _init_lora
  (miles/backends/megatron_utils/update_weight/update_weight_from_distributed/p2p.py)
  - Trigger path: actor.py:188 (passes is_lora=is_lora_enabled(args)) → actor.py:499 (calls
  update_weights()). For p2p + full-param, is_lora=False, so the assert passes but self.is_lora is
  never set → first weight sync hits line 321 → crash.

  Fix: add a class-level default is_lora: bool = False on the mixin (and simplify the getattr at
  mixin.py:330 back to self.is_lora).

gongyisheng:
> @gongyisheng Need to check these (potential issue):
> 
> * p2p full-parameter training might crash with AttributeError (reproduced)
> * mixin.py:321 — if not (self.is_lora and self._lora_base_synced): reads self.is_lora unconditionally
>   (miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py)
> * Root cause in p2p.py:48–64 — UpdateWeightP2P.**init** only does assert not is_lora; it never sets
>   self.is_lora and never calls _init_lora
>   (miles/backends/megatron_utils/update_weight/update_weight_from_distributed/p2p.py)
> * Trigger path: actor.py:188 (passes is_lora=is_lora_enabled(args)) → actor.py:499 (calls
>   update_weights()). For p2p + full-param, is_lora=False, so the assert passes but self.is_lora is
>   never set → first weight sync hits line 321 → crash.
> 
> Fix: add a class-level default is_lora: bool = False on the mixin (and simplify the getattr at mixin.py:330 back to self.is_lora).

@yushengsu-thu Agreed, fixed. please have a look

yushengsu-thu (approved):

yueming-yuan (approved):

PR #989, opened 2026-04-16
fix(rollout): guard round(None) in zero-std metric aggregation

DavidBellamy:
## Problem

`_compute_zero_std_metrics` in `miles/ray/rollout.py` crashes `RolloutManager.generate()` with a `TypeError` when any zero-std group's leading sample has a `None` reward:

```
ray.exceptions.RayTaskError(TypeError): ray::RolloutManager.generate()
  File "/root/miles/miles/ray/rollout.py", line 1266, in _compute_zero_std_metrics
    interesting_rewards = [str(round(g[0].get_reward_value(args), 1)) for g in interesting_sample_groups]
TypeError: type NoneType doesn't define __round__ method
```

This fires inside `_log_rollout_data`, **after** rollout collection and `dynamic_sampling_filter` have already accepted the batch. The trainer actor never receives the batch, so `optimizer.step()` never runs and async RL silently stalls.

Samples with `Status.ABORTED` carry `reward=None` (sandbox timeout, tool-invocation loops, agent LimitsExceeded, etc.), so any grouping of aborted trials where `_is_zero_std` returns True (including all-None groups) triggers this path on GRPO-style runs.

## Fix

Extract a small `_reward_label` helper that buckets `None` rewards under a dedicated `"none"` label instead of passing `None` to `round()`. Keeps the metric informative (`zero_std/count_none` shows the aborted-group count) and preserves existing behavior for numeric rewards.

## Test

Smoke-checked the helper locally against numeric, float-rounding, and `None` cases:

```python
assert _reward_label(MockSample(0.7)) == '0.7'
assert _reward_label(MockSample(0.0)) == '0.0'
assert _reward_label(MockSample(None)) == 'none'
assert _reward_label(MockSample(0.456)) == '0.5'
```

### DIFF
--- a/.github/workflows/maintain-deploy.yml
+++ b/.github/workflows/maintain-deploy.yml
@@ -60,10 +60,17 @@ jobs:
     if: needs.check.outputs.skip != 'true'
     runs-on: ubuntu-latest
     steps:
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          token: ${{ secrets.DEPLOY_TOKEN }}
+          token: ${{ steps.app-token.outputs.token }}
 
       - name: Configure git
         run: |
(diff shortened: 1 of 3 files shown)


guapisolo:
Thanks for reporting this issue. But close this since it's not a full refac.
A full refac is delivered in 2814, and add you as co-auth

PR #993, opened 2026-04-16
[OPD] [1/N] Decouple on-policy distillation from advantage estimators and add Megatron teacher mode

maocheng23:
## Summary

- **Refactors OPD to be orthogonal to advantage estimators**: OPD is no longer a separate `advantage_estimator` choice. Instead, it works as an additive KL penalty (`--use-opd --opd-kl-coef`) on top of any estimator (GRPO, PPO, REINFORCE++, etc.). This removes the old `on_policy_distillation` advantage estimator.
- **Adds Megatron-based teacher mode** (`--opd-type megatron`): Teacher model is loaded directly into Megatron using the same checkpoint mechanism as the reference model, computing teacher log-probs during the training forward pass. No external SGLang server needed when teacher shares architecture with the policy model.
- **Retains SGLang teacher mode** (`--opd-type sglang`): Existing flow where teacher log-probs come from an external SGLang server during rollout.

## Key changes

- `loss.py`: New `apply_opd_kl_to_advantages()` computes reverse KL (student - teacher) and adds it as a penalty to advantages in-place
- `actor.py`: Loads teacher checkpoint via `load_other_checkpoint("teacher", ...)` and runs teacher forward pass during training
- `arguments.py`: New `--use-opd`, `--opd-type`, `--opd-kl-coef`, `--opd-teacher-load`, `--opd-teacher-ckpt-step` args with validation
- `on_policy_distillation.py`: `post_process_rewards` now returns scalar rewards (0.0 for pure distillation) compatible with any advantage estimator
- `types.py`: Added `teacher_log_probs` field to `Sample`
- New example script `run-qwen3-8B-opd-megatron.sh` and updated README

## Test plan

- [ ] Run `run-qwen3-8B-opd.sh` (SGLang mode) end-to-end and verify training converges
- [ ] Run `run-qwen3-8B-opd-megatron.sh` (Megatron mode) end-to-end and verify teacher forward pass produces valid log-probs
- [ ] Verify `opd_reverse_kl` metric is logged correctly
- [ ] Test argument validation: `--use-opd` without `--opd-type`, `--opd-type megatron` without `--opd-teacher-load`, etc.
- [ ] Verify existing non-OPD training (GRPO, PPO) is unaffected

### DIFF
--- a/examples/on_policy_distillation/README.md
+++ b/examples/on_policy_distillation/README.md
@@ -1,17 +1,42 @@
 # On-Policy Distillation Example
 
-This example shows how to run **on-policy distillation** using miles. A small student (Qwen3-8B) is aligned to imitate a larger teacher (Qwen3-32B) by training only on the student's own rollouts and matching the teacher's token-level log-probabilities.
+This example shows how to run **on-policy distillation (OPD)** using miles. A small student (Qwen3-8B) is aligned to imitate a larger teacher (Qwen3-32B) by training only on the student's own rollouts and matching the teacher's token-level log-probabilities.
 
-In this example, the teacher model acts as a reward model (RM) by providing teacher log probabilities as the supervision signal.
+## Key Features
+
+- **OPD is orthogonal to advantage estimators**: OPD works as an additive KL penalty on top of any advantage estimator (GRPO, PPO, REINFORCE++, etc.), not as a separate estimator.
+- **Two teacher modes**:
+  - **sglang**: Teacher runs on an external SGLang server, teacher log-probs are obtained during rollout.
+  - **megatron**: Teacher is loaded directly into Megatron via `--opd-teacher-load`, teacher log-probs are computed during training forward pass.
+
+## Key Arguments
+
+| Argument | Description |
+|----------|-------------|
+| `--use-opd` | Enable on-policy distillation. Required flag to use OPD. |
+| `--opd-type` | Type of OPD: `sglang` or `megatron`. Required when `--use-opd` is set. |
+| `--opd-kl-coef` | OPD KL penalty coefficient (default: 1.0). |
+| `--opd-teacher-load` | Path to teacher checkpoint. **Required** when `--opd-type=megatron`, **must not be set** when `--opd-type=sglang`. |
+| `--opd-teacher-ckpt-step` | Optional checkpoint step for teacher model. |
+
+## Mode Comparison
+
+| Mode | Teacher Location | When to use |
+|------|------------------|-------------|
+| `sglang` | External SGLang server | Teacher has different architecture or larger than GPU memory |
+| `megatron` | Loaded into Megatron training | Teacher has same architecture as policy/ref model |
 
 ## Components
 
-- `on_policy_distillation.py` implements::
+- `on_policy_distillation.py` implements (for SGLang mode):
   - `reward_func` calls the teacher server (via `args.rm_url`) with every sample to obtain token-level logprobs.
   - `post_process_rewards` trims the teacher logprobs to the generated response span and writes the tensors back to each `Sample` to compute advantages.
 - `run-qwen3-8B-opd.sh` launches an SGLang teacher server, then submits a Ray job that runs `train.py`.
+- `run-qwen3-8B-opd-megatron.sh` uses Megatron-loaded teacher model (no external server needed).
 
 ## Running the example
+
+### Using SGLang Teacher (External Server)
 
 1. Download or prepare the required checkpoints and data.
 ```bash
@@ -30,9 +55,36 @@ PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py \
     --hf-checkpoint /root/Qwen3-8B \
     --save /root/Qwen3-8B_torch_dist
 ```
-3. run on-policy distillation:
+
+3. Run on-policy distillation:
 ```bash
 bash examples/on_policy_distillation/run-qwen3-8B-opd.sh
+```
+
+### Using Megatron Teacher (No External Server)
+
+1. Prepare student checkpoint (same as above).
+
+2. **IMPORTANT**: Convert your teacher model to Megatron format (change the path to your actual teacher):
+```bash
+# This example uses the same model as both student and teacher (for demonstration only)
+# In practice, use a different (stronger) model as the teacher!
+cd /root/miles
+source scripts/models/qwen3-8B.sh  # Or your teacher model config
+
+PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py \
+    ${MODEL_ARGS[@]} \
+    --hf-checkpoint /root/YourTeacherModel \
+    --save /root/YourTeacherModel_torch_dist
+```
+
+3. Edit `run-qwen3-8B-opd-megatron.sh` to update paths:
+   - Change `--opd-teacher-load` to your teacher model path
+   - Adjust `--opd-kl-coef` based on your task
+
+4. Run:
+```bash
+bash examples/on_policy_distillation/run-qwen3-8B-opd-megatron.sh
 ```
 
 
@@ -47,10 +99,25 @@ Using Qwen3-8B-Base model sfted on part of the [OpenThoughts3-1.2M](https://hugg
 
 
 
-
 # FAQ
-1. **Why are teacher logits computed via a sglang server instead of inside the training backend?**
-The teacher runs on an independent SGLang server that miles treats as a reward model. Hosting it inside Megatron/FSDP would require maintaining a second, fully configured training stack for the teacher.
+1. **Why are there two OPD modes?**
+   - `sglang` mode: The teacher runs on an independent SGLang server. This is useful when the teacher has a different architecture or is too large to load together with the policy model.
+   - `megatron` mode: The teacher is loaded into Megatron using the same parameter loading mechanism as the reference model. This requires the teacher to have the same architecture as the policy model.
+
+2. **How do I use Megatron-based teacher instead of SGLang server?**
+   Replace your OPD arguments:
+   ```bash
+   # Instead of:
+   --use-opd --opd-type sglang --opd-kl-coef 1.0
+   # Use:
+   --use-opd --opd-type megatron --opd-kl-coef 1.0 --opd-teacher-load /path/to/teacher_checkpoint
+   ```
+
+3. **What happens if I set wrong arguments?**
+   The system will raise clear errors:
+   - `--use-opd` without `--opd-type`: Error asking you to specify type
+   - `--opd-type megatron` without `--opd-teacher-load`: Error asking for teacher checkpoint
+   - `--opd-type sglang` with `--opd-teacher-load`: Error indicating conflict
 
 
 # References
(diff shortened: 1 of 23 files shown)


examples/on_policy_distillation/README.md
```diff
@@ -1,18 +1,43 @@
...
+
+| Argument | Description |
+|----------|-------------|
+| `--use-opd` | Enable on-policy distillation. Required flag to use OPD. |
+| `--opd-type` | Type of OPD: `sglang` or `megatron`. Required when `--use-opd` is set. |
+| `--opd-kl-coef` | OPD KL penalty coefficient (default: 1.0). |
+| `--opd-teacher-load` | Path to teacher checkpoint. **Required** when `--opd-type=megatron`, **must not be set** when `--opd-type=sglang`. |
+| `--opd-teacher-ckpt-step` | Optional checkpoint step for teacher model. |
+
+## Mode Comparison
```
Shi-Dong, line 22: I wonder whether using the `megatron` mode has certain intrinsic advantages over the `sglang` mode, e.g. perhaps the `megatron` mode produces more accurate teacher logprobs? If yes, I think it might be better to explain here in the doc. If no, I wonder whether we should just default to the sglang mode, since it offers more flexibility (allowing different architecture and larger teacher models).

examples/on_policy_distillation/README.md
```diff
@@ -1,18 +1,43 @@
...
+
+## Key Arguments
+
+| Argument | Description |
+|----------|-------------|
+| `--use-opd` | Enable on-policy distillation. Required flag to use OPD. |
+| `--opd-type` | Type of OPD: `sglang` or `megatron`. Required when `--use-opd` is set. |
+| `--opd-kl-coef` | OPD KL penalty coefficient (default: 1.0). |
+| `--opd-teacher-load` | Path to teacher checkpoint. **Required** when `--opd-type=megatron`, **must not be set** when `--opd-type=sglang`. |
+| `--opd-teacher-ckpt-step` | Optional checkpoint step for teacher model. |
```
Shi-Dong: I wonder what this flag is for: shouldn't `--opd-teacher-load` already specify the teacher ckpt?

miles/backends/training_utils/loss.py
```diff
@@ -276,6 +276,47 @@ def get_values(
...
+        raise ValueError(f"OPD with opd_type='{args.opd_type}' requires teacher_log_probs, but it is missing.")
+
+    device = student_log_probs[0].device
+    teacher_log_probs = [t.to(device=device) for t in teacher_log_probs]
+
+    reverse_kls = []
+    for i, adv in enumerate(advantages):
+        reverse_kl = student_log_probs[i] - teacher_log_probs[i]
+        advantages[i] = adv - args.opd_kl_coef * reverse_kl
+        reverse_kls.append(reverse_kl)
```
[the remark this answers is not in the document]
Shi-Dong: I think this comment is worth addressing. In addition, I think it'd be helpful to raise a `ValueError` if the length of `student_log_probs` is different from `teacher_log_probs` or `advantages`. If the lengths mismatch, the current implementation will not raise an alarm, which could lead to hard-to-debug consequences.

miles/utils/arguments.py
```diff
@@ -1844,6 +1891,37 @@ def miles_validate_args(args):
...
+                )
+            if not os.path.exists(args.opd_teacher_load):
+                raise FileNotFoundError(
+                    f"opd_teacher_load {args.opd_teacher_load} does not exist, please check the path."
+                )
+            if not os.path.exists(os.path.join(args.opd_teacher_load, "latest_checkpointed_iteration.txt")):
+                logger.info(
+                    f"opd_teacher_load {args.opd_teacher_load} does not have latest_checkpointed_iteration.txt, "
+                    "please make sure it is a valid megatron checkpoint directory."
+                )
```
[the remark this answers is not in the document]
Shi-Dong: Why not raise an error here?

examples/on_policy_distillation/README.md
```diff
@@ -1,18 +1,43 @@
...
+  - **sglang**: Teacher runs on an external SGLang server, teacher log-probs are obtained during rollout.
+  - **megatron**: Teacher is loaded directly into Megatron via `--opd-teacher-load`, teacher log-probs are computed during training forward pass.
+
+## Key Arguments
+
+| Argument | Description |
+|----------|-------------|
+| `--use-opd` | Enable on-policy distillation. Required flag to use OPD. |
+| `--opd-type` | Type of OPD: `sglang` or `megatron`. Required when `--use-opd` is set. |
+| `--opd-kl-coef` | OPD KL penalty coefficient (default: 1.0). |
```
Shi-Dong: For clarity, I'd suggest explaining here how this KL coefficient works, for example something like `opd_reward = task_reward - kl_coef * reverse_kl`.

miles/backends/megatron_utils/actor.py
```diff
@@ -579,6 +599,9 @@ def load_other_checkpoint(self, model_tag: str, path: str) -> None:
         if model_tag == "ref" and self.args.ref_ckpt_step is not None:
             self.args.ckpt_step = old_ckpt_step
 
+        if model_tag == "teacher" and self.args.opd_teacher_ckpt_step is not None:
+            self.args.ckpt_step = old_ckpt_step
+
```
Shi-Dong, line 555: I'm a bit confused by the combined effect of Lines 586-589 and Lines 602-604: seems that `self.args.ckpt_step` doesn't change at all?

examples/on_policy_distillation/README.md
maocheng23: Good question. Updated the README to explain that the main advantage of `megatron` mode is numerical consistency: teacher logprobs are computed in the same forward pass with the same kernels, dtypes, and tokenization as the student, which avoids the subtle drift you can get from a separate inference engine, and removes the second server. `sglang` mode is the right pick when the teacher has a different architecture or doesn't fit alongside the policy in training memory. Added a 'Which mode should I pick?' paragraph in d738992.

examples/on_policy_distillation/README.md
maocheng23: `--opd-teacher-load` points to the checkpoint *directory*, and `--opd-teacher-ckpt-step` optionally selects a specific iteration inside it (overriding `latest_checkpointed_iteration.txt`). This mirrors the existing `--ref-ckpt-step` flag. Clarified the description in d738992.

miles/backends/training_utils/loss.py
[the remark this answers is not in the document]
maocheng23: Agreed, both your point and gemini's. Added in d738992:
- A list-length check up front (`len(advantages) == len(student_log_probs) == len(teacher_log_probs)`).
- Per-sample shape checks for `student_log_probs[i]` vs `teacher_log_probs[i]`, and for `advantages[i]` vs `student_log_probs[i]` — the latter catches the GRPO-style scalar-broadcast trap gemini flagged.

miles/utils/arguments.py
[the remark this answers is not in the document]
maocheng23: Happy to tighten this — but flagging that the existing `ref_load` check directly above (line 1888) follows the same log-only pattern, so the new code is just mirroring the convention. Should I raise on missing `latest_checkpointed_iteration.txt` for both `opd_teacher_load` and `ref_load` in this PR for consistency, or keep both as warnings and tighten them in a follow-up? Happy either way.

examples/on_policy_distillation/README.md
maocheng23: Good call. Added the formula to the `--opd-kl-coef` row in d738992: `opd_advantage = base_advantage - opd_kl_coef * (student_logp - teacher_logp)`.

miles/backends/megatron_utils/actor.py
maocheng23: The save/restore is intentional, but you're right that it reads as a no-op at a glance. `load_checkpoint` (a few lines below) reads `self.args.ckpt_step` to decide which iteration to load, so we override it for the duration of that call and restore after. Added a 2-line comment above the override block in d738992 to make the intent obvious. Open to refactoring to a context manager if you'd prefer, but kept it minimal for this PR.

miles/utils/arguments.py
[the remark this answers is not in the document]
Shi-Dong: I'd recommend raising error for both `opd_teacher_load` and `ref_load`.

Shi-Dong (approved):
LGTM! Please address [this comment](https://github.com/radixark/miles/pull/993/changes#r3170871971) before merging.

Zhichenzzz (approved):
Great work! LGTM. Just a reminder of https://github.com/radixark/miles/actions/runs/25139515672/job/73685829929?pr=993 this CI failed.

miles/utils/types.py
Shi-Dong, line 160: Probably you also want to add an assertion for `teacher_log_probs` here?

Shi-Dong (commented):
I think you should also add the corresponding tests in `test_openai_endpoint_utils.py` for the extra `teacher_log_probs` field.

tests/fast/backends/training_utils/loss/test_loss_snapshot.py
yueming-yuan, line 1: suggest adding new OPD logic to the loss snapshot test (non-blocking, can be done later)

miles/backends/training_utils/loss.py
```diff
@@ -12,13 +12,72 @@
 from miles.utils.types import RolloutBatch
 
 
+def apply_opd_kl_to_advantages(
```
yueming-yuan: suggest to move this function into `loss_hub`, maybe `loss_hub/advantages.py` (?) or create a new file `loss_hub/opd.py`. We might want to keep this `loss.py` a lightweight entrypoint without  implementation details/utils

yueming-yuan (approved):

miles/utils/types.py
maocheng23: Good catch — added in 3150d875f. `Sample.validate()` now asserts `len(teacher_log_probs) == response_length` (mirroring the `rollout_log_probs` guard), and `strip_last_output_tokens` trims `teacher_log_probs` too so the assertion still holds after truncation.

I also added the merge-workflow coverage you asked for (same commit) in `tests/fast/rollout/generate_utils/test_openai_endpoint_utils.py`: `teacher_log_probs` propagation through `merge_samples` (zeros over the injected observation span), None-stays-None for non-OPD runs, and the new length-mismatch assertion surfaced via the merge.

miles/backends/training_utils/loss.py
maocheng23: Done in bb43049d9 — moved `apply_opd_kl_to_advantages` into a dedicated `loss_hub/opd.py`, so `loss.py` stays a lightweight entrypoint and just imports it for `compute_advantages_and_returns`.

tests/fast/backends/training_utils/loss/test_loss_snapshot.py
maocheng23: Added `tests/fast/backends/training_utils/loss/test_opd.py` in bb43049d9 with direct unit tests for the OPD loss path: weighted reverse-KL applied to advantages, the `opd_reverse_kl` metric, the `student_log_probs is None` short-circuit, and the length/shape guards.

I went with a focused unit test rather than a loss-*snapshot* case, since the snapshots live in the external `yueming-yuan/miles-artifacts` repo and a new case would need regenerating + committing there. Happy to add a snapshot case as a follow-up next time we refresh artifacts.

PR #994, opened 2026-04-17
[OPD] [2/N] Move to miles/rollout, add CI test and docs

maocheng23:
## Summary

Stacked on #993.

- Move `on_policy_distillation.py` from `examples/` to `miles/rollout/` as a first-class rollout component
- Update module path references in shell scripts and examples
- Add CI test (`test_qwen2.5_0.5B_opd_sglang.py`): Qwen2.5-0.5B self-distillation with SGLang OPD mode, 8 GPUs, bridge
- Add canonical OPD docs in `docs/advanced/on-policy-distillation.md` and link them from the current advanced docs index
- Support Rethinking OPD for SGLang mode via the weighted top-k reverse-KL reward path: `--opd-log-prob-top-k`, `--opd-top-k-strategy`, and `--opd-reward-weight-mode`
- Address review feedback by deduplicating the examples README into a docs pointer, citing the Thinking Machines OPD blog and Rethinking OPD paper, and tightening resource cleanup in the E2E test

## Test plan

- [x] `python -m py_compile miles/rollout/on_policy_distillation.py miles/backends/training_utils/loss.py miles/utils/types.py miles/ray/rollout.py miles/rollout/sglang_rollout.py miles/utils/arguments.py tests/fast/rollout/test_on_policy_distillation.py tests/e2e/short/test_qwen2.5_0.5B_opd_sglang.py`
- [x] Manual top-k helper assertions via `PYTHONPATH=. python - <<'PY' ...`
- [x] `python -m py_compile tests/e2e/short/test_qwen2.5_0.5B_opd_sglang.py`
- [x] `git diff --check`
- [x] `PYTHONPATH=. python tests/ci/run_suite.py --hw cpu --suite stage-a-fast --list-only`
- [x] `PYTHONPATH=. python tests/ci/run_suite.py --hw cuda --suite stage-b-short-8-gpu --list-only | rg "opd|Enabled|Hardware|Suite"`
- [ ] Run `test_qwen2.5_0.5B_opd_sglang.py` in GPU CI (label: `run-ci-short`)

Note: direct local `pytest tests/fast/rollout/test_on_policy_distillation.py -q` hung in this environment during collection with a protobuf mutex message, so I validated the helper assertions directly and verified CI registration/listing.


docs/advanced/on-policy-distillation.md
```diff
@@ -0,0 +1,111 @@
+# On-Policy Distillation
+
+On-policy distillation (OPD) enables a student model to learn from a larger teacher model by training on its own rollouts while matching the teacher's token-level log-probabilities. OPD is orthogonal to advantage estimators — it works as an additive KL penalty on top of any estimator (GRPO, PPO, REINFORCE++, etc.).
```
Shi-Dong: A few suggestions on this paragraph:

1. Remove "larger" in "larger teacher model": the teacher model isn't necessarily larger.
2. "training on its own rollouts while matching the teacher's token-level log-probabilities" is a bit vague. I'd recommend something like "the student model's reward is derived from the teacher model's token-level log-probabilities via reverse KL divergence."
3. "OPD is orthogonal to advantage estimators" sounds weird: OPD is a training recipe, whereas advantage estimator is a component in RL training, so they are not directly comparable.
4. Perhaps we should cite the blog post from Thinky here.

examples/on_policy_distillation/README.md
Shi-Dong, line 1: I think this README has a lot of overlap with `docs/en/advanced/on-policy-distillation.md`. I wonder whether we should just keep the copy under `docs`, and link to that copy in this doc.  In this way we don't have to maintain two docs when we update code.

docs/advanced/on-policy-distillation.md
```diff
@@ -24,6 +27,22 @@ Where $A_t$ is the original advantage from the base estimator (e.g., GRPO), $\la
...
+## Rethinking OPD Top-K Reward
+
+SGLang OPD supports the top-k token reward recipe from [Rethinking On-Policy Distillation](https://arxiv.org/abs/2604.13016). Set `--opd-log-prob-top-k` above zero to request student rollout top-logprobs, score the same sequence with the teacher, and aggregate a weighted reverse-KL estimate over a selected token set at each response position.
+
+The token set is controlled by `--opd-top-k-strategy`:
+
+| Strategy | Token set |
+|----------|-----------|
+| `only_stu` | Student top-k tokens, with teacher logprobs queried for those IDs. |
+| `only_tch` | Teacher top-k tokens, with student logprobs queried for those IDs. |
```
Shi-Dong: Very nit: I'd prefer full names `only-student` and `only-teacher` to the current ones. At the very least, we should use dash instead of underscore to make it consistent with `union-intersection`.

docs/advanced/on-policy-distillation.md
```diff
@@ -24,6 +27,22 @@ Where $A_t$ is the original advantage from the base estimator (e.g., GRPO), $\la
...
+| `intersection` | Tokens appearing in both top-k sets. |
+| `union` | Tokens appearing in either top-k set, with duplicates removed. |
+| `union-intersection` | Tokens appearing in exactly one top-k set. |
```
Shi-Dong: Perhaps `xor` would be a more intuitive name than `union-intersection`?

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+                    token_id,
+                    teacher_top_maps[i],
+                    teacher_on_student_maps[i],
+                    source="teacher",
+                )
+            )
+
+        weights = _reward_weights(student_logps, teacher_logps, weight_mode, normalize=normalize_weights)
+        reverse_kl = sum(
+            w * (s_logp - t_logp) for w, s_logp, t_logp in zip(weights, student_logps, teacher_logps, strict=False)
```
Shi-Dong: Still I'd recommend asserting that the zipped lists have equal lengths: `strict=False` is dangerous as it silently drops list mismatch bugs.

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+
+    ``--opd-log-prob-top-k>0`` follows the practical recipe from
+    "Rethinking On-Policy Distillation" by forming a top-k token set per
+    response position and storing a precomputed weighted reverse-KL estimate.
+    """
+    raw_rewards = [sample.get_reward_value(args) for sample in samples]
+    response_lengths = [sample.response_length for sample in samples]
+
+    if _get_opd_top_k(args) > 0:
+        for sample, reward in zip(samples, raw_rewards, strict=False):
```
Shi-Dong: Same for the zip occurrences in this function.

tests/fast/rollout/generate_utils/test_openai_endpoint_utils.py
Shi-Dong, line 264: Can you add an additional test for the `opd_student_top_logprobs` field in the metadata? Actually I think there might be issues with this field: it is merged via `_merge_equal_value` and might run into an assertion error, since student logprobs are different across samples.

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+        weights = _reward_weights(student_logps, teacher_logps, weight_mode, normalize=normalize_weights)
+        reverse_kl = sum(
+            w * (s_logp - t_logp) for w, s_logp, t_logp in zip(weights, student_logps, teacher_logps, strict=False)
+        )
+        reverse_kls.append(reverse_kl)
+
+    return torch.tensor(reverse_kls, dtype=torch.float32)
+
+
+async def reward_func(args, sample, **kwargs):
```
Shi-Dong: Might be better to have a return type annotation for this function.

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+        student_token_ids = _unique_ids(teacher_top)
+        reward_payload["student_on_teacher"] = await _post_json(
+            _student_score_url(args),
+            _score_payload(sample.tokens, token_ids=student_token_ids),
+        )
+
+    return reward_payload
+
+
+def post_process_rewards(args, samples: list[Sample], **kwargs):
```
Shi-Dong: Please also annotate return type here.

Shi-Dong (commented):
Mostly minor suggestions, but `opd_student_top_logprobs` might be a breaking issue.

Shi-Dong (approved):
LGTM!

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+                    teacher_on_student_maps[i],
+                    source="teacher",
+                )
+            )
+
+        weights = _reward_weights(student_logps, teacher_logps, weight_mode, normalize=normalize_weights)
+        reverse_kl = sum(
+            w * (s_logp - t_logp) for w, s_logp, t_logp in zip(weights, student_logps, teacher_logps, strict=True)
+        )
+        reverse_kls.append(reverse_kl)
```
Zhichenzzz, line 267: Could we change those two-for-loop into torch ops to save time?

miles/rollout/on_policy_distillation.py
```diff
@@ -0,0 +1,322 @@
...
+    student_top = _student_top_logprobs(sample, sample.response_length)
+
+    teacher_top_k = top_k if strategy != "only-student" else 0
+    teacher_token_ids = _unique_ids(student_top) if strategy in {"only-student", "union", "xor"} else None
+    teacher_payload = _score_payload(sample.tokens, top_k=teacher_top_k, token_ids=teacher_token_ids)
+    teacher_response = await _post_json(args.rm_url, teacher_payload)
+
+    reward_payload = {"teacher": teacher_response}
+    if strategy in {"only-teacher", "union", "xor"}:
+        teacher_top = _trim_input_field(teacher_response["meta_info"], "input_top_logprobs", sample.response_length)
```
Zhichenzzz: maybe reorganize these codes to make different strategies clear and avoid unused compute, e.g., only-teacher mode doesnt need student-top.

miles/backends/training_utils/loss_hub/opd.py
```diff
@@ -29,6 +29,32 @@ def apply_opd_kl_to_advantages(
...
 
+    precomputed_reverse_kls = rollout_data.get("opd_reverse_kl")
+    if precomputed_reverse_kls is not None:
+        if len(advantages) != len(precomputed_reverse_kls):
+            raise ValueError(
+                f"OPD length mismatch: advantages={len(advantages)}, "
+                f"opd_reverse_kl={len(precomputed_reverse_kls)}."
+            )
+
+        device = student_log_probs[0].device
```
Zhichenzzz: precomputed_reverse_kls[0].device or student_log_probs[0].device?

Zhichenzzz (approved):
most LGTM, some comments left.

miles/rollout/on_policy_distillation.py
maocheng23: Good point. I looked into this more closely. The final weighting/reduction could probably be partly tensorized, but the current hot path also includes strategy-specific token selection and fallback logprob lookup from SGLang’s returned top/token-id logprob structures.

I left the per-position aggregation loop as-is because fully tensorizing all strategies would require padding/masking ragged token sets for `intersection`/`union`/`xor`.

yueming-yuan (approved):

PR #1008, opened 2026-04-19
Add GenerateFnInput compatibility shim in sglang_rollout

Shi-Dong:
## Summary

Small compatibility shim so `generate_and_rm` can dispatch to the new single-argument `GenerateFnInput`-style custom generate functions (e.g. `miles/rollout/generate_hub/agentic_tool_call.py::generate`) in addition to the legacy 3-arg and 4-arg signatures.

Without the shim, any rollout wired to a `GenerateFnInput`-style function crashes immediately with:

```
TypeError: generate() takes 1 positional argument but 3 were given
```

because the old dispatcher always called the target as `custom_generate_func(args, sample, sampling_params[, evaluation=...])`.

The shim inspects the target's signature: if it has exactly one parameter, the four runtime values are packed into a `GenerateFnInput` dataclass and awaited as a single-arg call, and `output.samples` replaces `sample`. Otherwise it falls through to the existing 3-arg / 4-arg paths. Fully backwards-compatible with all existing custom generate functions.

## Test plan

- [x] Existing 3-arg and 4-arg custom generate functions still dispatch correctly (no behavior change — the `len(params) == 1` branch only fires for the new contract).
- [x] `miles/rollout/generate_hub/agentic_tool_call.py::generate` (single-arg `GenerateFnInput`) dispatches through the new branch without the `TypeError`, verified on a 2-node async agentic rollout against GLM-4.7-Flash on 2026-04-18 PT.

### DIFF
--- a/miles/rollout/sglang_rollout.py
+++ b/miles/rollout/sglang_rollout.py
@@ -15,7 +15,7 @@ from packaging.version import parse
 from tqdm import tqdm
 
 from miles.backends.megatron_utils.lora_utils import LORA_ADAPTER_NAME, is_lora_enabled
-from miles.rollout.base_types import RolloutFnEvalOutput, RolloutFnTrainOutput
+from miles.rollout.base_types import GenerateFnInput, RolloutFnEvalOutput, RolloutFnTrainOutput
 from miles.rollout.filter_hub.base_types import MetricGatherer, call_dynamic_filter
 from miles.utils import dumper_utils
 from miles.utils.async_utils import run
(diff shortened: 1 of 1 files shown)


Shi-Dong:
Superseded by #1016 which lands the same fix more cleanly: PR #1016 centralizes the legacy ↔ new `generate_function` adapter in `miles/rollout/inference_rollout/compatibility.py` (`LegacyGenerateFnAdapter` + `load_generate_function`), whereas this PR inlined the dispatch in `generate_and_rm`. PR #1016 is on main as commit 9a0036447. Closing.

PR #1009, opened 2026-04-19
fix(rollout/session): reseed instead of erroring when no assistant persisted

DavidBellamy:
When the agent retries with a divergent user message before the first assistant turn has been persisted, `_try_detect_and_rollback_to_assistant_checkpoint` raises `MessageValidationError` because the matched prefix contains no assistant checkpoint to roll back to. This kills trials that would otherwise recover on retry.

## Repro

Observed on two Harbor trials (`battery-charging-optimization` #6 and #7) aborted with:

```
litellm.BadRequestError: OpenAIException - Error code: 400 - {error: rollback failed: no assistant message found in the first 1 matched messages (stored has 2 messages, request has 3 messages)}
```

Chain of events:
1. Agent emits the first assistant turn, which hits `max_tokens` truncation and errors out before the response can be persisted.
2. Session state is therefore just `[system, user]` with `num_assistant=0`.
3. Agent retry path sends a new request whose user content diverges from the stored user message.
4. `match_len=1` (system matches, user diverges); no assistant in stored[:1], so `checkpoint_index=-1`, and the function raises.

## Fix

In the `checkpoint_index < 0` branch, if `num_assistant == 0` the stored state is a prompt-only seed whose first assistant turn never persisted. Clear the session and return `True` so the caller treats this as a fresh session. The existing raise path is preserved for cases where at least one assistant has been stored (rollback past an assistant turn remains disallowed).

The caller (`prepare_pretokenized`) early-returns `None` on the reseed signal, equivalent to the existing empty-session branch.

## Diff shape

- `_try_detect_and_rollback_to_assistant_checkpoint`: return type `None` -> `bool`
- New `num_assistant == 0` branch inside the existing `if checkpoint_index < 0:`
- Existing early returns get explicit `return False`
- `prepare_pretokenized` propagates the reseed signal

No behavior change for any case where at least one assistant has been persisted.

### DIFF
--- a/miles/rollout/session/linear_trajectory.py
+++ b/miles/rollout/session/linear_trajectory.py
@@ -77,7 +77,11 @@ class LinearTrajectory:
             return None
 
         # 1. Detect agent retries and roll back (at most one assistant step).
-        self._try_detect_and_rollback_to_assistant_checkpoint(request_messages)
+        #    Returns True iff the session was reseeded (see below); in that case
+        #    there is no stored history to merge against, so fall through to the
+        #    same None-return as an empty session.
+        if self._try_detect_and_rollback_to_assistant_checkpoint(request_messages):
+            return None
         # 2. Confirm the (possibly rolled-back) stored messages are a prefix of request,
         #    and that each appended message role is in tito_tokenizer.allowed_append_roles.
         try:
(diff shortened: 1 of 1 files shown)


guapisolo:
Thanks for contribution. I understood that this PR intended to handle "The agent retries with a divergent user message" case. But we think that current session server design should only handle append-only trajectory... and diverge should be limited to one step

PR #1010, opened 2026-04-19
feat(rollout/session): stub /v1/model_info on session server

DavidBellamy:
OpenAI-compatible clients (LiteLLM, openai-python) probe `/v1/model_info` on first connection for tokenizer hash and model-ID discovery. SGLang exposes this at `/model_info` (no `/v1/` prefix), so the existing session-proxy wildcard forwards to a non-existent SGLang endpoint and returns 404. Every new session emits a warning like:

```
Could not fetch tokenizer hash from
  http://.../sessions/<id>/v1/model_info: Client error 404 Not Found
```

## Fix

Add a specific route returning a minimal OpenAI-compatible stub with the served model name. Does not validate `session_id` because model_info is session-independent; the session path segment is kept for URL-shape consistency. Declared before the catch-all wildcard so it matches first.

## Scope

20-line addition, no behavior change for any existing route.

### DIFF
--- a/miles/rollout/session/sessions.py
+++ b/miles/rollout/session/sessions.py
@@ -246,6 +246,26 @@ def setup_session_routes(app, backend, args):
         finally:
             _inflight_chat["count"] -= 1
 
+    @app.get("/sessions/{session_id}/v1/model_info")
+    async def session_v1_model_info(session_id: str):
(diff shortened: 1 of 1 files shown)


guapisolo:
good. Could you add a CI test for this?

PR #1014, opened 2026-04-20
[ci] fix hf rate limit error by caching tokenizer loading

guapisolo:
### Background
HF rate limit 429 rate limit. 
e.g. https://github.com/radixark/miles/actions/runs/24480248250/job/71542821906

### Change
Simply do a cache to load tokenizer. 
- No invasion to training.
- Previously, the rate limit issue is fixed by setting secret HF_TOKEN, but that is only effective for PRs from branches in miles main repo. This fix also works for PR from fork repos.


maocheng23 (approved):
LGTM

PR #1016, opened 2026-04-20
Use load_generate_function in legacy sglang_rollout path

maocheng23:
## Summary

Replace the inline `inspect.signature`-based dispatch in `sglang_rollout.py::generate_and_rm` with the existing `load_generate_function` from `compatibility.py`.

This allows `GenerateFnInput`-style generate functions (e.g. `agentic_tool_call.generate`) to work on the legacy rollout path (without `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1`), while keeping full backward compatibility with existing 3-arg and 4-arg custom generate functions.

### Why not the approach in #1008?

PR #1008 adds an inline shim that `inspect.signature`-sniffs on every per-sample call and duplicates logic already in `compatibility.py`. This PR instead reuses the existing `load_generate_function` + `LegacyGenerateFnAdapter` infrastructure which:

- Detects legacy vs new signatures at load time (not per-call)
- Uses a more precise heuristic (`len(params) >= 3 and params[0] != "input"`) vs `len(params) == 1`
- Keeps a single source of truth for dispatch logic (DRY)
- Has no performance impact relative to the old code (same `load_function` + `inspect.signature` cost that was already there)

### Changes

- `miles/rollout/sglang_rollout.py`: Replace 6-line inline dispatch with 4-line unified `GenerateFnInput` call via `load_generate_function`
- Remove unused `import inspect`

## Test plan

- Existing 3-arg and 4-arg custom generate functions dispatch correctly through `LegacyGenerateFnAdapter` (no behavior change)
- New `GenerateFnInput`-style functions (e.g. `agentic_tool_call.generate`) work without `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1`
- Per-sample `generate_function_path` (from eval dataset configs) continues to work

### DIFF
--- a/miles/rollout/sglang_rollout.py
+++ b/miles/rollout/sglang_rollout.py
@@ -1,6 +1,5 @@
 import asyncio
 import copy
-import inspect
 import logging
 import uuid
 from argparse import Namespace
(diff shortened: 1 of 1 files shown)


Shi-Dong (approved):
LGTM!

yueming-yuan (approved):

PR #1022, opened 2026-04-20
Feat: Qwen3.6 RL support

Zhichenzzz:


miles_plugins/models/qwen3_6.py
```diff
@@ -0,0 +1,20 @@
+"""Qwen3.6 model spec.
+
```
guapisolo: Better add a comment here:
Notice:
1. qwen35 has `fp32` `linear_attn.A_log`. 
2. qwen36 reverted back this params to `bf16`... but in standard qwen36 forward impl, it still upcast `A_log` to fp32 during calculation. 
3. To reduce engineering complexity, we directly store it as fp32 (same as qwen35), which also align with SGLang impl.

guapisolo:
I have a small suggestion, what about directly use qwen35 structure import in scripts, and remove the newly added `miles_plugins/mbridge/qwen3_6.py`, `miles/backends/megatron_utils/megatron_to_hf/qwen3_6.py` and `miles_plugins/models/qwen3_6.py`? Since these files didn't include effective information. We can add the notice and cmts in qwen35 files. (like moe fuse and a_log precision)

tools/run_qwen3_6_35b_a3b_mtp_configs.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
guapisolo: good flywheel, but I suggest it should remove some personal info and be move to `tools/` for future model support usage?

tools/summarise_qwen36_results.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+# Summarise Qwen3.6-35B-A3B RL training logprob diff results.
```
guapisolo: Also for this

scripts/run-qwen3.6-27B.sh
```diff
@@ -0,0 +1,155 @@
...
+   --adam-beta1 0.9
+   --adam-beta2 0.98
+
+   --optimizer-cpu-offload
+   --overlap-cpu-optimizer-d2h-h2d
+   --use-precision-aware-optimizer
+)
+
+SGLANG_ARGS=(
+   # SGLang TP>1 produces garbage output for Qwen3.5-family (sgl-project/sglang#21039);
```
guapisolo: Better remove this information in this PR, also for qwen35, tp issues have already been fixed in 0.5.10

guapisolo (approved):
LGTM!

yueming-yuan (approved):

PR #1024, opened 2026-04-21
tito: support agent-layer-inserted assistant messages in append segmentation

DavidBellamy:
## Summary

`TITOTokenizer` (`miles/utils/chat_template_utils/tito_tokenizer.py`) is designed around the invariant that assistant turns are produced by the inference engine and arrive as pretokenized IDs, so the append path (`_split_appended_segments`, `tokenize_additional_non_assistant`) only handled `tool`, `user`, and `system` roles.

That invariant does not hold when the agent framework inserts non-generated assistant messages between tool calls. A concrete case: [terminus-2](https://github.com/SWE-agent/SWE-agent) emits self-reflection / planning assistant turns that are written back to the session but not produced by SGLang.

When such a turn is appended, `/sessions/{id}/v1/chat/completions` 500s:

\`\`\`
File \"miles/utils/chat_template_utils/tito_tokenizer.py\", line 121, in _split_appended_segments
    raise ValueError(f\"unsupported appended role for TITO segmentation: {role}\")
ValueError: unsupported appended role for TITO segmentation: assistant
\`\`\`

## Change

- \`_split_appended_segments\`: accept \`assistant\` alongside \`user\`/\`system\` as a valid single-message segment role.
- New \`_tokenize_assistant_segment\`: tokenizes the appended assistant turn against a \`[_DUMMY_SYSTEM, _DUMMY_USER]\` base context. Most chat templates require an assistant turn to follow a user/tool turn (not system directly), so we provide a stub user before it to render the correct boundary tokens.
- \`tokenize_additional_non_assistant\`: add the \`elif role == \"assistant\"\` branch. Function name kept for backward compatibility; docstring updated to reflect the broadened semantics (\"appended messages that were not generated by the inference engine\" — includes agent-layer assistant insertions).

## Opting in

Callers still need \`--tito-allowed-append-roles\` to include \`assistant\` so the outer \`assert_messages_append_only_with_allowed_role\` check accepts these turns. That CLI choice is already configurable per the existing \`allowed_append_roles\` argument on \`TITOTokenizer.__init__\`.

## Compat

- Default behavior unchanged: callers that don't pass \`assistant\` in \`--tito-allowed-append-roles\` still reject assistant appends at the outer assertion.
- No change to inference-engine-produced assistant turns: those still flow through the pretokenized IDs path and are merged via \`merge_tokens\`, not this path.
- No new dependencies.

## Testing

Reproduced 100% before this change and cleared completely after on an agentic RL workload using terminus-2 and shared-session TITO.


DavidBellamy:
Superseding with a branch hosted on LLM360/miles (see replacement PR). Closing to keep the review surface clean.

PR #1030, opened 2026-04-21
feat: rollout-data replay and all-gather CP for megatron runner

yueming-yuan:
Follow-up to the megatron runner stack (stacked on top of #780). Adds two features plus two small fixes to `miles.utils.debug_utils.run_megatron`.

### `fix test_run assertions to match shell-quoted export format`

Test-only. The CLI now shell-quotes env exports (`DUMPER_ENABLE='1'` instead of `DUMPER_ENABLE=1`); update the assertions in `test_run.py` accordingly.

### `fix run megatron debug`

- `worker/main.py`: register `--use-routing-replay` / `--use-miles-router` on the parser alongside the dataclass bridge. Without this the worker rejected these flags even though the CLI was passing them through.
- `get_model(..., wrap_with_ddp=script.run_backward)`: skip the DDP wrap for forward-only runs to avoid allocating gradient buffers (OOM on larger models).

### `update run megatron utils: support loading from rollout data and log probs print; support all-gather cp`

Main feature commit.

**`--rollout-data <path.pt>`**: load a rollout dump (as produced by `--debug-rollout-only`) and run Megatron forward on every sample's `tokens`, then compare the per-position logprobs against `sample.rollout_log_probs` (from SGLang). Details:

- All samples pad to a common length (multiple of 128) so the worker can iterate them one at a time with a shared padded seq_length.
- SGLang logprobs are saved as the comparator's baseline JSON (`sglang_logprobs/rank_0.json`).
- If samples have `rollout_routed_experts`, it is converted to the run_megatron replay format (`list[list[Tensor]]`, outer = per MoE layer) and written under `routing_replay/`, then fed into the worker via `--script-routing-replay-load-path`.
- After the worker finishes, `compare_logprobs` is invoked in report-only mode (threshold=1e10) to print the delta distribution.

**`--allgather-cp`**: new CLI flag plumbed through to `WorkerScriptArgs.allgather_cp`. Flips `prepare_batch` between zigzag CP (the existing path) and contiguous CP slicing (`rank k -> [k*local, (k+1)*local)`).

Worker-side adjustments in `worker/main.py`:
- Accept both `list[int]` and `list[list[int]]` in the token-ids file.
- Loop forward over sequences; accumulate logprob entries and emit a single `rank_{r}.json` at the end (instead of one-per-call).
- Only run top-k printing on the first sequence (noise otherwise).

### `support allgather cp replay in run-megatron`

Plumbs `allgather_cp` into `_load_replay`. When set, routing-replay tensors are sliced contiguously (`t[cp_rank*local : (cp_rank+1)*local]`); otherwise the existing zigzag path (via `miles.backends.training_utils.cp_utils.natural_to_zigzag_slice`) is kept. Adds `test_cp_allgather_slicing` to cover the new branch; the existing `test_cp_calls_zigzag_slice` still exercises the zigzag path.

## Test plan

- [ ] `pytest tests/fast/utils/debug_utils/run_megatron/ -q` (unit tests — includes new `test_cp_allgather_slicing`)
- [ ] `pytest tests/e2e/short/test_run_megatron.py -q` (the existing standalone-E2E test from #780 still passes)
- [ ] Manual: run `python -m miles.utils.debug_utils.run_megatron run --rollout-data <path.pt> ...` on a saved rollout and verify the logprob comparator report prints a sane distribution.
- [ ] Manual: run with `--allgather-cp --cp 2` on a 2-GPU setup and verify the routing replay loads without shape mismatches.


fzyzcjy (approved):
stamp w/o careful review b/c this is an isolated debug util

PR #1037, opened 2026-04-22
Prioritize session-server /health so heartbeats don't starve

Shi-Dong:
## Summary
- The Harbor agent server (harbor-private) heartbeat-polls every registered miles session server's `/health` every 30 s; after 3 consecutive failures (each with a 60 s HTTP timeout) it **bulk-cancels every in-flight trial** bound to that session (`Flushed N trial(s)` in the server log).
- We observed **~55% of completed v2-run trials coming back as `exit_status=Flushed, reward=0.0`** on our GLM-4.7-Flash SWE-bench training job, each preceded by `Health check failed (3/3)`. No session-server restart; just event-loop starvation blocking the async `/health` handler.
- This PR makes the session server's `/health` responsive even when the asyncio event loop is briefly CPU-blocked by tokenization inside `prepare_pretokenized` / `update_pretokenized_state`.

## Changes (all in `miles/rollout/session/sessions.py`)
1. Declare `/health` as a **synchronous** handler (`def`, not `async def`). Starlette/FastAPI dispatches sync handlers to the default threadpool, so `/health` no longer needs to wait for the asyncio event loop to drain the chat-completions queue.
2. Short-circuit the `debug_request_logger` middleware for `/health`. Heartbeat probes are high-frequency; skipping the two `logger.info(...)` calls + string formatting trims a bit more event-loop work per probe.

Net diff: **+12 / -1**, single file.

## Why not the other options we discussed
- **Run tokenization in a threadpool** (move `prepare_pretokenized` / `update_pretokenized_state` off the event loop): structurally the deepest fix, but larger surface area — left as a follow-up if heartbeats keep missing under truly sustained bursts.
- **Second uvicorn worker on a sibling port for heartbeats**: explicitly ruled out to keep the API surface unchanged.

## Default behavior preserved
- No API changes. Same `/health` response body.
- All other routes unchanged — the debug logging for `/sessions/*` still happens.
- No new ports, no new processes.

## Test plan
- [ ] Run a Flash agentic-async training job; confirm the aws-agent-server log stops emitting `Health check failed` bursts during rollout.
- [ ] Verify `curl http://<session-server>:<port>/health` returns `{"status":"ok", ...}` the same as before.
- [ ] Tests in `tests/fast/router/test_sessions.py` already exercise `/health` via `requests.get` — they are protocol-agnostic and should keep passing.

### DIFF
--- a/miles/rollout/session/sessions.py
+++ b/miles/rollout/session/sessions.py
@@ -40,8 +40,14 @@ def setup_session_routes(app, backend, args):
 
     registry = SessionRegistry(args, tokenizer, tito_tokenizer=tito_tokenizer)
 
+    # /health is defined as a sync function so Starlette dispatches it to the
+    # default threadpool rather than the asyncio event loop. The event loop is
+    # occasionally starved by CPU-bound tokenization in chat_completions, which
+    # makes async /health responses miss the agent-server's 60s heartbeat
+    # window and triggers Flushed-group cancellation on the agent server.
+    # Running /health in a threadpool worker sidesteps that queueing.
     @app.get("/health")
-    async def health():
+    def health():
         body = {"status": "ok"}
         if session_server_instance_id is not None:
             body["session_server_instance_id"] = session_server_instance_id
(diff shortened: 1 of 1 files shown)


Shi-Dong:
Closing in favor of the structural fix — offloading the CPU-bound tokenization (prepare_pretokenized / update_pretokenized_state) to a threadpool via starlette.concurrency.run_in_threadpool. The sync-def /health change in this PR is a micro-optimization at best and does not solve event-loop starvation, as gemini-code-assist correctly pointed out. Confirmed empirically on the Flash 5-node v3 run (same flush pattern, same ~80 min cadence). Opening the replacement PR next.

PR #1045, opened 2026-04-24
DeepSeek V4 RL support

yueming-yuan:
### This PR is an archive of the original DeepSeek V4 Day 0 support. See #1181 for the final version.

https://github.com/radixark/Megatron-LM/pull/28


maocheng23 (approved):

Zhichenzzz (approved):

guapisolo (approved):
GOAT!!!!!!

yushengsu-thu (approved):

harvenstar (approved):

ByronHsu:
GOAT

hebiao064:
could you please share the kl and reward dashboard if possible? thanks

miles_plugins/models/deepseek_v4/ops/kernel/tilelang_sparse_mla_fwd.py
```diff
@@ -0,0 +1,189 @@
...
+            for h_i in T.Parallel(H_per_block):
+                sumexp[h_i] = T.log2(sumexp[h_i]) + m_i[h_i] * sm_scale
+
+            T.copy(acc_o, Output[b_i, s_i, H0:H1, :])
+            T.copy(sumexp, Lse[b_i, s_i, H0:H1])
+
+    return main
+
+
+def sparse_mqa_fwd_interface(q, kv, attn_sink, topk_idxs, sm_scale=None, block_I=64, num_stages=2, threads=256):
```
henryxuxu0716, line 152: 请问到底是mla还是mqa呢

lilei199908: 本身dsa的mla和mqa就差不多，都走的mqa

examples/train_infer_mismatch_helper/mis.py
fzyzcjy, line 1: TODO: fix apply chat template, use their py instead of hf tokenizer

Zhichenzzz:
> could you please share the kl and reward dashboard if possible? thanks

Hi @hebiao064 We have some initial reasoning task reward curve for the reference! 
image

miles_plugins/models/deepseek_v4/ops/kernel/tilelang_sparse_mla_bwd.py
```diff
@@ -0,0 +1,282 @@
...
+
+    return postprocess_kernel
+
+
+@tilelang.jit(
+    out_idx=[-3],
+    pass_configs={
+        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
+        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
+        tilelang.PassConfigKey.TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE: True,
```
pengdurice, line 84: Hello @yueming-yuan, nice to meet you. This is Peng. I’m cherry-picking your code for DeepSeek V4 training. Thank you for the implementation!

I ran into a NaN issue during the backward pass, and after debugging it appears to come from enabling `TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE` in the sparse MLA/MQA backward kernel. In our setup, aggressive shared-memory merge can alias/corrupt the region used by `Q_shared`, which matches the failure mode described in this TileLang PR: https://github.com/tile-ai/tilelang/pull/2204.

Would you be okay with removing/disabling this pass config temporarily to unblock training until the upstream TileLang fix lands and we can use a version that includes it?

Thank you!

tang-t21:
Hi @yueming-yuan and folks, thanks for putting together the DeepSeek V4 support here.

I hit the same tilelang sparse MLA backward NaN issue while validating Bridge TP8/EP8 e2e training, and opened a small stacked fix here:

https://github.com/yueming-yuan/miles/pull/4

I also noticed @pengdurice had already reported what looks like the same root cause in this thread: https://github.com/radixark/miles/pull/1045#discussion_r3263224333. The behavior matches what we saw: forward loss stays finite, but backward produces NaN gradients when `TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE` is enabled in `tilelang_sparse_mla_bwd.py`. Disabling that pass makes the standalone backward test finite and lets the Bridge e2e run complete.

Validation summary from our side:
- standalone `test_sparse_mla_backward`: 5 passed after adding a finite-gradient assertion
- Bridge tilelang backend 3 iter: losses `0.9047399`, `1.2337190`, `0.5607164`; NaN iterations `0`

This is stacked on top of this PR branch, so the diff should just be the temporary pass-config removal plus the test guard. Thanks again!

miles_plugins/models/deepseek_v4/ops/kernel/tilelang_sparse_mla_bwd.py
yangbofun: > Hello @yueming-yuan, nice to meet you. This is Peng. I’m cherry-picking your code for DeepSeek V4 training. Thank you for the implementation!
> 
> I ran into a NaN issue during the backward pass, and after debugging it appears to come from enabling `TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE` in the sparse MLA/MQA backward kernel. In our setup, aggressive shared-memory merge can alias/corrupt the region used by `Q_shared`, which matches the failure mode described in this TileLang PR: [tile-ai/tilelang#2204](https://github.com/tile-ai/tilelang/pull/2204).
> 
> Would you be okay with removing/disabling this pass config temporarily to unblock training until the upstream TileLang fix lands and we can use a version that includes it?
> 
> Thank you!

It works when set TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE False. Thanks.

PR #1048, opened 2026-04-27
[TITO] refactor (1/N): Use (tito_model x allow_append_roles) to infer fixed chat template and kwargs, refactor chat template cpu tests, and add doc

guapisolo:
## Motivation
Originally, we use `autofix` to let miles use fixed chat template automatically. 
But implicitly infer chat template from `--hf-checkpoint` is not a good behavior, might cause issues. Also, this is not easy to use.

## How
- Now, the fixed chat template will be bond in `TITOTokenizer`, and infer by `(tito_model x allow_append_roles)`. User can directly use models already supported in the doc. 
  - `allow_append_roles` is the role set that can be appended after the first `assistant` message. 
- Tool call parser and reasoning parser need to be enabled in current TITO design to avoid unstable chat template behavior. And `TITOTokenizer` bond parser with the model, and be automatically applied in sglang param. 
- fix wrong tito tokenizer behavior in lacking `chat_template_kwargs`.
- Reorg tests. Let verify code path and CI use the same testing frame.
- Give docs to explain the whole design.


miles/rollout/session/sessions.py
```diff
@@ -35,6 +35,7 @@ def setup_session_routes(app, backend, args):
     tito_tokenizer = get_tito_tokenizer(
         tokenizer,
         tokenizer_type=getattr(args, "tito_model", "default"),
+        chat_template_kwargs=getattr(args, "apply_chat_template_kwargs", None),
```
guapisolo, line 38: bugfix here.

jybsuper (approved):
Synced offline. The workflow is comprehensive and robust. LGTM!

zyzshishui (approved):

PR #1052, opened 2026-04-28
[feat] Init true on policy with qwen_dense

maocheng23:
## Summary

Initial framework introducing **true-on-policy** for Qwen3-dense across SGLang, Megatron, and Miles. This is one of **three tightly-coupled PRs** that must land together — they share a single contract identifier `qwen3_dense_true_on_policy_v1` defined by a vendored schema in each repo.

**Companion PRs (must land in lockstep):**
- SGLang: sgl-project/sglang#23961
- Megatron-LM: radixark/Megatron-LM#29

## Target

Bit-identical (exact-zero) logprob parity between the SGLang rollout engine and the Megatron trainer for every scored response token at TP=1, TP>1, PP>1, and Ulysses CP for Qwen3-4B (dense).

## Design

Three-layer contract architecture:

1. **Miles** (launcher, this PR): picks contract, validates parallel layout against typed model profile, renders SGLang/Megatron/env args from a typed kernel policy, drives PPO logprob comparison.
2. **SGLang** (rollout numerical truth): produces logprobs using deterministic kernels.
3. **Megatron** (parity target): reproduces SGLang's numerics in a differentiable training forward via `SGLangSpecProvider` layer classes.

The contract object owns its own runtime policy — adding a new architecture (e.g. Qwen3-MoE) is one new contract object + one new model profile entry, not edits across three repos.

## In this PR (Miles)

- New `miles/true_on_policy/` package:
  - `schema.py` — vendored shared identity, byte-identical with SGLang and Megatron copies
  - `contracts.py` — `TrueOnPolicyContract` definition + lookup registry
  - `model_profiles.py` — `QWEN3_DENSE_PROFILE` (sole registered family)
  - `config.py` — `TrueOnPolicyConfig` → `TrueOnPolicyKernelPolicy` → `TrueOnPolicyLaunchPlan`; renders SGLang / Megatron / env args from typed contract
- `apply_true_on_policy_script_defaults(args)` and `build_true_on_policy_launch_plan(args)` entry points
- `scripts/run_qwen3_4b.py` launcher with `--true-on-policy=true` switch and Megatron model_provider routing through `config.true_on_policy_contract`
- `miles/utils/ppo_utils.py`:
  - `compute_log_probs(..., true_on_policy_mode=True)` — full-vocab gather → real-vocab truncate → fp32 log-softmax → gather
  - **NEW: `_ReplicatedLossAllGatherLastDim` autograd Function** (the grad_norm bug fix). Megatron's standard `all_gather_last_dim_from_tensor_parallel_region` has a reduce-scatter backward, which is correct when each rank contributes a distinct output gradient. In the true-on-policy logprob path *every TP rank computes the same scalar loss* from the gathered full vocabulary, so reduce-scatter sums TP_size identical gradients into the local logits and scales them by TP_size. Replaced with a typed autograd Function whose backward narrows (splits) the gradient — the correct inverse of the forward all-gather when the downstream loss is replicated.
  - This was visible as a ~1.8x grad_norm gap vs the off-policy baseline at clipfrac=0.
- `miles/backends/training_utils/loss.py` — `true_on_policy_mode` handling (dtype management, response-only logits slicing across CP modes)
- Phase 4 — stop emitting legacy `--use-sglang` and `--sglang-rl-on-policy-target` flags. Only emits `--true-on-policy-contract`.

## Validation

- ✅ CPU unit tests green (`tests/fast/true_on_policy/`, `tests/fast/utils/test_true_on_policy_logprobs.py`)
- ✅ Test added for `_ReplicatedLossAllGatherLastDim` backward (verifies narrow vs reduce-scatter)
- 🔴 **GPU exact-zero E2E gate not yet run** at TP=1, TP>1, PP>1, CP. This is the next task on the stack before adopting Qwen3-MoE / Qwen3-Next.

## Out of scope

- Qwen3-MoE / Qwen3-Next model profiles — additive after this stack lands; would be one new `TrueOnPolicyModelProfile` entry per family
- Alignment harness (layer-dump comparator, first-divergence comparator) — separate follow-up PR
- Model-family cross-check at config validation — recommended before contract #2

## Test plan

- [ ] CPU unit tests pass in CI
- [ ] GPU exact-zero E2E gate at TP=1
- [ ] GPU exact-zero E2E gate at TP=2
- [ ] grad_norm parity check (before/after `_ReplicatedLossAllGatherLastDim` fix)
- [ ] PPO training smoke at planned configs

Screenshot 2026-04-28 at 2 00 27 PM

Screenshot 2026-04-28 at 2 00 52 PM

Screenshot 2026-04-28 at 2 01 08 PM

Screenshot 2026-04-28 at 2 01 46 PM


miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -23,13 +23,17 @@
 
 def get_base_gpu_id(args, rank):
     num_gpus = min(args.num_gpus_per_node, args.rollout_num_gpus_per_engine)
-    if args.colocate:
+    if getattr(args, "colocate", False):
```
yueming-yuan: can we keep the original implementation here (just use `args.xxx` rather than `getattr`) to avoid silent errors?

miles/backends/training_utils/loss.py
```diff
@@ -552,6 +509,72 @@ def icepop_function(
     return pg_loss, loss_masks, metrics
 
 
+_POLICY_LOSS_DUMP_COUNTER = 0
```
yueming-yuan: nit: can we make this part more clean? (e.g., if the debug utils are not expected to be frequently used, we can remove them; if they are expected to be used in the future, can we move to a separate file, and just plug several lines in core logics?)

scripts/run_qwen3_4b.py
```diff
@@ -27,19 +32,41 @@ class ScriptArgs(U.ExecuteTrainConfig):
     train_fp8: bool = False
     enable_megatron_bridge: bool = False
     enable_mis: bool = False
+    tensor_model_parallel_size: int | None = None
```
yueming-yuan: nit: to keep the readability and reproducibility of these scripts, I might suggest not adding too many arguments and changing too much here, maybe we can just provide one verified config and let the user to edit the script according to their task (e.g. just one `--true-on-policy`, and any other necessary args; and we directly code the recommended settings in this file). This will also make it easier for us to maintain

miles/utils/arguments.py
```diff
@@ -1819,9 +1831,24 @@ def _resolve_eval_datasets(args) -> list[EvalDatasetConfig]:
     return eval_datasets
 
 
+def _maybe_enable_true_on_policy_sglang_cp_lm_head(args) -> None:
```
yueming-yuan: qq: why is this function empty?

miles/backends/training_utils/cp_utils.py
```diff
@@ -131,6 +131,34 @@ def sum_of_token(x: torch.Tensor) -> torch.Tensor:
...
+        zip(total_lengths, response_lengths, loss_masks, strict=False)
+    ):
+        max_seq_len = max_seq_lens[i] if max_seq_lens is not None else None
+        prompt_length = total_length - response_length
+        _, _, _, tokens_offset = get_logits_and_tokens_offset_with_cp(
+            total_length, response_length, qkv_format, max_seq_len
+        )
+        loss_mask_0 = loss_mask[tokens_offset[0][0] - prompt_length : tokens_offset[0][1] - prompt_length]
+        loss_mask_1 = loss_mask[tokens_offset[1][0] - prompt_length : tokens_offset[1][1] - prompt_length]
+        local_masks.append(torch.cat([loss_mask_0, loss_mask_1], dim=0))
```
Zhichenzzz: Could we extract L147–157 and the duplicate slicing at L98–109 into a shared `_slice_loss_mask_for_local_cp` helper, so both call sites share a single source of truth?

Zhichenzzz (commented):
Most backends LGTM! Thank you for the great work! Just a quick reminder, there is a small issue raised by ci test, https://github.com/radixark/miles/actions/runs/25076368286/job/73469945178?pr=1052#step:9:347

yueming-yuan (approved):

PR #1054, opened 2026-04-29
Use TE NVFP4 reference quantizer and add bitwise exact tests

zianglih:
This implementation should have better numerics and is less error-prone.

Current FlashInfer quantizer does not have `rn` for e4m3 scales and is not bitwise exact with TE, so not yet safe to use.

Redo the implementation for the following flags to mirror well tested mxfp8 implementation introduced in https://github.com/radixark/miles/pull/614:

- `--num-layers-at-start-in-bf16`
- `--num-layers-at-end-in-bf16`
- `--extra-high-precision-layers-hf`
- `--extra-high-precision-layers-megatron`


zianglih:
Hi @Zhichenzzz @yueming-yuan , let me know if you have any workloads that use the nvfp4 quantizers and want to test them. Thank you.

Zhichenzzz:
> Hi @Zhichenzzz @yueming-yuan , let me know if you have any workloads that use the nvfp4 quantizers and want to test them. Thank you.

Thank you @zianglih for the great work! Could you also share some accuracy validation experiments on top of this pr? In this case, we can reproduce and re-verify it. Thanks!

zianglih:
@Zhichenzzz  There is no E2E RL validation but the quantizer is bitwise exact with TE's ground truth reference implementation. You can check the new tests I have added.

Zhichenzzz (approved):
LGTM! Due to resource constraints, I verified this with kernel-level matching tests.

yueming-yuan (approved):

PR #1062, opened 2026-05-03
Random fully async agent example

jthomy:
Random fully async agent example as a minimal example for agentic RL(hicache, disaggregated setup, EP, etc).

EDIT: The patches are temporary necessary fixes, I believe they should be commited to Miles' SGLang branch instead.

Expected throughput:
image


lawrence-harmonic:
Purpose of patches:

> No tbo support (patch 00)

Fixed in https://github.com/sgl-project/sglang/pull/20114

> Weight sync with disagg wasn't pausing the engines (patch 01)

Should be fixed in https://github.com/sgl-project/sglang/pull/24462 (haven't verified)

>Mamba indices need to be on Cuda (patch 4)
Race bug where data is registered only on froward path but used in load/write streams resulting in corrupted pointers (patch 13)

Fixed in:
https://github.com/sgl-project/sglang/pull/22940
https://github.com/sgl-project/sglang/pull/23427

> OOM on startup although the host has enough RAM, removing strict NUMA policy (patch 15)
At least in our cluster, SGLang crashes with a certain probability on Inconsistency detected by ld.so: ../elf/dl-tls.c: 1044: _dl_add_to_slotinfo: Assertion `idx == 0' failed!  - added support to preload it which appears to fix it.
Weight sync with multi tokenizer pool (this is necessary for good speed if you need tokenization) appears to be broken with weight sync. In our case we can simply disable tokenizer init as we are using token in token out for our RL (more on policy anyway)

Needs upstream fix?

lawrence-harmonic:
> At least in our cluster, SGLang crashes with a certain probability on Inconsistency detected by ld.so: ../elf/dl-tls.c: 1044: _dl_add_to_slotinfo: Assertion `idx == 0' failed!  - added support to preload it which appears to fix it.

This turns out to be a rare bug in glibc, which we have produced a repro using only dlopen: https://gist.github.com/lawrence-harmonic/995d2ef964e2f40af1d81205d7285355

LD_PRELOAD is not the real fix, it only avoids the issue since the issue is sensitive to load order (dlopen must fail at multiple of 62). We will report the issue to glibc mailing list.

Remaining issues with this test script:

> OOM on startup although the host has enough RAM, removing strict NUMA policy (patch 15)

> Weight sync with multi tokenizer pool (this is necessary for good speed if you need tokenization) appears to be broken with weight sync. In our case we can simply disable tokenizer init as we are using token in token out for our RL (more on policy anyway)

jthomy:
I removed the patches. All patches are solved in https://github.com/sgl-project/sglang/pull/25201 and additionally for R3 in https://github.com/sgl-project/sglang/pull/24908

miles/backends/sglang_utils/sglang_engine.py
yueming-yuan, line 1: Suggest removing these environment hacks; the CIs are directly run in the docker so it should work without the patch
If you see these issues, they are usually caused by the incorrect usage of the environment, e.g. not the latest docker, directly do `pip install` without `--no-deps` and it breaks the dependency, etc.

jthomy: Thanks, we found that this is a bug in glibc, and bumping nixl-cu12 from 1.0.0 to 1.0.1 changes the load pattern slightly, avoiding the bug (just by changing the load pattern and being more lucky - it doesn't actually fix it), so I removed it here.
The bug occurs if the Nth dlopen fails for any N which is a multiple of 62, dlopen is intended to fail for plugin systems which check for existence of optional libs.

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -637,7 +637,7 @@ def _compute_server_args(
 
     if worker_type == "prefill":
         kwargs["disaggregation_mode"] = "prefill"
-        kwargs["load_balance_method"] = "round_robin"
+        kwargs.setdefault("load_balance_method", "round_robin")
```
Zhichenzzz, line 640:
```suggestion
        kwargs["load_balance_method"] = "round_robin"
```

Zhichenzzz (approved):
Most LGTM! Just a small nit here! Thanks @jthomy!

miles/ray/rollout.py
```diff
@@ -59,6 +59,11 @@
 # ---------------------------------------------------------------------------
 
 
+def _get_sglang_env_passthrough() -> dict[str, str]:
+    env_names = os.environ.get("MILES_SGLANG_ENV_PASSTHROUGH", "")
+    return {name: os.environ[name] for name in env_names.replace(",", " ").split() if name in os.environ}
+
+
```
Zhichenzzz:
```suggestion

```

miles/ray/rollout.py
```diff
@@ -139,6 +144,7 @@ def start_engines(self, port_cursors: dict[int, int] | None = None) -> tuple[lis
                     "SGLANG_ENABLE_STRICT_MEM_CHECK_DURING_IDLE": "false",
                 }.items()
             }
+            env_vars.update(_get_sglang_env_passthrough())
```
Zhichenzzz:
```suggestion
```

miles/ray/rollout.py
Zhichenzzz: Env vars in `extra_env_vars` are already forwarded to SGLang by Ray so this is redundant and can be removed.

miles/ray/rollout.py
Zhichenzzz: same as above

Zhichenzzz (changes requested):

miles/ray/rollout.py
jthomy: Thank you, sounds good.

yueming-yuan (approved):

Zhichenzzz (approved):

PR #1065, opened 2026-05-04
[TITO] model-support: add DeepSeek V4 TITO support 

zyzshishui:
## Summary

Adds TITO support for DeepSeek V4 (tested on Flash due to lack of gpu, but should also work for V4 pro).

Registers a new `deepseekv4` TITO family and wires it to SGLang’s DeepSeek V4 encoder path instead of the regular HF/Jinja chat template path.

**Note**: A temporary compatibility guard is included in `miles/utils/dumper_utils.py` because the current SGLang V4 support branch is not upstreamed and its dumper API is too old. We would remove it after https://github.com/radixark/miles/pull/1045 get merged.

### 1. DeepSeek V4 TITO tokenizer verifier

Verifies that the registered `deepseekv4` TITO tokenizer can incrementally merge appended tool messages and decode back to the same text as a full prompt render.

```bash
python scripts/tools/verify_chat_template.py \
  --model deepseek-ai/DeepSeek-V4-Flash \
  --tito-model deepseekv4 \
  --tito-allowed-append-roles tool \
  --thinking both
```

### 2. Fast tokenizer regression suite
Verifies the shared chat-template / TITO tokenizer utilities, including DeepSeek V4 prompt-id alignment against SGLang’s DSv4 encoder path.
```
pytest tests/fast/utils/chat_template_utils
```

### 3. DeepSeek V4 session-server e2e
Runs the real Miles session-server TITO path with SGLang inference. This verifies that Miles-owned input_ids, DeepSeek V4 prompt encoding, tool-call parsing, rollback, and accumulated token mismatch checks work together.

```
python scripts/tools/verify_session_tito_tokenizer.py \
  --hf-checkpoint deepseek-ai/DeepSeek-V4-Flash \
  --tito-model deepseekv4 \
  --tito-allowed-append-roles tool \
  --reasoning-parser deepseek-v4 \
  --tool-call-parser deepseekv4 \
  --tp-size 4 \
  --num-gpus 4
```

### DIFF
--- a/miles/utils/chat_template_utils/tito_tokenizer.py
+++ b/miles/utils/chat_template_utils/tito_tokenizer.py
@@ -766,6 +766,53 @@ class DeepSeekV32TITOTokenizer(TITOTokenizer):
 
 
 # ---------------------------------------------------------------------------
+# DeepSeek V4 implementation
+# ---------------------------------------------------------------------------
+
+
+class DeepSeekV4TITOTokenizer(TITOTokenizer):
+    """DeepSeek V4 — official encoder via sglang's ``encoding_dsv4``.
+
+    Like V3.2, V4 ships no jinja chat_template; miles' ``apply_chat_template``
+    routes any V4 tokenizer to the ``chat_template_utils.deepseek_v4`` bridge, and
+    TITO incremental tokenization rides that same bridge to stay byte-aligned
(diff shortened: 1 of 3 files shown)


Shi-Dong (approved):
LGTM

PR #753, opened 2026-03-18
[loss refactor] [1] snapshot tests + file structure

yueming-yuan:

PR #1068, opened 2026-05-04
feat: add chunked TP logprob loss

mouad-hpc:
## Summary

Adds an opt-in chunked TP logprob path that avoids materializing the full `[tokens, vocab / TP]` logits tensor on the actor's policy / SFT loss.

When `--use-chunked-tp-logprob-loss` is set, the actor's `output_layer.forward` is patched to return hidden states. The loss path picks up the `ActorOutputProjection` adapter via `args.actor_projection` and replays the local output projection in sequence chunks before calling the existing TP-aware logprob / entropy helpers.

Result on Qwen3.5-35B / 122B: **~20% step-time reduction**, with no model-family-specific assumptions beyond the Megatron causal-LM actor output_layer interface.

## Changes

- Add `--use-chunked-tp-logprob-loss` and `--chunked-tp-logprob-seq-chunk-size` (default 256).
- New `miles/backends/megatron_utils/chunked_tp_logprob.py`:
  - `ActorOutputProjection` adapter encapsulates the output_layer monkey-patch + `.linear(hidden)` replay + sequence-parallel gather.
  - `setup_chunked_tp_logprob(model, args, role)` installs the adapter and stashes it on `args.actor_projection`. No-op when disabled or for non-actor roles.
- `loss_hub/logit_processors.py::get_log_probs_and_entropy` picks up `args.actor_projection` and branches to a hidden-state replay path that calls `projection.linear` per sequence chunk. Signature unchanged.
- `loss_hub/math_utils.py::calculate_log_probs_and_entropy` gains a `need_entropy_grad` switch (default `True`, preserves pre-PR behavior); the bypass / regular paths pass `need_entropy_grad=with_entropy`.
- Single-line wiring in `miles/backends/megatron_utils/model.py` after `get_model(...)` in `setup_model_and_optimizer`.

## Constraints

`--use-chunked-tp-logprob-loss` currently requires:
- `--qkv-format bshd`
- `--context-parallel-size 1`
- `--allgather-cp` off
- `--true-on-policy-mode` off

## Validation

- `python -m py_compile` on every touched file.
- `ruff check` on every touched file.
- `git diff --check`.
- Ran with miles stack 3 times with Qwen 3.5 122B and Qwen 3.5 35B


Zhichenzzz:
@mouad-hpc Thanks for PR! I noticed three times of CI-megatron failed, could you check that and fix the bug first? Thank you!

mouad-hpc:
> @mouad-hpc Thanks for PR! I noticed three times of CI-megatron failed, could you check that and fix the bug first? Thank you!

@Zhichenzzz All done if you could take another look!

mouad-hpc:
@Zhichenzzz Could we check the CI, it seems to be the CI issue currently is due to a HF token issue? 
image

miles/backends/megatron_utils/model.py
```diff
@@ -48,6 +54,18 @@
 from .lora_utils import save_lora_checkpoint
 
 
+def _find_output_layer(model_chunk: torch.nn.Module) -> torch.nn.Module | None:
```
Zhichenzzz: Is there a cleaner way to do this? Falling back from `get_attr_wrapped_model` to a brute-force .modules() scan feels fragile.

mouad-hpc: Probably unwrapping to an expected shape/list, will work on a quick fix

mouad-hpc: @Zhichenzzz  Any updates?

Zhichenzzz: Hi Mouad, we need to merge this loss func refactoring first https://github.com/radixark/miles/pull/753 and then merge your pr. Thanks!

mouad-hpc: sounds good

PR #1069, opened 2026-05-05
[fix] fix truncate function return routed expert

### LINKED ISSUE #861: `_truncate_sample_output` does not truncate `rollout_routed_experts`

zyzshishui:
Replace the inline _truncate_sample_output helper with Sample.strip_last_output_tokens. The helper takes the count to strip from the end, so pass overshoot (= response_length - allowed_output), not allowed_output itself.

Fix #861

### DIFF
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -227,23 +227,9 @@ def truncate_samples_by_total_tokens(
         if allowed_output <= 0:
             break
 
-        _truncate_sample_output(sample, allowed_output, tokenizer)
+        sample.strip_last_output_tokens(overshoot, tokenizer)
+        sample.status = Sample.Status.TRUNCATED
         result.append(sample)
         break
 
     return result
-
-
-def _truncate_sample_output(sample: Sample, keep_tokens: int, tokenizer) -> None:
-    """Truncate a sample's output in-place to exactly ``keep_tokens`` tokens."""
-    prompt_len = len(sample.tokens) - sample.response_length
-    kept_ids = sample.tokens[prompt_len : prompt_len + keep_tokens]
-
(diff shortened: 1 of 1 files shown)


guapisolo:
@codex

guapisolo (approved):
LGTM.

PR #1074, opened 2026-05-05
fix: skip linux-only deps on macOS in requirements.txt

fzyzcjy:
want to install miles locally (macos) for ide (not for running)


guapisolo (approved):
Approved. But interestingly, my mac run `pip install -r requirements.txt` smoothly without error (

PR #1075, opened 2026-05-05
refactor: mark create_cpu_replica as private

fzyzcjy:
## Summary
- `UpdateWeightP2P.create_cpu_replica` is only called internally via `self.create_cpu_replica(...)` in `connect_rollout_engines` (same class, same file). No external callers.
- Rename to `_create_cpu_replica` to follow the project convention of prefixing private members with `_`.

## Test plan
- [ ] No external imports/usages broken (verified via grep — only one call site, in the same class).


guapisolo (approved):
approve nit.

PR #1087, opened 2026-05-07
[lora] avoid merge adapter weights for base weight sync

gongyisheng:
noticed the base weight sync issue when lora is enabled, base weight will be exported with merged lora weights, causing correctness issue


yushengsu-thu (approved):
sync offline

PR #1092, opened 2026-05-08
chore: [lora] avoid merge adapter weights

gongyisheng:
follow up of this PR: https://github.com/radixark/miles/pull/1087
due to the hf param name changes after disable merge_adapter_weights, need to remove the ".base_layer." part so that it can be successfully synced to sglang.


yushengsu-thu (approved):

yushengsu-thu:
fixed the reverted problem: https://github.com/radixark/miles/pull/1091

PR #1108, opened 2026-05-11
[CI] tito-logprob: restore qwen3 / glm47 family sweep via CONFIGS loop

guapisolo:
## Why

The old j2-generated \`pr-test.yml\` dispatched this test twice via \`strategy.matrix.info\`: once with \`SESSION_TEST_MODEL_FAMILY=qwen3\`, once with \`SESSION_TEST_MODEL_FAMILY=glm47\`. The new \`run_suite.py\` framework has no variant dimension — every test is a single \`(suite, num_gpus, est_time)\` triple — so the \`glm47\` variant has been silently dropped since the CI refactor.

## What

Restore coverage by sweeping \`CONFIGS=[\"qwen3\", \"glm47\"]\` inside the test file itself, the same in-file loop pattern already used by \`test_qwen3_30B_A3B_r3.py\`. \`SESSION_TEST_MODEL_FAMILY\` now becomes a single-family override, primarily for local debugging.

- \`MODEL_FAMILY\` module constant → \`CONFIGS\` list + \`_resolve_configs\` helper.
- \`prepare\` / \`execute\` / \`_get_config\` now take \`model_family\` as a parameter instead of reading it from the module-level constant.
- \`__main__\` loops over \`_resolve_configs()\`: \`prepare(fam)\` + \`execute(fam)\` per family.
- \`est_time\` bumped 600 → 1200 since two families run sequentially in one job.

## Verification

Real GPU runs (4× H200 free locally):

- \`SESSION_TEST_MODEL_FAMILY=glm47 python tests/e2e/sglang/test_tito_logprob_equivalence.py\` — 16 samples, \`Logprob diff mean=0.000000\`, \`tito_session_mismatch_rate=0.0\`, Ray job succeeded.
- \`SESSION_TEST_MODEL_FAMILY=qwen3 python tests/e2e/sglang/test_tito_logprob_equivalence.py\` — same: 16 samples, all zero diff.
- Full CONFIGS loop (no override) — both families ran back-to-back in one process, 32 samples total verified zero diff, both Ray jobs succeeded. Sglang + Ray teardown between configs is handled by \`execute_train\`'s built-in \`pkill -9 sglang; ray stop --force\` at the start of each call.

## Stacked PR

Base: pr/2-ci-stage-b-sglang-8gpu (#1107). Depends on the 8-GPU suite rename being in place; otherwise this register would not match any active suite.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -76,7 +76,7 @@ jobs:
       execute_command: "python tests/ci/run_suite.py --hw cuda --suite stage-b-fast-1-gpu"
     secrets: inherit
 
-  # Stage B: SGLang patch tests (1 GPU)
+  # Stage B: SGLang patch tests (8 GPU)
   stage-b-sglang:
     needs: [stage-a-fast]
     if: |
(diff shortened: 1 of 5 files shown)


guapisolo:

PR #1109, opened 2026-05-11
[CI] r3-router: restore qwen3_30b_a3b / glm47_flash family sweep via CONFIGS loop

guapisolo:
## Why

Same regression as \`tito_logprob_equivalence\` (#TBD): the old matrix ran this test with \`ROUTER_EQ_MODEL_FAMILY=qwen3_30b_a3b\` and \`=glm47_flash\`; after the CI refactor only the \`qwen3_30b_a3b\` default still runs because the framework has no variant dimension.

## What

Loop over \`CONFIGS=[\"qwen3_30b_a3b\", \"glm47_flash\"]\` in \`__main__\` and thread \`model_family\` through \`prepare\` / \`execute\` / \`_variant_dir\` so dump paths stay distinct per \`(family, router-variant)\` pair. The inner \`(miles, sgl)\` router-variant compare still happens once per family.

\`ROUTER_EQ_MODEL_FAMILY\` becomes a single-family override for local debugging.

\`est_time\` bumped 600 → 1200.

## Verification

Real GPU run (4× H200 free locally):

- \`ROUTER_EQ_MODEL_FAMILY=glm47_flash python tests/e2e/sglang/test_r3_router_equivalence.py\` — both inner variants (\`miles\` router, \`sgl\` router) ran sequentially, dumps written to \`/tmp/router-eq/glm47_flash/{miles,sgl}/dump.jsonl\`, \`_assert_records_equal\` passed byte-for-byte across 10 samples: \`[router-eq] model_family=glm47_flash variants miles/sgl match across 10 samples\`. Both Ray jobs succeeded.

(The \`qwen3_30b_a3b\` path is structurally symmetric; not run locally because Qwen3-30B-A3B is ~60 GB and not in the cluster cache.)

## Stacked PR

Base: pr/2-ci-stage-b-sglang-8gpu (#1107).

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -76,7 +76,7 @@ jobs:
       execute_command: "python tests/ci/run_suite.py --hw cuda --suite stage-b-fast-1-gpu"
     secrets: inherit
 
-  # Stage B: SGLang patch tests (1 GPU)
+  # Stage B: SGLang patch tests (8 GPU)
   stage-b-sglang:
     needs: [stage-a-fast]
     if: |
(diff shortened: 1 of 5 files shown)


guapisolo:

PR #1111, opened 2026-05-11
[CI] glm47-flash-ckpt: run save + async_save back-to-back

guapisolo:
## Why

The old matrix dispatched this file twice — once without \`--async-save\`, once with — via \`test_file\` argv injection. The new \`run_suite.py\` framework drives each test as \`python3 <file>\` with no argv, so only the synchronous \`save\` mode was running.

## What

Mirror the \`test_qwen3_4B_ckpt.py\` pattern: drop the \`argparse\` and run \`save → load → async_save → load\` sequentially in one process. Both ckpt modes now share one model download / one runner.

\`est_time\` bumped 1200 → 2400 (two ckpt-mode roundtrips instead of one).

## Verification

Not run locally — needs 8 GPUs exclusively + ~120 GB \`_torch_dist\` disk write for the megatron→fsdp conversion. CI is the right place.

The argument-capture smoke confirmed all four \`execute()\` calls produce the expected \`train_args\`: \`save\` with \`--save-interval 2\` + \`--ci-save-model-hash\`, \`load\` with \`--ckpt-step <N>\` + \`--ci-check-model-hash\`, \`async_save\` with the additional \`--async-save\` + \`--use-persistent-ckpt-worker\` flags.

## Stacked PR

Base: pr/2-ci-stage-b-sglang-8gpu (#1107). The suite (stage-c-ckpt-8-gpu) is unchanged, but the branch is rebased onto pr/2 to keep the PR series consistent.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -76,7 +76,7 @@ jobs:
       execute_command: "python tests/ci/run_suite.py --hw cuda --suite stage-b-fast-1-gpu"
     secrets: inherit
 
-  # Stage B: SGLang patch tests (1 GPU)
+  # Stage B: SGLang patch tests (8 GPU)
   stage-b-sglang:
     needs: [stage-a-fast]
     if: |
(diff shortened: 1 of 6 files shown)


guapisolo:

PR #1121, opened 2026-05-13
[loss refactor] [2] code style improvements

yueming-yuan:
## Summary

Follow-up to #753

### DIFF
--- a/miles/backends/training_utils/loss_hub/logits.py
+++ b/miles/backends/training_utils/loss_hub/logits.py
@@ -3,7 +3,7 @@ from collections.abc import Iterator
 
 import torch
 
-from miles.backends.training_utils.cp_utils import _allgather_cp_redistribute, get_logits_and_tokens_offset_with_cp
+from miles.backends.training_utils.cp_utils import allgather_cp_redistribute, get_logits_and_tokens_offset_with_cp
 from miles.backends.training_utils.parallel import get_parallel_state
 from miles.utils.ppo_utils import calculate_log_probs_and_entropy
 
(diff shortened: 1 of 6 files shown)


miles/backends/training_utils/loss_hub/losses.py
```diff
@@ -1,17 +1,31 @@
...
-from .logits import get_log_probs_and_entropy, get_values
+
+class LossFunction(Protocol):
+    def __call__(
+        self,
+        args: Namespace,
+        batch: RolloutBatch,
+        logits: torch.Tensor,
+        sum_of_sample_mean: Callable[[torch.Tensor], torch.Tensor],
+    ) -> tuple[torch.Tensor, dict[str, torch.Tensor]]: ...
```
Shi-Dong: Might be better to replace `...` with a docstring for this protocol.

miles/backends/training_utils/loss_hub/logits.py
```diff
@@ -3,11 +3,10 @@
 
 import torch
 
+from miles.backends.training_utils.cp_utils import _allgather_cp_redistribute, get_logits_and_tokens_offset_with_cp
```
Shi-Dong: I think a function is meant to be a module-private function if its name starts with an underscore. If we are to use `_allgather_cp_redistribute` in this module, we'd better rename it.

Shi-Dong (approved):
LGTM!  Two tiny nits.

Zhichenzzz (approved):
Lgtm!

PR #1133, opened 2026-05-14
fix: support Qwen3-VL Megatron training inputs

nikhilbarhate99:
## Summary
Hi, I ran into an issue while trying to run VLM training with Qwen3-VL. There were two issues:
  1. `get_rollout_data()` assumed every value in multimodal_train_inputs was already a tensor and called `.to(device)` on it. Qwen3-VL processor outputs can include list-backed values, so that path was breaks for VLM data.
  2. The Megatron forward/logprob/train path passed every multimodal key into the Megatron model. Qwen3-VL emits `mm_token_type_ids`, which is valid for the HF/SGLang side, but Megatron’s Qwen3-VL bridge does not accept that kwarg. That caused the training/logprob forward to fail once rollout data was handed back to Megatron.

 
This fix converts non-tensor multimodal values to CUDA tensors and filters `mm_token_type_ids` for the Megatron backend, while keeping the actual image/video tensors needed by the bridge model.

### DIFF
--- a/miles/backends/training_utils/data.py
+++ b/miles/backends/training_utils/data.py
@@ -18,6 +18,18 @@ from .parallel import get_parallel_state
 logger = logging.getLogger(__name__)
 
 
+def _move_multimodal_value_to_device(value, device: torch.device | int):
+    if torch.is_tensor(value):
+        return value.to(device=device)
+    return torch.as_tensor(value, device=device)
+
+
+def _skip_multimodal_training_key(args: Namespace, key: str) -> bool:
+    # Qwen3-VL processors emit mm_token_type_ids for HF/SGLang, but Megatron's
+    # Qwen3-VL bridge model consumes image/video tensors directly.
+    return getattr(args, "train_backend", None) == "megatron" and key == "mm_token_type_ids"
+
+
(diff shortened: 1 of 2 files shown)


coding-famer:
I believe a better solution would be to modify the processor side like:
https://github.com/THUDM/slime/blob/41dc3b6d21d3c75b212965077a1cc4117932f06d/slime/utils/processing_utils.py#L22-L40

miles/backends/training_utils/data.py
```diff
@@ -18,6 +18,18 @@
...
+def _move_multimodal_value_to_device(value, device: torch.device | int):
+    if torch.is_tensor(value):
+        return value.to(device=device)
+    return torch.as_tensor(value, device=device)
+
+
+def _skip_multimodal_training_key(args: Namespace, key: str) -> bool:
+    # Qwen3-VL processors emit mm_token_type_ids for HF/SGLang, but Megatron's
+    # Qwen3-VL bridge model consumes image/video tensors directly.
+    return getattr(args, "train_backend", None) == "megatron" and key == "mm_token_type_ids"
```
Jeffjeno, line 30: actually, it also met the problem in fsdp backend.

PR #1142, opened 2026-05-16
[TITO] Refactor/session verify args consume miles params Namespace

guapisolo:
## Summary
Session-verify wrapper consumes miles' `parse_args` Namespace instead of its own argparse.

## Motivation
`scripts/tools/verify_session_tito_tokenizer.py` maintained a second argparse with self-named parameters, so its flags drifted from miles' canonical names and multi-machine knobs (`--actor-num-nodes`, `--actor-num-gpus-per-node`) were not reachable. Consuming `miles.utils.arguments.parse_args` makes the wrapper speak miles' canonical flags; the only wrapper-only flag left is `--assistant-text-threshold`. The session-verify behavior, the model registry, and the emitted `train_args` are unchanged.

## Before / After
- **Before:** the CLI owned a second argparse whose parameter names were self-defined, not miles' canonical flags.
- **After:** the CLI calls `parse_args(add_custom_arguments=session_verify_extras)`, consuming miles' canonical flags.
- `run_session_verify(args: Namespace)` replaces the old multi-kwarg signature.
- **What moved where:** `build_train_args` becomes `namespace_to_train_args(ns)`, a pure Namespace -> string adapter over miles-canonical fields.
- **What moved where:** wrapper-invariant defaults centralize in `SESSION_VERIFY_INVARIANT_ARGS`, consumed identically by the CLI, the fast tests, the e2e `_common.py` driver.

## Behavior Preservation
- **How we know:** `test_session_verify_runner.py` asserts `namespace_to_train_args` keeps `--ci-test` enabled on the fsdp debug-rollout path.
- It also asserts the adapter honors default plus model-specific `rollout_max_response_len`.
- The refactor leaves every `test_session_server_multi_role/` `ModelConfig` row structurally unchanged.
- The sole field change is qwen3's per-family `assistant_text_threshold`, raised from the 0.2 default to `0.25`.
- qwen3's TITO assistant_text roundtrip drifts to 0.203 in CI, just past the old 0.2 gate.
- That gate is test-only; it does not alter the wrapper, adapter, or emitted `train_args`.

## Verification
- **Existing test suite:** `tests/fast/utils/test_utils/test_session_verify_runner.py` passes.

## Review Focus
- Scrutinize `namespace_to_train_args` for any miles-canonical field whose default differs from the old wrapper param, which would silently shift the emitted `train_args`.
- Scrutinize `SESSION_VERIFY_INVARIANT_ARGS` against miles `parse_args` defaults for `set_defaults` collisions.

### DIFF
--- a/miles/utils/test_utils/session_verify_runner.py
+++ b/miles/utils/test_utils/session_verify_runner.py
@@ -77,6 +77,7 @@ SESSION_VERIFY_INVARIANT_ARGS: dict[str, Any] = {
     "custom_agent_function_path": "miles.utils.test_utils.session_verify_agent.run_agent",
     "use_session_server": True,
     "debug_rollout_only": True,
+    "ci_test": True,
     "colocate": True,
     "train_backend": "fsdp",
 }
@@ -177,6 +178,8 @@ def _namespace_to_train_args(ns: argparse.Namespace) -> str:
         parts.append("--use-session-server")
     if ns.debug_rollout_only:
         parts.append("--debug-rollout-only")
+    if ns.ci_test:
+        parts.append("--ci-test")
     if ns.colocate:
         parts.append("--colocate")
     return " ".join(parts) + " "
@@ -214,9 +217,7 @@ def run_session_verify(args: argparse.Namespace) -> None:
     args.sglang_reasoning_parser, args.sglang_tool_call_parser = resolve_reasoning_and_tool_call_parser(
         args.tito_model, args.sglang_reasoning_parser, args.sglang_tool_call_parser
     )
-    args.tito_allowed_append_roles = sorted(
-        set(r.lower() for r in args.tito_allowed_append_roles) | {"tool"}
-    )
+    args.tito_allowed_append_roles = sorted(set(r.lower() for r in args.tito_allowed_append_roles) | {"tool"})
 
     _ensure_prompt_data()
     _clear_proxy_env()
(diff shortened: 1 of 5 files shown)


tests/fast/utils/test_utils/test_session_verify_runner.py
```diff
@@ -6,34 +7,51 @@
 
 import pytest
 
-from miles.utils.test_utils.session_verify_runner import _assert_session_verify_metrics, build_train_args
+from miles.utils.test_utils.session_verify_runner import (
+    SESSION_VERIFY_INVARIANT_ARGS,
+    _assert_session_verify_metrics,
+    _namespace_to_train_args,
```
Shi-Dong: If these two functions are used across modules, better remove the leading underscore from their names.

miles/utils/test_utils/session_verify_runner.py
```diff
@@ -78,139 +125,91 @@ def _clear_proxy_env() -> None:
...
+      disagree with the bound values raise ``ValueError`` here, before any
+      GPU work starts.
+    - ``args.hf_checkpoint`` is replaced with the local download path so the
+      composed train_args points at the downloaded model, not the HF id.
+    - ``args.tito_allowed_append_roles`` is normalized (lowercase, dedup,
+      ensure ``'tool'`` is in) to match the schedule contract in
+      ``session_verify_agent._SUPPORTED_ROLE_SURFACES``.
     """
     import miles.utils.external_utils.command_utils as U
     from miles.utils.chat_template_utils import resolve_reasoning_and_tool_call_parser
```
Shi-Dong: I'd suggest moving all imports to the beginning.

Shi-Dong (approved):
LGTM. Please address the comment before merging.

PR #1192, opened 2026-05-25
[cleanup] Remove prefill logprobs recomputation path

maocheng23:
Stacked on top of radixark/miles#1088.

## Summary

Removes the unused prefill-logprob recomputation path from Miles.

This was split out from the Qwen3 MoE SP+PP true-on-policy support PR so the functional SP+PP support and the controversial cleanup can be reviewed independently.

## Main Changes

- Remove the `--recompute-logprobs-via-prefill` CLI flag and validation.
- Remove the rollout hooks that called SGLang prefill scoring after generation.
- Delete the SGLang prefill scoring helper and its dedicated fast tests.
- Remove test assertions that only existed to verify the flag was absent from true-on-policy launch plans.
- Keep `--sglang-enable-prefill-only-deterministic-inference` available for any future model that genuinely needs prefill-only deterministic inference.

## Validation

Local split checks:

- [x] `git diff --check` for radixark/miles#1088 after the split
- [x] `git diff --check` for this stacked removal PR
- [x] `python -m py_compile` over changed Python files in radixark/miles#1088 after the split
- [x] `python -m py_compile` over changed Python files in this stacked removal PR

The prior focused pytest run for the affected true-on-policy/script tests hung locally without producing failures, so I stopped it and did not use it as validation evidence.


Zhichenzzz (approved):
LGTM! Thanks.

PR #1199, opened 2026-05-26
[2/12] Add Qwen3 MoE true-on-policy contract schema

maocheng23:
Stacked split of #1059, rebuilt after rebasing onto latest `main`.

Base branch: `split/pr1059-01-ci-gates`
Head branch: `split/pr1059-02-contract-schema`

This is PR 2 of 12. The stack is cumulative; the final branch is the rebased equivalent of the original PR head without stale-base reversions.

### DIFF
--- a/miles/true_on_policy/contracts.py
+++ b/miles/true_on_policy/contracts.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 
 from .schema import (
     QWEN3_DENSE_TRUE_ON_POLICY_V1_SCHEMA,
+    QWEN3_MOE_TRUE_ON_POLICY_V1_SCHEMA,
     KernelContract,
     LogprobContract,
     ModelFamily,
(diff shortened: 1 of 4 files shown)


miles/true_on_policy/contracts.py
```diff
@@ -50,10 +51,24 @@ def kernel_policy_kwargs_for(
         self,
         *,
         train_backend: str,
-        sglang_target: str,
+        parallel_layout=None,
```
yueming-yuan: qq, when will we use parallel_layout? why do we use this branch
```
if parallel_layout is not None:
...
else:
...
```

maocheng23: Moved this hunk into #1200. In this PR, `kernel_policy_kwargs_for` is back to the legacy `sglang_target: str` path, so there is no unused `parallel_layout` branch here anymore. PR #1200 now introduces `parallel_layout` and wires it from `TrueOnPolicyConfig.build_kernel_policy()`.

yueming-yuan (approved):

PR #1200, opened 2026-05-26
[3/12] Build true-on-policy launch config

maocheng23:
Stacked split of #1059, rebuilt after rebasing onto latest `main`.

Base branch: `split/pr1059-02-contract-schema`
Head branch: `split/pr1059-03-contract-config`

This is PR 3 of 12. The stack is cumulative; the final branch is the rebased equivalent of the original PR head without stale-base reversions.

### DIFF
--- a/miles/true_on_policy/config.py
+++ b/miles/true_on_policy/config.py
@@ -35,7 +35,10 @@ class TrueOnPolicyParallelLayout:
     train_tensor_parallel_size: int
     train_context_parallel_size: int
     train_pipeline_parallel_size: int
+    train_expert_model_parallel_size: int
+    train_expert_tensor_parallel_size: int
     rollout_num_gpus_per_engine: int
+    rollout_expert_parallel_size: int = 1
 
     @property
     def uses_train_tp(self) -> bool:
@@ -50,8 +53,20 @@ class TrueOnPolicyParallelLayout:
         return self.train_pipeline_parallel_size > 1
 
     @property
+    def uses_train_ep(self) -> bool:
+        return self.train_expert_model_parallel_size > 1
+
+    @property
+    def uses_train_expert_tp(self) -> bool:
+        return self.train_expert_tensor_parallel_size > 1
+
+    @property
     def uses_rollout_tp(self) -> bool:
         return self.rollout_num_gpus_per_engine > 1
+
+    @property
+    def uses_rollout_ep(self) -> bool:
+        return self.rollout_expert_parallel_size > 1
 
 
 @dataclass(frozen=True)
@@ -68,6 +83,14 @@ class TrueOnPolicyKernelPolicy:
     batch_invariant_mode: bool
     tp_invariant_row_linear: bool
     deterministic_tp_allreduce: bool
+    deterministic_moe_routing: bool
+    moe_topk_tiebreak: str | None
+    deterministic_moe_dispatch: bool
+    deterministic_moe_combine: bool
+    ep_invariant_moe: bool
+    sglang_attention_data_parallel_size: int
+    disable_sglang_cuda_graph: bool
+    rollout_tp_size: int
 
     def build_sglang_args(self) -> TrueOnPolicyArgList:
         values = [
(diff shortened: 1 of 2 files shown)


miles/true_on_policy/config.py
```diff
@@ -189,28 +242,72 @@ def validate(self) -> None:
...
+        self._validate_megatron_train_topology()
+
+    def _validate_megatron_train_topology(self) -> None:
+        if self.train_backend != "megatron" or self.train_world_size is None:
+            return
+
+        model_parallel_size = (
+            self.tensor_model_parallel_size * self.pipeline_model_parallel_size * self.context_parallel_size
+        )
+        if self.train_world_size % model_parallel_size != 0:
```
yueming-yuan: Curious why we need this check, i guess it is already required by megatron? is there anything special

miles/true_on_policy/config.py
```diff
@@ -189,28 +242,72 @@ def validate(self) -> None:
...
+            raise ValueError(
+                "Megatron true-on-policy train world size must be divisible by "
+                "TP * PP * CP "
+                f"({self.train_world_size} % {model_parallel_size} != 0)."
+            )
+
+        train_data_parallel_size = self.train_world_size // model_parallel_size
+        if self.expert_model_parallel_size > train_data_parallel_size:
```
yueming-yuan: if we already check `train_data_parallel_size % self.expert_model_parallel_size == 0` below, do we need this?

miles/true_on_policy/config.py
```diff
@@ -189,28 +242,72 @@ def validate(self) -> None:
...
+                "Megatron MoE true-on-policy requires train DP to be divisible by EP "
+                f"({train_data_parallel_size} % {self.expert_model_parallel_size} != 0)."
+            )
 
     def build_kernel_policy(self) -> TrueOnPolicyKernelPolicy:
+        policy_kwargs = self.contract.kernel_policy_kwargs_for(
+            train_backend=self.train_backend,
+            parallel_layout=self.parallel_layout,
+        )
+        policy_kwargs["rollout_tp_size"] = self.rollout_num_gpus_per_engine
```
yueming-yuan, line 283: Curious how we process the DP? If dp attn is not supported, why do we set https://github.com/radixark/miles/pull/1200/changes#diff-30181e1a2ccc00c3d65a969d05545bb1549739afd7575d36e5522715cb6658e7R106-R113

maocheng23: For this Qwen3-MoE true-on-policy contract, we do not enable SGLang DP attention. `sglang_attention_data_parallel_size` is pinned to `1`, so `build_sglang_args()` does not emit `--sglang-data-parallel-size` or `--sglang-enable-dp-attention`. The branch is only there for a future parity-gated DP-attention contract; the current verified path stays on EP-only rollout.

miles/true_on_policy/config.py
maocheng23: Agreed, removed the separate `EP > DP` branch. It was only a clearer diagnostic; for the valid positive-DP case, `train_data_parallel_size % expert_model_parallel_size != 0` already catches the invalid EP layout.

miles/true_on_policy/config.py
maocheng23: Agreed, removed the explicit `train_world_size % (TP * PP * CP)` error from this PR and will let Megatron own that topology validation. The true-on-policy config now only derives train DP for the remaining MoE EP divisibility check.

maocheng23:
Moved the `parallel_layout` branch from #1199 into this PR. Here it is used by `TrueOnPolicyConfig.build_kernel_policy()`, which passes `self.parallel_layout` into `kernel_policy_kwargs_for()`. The reason for the branch is that `sglang_target` can only represent the old dense `fsdp` vs `fsdp_tp` distinction, while Qwen3-MoE needs the full topology to derive train expert TP and train/rollout EP invariance flags. The fallback remains for the legacy caller path.

yueming-yuan (approved):

PR #963, opened 2026-04-09
Support DeepSeek v3.2

zianglih:
@humansand

## Summary

This PR adds DeepSeek V3.2 support for Miles training and rollout, including the 5-layer debug/e2e workload used to validate MXFP8 mixed-BF16 training with SGLang rollout.

## Changes

- Add DeepSeek V3.2 model launch/config entries and `scripts/run_deepseek_v32.py` for prepare/train workflows.
- Add MBridge support for `deepseek_v32`, including DSA indexer weight mappings and V3.2/GLM-5 RoPE format handling.
- Add runtime HF config compatibility registration for clean `deepseek_v32` checkpoints, so we no longer mutate checkpoints with local shim files or patched `config.json` fields.
- Treat missing `indexer_rope_interleave` as non-interleaved/`False`; GLM-5 keeps its explicit interleaved setting from config.
- Add `--freeze-indexer` and supporting argument/config plumbing.
- Add MXFP8 mixed-BF16 defaults for V3.2 KV projection layers and torch-dist conversion pipeline-model-parallel handling.
- Add a B200 e2e test for `Pinaster/DeepSeek-V3.2-5layer` covering BF16 prepare, MXFP8 conversion, Megatron conversion, SGLang rollout, mixed-BF16 MXFP8 training, and fault-tolerance recovery.

## Validation

- Local hygiene: `pre-commit run --all-files` passed.
- Clean-checkpoint B200 devbox run passed after deleting all generated model/checkpoint directories and rerunning:
  `PYTHONPATH=/root/miles python3 tests/ci/run_suite.py --hw cuda --suite stage-c-8-gpu-b200 --labels run-ci-megatron`
- Result: `Test Summary: 1/1 passed`, runtime `1488.07s`.
- Metrics from the run: max `train/train_rollout_logprob_abs_diff=0.024676986038684845`, max `train/kl_loss=0.0006870048237033188`, `train/ppo_kl=0.0`.
- Post-cleanup smoke check on the clean BF16 checkpoint: `AutoConfig` resolves `DeepseekV32Config deepseek_v32 False`, and `AutoTokenizer` loads `LlamaTokenizer`.

PR #1214, opened 2026-05-26
[refactor] hf_config: single entry point for HF config loading with alias + overrides

yueming-yuan:
## Summary

Add `miles/utils/hf_config.py` exposing `load_hf_config` as the single entry point for loading HF configs from local checkpoints. It:

- Registers miles-local `model_type` aliases (e.g. `deepseek_v32` -> `DeepseekV3Config`) before calling `AutoConfig`, so checkpoints don't need to be mutated and callers don't need to remember to register.
- Falls back to a namespace object built from `config.json` when `AutoConfig` can't parse the model_type (carried over from `hf_attention._load_hf_config`).
- Accepts an `overrides` dict applied via `setattr` after loading — patch fields like `max_position_embeddings` without touching the checkpoint.

Replaces 9 direct `AutoConfig.from_pretrained` call sites with `load_hf_config`, and removes the duplicated `_load_hf_config` from `miles_plugins/models/hf_attention.py`.

## Why

Until now, supporting a new `model_type` that transformers doesn't know about (V3.2 today, others later) required adding a `register_hf_config_compat()` call at every entry-point that loads HF configs. Easy to forget; failure mode is a confusing `AutoConfig` error far from the missing call site.

Binding register-then-load into one helper turns "forgetting to register" from a latent runtime bug into a non-issue.

## Relation to #963

#963 introduced `miles/utils/hf_config_compat.py` plus `register_hf_config_compat()` calls in 6 modules. This PR supersedes that file. When #963 rebases onto this branch, those 6 scattered `register_hf_config_compat()` calls become unnecessary and should be dropped.

## Test plan

- [x] \`pytest tests/utils/test_hf_config.py\` — 8 tests pass locally, covering:
  - \`overrides\` apply via setattr
  - \`trust_remote_code\` default and kwarg passthrough
  - Namespace fallback (incl. nested \`text_config\`)
  - Idempotent repeated calls
  - End-to-end \`deepseek_v32\` alias loading via real \`AutoConfig\` + DeepseekV3Config defaults
- [x] \`pre-commit run --files <all changed>\` passes

### DIFF
--- a/miles/backends/fsdp_utils/actor.py
+++ b/miles/backends/fsdp_utils/actor.py
@@ -8,12 +8,12 @@ import torch
 import torch.distributed as dist
 from ring_flash_attn import update_ring_flash_attn_params
 from tqdm import tqdm
-from transformers import AutoConfig
 
 from miles.ray.train_actor import TrainRayActor
 from miles.utils import train_dump_utils, train_metric_utils
 from miles.utils.context_utils import with_defer
 from miles.utils.distributed_utils import get_gloo_group
+from miles.utils.hf_config import load_hf_config
 from miles.utils.memory_utils import clear_memory, print_memory
 from miles.utils.processing_utils import load_processor, load_tokenizer
 from miles.utils.ray_utils import Box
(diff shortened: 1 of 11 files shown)


yueming-yuan:
Closing in favor of a stacked pair: (1) remove the now-outdated qwen3.5 HF config fallback in hf_attention, then (2) reintroduce hf_config with overrides + alias mechanism on top.

PR #1215, opened 2026-05-26
hf_attention: drop outdated namespace fallback for qwen3.5

yueming-yuan:
## Summary

Remove the custom \`_load_hf_config\` namespace fallback in \`miles_plugins/models/hf_attention.py\`. It existed to load qwen3.5 checkpoints when transformers didn't yet recognize the model_type. Recent transformers releases ship \`qwen3_5\` / \`qwen3_5_moe\` / \`qwen3_next\` natively (verified on transformers 5.3.0), so the fallback is dead weight.

## Changes

- Delete \`_load_hf_config\` from \`hf_attention.py\` (function + json/os imports it required)
- Replace 4 call sites with direct \`AutoConfig.from_pretrained(..., trust_remote_code=True)\`:
  - \`HuggingfaceAttention.__init__\`
  - \`qwen3_5.py:get_qwen3_5_layer_spec\`
  - \`qwen3_next.py:get_qwen3_next_layer_spec\`
  - \`tools/convert_hf_to_torch_dist.py\` (drop the try/except around AutoBridge.from_pretrained)

## Why

The namespace fallback wasn't a real Config instance:
- \`isinstance(cfg, SomeHFConfig)\` returned False
- No \`.to_dict\` / \`.save_pretrained\` / other Config methods
- Nested dicts (e.g. \`quantization_config\`) only unwrapped one level

Now that qwen3.5 is upstream, failing loud via the standard \`Unrecognized model_type\` error is preferable to silently routing to a half-shaped namespace.

## Test plan

- [x] \`pre-commit run --files <changed>\` passes
- [x] Confirmed \`transformers 5.3.0\` lists \`qwen3_5\`, \`qwen3_5_moe\`, \`qwen3_next\` in \`CONFIG_MAPPING_NAMES\`

PR #1216, opened 2026-05-26
[refactor] hf_config: single entry point with alias registration and overrides

yueming-yuan:
> **Stacked on top of #1215** — review/merge that first.

## Summary

Add `miles/utils/hf_config.py` exposing `load_hf_config` as the single entry point for loading HF configs from local checkpoints. It:

- Registers miles-local `model_type` aliases (e.g. `deepseek_v32` -> `DeepseekV3Config`) before calling `AutoConfig`, so checkpoints don't need to be mutated and callers don't need to remember to register.
- Accepts an `overrides` dict applied via `setattr` after loading — patch fields like `max_position_embeddings` without touching the checkpoint.

Replaces 9 direct `AutoConfig.from_pretrained` call sites with `load_hf_config`.

## Why

Supporting a new `model_type` that transformers doesn't know about (V3.2 today, others later) used to require adding a `register_hf_config_compat()` call at every entry-point that loads HF configs. Easy to forget; failure mode is a confusing `AutoConfig` error far from the missing call site.

Binding register-then-load into one helper turns "forgetting to register" from a latent runtime bug into a non-issue. Unknown `model_type` now fails loud — the fix is "add an entry to `_CONFIG_ALIASES`," not "add a register call somewhere."

## Relation to #963

#963 introduced `miles/utils/hf_config_compat.py` plus six scattered `register_hf_config_compat()` calls. This PR supersedes that file. When #963 rebases on top of this branch, those six register calls become unnecessary and should be dropped.

## Test plan

- [x] `pytest tests/utils/test_hf_config.py` — 6 tests pass locally:
  - `overrides` apply via setattr
  - `trust_remote_code` default and kwarg passthrough
  - Idempotent repeated calls
  - Unknown model_type raises (no silent degradation)
  - End-to-end `deepseek_v32` alias loading via real `AutoConfig` + DeepseekV3Config defaults
- [x] `pre-commit run --files <changed>` passes

### DIFF
--- a/tools/convert_fsdp_to_hf.py
+++ b/tools/convert_fsdp_to_hf.py
@@ -6,8 +6,10 @@ import time
 
 import torch
 import torch.distributed.checkpoint as dist_cp
-from transformers import AutoConfig, AutoModelForCausalLM
+from transformers import AutoModelForCausalLM
 from typing_extensions import override
+
+from miles.utils.hf_config import load_hf_config
 
 
 class UnpicklerWrapper(pickle.Unpickler):
@@ -117,7 +119,7 @@ def _convert_fsdp_to_hf(
 
     tensor_items = {k: v for k, v in state_dict.items() if isinstance(v, torch.Tensor)}
 
-    config = AutoConfig.from_pretrained(origin_hf_dir, trust_remote_code=True)
+    config = load_hf_config(origin_hf_dir)
     hf_model = AutoModelForCausalLM.from_config(config)
     target_keys = set(hf_model.state_dict().keys())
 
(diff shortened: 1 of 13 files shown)


guapisolo (approved):
LGTM.

Zhichenzzz (approved):

tools/convert_fsdp_to_hf.py
```diff
@@ -117,7 +119,7 @@ def _convert_fsdp_to_hf(
 
     tensor_items = {k: v for k, v in state_dict.items() if isinstance(v, torch.Tensor)}
 
-    config = AutoConfig.from_pretrained(origin_hf_dir, trust_remote_code=True)
+    config = load_hf_config(origin_hf_dir)
```
maocheng23, line 122: Can you please check this?
`load_hf_config()` returns a dynamic `DeepseekV32Config` subclass, but `AutoModelForCausalLM.from_config(config)` requires an exact registered config class. For `deepseek_v32`, this still raises `ValueError: Unrecognized configuration class`.

maocheng23 (approved):
Approved with one question

PR #1217, opened 2026-05-26
Change `get_deepscaler_rule_based_reward` fallback to full sequence

zianglih:
https://github.com/radixark/miles/pull/838

This model has the following format (https://huggingface.co/jdopensource/JoyAI-LLM-Flash/blob/main/chat_template.jinja):
```

(RolloutManager pid=1083890) [2026-05-05 04:06:01] sglang_rollout.py:425 - First rollout sample: ['<|begin▁of▁sentence|>You are JoyAI , a large language model trained by JD（京东）that can interact with a computer to solve tasks. Answer as concisely as possible.<|User|>Solve the following math problem step by step. The last line of your response should be of the form Answer: \\boxed{$Answer} where $Answer is the answer to the problem.\n\n记\n$$\na=\\lim _{n \\rightarrow+\\infty} \\sum_{k=0}^{n-1} \\frac{120}{\\sqrt{n^{2}+k n}}\n$$\n求取整 $[a]$ 的值.\n\nRemember to put your answer on its own line after "Answer:".<|Assistant|><|end_of_thought|>Alright, let\'s go step-by-step.\n\nWe have:\n\n$$\na = \\lim_{n \\to \\infty} \\sum_{k=0}^{n-1} \\frac{120}{\\sqrt{n^2 + kn}}\n$$\n\n---\n\n**Step 1: Interpret the sum**\n\nWe can rewrite the general term:\n\n$$\n\\frac{120}{\\sqrt{n^2 + kn}} = \\frac{120}{n \\sqrt{1 + \\frac{k}{n}}}\n$$\n\nbecause $\\sqrt{n^2 + kn} = n \\sqrt{1 + \\frac{k}{n}}$.\n\nSo:\n\n$$\na = \\lim_{n \\to \\infty} \\sum_{k=0}^{n-1} \\frac{120}{n} \\cdot \\frac{1}{\\sqrt{1 + \\frac{k}{n}}}\n$$\n\n---\n\n**Step 2: Recognize a Riemann sum**\n\nThis is a Riemann sum for the function $f(x) = \\frac{1}{\\sqrt{1+x}}$ on the interval $[0,1]$, with $n$ equal subintervals, right endpoints? Let\'s check:\n\nThe points are $x_k = \\frac{k}{n}$, for $k = 0, 1, \\dots, n-1$.\n\nActually, $k=0$ gives $x_0 = 0$, $k=n-1$ gives $x_{n-1} = \\frac{n-1}{n} \\to 1^-$.\n\nYes, it\'s a right-endpoint approximation if we consider $t_k = k/n$, but careful: $k$ runs from $0$ to $n-1$, so $x_k = k/n$ ranges from $0$ to $(n-1)/n$, which is slightly less than 1. That\'s standard for a Riemann sum over $[0,1]$ with $n$ partitions of width $1/n$, taking sample points at the left endpoints? Wait — check width:\n\nTerm is $\\frac{120}{n} \\cdot \\frac{1}{\\sqrt{1 + k/n}}$.\n\nHere $k/n$ is indeed the left endpoint if we think of subinterval $[k/n, (k+1)/n]$, but the function evaluated at $k/n$. That\'s a left Riemann sum for $f(x) = 1/\\sqrt{1+x}$ on $[0,1]$ with $n$ subintervals. But as $n \\to \\infty$, left or right sum doesn\'t matter for continuous functions.\n\nSo:\n\n$$\n\\sum_{k=0}^{n-1} \\frac{120}{n} \\cdot \\frac{1}{\\sqrt{1 + k/n}} \\to 120 \\int_0^1 \\frac{1}{\\sqrt{1+x}} \\, dx\n$$\n\n---\n\n**Step 3: Compute the integral**\n\nLet $u = 1+x$, $du = dx$, when $x=0$, $u=1$; when $x=1$, $u=2$.\n\n$$\n\\int_0^1 \\frac{1}{\\sqrt{1+x}} \\, dx = \\int_1^2 u^{-1/2} \\, du\n$$\n\n$$\n= \\left[ 2 u^{1/2} \\right]_{1}^{2} = 2\\sqrt{2} - 2\\sqrt{1} = 2\\sqrt{2} - 2\n$$\n\n---\n\n**Step 4: Multiply by 120**\n\n$$\na = 120 \\times (2\\sqrt{2} - 2) = 240\\sqrt{2} - 240\n$$\n\n---\n\n**Step 5: Check simplification**\n\nFactor $240$:\n\n$$\n240(\\sqrt{2} - 1)\n$$\n\nThat\'s the exact value.\n\n---\n\n**Step 6: Verify no missing factor**\n\nThe original sum: $ \\frac{120}{\\sqrt{n^2 + kn}} = \\frac{120}{n\\sqrt{1+k/n}} $, so yes $ \\frac{120}{n} \\times \\frac{1}{\\sqrt{1+k/n}} $, sum from $k=0$ to $n-1$ approximates $120 \\int_0^1 \\frac{dx}{\\sqrt{1+x}}$.\n\nEverything matches.\n\n---\n\nAnswer: \\boxed{240\\sqrt{2} - 240}<|end▁of▁sentence|>'], label: 90, reward: 0
```
image

### DIFF
--- a/miles/rollout/rm_hub/deepscaler.py
+++ b/miles/rollout/rm_hub/deepscaler.py
@@ -7,7 +7,7 @@ def get_deepscaler_rule_based_reward(response, label):
     elif "###Response" in response:
         model_solution = response.split("###Response")[1]
     else:
-        return 0
+        model_solution = response
 
     model_answer = extract_answer(model_solution)
     if model_answer is None:


zianglih:
let me see if we can explicitly handle `<|end_of_thought|>` per gemini suggestion

zianglih:
Closing since identical to https://github.com/THUDM/slime/pull/1170 , we should use `"--rm-type math "`.

PR #1224, opened 2026-05-27
ci: remove broken release-docs workflow

Zhichenzzz:
## What

Remove `.github/workflows/release-docs.yaml`.

## Why

The workflow still references files that were deleted in the docs cleanup (#1097, commit `bce1d21a5`):

- `docs/requirements.txt` — `pip install -r docs/requirements.txt` fails with `No such file or directory`
- `docs/build.sh` — the **Build documentation** step calls `bash ./build.sh en/zh`, which no longer exists

Every push touching `docs/**` therefore fails at the **Install dependencies** step (see [this run](https://github.com/radixark/miles/actions/runs/26492279119/job/78012453290)). The `apt-get` step is also missing `sudo`.

Since the docs were migrated away from the old sphinx setup, the workflow is fully out of sync. Removing it stops the failing CI noise; docs publishing can be reintroduced when it's reworked for the new docs structure.


guapisolo (approved):
LGTM

PR #1232, opened 2026-05-27
[feat] Gemma-4 model support

Zhichenzzz:
## Summary
Single-node RL training support for **Gemma-4 26B-A4B-it**

### What this adds
- **Model config + launch script** (`scripts/models/gemma-4-26b-a4b-it.sh`, `scripts/run-gemma-4-26b-a4b-it.sh`): dapo-math, GRPO, sglang **triton** rollout backend (gemma-4 has head_dim=512 global layers), `--use-rollout-routing-replay`.
- **Route to the Megatron-Bridge LLM provider** via a locally-rewritten config + `--megatron-to-hf-mode bridge`.
- **HF↔Megatron weight-iterator wiring** for the bridge path, plus `hf_validate_args` / model-arg fixes.
- **Median + p95-trimmed `train_rollout_logprob_abs_diff` metrics** (surface the distribution, not just the mean).

### Result
 train_rollout_logprob_abs_diff 0.006–0.016 in gsk8m RL training within 20 steps


miles/rollout/rm_hub/deepscaler.py
```diff
@@ -7,7 +7,11 @@ def get_deepscaler_rule_based_reward(response, label):
     elif "###Response" in response:
         model_solution = response.split("###Response")[1]
     else:
-        return 0
+        # Models that do not emit </think>/###Response markers (e.g. Gemma-4,
```
yueming-yuan: I think we discussed this and decided not to change this one..

yueming-yuan (approved):

PR #1234, opened 2026-05-27
fix: scope step-key strip to caller's exact key

mouad-hpc:
Previously mlflow_utils and TensorboardBackend stripped any metric ending in
"/step" before logging, which silently dropped legitimate user metrics that
happened to share the suffix. Now the dispatcher pulls the literal step_key
named by the caller, passes both step and step_key to each backend, and lets
each backend decide what to do with the step axis:

  - wandb: full dict (define_metric pairing already handles step_key)
  - mlflow: full dict + explicit step= (logs step_key as a metric too)
  - prometheus: full dict
  - tensorboard: drop only the exact step_key, then pass step=


yushengsu-thu (approved):

yushengsu-thu:
synced

yushengsu-thu (approved):

yushengsu-thu (approved):
please fix ci

mouad-hpc:
> please fix ci

CI problem disappeared after I updated the branch

Zhichenzzz (approved):

PR #1225, opened 2026-05-27
Add Terminal-Bench eval driver targeting miles_agent_server (DeepSeek-V4-Pro example)

Shi-Dong:
## Summary

Adds `examples/eval/terminal_bench_via_agent_server/`: a small, pure-client example that evaluates an OpenAI-compatible LLM API (default DeepSeek-V4-Pro) on the full 89-task Terminal-Bench 2.0 benchmark by driving a running `miles_agent_server` over its `/run` HTTP endpoint.

The `miles_agent_server.py` used here is the one shipped on the [`shi/rebase-on-upstream-v0.7.0`](https://github.com/radixark/harbor-private/tree/shi/rebase-on-upstream-v0.7.0) branch of `harbor-private` (which forward-ports harbor-private's customizations onto upstream Harbor v0.7.0). The server's `README.md` covers setup.

## How this differs from `examples/eval/terminal_bench/`

| | `terminal_bench/` (existing) | `terminal_bench_via_agent_server/` (this PR) |
|---|---|---|
| Eval backend | Terminal-Bench's `tb` CLI run as a host subprocess | An already-running `miles_agent_server` over HTTP |
| Concurrency | `tb`'s `--n-concurrent` | client-side `--max-concurrent` + the agent server's `--max-concurrent` |
| Reusability | one process per eval | a single long-lived server handles many evals (and training rollouts) |
| Closer to | local one-shot benchmarking | RL training-loop rollouts that hit the same server |

## What's in this PR

- `eval_tb_deepseek_v4_pro.py` — argparse-based async client, capped at `--max-concurrent` in-flight `/run` requests, streams per-trial rows to a `.jsonl` and writes an aggregated `.parquet` (via `polars`) at the end.
- `README.md` — prerequisites (running agent server, populated `$HARBOR_TASKS_DIR`, exported `DEEPSEEK_API_KEY`) plus usage, tuning notes, and a small `polars` snippet for post-hoc aggregation.

## Security

The LLM API key is read **only** from `$DEEPSEEK_API_KEY` (default; override with `--api-key-env`). It is never hard-coded, never put on the command line, and never written to disk by this script. The example uses `<your-key>` placeholders throughout the README.

## Test plan

- [ ] Start a `miles_agent_server` on the `shi/rebase-on-upstream-v0.7.0` branch of `harbor-private` per its README.
[...]

PR #1236, opened 2026-05-28
Add 2-node TB2 training example targeting the harbor-private agent server

Shi-Dong:
## Summary

Adds `examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/` — the **training-side companion** to `examples/eval/terminal_bench_via_agent_server/` (merged in #1225). Both examples target the `/run` endpoint of the [`miles_agent_server`](https://github.com/radixark/harbor-private/blob/shi/rebase-on-upstream-v0.7.0/miles_agent_server.py) shipped on the [harbor-private branch `shi/rebase-on-upstream-v0.7.0`](https://github.com/radixark/harbor-private/tree/shi/rebase-on-upstream-v0.7.0); this one drives it from the existing `run-glm47-flash-agentic-async.py` launcher instead of a one-shot Python client.

End-to-end picture:

```
trainer-side (rcli pod, this PR)               agent-server (external host)
+--------------------+                         +-----------------------------+
|  Megatron actor    |  weights -> sglang ---->|                             |
|  RolloutManager    |                         |   miles_agent_server.py     |
|     ^              |  POST /run         ---->|     (harbor-private branch) |
|     | GRPO updates |                         |     spawns Docker -> agent  |
+-----+--------------+                         |     -> verifier -> reward   |
                                               +-----------------------------+
```

## What's in this PR

- `README.md` — full 2-node recipe (prerequisites, dataset symlink, training branch with the threadpool fix, exact launcher invocation with all defaults, sanity-check sequence, tuning notes).
- `launch.sh` — thin bash wrapper around `run-glm47-flash-agentic-async.py` that pre-fills the 2-node defaults and threads a `<run-tag>` positional argument through `--save-dir`, `--save-traces-dir`, and `--wandb-run-name` so multiple attempts don't collide. Asserts that `/root/swe_train.jsonl` (the launcher's hardcoded prompt-data path) is symlinked, and exposes `RCLI_JOB_NAME` + `AGENT_SERVER_URL` env overrides for the few values that change across jobs.

**No new launcher logic.** The 5-node baseline in `run-glm47-flash-agentic-async.py` is unchanged; this PR is documentation + a small bash convenience wrapper.

## Why a separate 2-node example?

The 5-node config in `run-glm47-flash-agentic-async.py` is designed for production scale (4 rollout nodes, 1 trainer). For developers who want to:

- Smoke-test a new agent-server build (like the harbor-private rebase) without booking 5 nodes
- Iterate on training-loop changes with 1/2 the GPU cost
- Reproduce results on a smaller cluster

…the 2-node `--num-nodes 2 --train-num-nodes 1` layout (1 trainer + 1 rollout) is the minimum useful split. This PR captures the canonical recipe so the next person doesn't have to re-derive it.

## Test plan

- [ ] `rcli job new --name flash-pr-train --num-nodes 2 ...` on `sci-h200`.
- [ ] Start a `miles_agent_server` on the harbor-private branch on an external host (we use `aws-agent-server`), with the 23 TB2 variance tasks populated under `$HARBOR_TASKS_DIR`.
- [ ] Inside the head pod: checkout `shi/260421-glm47-flash-2node@83bcfa6`, symlink the 23-task TB2 jsonl to `/root/swe_train.jsonl`, run `bash launch.sh pr-smoke`.
- [ ] Confirm: Ray `RUNNING`, wandb URL appears, first `rollout 0:` / `step 0:` land within ~60 min, then 10 consecutive iters complete cleanly.
- [ ] Post training-side metrics (rollout/raw_reward, step/kl_loss, step/grad_norm) + agent-server-side dashboard snapshot back to this PR as evidence.

I'm executing the test plan now and will follow up with results.

### DIFF
--- a/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
+++ b/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
@@ -4,13 +4,26 @@
 # shi/rebase-on-upstream-v0.7.0 as the rollout backend.
 #
 # See README.md for prerequisites (running agent server, populated
-# /root/swe_train.jsonl, the threadpool-fix training branch).
+# /root/swe_train.jsonl, a current miles checkout).
 #
 # Usage:
 #   bash launch.sh <run-tag>            # e.g. bash launch.sh pr-smoke
 #
 # <run-tag> is threaded through --save-dir, --save-traces-dir, and
 # --wandb-run-name so multiple attempts don't collide.
+#
+# Optional env overrides (sensible defaults shown):
+#   AGENT_SERVER_URL=http://agent-server:8080
+#       Base URL of the running miles_agent_server.
+#   ROUTER_EXTERNAL_HOST=
+#       Hostname/FQDN that the trainer's session-server advertises back
+#       to the agent server, so the agent server can call into the
+#       trainer for tokenization. Required when the agent server cannot
+#       reach the trainer at the trainer's default service name (e.g. an
+#       off-cluster agent server with an explicit ingress/proxy hostname).
+#       Leave unset to omit --router-external-host.
+#   WANDB_TEAM=
+#       Wandb team to log to. Leave unset to omit --wandb-team.
 
 set -euo pipefail
 
@@ -20,40 +33,34 @@ if [ "$#" -lt 1 ]; then
 fi
 RUN_TAG="$1"
 
-# Resolve where the launcher script lives. We expect to be invoked from
-# the miles repo root (cwd-sensitive: run-glm47-flash-agentic-async.py
-# uses relative paths to other tooling and the prompt-data symlink).
 LAUNCHER="examples/experimental/swe-agent-v2/run-glm47-flash-agentic-async.py"
 if [ ! -f "$LAUNCHER" ]; then
     echo "error: $LAUNCHER not found. Run this script from the miles repo root." >&2
     exit 66
 fi
 
-# The rcli job name is used to construct the Tailscale-ingress FQDN that
-# the rollout router advertises back to the external agent server.
-# Default matches what we use; override via $RCLI_JOB_NAME.
-RCLI_JOB_NAME="${RCLI_JOB_NAME:-shidong-flash-pr-train}"
-
-# Dataset symlink check (the launcher hard-codes /root/swe_train.jsonl).
 if [ ! -e /root/swe_train.jsonl ]; then
     echo "error: /root/swe_train.jsonl missing. See README.md step 'Launch'." >&2
     exit 65
 fi
 
-# Make sure the rcli job's namespace can route to the external agent server
-# via the in-cluster ts-egress-aws-agent-server ExternalName.
-AGENT_SERVER_URL="${AGENT_SERVER_URL:-http://ts-egress-aws-agent-server:8080}"
+AGENT_SERVER_URL="${AGENT_SERVER_URL:-http://agent-server:8080}"
+ROUTER_EXTERNAL_HOST="${ROUTER_EXTERNAL_HOST:-}"
+WANDB_TEAM="${WANDB_TEAM:-}"
 
-python "$LAUNCHER" \
-    --num-nodes 2 --train-num-nodes 1 --skip-prepare \
-    --max-seq-len 65536 \
-    --save-dir "/workspace/GLM-4.7-Flash_2node_tb2_${RUN_TAG}/" \
-    --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces" \
-    --rollout-batch-size 4 --n-samples-per-prompt 8 --global-batch-size 32 \
-    --save-interval 5 \
-    --agent-server-url "$AGENT_SERVER_URL" \
-    --router-external-host "${RCLI_JOB_NAME}-ts-ingress.tail134ba0.ts.net" \
-    --wandb-project glm47-flash-agentic-async \
-    --wandb-team ch271828n-team \
-    --wandb-run-name "$RUN_TAG" \
+ARGS=(
+    --num-nodes 2 --train-num-nodes 1 --skip-prepare
+    --max-seq-len 65536
+    --save-dir "/workspace/GLM-4.7-Flash_2node_tb2_${RUN_TAG}/"
+    --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces"
+    --rollout-batch-size 4 --n-samples-per-prompt 8 --global-batch-size 32
+    --save-interval 5
+    --agent-server-url "$AGENT_SERVER_URL"
+    --wandb-project glm47-flash-agentic-async
+    --wandb-run-name "$RUN_TAG"
+)
+[ -n "$ROUTER_EXTERNAL_HOST" ] && ARGS+=(--router-external-host "$ROUTER_EXTERNAL_HOST")
+[ -n "$WANDB_TEAM" ] && ARGS+=(--wandb-team "$WANDB_TEAM")
+
+python "$LAUNCHER" "${ARGS[@]}" \
     2>&1 | tee "/tmp/flash-2node-launch-${RUN_TAG}.log"
--- a/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
+++ b/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# Launch a 2-node GLM-4.7-Flash agentic-async training run that uses the
+# miles_agent_server from the harbor-private branch
+# shi/rebase-on-upstream-v0.7.0 as the rollout backend.
+#
+# See README.md for prerequisites (running agent server, populated
+# /root/swe_train.jsonl, the threadpool-fix training branch).
+#
+# Usage:
+#   bash launch.sh <run-tag>            # e.g. bash launch.sh pr-smoke
+#
+# <run-tag> is threaded through --save-dir, --save-traces-dir, and
+# --wandb-run-name so multiple attempts don't collide.
+
+set -euo pipefail
+
+if [ "$#" -lt 1 ]; then
+    echo "usage: $0 <run-tag>" >&2
+    exit 64
+fi
+RUN_TAG="$1"
+
+# Resolve where the launcher script lives. We expect to be invoked from
+# the miles repo root (cwd-sensitive: run-glm47-flash-agentic-async.py
+# uses relative paths to other tooling and the prompt-data symlink).
+LAUNCHER="examples/experimental/swe-agent-v2/run-glm47-flash-agentic-async.py"
+if [ ! -f "$LAUNCHER" ]; then
+    echo "error: $LAUNCHER not found. Run this script from the miles repo root." >&2
+    exit 66
+fi
+
+# The rcli job name is used to construct the Tailscale-ingress FQDN that
+# the rollout router advertises back to the external agent server.
+# Default matches what we use; override via $RCLI_JOB_NAME.
+RCLI_JOB_NAME="${RCLI_JOB_NAME:-shidong-flash-pr-train}"
+
+# Dataset symlink check (the launcher hard-codes /root/swe_train.jsonl).
+if [ ! -e /root/swe_train.jsonl ]; then
+    echo "error: /root/swe_train.jsonl missing. See README.md step 'Launch'." >&2
+    exit 65
+fi
+
+# Make sure the rcli job's namespace can route to the external agent server
+# via the in-cluster ts-egress-aws-agent-server ExternalName.
+AGENT_SERVER_URL="${AGENT_SERVER_URL:-http://ts-egress-aws-agent-server:8080}"
+
+python "$LAUNCHER" \
+    --num-nodes 2 --train-num-nodes 1 --skip-prepare \
+    --max-seq-len 65536 \
+    --save-dir "/workspace/GLM-4.7-Flash_2node_tb2_${RUN_TAG}/" \
+    --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces" \
+    --rollout-batch-size 4 --n-samples-per-prompt 8 --global-batch-size 32 \
+    --save-interval 5 \
+    --sglang-mem-fraction-static 0.72 \
+    --agent-server-url "$AGENT_SERVER_URL" \
+    --router-external-host "${RCLI_JOB_NAME}-ts-ingress.tail134ba0.ts.net" \
+    --wandb-project glm47-flash-agentic-async \
+    --wandb-team ch271828n-team \
+    --wandb-run-name "$RUN_TAG" \
+    2>&1 | tee "/tmp/flash-2node-launch-${RUN_TAG}.log"
--- a/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
+++ b/examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
@@ -51,7 +51,6 @@ python "$LAUNCHER" \
     --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces" \
     --rollout-batch-size 4 --n-samples-per-prompt 8 --global-batch-size 32 \
     --save-interval 5 \
-    --sglang-mem-fraction-static 0.72 \
     --agent-server-url "$AGENT_SERVER_URL" \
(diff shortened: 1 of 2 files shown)


Shi-Dong:
## Test plan attempted — blocked by an unrelated docker image dependency conflict

### What worked

- `rcli job new --name flash-pr-train --num-nodes 2 ...` on `sci-h200` → job `shidong-flash-pr-train` (k8s id `01KSPN95BGZCPAP4DFF5R2KMET`) up, pods Running. (One transient asyncpg drop in the L3's connection pool on the first attempt; second attempt succeeded.)
- Restarted the rebased `miles_agent_server` (from `harbor-private` branch `shi/rebase-on-upstream-v0.7.0` @ `d06622b`) on `aws-agent-server:8080`. `/health` clean.
- Head pod setup: `fish` + `source /workspace/radixark-shi/radix_cli/k8s_startup.fish` + dataset symlink `/cluster_personal/job_workspaces/shidong-flash-5node/tb2_train.jsonl → /root/swe_train.jsonl` (23 records).
- Reset `origin` from the broken Mac local-path default (`file:///Volumes/x10/miles`) to GitHub HTTPS via the `$GITHUB_URL` PAT, then `git fetch + checkout shi/260421-glm47-flash-2node @ 83bcfa6` (threadpool fix verified at `miles/rollout/session/sessions.py:7,173,245`).
- Fetched this PR's `launch.sh` and ran it with `RUN_TAG=260527-glm47-flash-tb2-async-2node-pr-train-smoke`, `RCLI_JOB_NAME=shidong-flash-pr-train`. Ray submission succeeded (job id `raysubmit_T6XnFZUJtxNeWkK1`).

### What blocked (not PR-related)

The Ray training entrypoint died at `train_async.py` argument parsing, calling `AutoConfig.from_pretrained("/models/zai-org/GLM-4.7-Flash", trust_remote_code=True)`:

```
huggingface_hub.errors.HFValidationError:
  Repo id must be in the form 'repo_name' or 'namespace/repo_name':
  '/models/zai-org/GLM-4.7-Flash'.
```

Root cause: **`transformers==5.6.0`** on the `radixark/miles:dev` image went through `cached_files` → `hf_hub_download` → `validate_repo_id` even for absolute local paths, so the call rejects any `--hf-checkpoint` starting with `/`. Previous successful runs on this codebase used transformers 4.x where `cached_files` short-circuits on local paths.

Downgrading to `transformers==4.57.6` resolves the local-path issue **but** triggers the symmetric breakage on the other side:

```
sglang 0.5.13.dev20+gd9a0158 requires transformers==5.6.0
megatron-bridge 0.5.0+6fde1c85 requires transformers<=5.6.0,>=5.5.0
```

So the `radixark/miles:dev` image's pinned dependency graph is currently self-conflicting for this launcher's local-path code path. Neither version satisfies both layers.

This is **not** anything that the launcher wrapper in this PR can fix — `launch.sh` is just a thin wrapper around the existing `run-glm47-flash-agentic-async.py`. The fix lives in one of:

1. Update `train_async.py` / arg validation to handle absolute local paths without going through transformers' Hub validator (e.g. pre-resolve via `Path(args.hf_checkpoint).resolve()` and dispatch on `is_dir()`).
2. Re-pin the image's transformers to a version that does local-path short-circuit (and re-pin sglang/megatron-bridge to match).
3. Make `--hf-checkpoint` accept a HuggingFace repo id (`zai-org/GLM-4.7-Flash`) and download to a local cache.

I also hit one fresh-pod path gotcha worth flagging in the README: `/models/zai-org/GLM-4.7-Flash` doesn't exist on a freshly-allocated pod even though it's the launcher default — the actual local path on this cluster is `/models/GLM-4.7-Flash` (no `zai-org/`). I worked around it with `ln -sf /models/GLM-4.7-Flash /models/zai-org/GLM-4.7-Flash`. I'll update the PR's README to document this prerequisite.

### Agent-server validation status

The rebased `miles_agent_server` itself does **not** need this 2-node run to be validated end-to-end — the earlier 5-node run from this session ([wandb gaurk8q5](https://wandb.ai/ch271828n-team/glm47-flash-agentic-async/runs/gaurk8q5)) already ran 41 training iters through it, with 942+ trials over ~13 hours, 90.4% prefill cache hit, and harbor-private's `poll_steps` length-limit wrapper firing 149× as designed. The 2-node PR adds a smaller-scale recipe targeting the same server, which we *can't* validate via training right now because of the image dep conflict above. The recipe itself (PR docs + `launch.sh`) is structurally correct; everything up to the training entrypoint worked.

### Teardown

Tearing down `shidong-flash-pr-train` (rcli job → helm → pods) now to release the 2 H200 nodes.

Shi-Dong:
## Correction — earlier transformers diagnosis was wrong

On a fresh `radixark/miles:dev` 2-node rcli job (`shidong-flash-debug`), I re-ran the experiment described in the previous comment and the **transformers / `HFValidationError` story doesn't hold up**.

### Standalone python on the same image

With the symlink in place, no env tweaks, no downgrade:

```
transformers: 5.6.0
huggingface_hub: 1.15.0
HF / TRANSFORMERS env vars: (none)

/models/zai-org/GLM-4.7-Flash (symlink) → SUCCESS: Glm4MoeLiteConfig
/models/GLM-4.7-Flash (real path)       → SUCCESS: Glm4MoeLiteConfig
```

So `transformers==5.6.0` handles absolute local paths just fine in this image. My earlier "transformers 5.x removed local-path short-circuit" hypothesis was wrong — I jumped to it without verifying.

### Reproducing the launch on this fresh pod

Same training branch (`shi/260421-glm47-flash-2node @ 83bcfa6`), same launch.sh, same model symlink, same dataset symlink. This time `parse_args` succeeds, the launcher gets past `run-glm47-flash-agentic-async.py`, Ray submits the job (`raysubmit_pYfvzD4tVQfpj7CC`), and the failure now lands at:

```
ray.exceptions.RayTaskError(TypeError): ray::MegatronTrainRayActor.update_weights()

File "miles/utils/distributed_utils.py", line 78, in init_process_group
    pg, _ = _new_process_group_helper(
TypeError: _new_process_group_helper() got an unexpected keyword argument 'pg_options'
```

This is a `torch.distributed` API mismatch. `torch._C._distributed_c10d._new_process_group_helper` removed/renamed the `pg_options` keyword in a recent torch version; `miles/utils/distributed_utils.py:78` still passes it. The two are out of sync inside the current `radixark/miles:dev` image.

### What this means

Two corrections to the earlier comment:

1. **The HFValidationError was probably specific to the previous pod's state** (most likely my own `pip install "transformers<5"` polluted the env between the two attempts, and the second attempt picked up the downgrade for parse_args but then transformers 4.x had a different issue I misread). On a clean pod it doesn't reproduce.
2. **The actual blocker for training right now is a torch.distributed API drift**, not transformers. `miles/utils/distributed_utils.py:78` passes a kwarg the current torch doesn't accept.

This is still not in the PR's scope — `launch.sh` and the README don't introduce or interact with `pg_options`. It's an image-internal version pin issue.

### Where this leaves the PR

- PR #1236's docs + `launch.sh` are still structurally correct; the recipe matches what would have worked when the image was last in a self-consistent state.
- The eval-side PR #1225 (where I did run 89 trials against the same rebased agent server) already validated the agent server itself end-to-end. The training-side validation here is blocked by the image dep churn, not by anything in this PR.

I'm leaving the `shidong-flash-debug` 2-node job up in case you want to poke at the torch version issue. If not, I can tear it down whenever.

Shi-Dong:
## Update — `pg_options` bug fixed in-place, but training is still blocked one layer deeper

### Bug nailed

`miles/utils/distributed_utils.py:77` on training branch `shi/260421-glm47-flash-2node @ 83bcfa6`:

```python
# BROKEN: lexicographic string comparison
pg_options_param_name = "backend_options" if str(torch.__version__) >= "2.6" else "pg_options"
```

With `torch.__version__ == "2.11.0+cu129"`, `"2.11.0+cu129" >= "2.6"` is **False** (char-by-char: `'1' < '6'` at index 2), so it picks `pg_options` — which torch ≥ 2.10 no longer accepts. Hence the `TypeError: _new_process_group_helper() got an unexpected keyword argument 'pg_options'`.

**`origin/main` is already correct**: it uses `parse(torch.__version__) >= parse("2.6")` from `packaging.version`. So no new fix-PR is needed — the bug only lives on the older training branch this playbook still points at.

After applying the correct comparison in-place (`tuple(...split(".")[:2]) >= (2, 6)`), the launch goes past the TypeError and reaches wandb init: `wandb run 4rqum8hq`.

### But training still doesn't run

After the patch:
- Ray job submits successfully (`raysubmit_CZ7pSa3bTp33bUFs`, status RUNNING for >2h)
- sglang loaded the model and went idle around 09:21:43 UTC
- `MegatronTrainRayActor.update_weights()` succeeded (the patch worked)
- wandb run initialized cleanly
- Head-pod → `ts-egress-aws-agent-server:8080/health` returns `{"status":"ok"}` (connectivity is fine)
- Agent server `/api/sessions` is still `[]` (zero `/run` requests ever dispatched)
- RolloutManager has been emitting `Warning: No progress for 30.0s. Queue size: 0, Collected: 0/4` continuously since ~09:48 UTC

So **every component is individually healthy**, the pipeline just never starts producing rollouts. `--rollout-function-path fully_async_rollout.generate_rollout_fully_async` resolves to a real function (line 333 of `examples/fully_async/fully_async_rollout.py`), but its queue never populates.

This is a second, deeper layer of training-branch staleness against the current `radixark/miles:dev` image — beyond `pg_options`, something else in the async-rollout dispatch path between the older training branch and the newer image is mismatched. I couldn't bottom this out without diving deeper into miles's async-rollout queueing internals than is in this PR's scope.

### What this PR validates

- `launch.sh` runs cleanly end-to-end up through Ray submission, wandb init, and `update_weights()` — its arg surface is correct.
- The harbor-private agent-server itself works under training-style load — already validated end-to-end by PR #1225's 89-trial run earlier.

### What it can't validate today

- Whether the rebased harbor-private agent server actually serves a complete training iter (10 iters of step lines) from this 2-node configuration. The block is upstream of the agent server — in miles training-branch / image dep mismatch.

### Recommendation

Two paths to unblock:
1. **Update miles training-branch playbook** to a branch that has both the threadpool fix (currently on `shi/260421-glm47-flash-2node`) **and** the `pg_options` fix (currently on `origin/main`) **and** whatever else has drifted to make async-rollout dispatch work with the current image.
2. **Rebuild `radixark/miles:dev`** with pins compatible with the existing training branch.

Both are outside this PR's scope. The PR's docs + `launch.sh` themselves are correct.

I've stopped the stuck Ray job (`raysubmit_CZ7pSa3bTp33bUFs`); the rcli pods are still up if anyone wants to dive into the queue-empty mystery.

Shi-Dong:
## Validated end-to-end on `origin/main`

Per Shi's suggestion I re-ran the recipe with `origin/main` instead of an older topic branch. **Two small launcher-args drift fixes** were needed (now in commit `ff3007c43`); after that the run goes all the way through to actual training.

### What changed in this commit

- **`launch.sh`**: drop `--sglang-mem-fraction-static 0.72`. Main's `run-glm47-flash-agentic-async.py` no longer accepts the flag (sglang memory-fraction is set internally), so passing it triggers `No such option '--sglang-mem-fraction-static'`.
- **`README.md`**: switch the "training branch" prerequisite from a specific topic branch (`shi/260421-glm47-flash-2node`) to `origin/main`. Older topic branches accumulate stale launcher flags and out-of-date glue (e.g. `miles/utils/distributed_utils.py:77` was using a lexicographic string compare like `str(torch.__version__) >= "2.6"` which evaluates `False` for `2.10`/`2.11` and trips a `_new_process_group_helper(... pg_options=...)` `TypeError`; main fixed it with `parse(torch.__version__) >= parse("2.6")`).

### What I observed on this validation run

`wandb` run: https://wandb.ai/ch271828n-team/glm47-flash-agentic-async/runs/uvdni5mi

| step | UTC | raw_reward (prev rollout) | kl_loss | grad_norm | rollout_kl |
|---|---|---|---|---|---|
| 0 | 13:51:28 | 0.4375 | 0.000 | 1.23 | — |
| 1 | 14:05:38 | 0.1875 | 0.0116 | 0.50 | 0.029 |
| 2 | ~14:23 | 0.4063 | 0.0089 | 0.43 | 0.020 |
| 3 | ~14:40 | — | 0.0074 | 0.37 | 0.016 |
| 4 | ~14:57 | — | 0.0079 | 0.40 | 0.019 |
| 5 | ~15:14 | — | 0.0075 | 0.47 | 0.017 |
| 6 | ~15:31 | — | 0.0088 | 0.46 | 0.019 |

Cadence ~16 min/iter (within memory's expected 10-20 min). `kl_loss` stable in the 0.007-0.012 band (well below the lr=1e-5 collapse signature). `grad_norm` healthy 0.37-1.23. `ess_ratio=1.0` throughout. `pg_loss` ~0 — expected on this 23-prompt variance band with GRPO.

Agent-server side, session `260528-02-17-45`:
- 256 `/run` requests handled through step 6
- 227 completed (verifier-graded), 11 errored, 18 in flight
- The rebased harbor-private agent server keeps pace cleanly under the trainer's batch cadence

### Validation summary for this PR

- The recipe end-to-end works on `main` against the rebased harbor-private agent server.
- `launch.sh` and the README are now compatible with main's launcher.
- The agent server itself is unchanged from the eval-side validation (PR #1225, 89-trial smoke).

I'm leaving the 2-node devbox up long enough to confirm step 9 lands cleanly (ETA ~+45 min), then tearing down.

examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
```diff
@@ -0,0 +1,59 @@
...
+
+python "$LAUNCHER" \
+    --num-nodes 2 --train-num-nodes 1 --skip-prepare \
+    --max-seq-len 65536 \
+    --save-dir "/workspace/GLM-4.7-Flash_2node_tb2_${RUN_TAG}/" \
+    --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces" \
+    --rollout-batch-size 4 --n-samples-per-prompt 8 --global-batch-size 32 \
+    --save-interval 5 \
+    --agent-server-url "$AGENT_SERVER_URL" \
+    --router-external-host "${RCLI_JOB_NAME}-ts-ingress.tail134ba0.ts.net" \
```
Shi-Dong: Don't expose our tailnet suffix in this example. Use a generic host name.

examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
```diff
@@ -0,0 +1,59 @@
...
+    --wandb-project glm47-flash-agentic-async \
+    --wandb-team ch271828n-team \
```
Shi-Dong: Use a generic wandb team name.

examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
```diff
@@ -0,0 +1,59 @@
...
+LAUNCHER="examples/experimental/swe-agent-v2/run-glm47-flash-agentic-async.py"
+if [ ! -f "$LAUNCHER" ]; then
+    echo "error: $LAUNCHER not found. Run this script from the miles repo root." >&2
+    exit 66
+fi
+
+# The rcli job name is used to construct the Tailscale-ingress FQDN that
+# the rollout router advertises back to the external agent server.
+# Default matches what we use; override via $RCLI_JOB_NAME.
+RCLI_JOB_NAME="${RCLI_JOB_NAME:-shidong-flash-pr-train}"
```
Shi-Dong: Remove the RCLI_JOB_NAME env var (RCLI is an internal tool).

examples/experimental/swe-agent-v2/train-tb2-via-harbor-private-server/launch.sh
```diff
@@ -0,0 +1,66 @@
...
+
+AGENT_SERVER_URL="${AGENT_SERVER_URL:-http://agent-server:8080}"
+ROUTER_EXTERNAL_HOST="${ROUTER_EXTERNAL_HOST:-}"
+WANDB_TEAM="${WANDB_TEAM:-}"
+
+ARGS=(
+    --num-nodes 2 --train-num-nodes 1 --skip-prepare
+    --max-seq-len 65536
+    --save-dir "/workspace/GLM-4.7-Flash_2node_tb2_${RUN_TAG}/"
+    --save-traces-dir "/workspace/flash-2node-traces-${RUN_TAG}/traces"
```
Shi-Dong: Make these paths generic as well.

maocheng23 (approved):
Overall LGTM, but can we add a WANDB curve onto it?

PR #1237, opened 2026-05-28
[refactor] hf_tokenizer: route AutoTokenizer call sites through load_tokenizer

yueming-yuan:
> **Stacked on top of #1216** — review/merge that first.

## Summary

Extends #1216 from `AutoConfig` to `AutoTokenizer`. `AutoTokenizer.from_pretrained` falls back to `AutoConfig` to resolve `tokenizer_class` from `model_type` when `tokenizer_config.json` doesn't declare one, so it inherits the same unknown-`model_type` failure mode (e.g. `deepseek_v32`).

Builds on `bb0d43270` (which added `register_hf_config_aliases()` inside `load_tokenizer`) by also routing the 5 remaining bare `AutoTokenizer.from_pretrained` call sites through `load_tokenizer`:

- `tools/convert_to_hf.py`
- `tools/convert_hf_to_int4.py`
- `tools/convert_hf_to_hf_int4.py`
- `miles/utils/debug_utils/run_megatron/cli/prompt_utils.py`
- `miles/utils/debug_utils/run_megatron/worker/top_k_print.py`

Also retargets the `test_top_k_print` mock from `transformers.AutoTokenizer` to `miles.utils.processing_utils.load_tokenizer` (same class of patch-target issue as commit `ab2d103df` fixed for the native-support test) and updates the `register_hf_config_aliases` docstring to mention `load_tokenizer`.

## Why

#1216 collapses scattered `AutoConfig.from_pretrained` calls into a single `load_hf_config` entry point that registers miles aliases first. This PR finishes the symmetric job for tokenizers — making `load_tokenizer` the only tokenizer entry point in miles. The remaining direct `register_hf_config_aliases()` call is the one in `miles/backends/megatron_utils/initialize.py` before megatron's third-party `_build_tokenizer`, which we cannot route through `load_tokenizer`.

## Test plan

- [x] `pytest tests/utils/test_hf_config.py` — 9 tests pass
- [x] `pre-commit run --files <changed>` passes
- [ ] CI: `tests/fast/utils/debug_utils/run_megatron/worker/test_top_k_print.py` (requires sglang, not installable locally)

PR #1243, opened 2026-05-28
[ci] precommit: ban bare AutoConfig/AutoTokenizer.from_pretrained

yueming-yuan:
> **Stacked on top of #1237** — review/merge that first.

## Summary

Add a `pygrep` pre-commit hook that bans bare `AutoConfig.from_pretrained` and `AutoTokenizer.from_pretrained` calls. Forces all checkpoint-loading code through miles's wrappers (`load_hf_config` / `load_tokenizer`), so the model_type alias registration in #1216 stays load-bearing instead of being a thing every new caller has to remember.

## Why

#1216 / #1237 made `load_hf_config` and `load_tokenizer` the single entry points so that callers don't have to remember to `register_hf_config_compat()` for non-native `model_type`s (e.g. `deepseek_v32`). That guarantee only holds if nobody bypasses the wrappers. This hook makes a bypass a CI failure with a pointer to the right fix, instead of a runtime KeyError far from the offending line.

## Exclusions

- `miles/utils/hf_config.py`, `miles/utils/processing_utils.py` — the wrappers themselves
- `tests/` — mocks/patches legitimately reference these symbols
- `miles_plugins/`, `examples/` — same precedent as the existing `ban-mpu-get` hook

## Other changes

`miles/utils/arguments.py`: two stale references in a docstring and a comment pointed at `AutoConfig.from_pretrained` / `AutoTokenizer.from_pretrained`. Updated to the new wrapper names so the hook passes and the docs stop misleading.

## Test plan

- [x] `pre-commit run --all-files` passes on the diff
- [x] Verified hook fires on a synthetic violation: `Failed: x = AutoConfig.from_pretrained('foo')` returns exit code 1

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,4 +59,4 @@ repos:
         language: pygrep
         entry: '(AutoConfig|AutoTokenizer)\.from_pretrained'
         types: [python]
-        exclude: '(miles/utils/hf_config\.py|miles/utils/processing_utils\.py|tests/|miles_plugins/|examples/)'
+        exclude: '(miles/utils/hf_config\.py|miles/utils/processing_utils\.py|tests/|test_utils/|miles_plugins/|examples/)'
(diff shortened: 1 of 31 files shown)


guapisolo (approved):
LGTM.

guapisolo:
btw, shall we also excluded the test file for hf tokenizer?

yueming-yuan:
> btw, shall we also excluded the test file for hf tokenizer?

done

PR #1181, opened 2026-05-22
deepseek v4 model support

yueming-yuan:
ci-image-tag: test-20260601
ci-megatron-pr: #28
ci-sglang-pr: #27306

CI passed in 8eac60ea6508cfbbc2fd2a1d1230d10c5d84e6b5 commit.

PR #1248, opened 2026-05-29
support `--allgather-cp` for `bshd` tokens/loss + replay data

yueming-yuan:
## What

Extend the existing `allgather_cp` (contiguous context-parallel) data path — which main already supports for the `thd` qkv_format — to the `bshd` qkv_format and to replay data:

- **`get_batch` (`training_utils/data.py`)** — bshd `tokens` and `loss_masks`: pad each sample to `max_seqlen`, then take this CP rank's contiguous chunk, instead of the strided `slice_with_cp`.
- **replay-list register path (`megatron_utils/actor.py`)** — `replay_data` for both `bshd` and `thd`: slice/pad contiguously when `allgather_cp` so replayed per-token data matches the contiguous token layout. Previously this path always used the strided `slice_with_cp`, which is inconsistent with the contiguous token/loss layout under `allgather_cp`.

## Why

`allgather_cp` already exists in main and is used (e.g. `run_deepseek_v32.py`, `run_glm5_744b_a40b.py`) — but only the `thd` path had contiguous slicing in `get_batch`, and the replay path had none. This fills the `bshd` + replay gaps so contiguous CP is consistent across tokens, loss masks, and replay data.

## Contract

Contiguous CP requires `max_seqlen % cp_size == 0`. The caller pads `max_seq_len` to a `cp_size` multiple; the `assert`s enforce the contract (fail-fast). Model-agnostic — gated only on `args.allgather_cp` / `cp_size`. When `allgather_cp` is off, the code falls through to the existing `slice_with_cp` path, so **no behavior change** for current configs.

## Notes

- Split out of the DeepSeek-V4 work (#1181). V4 is the first consumer of `bshd` + `allgather_cp`; it supplies the `cp_size`-aligned padding (kept in #1181) that satisfies the divisibility contract here.
- No existing **non-V4** config combines `bshd` + `allgather_cp`, so this path isn't exercised by current CI models on its own; correctness is covered downstream by the V4 runs that turn it on.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -291,12 +291,7 @@ class MegatronTrainRayActor(TrainRayActor):
                     assert max_seqlen % cp_size == 0, f"max_seqlen {max_seqlen} must be divisible by cp_size {cp_size}"
                     local_len = max_seqlen // cp_size
                     start = cp_rank * local_len
-                    replay_data = [
-                        pad_func(r, max_seqlen - r.size(0))[start : start + local_len]
-                        if max_seqlen > r.size(0)
-                        else r[start : start + local_len]
-                        for r in replay_data
-                    ]
+                    replay_data = [pad_func(r, max_seqlen - r.size(0))[start : start + local_len] for r in replay_data]
                 else:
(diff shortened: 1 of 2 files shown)


maocheng23 (approved):

yueming-yuan:
i reverted the gemini suggested changes on padding because personally prefer easier to read and avoid silent errors

PR #1252, opened 2026-05-29
Retrieve indexer topk from rollout

yueming-yuan:
Part 4/9 of the indexer-replay (R3) rollout-replay stack.

Extract the indexer top-k from the rollout response into the replay batch (mirrors routed-experts retrieval).

Stacked on #1251 (base `indexer_replay/3`).

PR #1253, opened 2026-05-29
Register indexer replay manager in miles

yueming-yuan:
Part 5/9 of the indexer-replay (R3) rollout-replay stack.

Add IndexerReplayManager + the get_topk_fn replay mechanism. Pad/masked tokens (all-invalid rows) are filled with arange so they stay finite (avoids all -inf index_score -> softmax NaN -> 0*NaN gradient contamination); replay_indices_are_token_positions rebases per-sample 0-based rollout indices onto the packed training sequence.

Stacked on #1252 (base `indexer_replay/4`).


miles/backends/training_utils/replay_data.py
```diff
@@ -71,6 +89,12 @@ def pad_func(data, pad):
             replay_data = replay_data.reshape(batch_size * seqlen, num_layers, topk)
         else:
             replay_data = [slice_with_cp(r, pad_func, qkv_format) for r in replay_data]
+            if indices_are_token_positions:
```
maocheng23: This offset would be wrong with zigzag CP?

yueming-yuan: yeah cp is not fully supported, let me disable first

maocheng23 (approved):
LGTM, cp part please add a todo?

PR #1177, opened 2026-05-22
Move miles_megatron_plugins into the miles repo as a top-level package

Shi-Dong:
## Summary
- Add `miles_megatron_plugins/` at the top level of the miles repo (`__init__.py` + `true_on_policy/` with 13 modules; ~1400 lines).
- Update `setup.py` `find_packages(include=[...])` to expose the new package.
- Update `pyproject.toml` `[tool.isort]` (`known_first_party`, `src_paths`) for consistency with `miles` / `miles_plugins`.
- Bump the `third_party/Megatron-LM` submodule pointer to `c3a94af51` (the matching deletion on `radixark/Megatron-LM` — see [radixark/Megatron-LM#32](https://github.com/radixark/Megatron-LM/pull/32)).
- Temporarily set `.gitmodules` `branch = miles/move-plugins-to-miles` for `third_party/Megatron-LM`. A follow-up commit flips back to `branch = miles-main` once #32 merges.

## Why
`miles_megatron_plugins` is semantically miles' code that Megatron-LM imports back (true-on-policy contracts, sglang backend, matmul / attention / norm replacements). Today it lives inside the `radixark/Megatron-LM` fork and is only on `sys.path` because production sets `PYTHONPATH=/root/Megatron-LM`. That PYTHONPATH workaround doesn't survive packaging miles as a wheel for the eventual `pip install miles` goal.

Moving the package into miles makes it part of miles' regular Python package. The `from miles_megatron_plugins...` imports inside Megatron-LM source (transformer_layer, transformer_block, attention, gpt_model, gpt_layer_specs, distributed_data_parallel, tensor_parallel/{layers, matmul_tp_inv}, models/gpt/{gpt_model, gpt_layer_specs}, extensions/sglang) continue to work because miles is always installed before `megatron.core` is imported — both in production (Dockerfile installs miles last) and in CI (each pr-test job runs `Install miles` after the container starts).

This is Phase 1 of the broader plan toward `pip install miles`. Phase 2 (separate PR) will broaden Megatron-LM's `packages.find.include` so `pip install -e Megatron-LM` exposes `megatron.training` / `megatron.legacy` / `megatron.rl`, removing the remaining piece of the CPU runner's PYTHONPATH hack.

## What does NOT change
[...]

PR #1175, opened 2026-05-22
fix: nemotron H megatron bridge bug, recover test_dumper.py & test_run_megatron.py

guapisolo:
(cc generated)
## Summary
Fix global Nemotron-H shims and restore the short Megatron dumper checks.

## Symptom & Reproduction
- **Symptom:** `tests/e2e/short/test_dumper.py` failed before dump comparison because the Megatron source patcher could not find its target text in `TransformerLayer._forward_attention` / `TransformerLayer._forward_mlp`.
- **Reproduction:** `CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 python tests/e2e/short/test_dumper.py run --mode tp2_pp2_cp2_ep2_etp2`
- **Linked log:** `PatchApplicationError: match text not found in source: inference_context = deprecate_inference_params(...)`

## Root Cause
`miles_plugins/megatron_bridge/nemotron_h.py` installed Nemotron-H compatibility shims at plugin import time. A regular Megatron training actor importing the bridge plugins therefore saw wrapped `TransformerLayer._forward_attention` / `_forward_mlp` methods, so the dumper source patcher matched against wrapper source instead of the original Megatron methods. The dumper patch config also still matched `residual = hidden_states` after Megatron changed that assignment to `residual = getattr(self, "_sglang_pre_mlp_residual", hidden_states)`. Separately, `miles.utils.debug_utils.run_megatron.worker.batch.prepare_batch` needed CP slicing from worker-local `cp_rank` / `cp_size`, while `miles.backends.training_utils.cp_utils.slice_with_cp()` only read the process-global `get_parallel_state()`.

## Fix
The Nemotron-H plugin now registers the bridge subclass at import time and installs the Nemotron-H-only shims inside `MilesNemotronHBridge.provider_bridge()`. The THD and BSHD dumper patch configs now match Megatron's current `_forward_mlp` residual assignment. `slice_with_cp()` accepts an explicit `parallel_state` for the run_megatron worker path, and the disabled short CI registrations for `test_dumper.py` and `test_run_megatron.py` are restored.

## Verification
- **New / updated test:** `python -m pytest tests/fast/backends/training_utils/test_ulysses_cp_utils.py -q` passed: 5 tests.
- **New / updated test:** `python -m pytest tests/fast/utils/debug_utils/run_megatron/worker/test_batch.py -q` passed: 60 tests.
[...]

PR #1259, opened 2026-05-29
Fix stage-c GPU CI: test_dumper entrypoint

fzyzcjy:
Makes `main`'s `stage-c` GPU CI green again. Three independent pre-existing failures — none introduced by this PR. They surface now only because the GPU stages run **after** the CPU stages go green, so on most PRs they were masked by earlier-stage gating.

## 1. test_dumper — `Missing command` (`e33646ec4`)

`tests/e2e/short/test_dumper.py` is a Typer app (subcommands `run` / `compare`) whose `__main__` calls `app()`. The CUDA runner launches every registered test as bare `python3 <file>` (`tests/ci/ci_utils.py` → `run_unittest_files`), so with no subcommand Typer exits with `Missing command` (exit 2). It was disabled before and re-enabled in #1175 without wiring a bare-invocation entrypoint.

**Fix:** add `@app.callback(invoke_without_command=True)` that runs the full pipeline for every `CONFIGS` mode when no subcommand is given. `run` / `compare` stay for manual single-mode use. No assert/threshold logic changed.

## [fixed in other prs] 2. miles_megatron_plugins — `ModuleNotFoundError` (`a0c297190`)

The CUDA precision tests `tests/e2e/precision/test_hf_attention_cp_relayout.py` and `test_qwen3_5_cp_correctness.py` fail with:

```
ModuleNotFoundError: No module named 'miles_megatron_plugins'
```

### Why this is broken on `main` right now (cross-repo desync)

CI checks out the **Megatron-LM `miles-main`** branch fresh on every run. On **2026-05-17**, Megatron-LM commit `23924a0b6` ("[feat] Init true on policy with qwen_dense", @maocheng23) added to `megatron/core/transformer/transformer_layer.py`:

```python
from miles_megatron_plugins.true_on_policy.contracts import resolve_true_on_policy_runtime_policy
```

But the matching `miles_megatron_plugins` package was **never landed in the miles repo** — it currently lives only in @shi's in-flight migration PR #1177 (*Move miles_megatron_plugins into the miles repo*), which is ~36 commits behind `main` with WIP `REVERT BEFORE MERGE` commits.

So since 2026-05-17, **any** miles CI that pulls latest `miles-main` and imports `megatron.core.transformer` is missing the package → these precision tests fail on **every branch, `main` included**. It went unnoticed only because `stage-c` rarely runs (GPU stages are gated behind green CPU stages). The Megatron-LM side (the import) landed; the miles side (the package) did not.

**Fix:** vendor the `miles_megatron_plugins/` package as a top-level dir. It is self-contained — imports only `torch` + `megatron.core.*`, no submodule deps — and the existing `find_packages(include=["miles*"])` already packages it (also registered for isort). It is the deterministic "true-on-policy" backend for Megatron Core (CP zigzag relayout, sglang-matching attention/matmul/norm/rope), i.e. the code `transformer_layer.py` already depends on. Sourced from @shi's #1177 branch and reformatted to current `main` pre-commit.

> ⚠️ **Overlaps #1177.** This is the *package half* of @shi's migration, landed standalone to unblock CI. Whichever lands first, the other rebases. cc @shi.

## [fixed in other prs] 3. Pre-existing pre-commit debt (`7fbf22ea2`)

`pre-commit run --all-files` over `main` flags `scripts/run_qwen3_4b_npu.py` + `examples/eval/terminal_bench_via_agent_server/eval_tb_deepseek_v4_pro.py` (pure ruff/black formatting). `main`'s pre-commit is currently red on these, which also fails this PR's `Run pre-commit` check, so they are swept here.

## Verification

Local: bare `test_dumper` dispatches into the pipeline (no `Missing command`); `pre-commit run --all-files` is clean. Full 8-GPU correctness is verified by this PR's own `stage-c` run.


fzyzcjy:
pre commit -> @yueming-yuan 's pr

megatron -> @maocheng23 checking

.github/workflows/_run-ci.yml
```diff
@@ -146,12 +146,14 @@ jobs:
           cd /sgl-workspace/sglang && git reset --hard HEAD && git clean -fd && git fetch origin "$SGLANG_PR" && git checkout -f FETCH_HEAD && git log --oneline -1 && pip install -e python --no-deps --break-system-packages
           cd /root/Megatron-LM && git reset --hard HEAD && git clean -fd && git fetch origin "$MEGATRON_PR" && git checkout -f FETCH_HEAD && git log --oneline -1 && pip install -e . --no-deps --break-system-packages
           cd $GITHUB_WORKSPACE && pip install -e . --no-deps --break-system-packages
+          pip install polars --break-system-packages
```
fzyzcjy, line 149: can remove this line after docker image update

Shi-Dong (approved):
LGTM!

PR #1264, opened 2026-05-29
Support atomic weight update groups

yueming-yuan:
## Summary

Add generic infrastructure for atomic weight update groups in the direct Megatron-to-HF weight iterator.

The iterator now asks `megatron_to_hf.get_atomic_update_groups(model_name, param_infos)` for exact `AtomicUpdateGroup(key, names)` specs. It first builds indivisible update units from those specs, validates group invariants explicitly, and then packs the units into `update_weight_buffer_size` buckets.

The default implementation returns no groups, so existing models keep the same bucket behavior.

This PR intentionally does not add any DeepSeek-V4-specific grouping keys.

## Validation

- `python -m compileall miles/backends/megatron_utils/update_weight/hf_weight_iterator_direct.py miles/backends/megatron_utils/megatron_to_hf/__init__.py`
- `python -m pytest --confcutdir=tests/fast/backends/megatron_utils tests/fast/backends/megatron_utils/test_hf_weight_iterator_direct.py -q`
- `pre-commit run --files miles/backends/megatron_utils/megatron_to_hf/__init__.py miles/backends/megatron_utils/update_weight/hf_weight_iterator_direct.py`

Note: running the test through the repository root conftest in this local environment fails before test collection because local `sglang` is not installed (`ModuleNotFoundError: No module named 'sglang'`). The targeted command above isolates this pure unit test from that global fixture import chain.

### DIFF
--- a/miles/backends/megatron_utils/megatron_to_hf/__init__.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/__init__.py
@@ -16,7 +16,7 @@ from .qwen3moe import convert_qwen3moe_to_hf
 @dataclass(frozen=True)
 class AtomicUpdateGroup:
     key: str
-    names: tuple[str, ...]
+    suffixes: tuple[str, ...]
 
 
 # TODO unify w/ `convert_to_hf`
--- a/miles/backends/megatron_utils/megatron_to_hf/__init__.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/__init__.py
@@ -27,6 +27,10 @@ def convert_to_hf(args, model_name, name, param, quantization_config=None):
     return quantize_params(args, name, converted_named_tensors, quantization_config)
 
 
+def get_atomic_update_group(model_name, name):
+    return None
+
+
 # TODO optimize
 _cached_tensors = {}
 
--- a/miles/backends/megatron_utils/megatron_to_hf/__init__.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/__init__.py
@@ -35,7 +35,7 @@ def convert_to_hf(args, model_name, name, param, quantization_config=None):
     return quantize_params(args, name, converted_named_tensors, quantization_config)
 
 
-def get_atomic_update_groups(model_name, param_infos) -> list[AtomicUpdateGroup]:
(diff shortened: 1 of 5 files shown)


miles/backends/megatron_utils/megatron_to_hf/__init__.py
```diff
@@ -27,6 +27,10 @@ def convert_to_hf(args, model_name, name, param, quantization_config=None):
     return quantize_params(args, name, converted_named_tensors, quantization_config)
 
 
+def get_atomic_update_group(model_name, name):
+    return None
```
yueming-yuan: This will be filled in deepseek v4 PR. just a placeholder here to review this infra-only PR.

maocheng23 (commented):
This only works for colocate? How about the disagg case then

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
```diff
@@ -74,22 +88,34 @@ def _gather_and_update_expert_weights(
...
 
         if named_tensors:
             self._update_expert_bucket_weights(named_tensors, update_bucket_weight_func, pbar)
 
+    def _get_weight_transfer_update_units(self, is_expert: bool) -> list[list[tuple[str, torch.Tensor]]]:
+        named_tensors = list(collect_named_tensors_for_weight_transfer(self.args, self.model, is_expert=None))
+        atomic_update_groups = get_atomic_update_groups(self.args, self.model_name)
+        update_units = get_named_value_update_units(named_tensors, atomic_update_groups)
+        assert_named_value_update_units_are_homogeneous(update_units)
+        return [unit for unit in update_units if (".experts." in unit[0][0]) == is_expert]
```
Zhichenzzz: will `unit[0][0]` always work in all cases?

yueming-yuan: Addressed in `4a7c08824`: update units are non-empty by construction, and `_is_expert_update_unit` now asserts that invariant before reading the first name. The homogeneity check still guarantees the first name determines the whole unit expert/non-expert class.

Zhichenzzz (approved):

PR #1268, opened 2026-05-30
Fix Qwen3-VL THD packed mRoPE positions

maocheng23:
## Summary
- move the existing Qwen3-VL Bridge rotary compatibility monkey patch into a small helper
- synthesize packed `[3, 1, total_T]` mRoPE `position_ids` for Miles THD batches by splitting on `packed_seq_params.cu_seqlens_q` and calling Qwen3-VL `get_rope_index` per original sequence
- add focused tests for per-segment packed mRoPE construction, local-length mismatch skipping, and install-time patching

## Scope
This is independent from #1244. It is based directly on `main` and only contains the Qwen3-VL THD mRoPE fix.

## Tests
- `uvx ruff check miles/backends/megatron_utils/qwen_vl_packed_mrope.py tests/fast/backends/megatron_utils/test_qwen_vl_packed_mrope.py`
- `python -m py_compile miles/backends/megatron_utils/qwen_vl_packed_mrope.py tests/fast/backends/megatron_utils/test_qwen_vl_packed_mrope.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q -c /dev/null -p no:cacheprovider --confcutdir=tests/fast/backends/megatron_utils tests/fast/backends/megatron_utils/test_qwen_vl_packed_mrope.py`
- `uvx pre-commit run --all-files --show-diff-on-failure --color=never`

### DIFF
--- a/miles/backends/megatron_utils/__init__.py
+++ b/miles/backends/megatron_utils/__init__.py
@@ -21,21 +21,9 @@ except ImportError:
     logging.warning("deep_ep is not installed, some functionalities may be limited.")
 
 try:
-    from megatron.bridge.models.qwen_vl.modelling_qwen3_vl.text_model import (
-        Qwen3VLMoETextRotaryEmbedding,
-        Qwen3VLTextRotaryEmbedding,
-    )
+    from .qwen_vl_packed_mrope import install_qwen_vl_packed_mrope_patch
 
-    def patch_rotary_embedding(cls):
-        _original_forward = cls.forward
-
-        def _patched_forward(self, *args, packed_seq_params=None, **kwargs):
-            return _original_forward(self, *args, **kwargs)
-
-        cls.forward = _patched_forward
-
-    patch_rotary_embedding(Qwen3VLTextRotaryEmbedding)
-    patch_rotary_embedding(Qwen3VLMoETextRotaryEmbedding)
+    install_qwen_vl_packed_mrope_patch()
 except ImportError:
     pass
 
--- a/miles/backends/megatron_utils/qwen_vl_packed_mrope.py
+++ b/miles/backends/megatron_utils/qwen_vl_packed_mrope.py
@@ -1,0 +1,197 @@
+"""Local Qwen3-VL Bridge shims for Miles THD packed batches."""
+
+from __future__ import annotations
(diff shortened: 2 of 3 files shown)


miles/backends/megatron_utils/__init__.py
```diff
@@ -21,21 +21,9 @@ def new_init(self, *args, **kwargs):
...
-        _original_forward = cls.forward
-
-        def _patched_forward(self, *args, packed_seq_params=None, **kwargs):
-            return _original_forward(self, *args, **kwargs)
-
-        cls.forward = _patched_forward
-
-    patch_rotary_embedding(Qwen3VLTextRotaryEmbedding)
-    patch_rotary_embedding(Qwen3VLMoETextRotaryEmbedding)
+    install_qwen_vl_packed_mrope_patch()
```
TSunny007, line 26: Rename this and the corresponding files as `qwen3_vl...`?

TSunny007:
Looking at #1272, I think their analysis is right — `forward` resets `position_ids = None` internally, so injecting it via kwargs here won't survive the reset. Posting my CP-padded note over there instead.

TSunny007:
Screenshot 2026-06-01 at 3 26 24 PM


Can confirm that this stabilizes our sequence packed runs and the rewards we collect after this change

Zhichenzzz:
> Nice — cleaner than what we have locally.
> 
> One gap: `_try_build_packed_mrope_position_ids` uses `cu_seqlens_q`, but with CP + THD packing the input_ids are laid out against `cu_seqlens_q_padded` (see bridge's own `thd_to_bshd` at `modelling_qwen3_vl/model.py:271-275`). Right now this no-ops in that case and CP+packing regresses to the broken cross-sample positions.
> 
> Plan to handle that here, or as a follow-up? We have it working in https://github.com/GymPod/miles/pull/23 and can fold it in once this lands so we can drop our local patch.

could you try this PR patch? https://github.com/radixark/miles/pull/1272. btw, I have no access to your GymPod forked miles lol

Zhichenzzz:
Opened another PR #1272 to solve the problems, will close this PR.

PR #1270, opened 2026-06-01
ci: re-enable test_run_megatron parallelism-equivalence test

Shi-Dong:
## Summary

Re-enables `tests/e2e/short/test_run_megatron.py`, which was disabled with `disabled="Jiajun fix later"` and a `# FIXME: @guapisolo add back after fixing CI issue` comment.

The disable was a stale drive-by introduced in #1219 — the test passes today. The most likely underlying fix is #1227's `cp=SimpleNamespace` change, which addressed the context-parallel argument plumbing the comparison run depends on.

This test runs the `tp1` baseline vs `tp2/pp2/cp2` target via `run-and-compare` and asserts the Megatron parallelism layouts produce equivalent forward/backward outputs (logprob comparator + tensor-diff comparator, strict zero-failure criterion). No assert logic is touched — only the registration marker is restored.

## Test plan

- [x] Reproduced end-to-end on an 8×H200 devbox (image `radixark/miles:dev`): `python tests/e2e/short/test_run_megatron.py run --mode tp1_vs_tp2pp2cp2`.
- [x] **Logprob Comparison: PASSED** — Max abs diff `4.13e-01`, Mean abs diff `4.40e-02` ≤ `0.06` threshold.
- [x] **Tensor diff summary: total 44, passed 30, failed 0, skipped 14, errored 0** — zero failures, satisfying the file's strict "all-passed" contract.
- [x] `run_suite.py --hw cuda --suite stage-c-8-gpu-h100 --match-all-labels --list-only` now lists the test under **Enabled** (previously Skipped).

### DIFF
--- a/tests/e2e/short/test_run_megatron.py
+++ b/tests/e2e/short/test_run_megatron.py
@@ -35,8 +35,7 @@ NUM_LAYERS: int = 5
 
 _RUN_DIR: Path = Path(tempfile.mkdtemp(prefix="test_run_megatron_"))
 
-# FIXME: @guapisolo add back after fixing CI issue
-register_cuda_ci(est_time=2000, suite="stage-c-8-gpu-h100", labels=["short"], disabled="Jiajun fix later")
+register_cuda_ci(est_time=2000, suite="stage-c-8-gpu-h100", labels=["short"])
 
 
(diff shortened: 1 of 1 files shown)


guapisolo (approved):
LGTM

PR #1272, opened 2026-06-01
feat: Fix Qwen3-VL THD packed mRoPE positions

Zhichenzzz:
## Problem

Under THD sequence packing, several variable-length sequences are concatenated into one
row `input_ids = [1, total]`, with `packed_seq_params.cu_seqlens_q` marking the segment
boundaries. Megatron-Bridge's `Qwen3VLModel.forward` resets `position_ids = None` and then
recomputes MRoPE positions via the module-level `get_rope_index` over the **whole packed
row**, so positions run monotonically across segment boundaries instead of restarting per
sequence.

This is harmless for text (RoPE is relative — a constant per-segment offset cancels), but
**wrong for multimodal**: image/video `(t, h, w)` grid positions do not commute with a
whole-row offset, so the positions become bogus for packed VL training.

## Why not just inject `position_ids`

Passing `position_ids` into `forward` from the outside does not work: `forward` discards it
at the `position_ids = None` reset. Also, the current Bridge `Qwen3VLModel` has **no
`get_rope_index` method** (it calls the module-level free function), so wiring that assumes a
`get_rope_index` method does not attach.

## Fix (miles-side, no Megatron-Bridge edit)

Hijack the internal call instead: wrap `Qwen3VLModel.forward` to build correct per-segment
positions (slice the `[1, total]` row by `cu_seqlens_q`; text-only segment → linear `0..L`,
media segment → per-segment `get_rope_index` with the matching grid slice; concat to
`[3, 1, total]`) and stash them on a thread-local; patch the module-level `get_rope_index`
to return the stash. The reset is then harmless — the following `get_rope_index` call yields
our positions, which flow through `preprocess_packed_seqs` (THD/CP) unchanged. Only the
position values change; tensor shapes and the downstream flow are identical.

Guarded to THD single-row packed input; otherwise Bridge runs normally (e.g. CP-sharded or
`[batch, seq]` falls back). Replaces the prior 1-line rotary-only monkey-patch (the rotary
signature shim is kept).

## Validation

- Per-segment position check: packed multimodal positions equal running each segment
  unpacked exactly (e.g. seg B t-channel `[0,1,2,...]` vs the bug's `[43,44,45,...]`).
- End-to-end Qwen3-VL RL (geo3k, `Qwen3-VL-2B-Instruct`, megatron + THD, 8×H200): job
  succeeds, 8 training steps with a weight update each, gradients flow, and the
  train↔rollout logprob abs-diff stays small/stable (~0.011–0.016) throughout.

### DIFF
--- a/miles/backends/megatron_utils/__init__.py
+++ b/miles/backends/megatron_utils/__init__.py
@@ -21,25 +21,6 @@ except ImportError:
     logging.warning("deep_ep is not installed, some functionalities may be limited.")
 
 try:
-    from megatron.bridge.models.qwen_vl.modelling_qwen3_vl.text_model import (
-        Qwen3VLMoETextRotaryEmbedding,
-        Qwen3VLTextRotaryEmbedding,
-    )
-
-    def patch_rotary_embedding(cls):
-        _original_forward = cls.forward
-
-        def _patched_forward(self, *args, packed_seq_params=None, **kwargs):
-            return _original_forward(self, *args, **kwargs)
-
-        cls.forward = _patched_forward
-
-    patch_rotary_embedding(Qwen3VLTextRotaryEmbedding)
-    patch_rotary_embedding(Qwen3VLMoETextRotaryEmbedding)
-except ImportError:
-    pass
-
-try:
     import miles_plugins.megatron_bridge  # noqa: F401
 except Exception as _e:  # best-effort; not every environment uses megatron.bridge
     logging.warning("miles megatron.bridge plugins failed to load: %s", _e)
--- a/miles_plugins/models/qwen3_vl.py
+++ b/miles_plugins/models/qwen3_vl.py
@@ -1,0 +1,160 @@
+"""miles-side fix for Qwen3-VL THD packed mRoPE positions (no Megatron-Bridge edit).
+
+Bridge's Qwen3VLModel.forward resets position_ids=None and recomputes via the
+module-level get_rope_index over the whole [1, total] packed row, so MRoPE positions
+don't restart per packed segment (wrong for multimodal). We hijack that call: stash
+correct per-segment positions and have a patched get_rope_index return them.
+"""
+
+from __future__ import annotations
+
+import importlib
+import logging
+import threading
+
+import torch
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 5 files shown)


TSunny007:
`_build_packed_positions` uses `cu_seqlens_q` and bails when `cu[-1] != flat.numel()`. With CP + THD packing the input_ids are laid out against `cu_seqlens_q_padded` (see bridge's `thd_to_bshd` at `modelling_qwen3_vl/model.py:271-275`), so this no-ops in that case and CP+packing falls back to the broken cross-sample positions.

Plan to handle that here, or as a follow-up?

miles/backends/megatron_utils/__init__.py
```diff
@@ -21,23 +21,12 @@ def new_init(self, *args, **kwargs):
...
-        def _patched_forward(self, *args, packed_seq_params=None, **kwargs):
-            return _original_forward(self, *args, **kwargs)
-
-        cls.forward = _patched_forward
-
-    patch_rotary_embedding(Qwen3VLTextRotaryEmbedding)
-    patch_rotary_embedding(Qwen3VLMoETextRotaryEmbedding)
-except ImportError:
-    pass
+    install_qwen_vl_packed_mrope_patch()
```
TSunny007: Can you name this `qwen3_vl`

TSunny007:
Screenshot 2026-06-01 at 3 26 24 PM

Our rewards are looking good with this change on sequence packed examples

Zhichenzzz:
> `_build_packed_positions` uses `cu_seqlens_q` and bails when `cu[-1] != flat.numel()`. With CP + THD packing the input_ids are laid out against `cu_seqlens_q_padded` (see bridge's `thd_to_bshd` at `modelling_qwen3_vl/model.py:271-275`), so this no-ops in that case and CP+packing falls back to the broken cross-sample positions.
> 
> Plan to handle that here, or as a follow-up?

Thanks for raising the issue here. I will fix this in later VL PRs.

miles_plugins/models/qwen3_vl.py
yueming-yuan, line 1: I would prefer to move this to `/miles_plugins/models/qwen3-vl` (or similar place), given we don't want to put patches in miles core logic..

yueming-yuan (approved):
approved (and better to clean up the patches in the future)

PR #1279, opened 2026-06-02
Fix flaky real-Ray stop_cell tests by polling for actor death

fzyzcjy:
## Problem

`tests/fast/ray/rollout/real_ray/test_rollout_manager.py` has flaky failures (e.g. `test_stop_cell_kills_target_engine_only`): `DID NOT RAISE (RayActorError, RayTaskError)`.

## Root cause

`RolloutManager.stop_cell` → `server_group.stop_engines` kills the engine actor via `ray.kill`, which tears the actor down **asynchronously**. The tests asserted a single `health_generate` call raises immediately afterwards, racing the kill: when the actor is not yet dead the mock returns `True` and `pytest.raises` sees DID NOT RAISE. The fixture uses a real Ray cluster (not local_mode), so the race is genuine.

## Fix

Replace the three single-shot kill-then-assert sites with `_assert_engine_dies`, which polls the health call until it raises a Ray actor error within a deadline (failing only if the actor never dies).

### DIFF
--- a/tests/fast/ray/rollout/real_ray/test_rollout_manager.py
+++ b/tests/fast/ray/rollout/real_ray/test_rollout_manager.py
@@ -126,9 +126,11 @@ async def _assert_engine_dies(actor_handle, *, deadline_s: float = 15.0, poll_in
     deadline = time.monotonic() + deadline_s
     while True:
         try:
-            await actor_handle.health_generate.remote(timeout=1.0)
+            ray.get(actor_handle.health_generate.remote(timeout=1.0), timeout=5.0)
         except (ray.exceptions.RayActorError, ray.exceptions.RayTaskError):
             return
+        except ray.exceptions.GetTimeoutError:
+            pass
         if time.monotonic() >= deadline:
             pytest.fail(f"engine actor still alive {deadline_s}s after stop_cell")
         await asyncio.sleep(poll_interval_s)
--- a/tests/fast/ray/rollout/real_ray/test_rollout_manager.py
+++ b/tests/fast/ray/rollout/real_ray/test_rollout_manager.py
@@ -123,24 +123,12 @@ def _make_test_args(tmp_path, *, models: list[tuple[str, bool]]):
 
 
 async def _assert_engine_dies(actor_handle, *, deadline_s: float = 15.0, poll_interval_s: float = 0.2) -> None:
-    """Poll ``actor_handle`` until a health call raises a Ray error.
-
-    ``stop_cell`` tears the actor down via ``ray.kill``, which is asynchronous:
-    the actor process may briefly outlive the call, so a single health check
-    right after ``stop_cell`` races the kill and can still reach a live actor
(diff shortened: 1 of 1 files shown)


yueming-yuan (approved):

fzyzcjy:
Verified locally on a remote H200 job (real Ray cluster, mock engines):

- **Stress loop** — ran `TestStartStopCell` + `TestCellDispatchAcrossModels` 10× in a row: **10/10 green (5 passed each)**, no flakiness.
- **Full suite** — `tests/fast/ray/rollout/real_ray/`: **32 passed** in ~39s with the final (review-addressed) code.

The previously flaky `test_stop_cell_kills_target_engine_only` (and the two sibling `stop_cell` assertions) now poll until the actor is confirmed dead instead of asserting on a single health call that raced `ray.kill`'s async teardown.

Addressed the review feedback: switched to a `time.monotonic()` deadline and moved the blocking `ray.get` onto a thread via `asyncio.to_thread` so it no longer blocks the event loop.

fzyzcjy:
## Summary of the fix and CI investigation

### Original flaky test
`stop_cell` → `server_group.stop_engines` kills the engine actor via `ray.kill`, which tears the actor down **asynchronously**. The three `stop_cell` tests asserted a single `health_generate` call raised *immediately* afterwards, racing the kill — when the actor wasn't dead yet the mock returned `True` and `pytest.raises` saw `DID NOT RAISE`. (The fixture uses a real Ray cluster, not local_mode, so the race is genuine.)

**Fix:** replaced the three single-shot assertions with a polling helper `_assert_engine_dies` that retries the health call until it raises a Ray actor error within a deadline.

### Why it took a few iterations
1. First helper used `ray.get` + a summed-interval timer.
2. Per review, switched to a `time.monotonic()` deadline and moved the blocking `ray.get` onto a thread via `asyncio.to_thread` — this **passed** `stage-a-cpu`.
3. Tried the cleaner native `await <ObjectRef>` form. It passed locally but **failed `stage-a-cpu`**. Root cause: the CPU CI runner uses **uvloop**, and under uvloop `await`-ing the ObjectRef of a just-`ray.kill`-ed actor **never resolves**, hanging the test. Reproduced the hang in-container with `uvloop.install()`.
4. **Final:** poll with a blocking `ray.get(...)` + `time.monotonic()` deadline. It's event-loop-policy independent and matches the `ray.get(...health_generate.remote()...)` pattern already used throughout this file. Verified 5/5 under uvloop, 10/10 under the default loop, and full `real_ray` suite (32 passed).

### Unrelated pre-commit failure
A `ruff F811 Redefinition of IndexerReplayManager` (`replay_base.py`) appeared in `Run pre-commit`. This branch doesn't touch that file; it was a transient duplicate produced by the PR merge ref against an in-flux `main` (replay code churn). Merging current `origin/main` resolved it.

### Result
`stage-a-cpu / run-cpu`, `stage-b-cpu / run-cpu`, `Run pre-commit`, CodeQL and both Analyze checks are all green.

PR #1284, opened 2026-06-03
Nemotron RL support

Zhichenzzz:
ci-sglang-pr: #27110
ci-megatron-pr: #51

https://github.com/sgl-project/sglang/pull/27110
https://github.com/radixark/Megatron-LM/pull/51


miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
```diff
@@ -269,6 +269,10 @@ def update_weights(self) -> None:
                 refs, long_lived_tensors = self._send_base_params(hf_named_tensors)
                 results = ray.get(refs)
                 _check_weight_sync_results(results, is_lora=False)
+                # Reap CUDA-IPC pending frees; under torch_memory_saver
+                # offload nothing else triggers the lazy reap (leaks ~full
+                # export size per round otherwise).
+                torch.cuda.ipc_collect()
```
yueming-yuan: is this needed? in my understanding ipc should be automatically collected in weight update (so other models work well)

miles/backends/megatron_utils/actor.py
```diff
@@ -745,6 +745,9 @@ def update_weights(self, info: "EnginesAndLock") -> None:
         with torch_memory_saver.disable() if self.args.offload_train else nullcontext():
             print_memory("before update_weights")
             self.weight_updater.update_weights()
+            # Reap CUDA-IPC pending frees from any updater path; under
+            # torch_memory_saver offload nothing else triggers the lazy reap.
+            torch.cuda.ipc_collect()
```
yueming-yuan: (same q here)

tools/convert_hf_to_torch_dist_bridge.py
yueming-yuan, line 1: qq: why do we need this script? given that we usually just load hf weight directly when we use megatron bridge

yueming-yuan: (I removed this file, but correct me if it is needed)

Zhichenzzz: should remove it

miles/backends/megatron_utils/update_weight/update_weight_from_tensor.py
Zhichenzzz: repeated ipc collect, will delete this and that one in sglang

yueming-yuan (approved):

PR #1294, opened 2026-06-04
[AMD] Merge MI300/MI350-5 Dockerfiles

JessicaJiang-123:
## Summary

This PR consolidates the two separate ROCm Dockerfiles (one for gfx942 / MI300-MI325, one for gfx950 / MI350-MI355X) into a single parameterized `docker/Dockerfile.rocm`. It builds on a prebuilt `lmsysorg/sglang:${SGLANG_IMAGE_TAG}` base and selects the target via build args (`GPU_ARCH`, `SGLANG_IMAGE_TAG`), with per-variant configuration moved into `docker/build.py`. ~~The PR also adds a scheduled GitHub Actions workflow that builds and pushes both ROCm variants daily.~~

Nightly image build for this Dockerfile lives in sglang: https://github.com/sgl-project/sglang/pull/27765

## Changes

- **Single merged ROCm Dockerfile** (`docker/Dockerfile.rocm`): one parameterized build for both arches.
  - gfx950 / MI350-MI355X uses base tag `v0.5.10-rocm720-mi35x` (default).
  - gfx942 / MI300-MI325 uses base tag `v0.5.10-rocm700-mi30x` via build args.
- **`docker/build.py`**: adds `rocm-mi350` (gfx950) and `rocm-mi300` (gfx942) variants.
- ~~**Daily workflow** (`.github/workflows/docker-build-amd-rocm.yml`): scheduled build/push of both ROCm
  variants via `build.py`.~~
- **Install `nvidia-modelopt[torch]>=0.37.0` on the ROCm image**: Megatron-Bridge
  (`megatron.bridge.models.gpt_provider`) and Megatron post-training hard-import `modelopt`, which
  previously caused a runtime `ModuleNotFoundError: No module named 'modelopt'`.
- **Scope `NVTE_FUSED_ATTN=0` to the Transformer Engine build step only**.
- **Pin CLI stack**: `typer==0.25.1` and `click==8.2.1` for compatibility.
- **Python 3.10 StrEnum shim**: the ROCm sglang base is Py3.10, but `StrEnum` is Py 3.11+; install `backports.strenum` and make those imports version-guarded.

Co-authored-by: @XinyuJiangCMU


docker/Dockerfile.rocm_MI300
```diff
@@ -1,286 +0,0 @@
...
-#########################################
-###6. Install torch_memory_saver#########
-#########################################
-RUN pip install git+https://github.com/fzyzcjy/torch_memory_saver.git@64a92e1d7fb822ea4af5579c8cebb162692c531c --no-cache-dir --force-reinstall
-#########################################
-#########################################
-
-
-#######################################
-####7. Install Apex for ROCm###########
```
bingxche, line 138: I noticed that in old mi300 dockerfile, we install rocm apex, but it is not shown in the new dockerfile, is this expected?

docker/build.py
```diff
@@ -47,6 +47,24 @@
...
+    "rocm-mi350": {
+        "image": "rocm/sgl-dev",
+        "tag_postfix": "-rocm720-mi35x",
+        "dockerfile": "docker/Dockerfile.rocm",
+        "build_args": {
+            "GPU_ARCH": "gfx950",
+            "SGLANG_IMAGE_TAG": "v0.5.10-rocm720-mi35x",
+        },
+    },
+    "rocm-mi300": {
```
bingxche, line 60: This new MI300 variant now builds from the shared `docker/Dockerfile.rocm`, but the deleted MI300-specific Dockerfile installed ROCm Apex while the shared Dockerfile does not. Miles still has runtime paths that import `apex.transformer.functional` (for example the GLM5 model code), so unless `lmsysorg/sglang:v0.5.10-rocm700-mi30x` already includes Apex, the `rocm-mi300` image can build successfully but fail at runtime. Could we either add the ROCm Apex install back for this variant, or include a smoke test proving Apex is available in the MI300 image?

docker/Dockerfile.rocm
```diff
@@ -29,13 +34,12 @@ ARG RCCL_TESTS_PATH=projects/rccl-tests
...
 
-# ======================================== Setup =============================================
-
+# ======================================== Setup ================================================
 WORKDIR /root/
 
 ENV MAX_JOBS=${MAX_JOBS}
 
-# Build configuration for MI350 / gfx950.
+# Build configuration for the selected arch (gfx942 / gfx950).
```
bingxche, line 42: Since this Dockerfile is now multi-arch, the AITER sed patches later in the file also run for `GPU_ARCH=gfx942` / ROCm 7.0. Earlier in the PR history these looked scoped to the ROCm 7.2 / gfx950 image-toolchain combination. If they are only needed for MI350, can we gate them behind `if [ "${GPU_ARCH}" = "gfx950" ]; then ... fi`? Otherwise please add MI300 build/runtime validation showing the patched AITER behavior is safe for gfx942 too.

.github/workflows/docker-build-amd-rocm.yml
```diff
@@ -0,0 +1,115 @@
+name: Docker Build & Push (AMD ROCm)
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docker/Dockerfile.rocm'
+      - 'docker/amd_patch/**'
```
bingxche: Do you think we also need include `docker/build.py`, `requirements.txt`, and this workflow file in `paths`? Changes there maybe affect the ROCm image build but currently would not trigger a main-branch rebuild until the schedule runs.

docker/Dockerfile.rocm_MI300
XinyuJiangCMU: Verified, the base image already bundles ROCm Apex, so the shared Dockerfile.rocm doesn't need to reinstall it.

docker/build.py
XinyuJiangCMU: Verified on gfx942: `from apex.transformer.functional import fused_apply_rotary_pos_emb_thd` imports successfully and the GLM5 path runs fine.

docker/Dockerfile.rocm
XinyuJiangCMU: Thanks for catching this. Fixed.

.github/workflows/docker-build-amd-rocm.yml
XinyuJiangCMU: Moot now.

XinyuJiangCMU:
@zyzshishui @guapisolo Could you take a look when you get a chance? Thanks!

HaiShaw (commented):
Can we use sglang image as docker base? If so, there is then no need to rebuild and install aiter.

XinyuJiangCMU:
> Can we use sglang image as docker base? If so, there is then no need to rebuild and install aiter.

Yes, the sglang base image already includes gfx950 prebuilt aiter with the needed patches. I removed the redundant aiter uninstall, rebuild, and patch block.

HaiShaw (approved):
LGTM

guapisolo (approved):
LGTM.

PR #1302, opened 2026-06-06
[feat] balance data by FLOPs

yueming-yuan:
## Upstream Slime PRs

This migrates the balance-data-by-workload behavior from THUDM/slime:

- THUDM/slime#2017: https://github.com/THUDM/slime/pull/2017 by @HaoDong0027
- THUDM/slime#2028: https://github.com/THUDM/slime/pull/2028 by @zhuzilin
- THUDM/slime#2029: https://github.com/THUDM/slime/pull/2029 by @zhuzilin


tests/fast/ray/rollout/conftest.py
```diff
@@ -44,6 +44,19 @@ def make_args(**overrides: Any) -> Namespace:
         use_dynamic_global_batch_size=False,
         disable_rollout_trim_samples=False,
         balance_data=False,
+        balance_by_flops=False,
```
maocheng23, line 47: I thought here we only need row 47?

miles/ray/rollout/train_data_conversion.py
```diff
@@ -129,7 +130,8 @@ def split_train_data_by_dp(args, data, dp_size):
     data["total_lengths"] = total_lengths
 
     if args.balance_data:
```
maocheng23, line 132: So here we want to use FLOP balance both w and w/o the --balance-by-flops?

miles/backends/training_utils/data.py
```diff
@@ -438,7 +446,14 @@ def _generate_data_iterator(rollout_data, micro_batch_size, micro_batch_indices=
         for i, num_mbs in enumerate(num_microbatches):
             start, end = i * num_local_gbs, (i + 1) * num_local_gbs
             samples = rollout_data["total_lengths"][start:end]
-            partitions = get_seqlen_balanced_partitions(samples, num_mbs, equal_size=False)
+            if args.balance_by_flops:
+                if num_mbs >= len(samples):
+                    partitions = [[idx] for idx in range(len(samples))]
```
maocheng23, line 451: Do we need to add on paddings here? if num_mbs>len(samples) and we iterate to from 0 to num_mbs, it will be indexError

tests/fast/ray/rollout/conftest.py
maocheng23: Nvm I see the key here, we need add on all these flags to enable the balance_by_flops, but this means we need to add an assert to make sure all params exist when enable this flag?

PR #1308, opened 2026-06-08
fix(qwen3-vl): per-segment mRoPE + vision under CP + THD packing

### LINKED ISSUE #1296: Qwen3-VL THD packed mRoPE: handle context parallelism (cu_seqlens_q_padded layout)
Follow-up to #1272 (cc @TSunny007).

`_build_packed_positions` in `miles/backends/megatron_utils/qwen3_vl_packed_mrope.py` keys on `cu_seqlens_q` and bails when `cu[-1] != flat.numel()`, falling back to the dense `get_rope_index`. Under **context parallelism + THD packing** the input_ids are laid out against `cu_seqlens_q_padded` (see bridge's `thd_to_bshd`, `modelling_qwen3_vl/model.py:271-275`), so the guard no-ops and CP+packing uses the broken cross-sample positions.

#1272 fixes the non-CP packed case (validated) and leaves CP as an explicit logged fallback. This issue tracks the CP+packing fix: reconstruct per-segment MRoPE positions for the CP / padded (zigzag) layout — e.g. compute positions on the full sequence before CP sharding and slice consistently with the tokens.

Zhichenzzz:
Follow-up to #1272 (stacked on its branch, so this diff shows only the #1296 changes — **merge after #1272**).

## What
Makes Qwen3-VL train end-to-end under **context parallelism + THD sequence packing** in bridge mode.

1. **Per-segment mRoPE under CP** — when the THD row is CP-sharded (zigzag), `_build_packed_positions` all-gathers the per-rank rows, de-interleaves to the full row (`_reassemble_full_row`, unit-tested in `tests/fast/test_qwen3_vl_cp_mrope.py`), rebuilds per-segment MRoPE, and re-slices into this rank's zigzag layout.
2. **Don't double-shard** — when the input is already CP-local, the bridge's internal `preprocess_packed_seqs` is wrapped to an identity that returns miles' full-cu `packed_seq_params` (CP attention still sees the full cu; the data isn't re-split).
3. **CP-local vision embeds** — `select_local_vision_embeds` maps each rank's local vision tokens to the matching slice of the full vision-tower output (and deepstack). Cooperates with a small hook in megatron-bridge (separate PR to radixark/Megatron-Bridge).
4. Wires `calculate_per_token_loss` into the bridge provider (Qwen3-VL asserts on it under CP) and a defensive `AllGatherVisionEmbeddings.apply` kwarg shim.

## Validation (e2e)
Qwen3-VL-2B geo3k, **CP=2 TP=4, 8×H200, THD packed, bridge mode**: stable over 3 steps, `train_rollout_logprob_abs_diff` 0.0141 → 0.0146 (healthy, == non-CP 0.011–0.016), `rollout/raw_reward` ~0.4, no crashes.

## Depends on
- #1272 (base of this PR)
- megatron-bridge vision hook (next PR to radixark/Megatron-Bridge) — until it lands, the miles-side monkeypatch provides the hook.

Fixes #1296

---
## Update: cleanup pass + re-validation

- helpers' first parameter renamed `self` → `model` (they are free functions, not methods)
- the `preprocess_packed_seqs` identity wrapper forwards `*args/**kwargs` instead of hard-coding the upstream signature
- the patch warns once at install when the running megatron-bridge lacks the `_miles_select_local_vision_embeds` hook (instead of silently mis-placing vision embeddings under CP); points at the matching Megatron-Bridge patch

Re-validated after cleanup: unit tests 6/6; Qwen3-VL-2B CP2 TP4 THD geo3k RL — `train_rollout_logprob_abs_diff` 0.0127–0.0131 (same healthy band as the original validation), coherent rollouts, no hook warning with the bridge patch installed.

Note on stacking: this branch contains #1272; the cleanup lives here because this PR owns the final shape of `qwen3_vl_packed_mrope.py`.

---
## Update: vision-embed plumbing removed (−53 lines)

Megatron-Bridge PR #9 now selects CP-local vision embeddings **natively** inside `Qwen3VLModel` (no override hook), so this PR no longer carries `select_local_vision_embeds`, the local→full position mapping, or any hook installation. What remains on the miles side: per-segment mRoPE position reconstruction and the `preprocess_packed_seqs` identity wrapper (both still needed — they are about positions, not embeddings), plus a warning when the running bridge lacks the native support.

Re-validated end-to-end after the removal (Qwen3-VL-2B, CP2 TP2, THD, geo3k RL): `train_rollout_logprob_abs_diff` 0.0130, coherent rollouts, no warnings, no crashes.

### DIFF
--- a/miles/backends/megatron_utils/model_provider.py
+++ b/miles/backends/megatron_utils/model_provider.py
@@ -98,6 +98,8 @@ def get_model_provider_func(
         provider.expert_tensor_parallel_size = args.expert_tensor_parallel_size
         provider.sequence_parallel = args.sequence_parallel
         provider.context_parallel_size = args.context_parallel_size
+        # CP>1 VL models assert this; bridge configs skip core_transformer_config_from_args.
+        provider.calculate_per_token_loss = args.calculate_per_token_loss
         provider.attention_softmax_in_fp32 = args.attention_softmax_in_fp32
         provider.variable_seq_lengths = args.variable_seq_lengths
         if hasattr(args, "moe_token_dispatcher_type"):
--- a/miles_plugins/models/qwen3_vl.py
+++ b/miles_plugins/models/qwen3_vl.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import importlib
 import logging
 import threading
+from typing import NamedTuple
 
 import torch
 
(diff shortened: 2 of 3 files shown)


yueming-yuan (approved):
approved to unblock, and we may need to unify all the CP utils in the training backend refactor

PR #1313, opened 2026-06-09
RDT weight sync: GPU->GPU zero copy transfer through SGLang Ray actor backend

xyuzh:
ci-sglang-pr: #27723

## Summary

RDT weight sync shares P2P and nccl distributed weight sync bucketed all-gather + HF conversion pipeline, moving the bucket payload over **NIXL** (`ray.put(_tensor_transport="nixl")` + RDMA pull) instead of holding a full GPU replica of the rollout model. We have measured faster weight sync speed across models than P2P RDMA and NCCL. 

**CORRECTNESS**: Weight equality verified with `--check-weight-update-equal`.

## Performance

**RDT / NIXL is the fastest of the three weight-sync transports** — a per-rank, zero-copy **GPU→GPU** RDMA pull with **no CPU model replica**.

image

*The trainer all-gathers params **bucket-by-bucket** into a small, reusable **fixed-size** GPU staging bucket (no full replica); the first `rollout_num_gpus` ranks are the transfer sources. Each flush, every rollout rank issues a concurrent NIXL RDMA pull (`set_target_for_ref` → `param.data`) of its shard over NVLink (same-node) or EFA/LIBFABRIC (cross-node) — no NCCL broadcast, no host bounce.*

### NCCL broadcast vs RDMA P2P vs RDT / NIXL

Three transports, all moving the same trainer → rollout weights:

- **NCCL broadcast** — serial, lock-step per-engine broadcast; weights stay on the trainer GPUs.
- **RDMA P2P (Mooncake TransferEngine)** — per-rank RDMA pull, but stages a **full model replica in CPU** on every trainer rank.
- **RDT / NIXL** *(this PR)* — per-rank, zero-copy **GPU→GPU** RDMA pull (Ray-native), **no CPU replica**.

Per-sync `update_weights` on the same model set (seconds, lower is better; **bold = fastest** per row):

| Model | Total params | Train → Infer | NCCL | RDMA P2P | **RDT / NIXL** |
|---|---|---|---|---|---|
| GLM-Z1-9B-0414 (dense) | 9B | TP2·CP2 → TP4 · 1 node | 0.69 | 0.71 | **0.40** |
| Moonlight-16B-A3B (MLA) | 16B (3B act) | TP2·EP8 → TP8·EP8 · 1 node | 1.48 | 1.07 | **1.00** |
| GLM-4.7-Flash (MLA MoE) | 30B (3B act) | TP4·EP8 → TP4·EP4 · 1 node | 2.51 | 4.23 | **1.90** |
| Qwen3-30B-A3B (MoE) | 30B (3B act) | TP4·EP8 → TP8·EP8 · 2 nodes | 2.67 | 2.16 | **1.60** |
| GLM-4.5-Air (MoE) | 106B (12B act) | TP1·PP4·EP8 → TP8·EP8 · 4 nodes | 5.00 | 2.64 | **2.30** |

*NCCL and RDMA P2P (Mooncake TransferEngine) measured on the same harness; **RDT / NIXL** from miles validation on Anyscale H100 / EFA, byte-equal every sync (`--check-weight-update-equal`; **RDT is fastest on 4 of 5 models — up to ~2.2× over NCCL — with no CPU replica**; the Mooncake RDMA path edges it only on Qwen3-30B, and is itself slower than plain NCCL on two models (GLM-Z1-9B, GLM-4.7-Flash).*

- **GPU→GPU, no CPU replica:** RDT pulls straight from the trainer GPU staging bucket into rollout `param.data`; the Mooncake RDMA path stages a full model copy in CPU per trainer rank.
- **Bandwidth:** ~150 GB/s over NVLink (same-node); ~23 GiB/s per source / ~92 GiB/s aggregate over EFA (GPUDirect RDMA, no host bounce).
- **Scales flat:** 2→4 nodes adds zero sync overhead — per-source shard and transfer are unchanged.
- Correct across the full architectural range — dense, coarse MoE, and MLA + fine-grained MoE (incl. latent-attention and fused-expert tensors).

## Design

`UpdateWeightFromRDT` inherits `DistBucketedWeightUpdateMixin` (same bucketed TP/EP all-gather + HF conversion as P2P). Each engine rank is backed by a small fixed-size GPU staging bucket on its source trainer rank; per flush, `load_weights` writes the TP-rank-correct sglang shard into bucket views, which are shipped via `ray.put(views, _tensor_transport="nixl")` and pulled into `param.data` on each `SchedulerActor`.

- **Bucket allocation (the throughput fix):** the staging bucket must **not** live in an expandable (VMM / `cuMemCreate`) segment — VMM memory cannot export legacy CUDA-IPC handles, so UCX's `cuda_ipc` lane silently drops and NIXL falls back to software-emulated RMA over eager TCP fragments (~0.3 GB/s vs ~150 GB/s over NVLink). Bisected to `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` on the source; the bucket is now allocated with `expandable_segments` temporarily forced off.
- **NIXL registration pinning:** bucket and param registrations are pinned for the process lifetime (`ray.experimental.register_nixl_memory`) — Ray otherwise ties registration to `ObjectRef` lifetime, so every flush re-pinned GiBs of GPU memory and invalidated the remote-agent cache, forcing a re-handshake.
- **LIBFABRIC backend preference** (`prefer_libfabric_nixl_backend`): validates a realistic-size CUDA registration before committing, because EFA accepts small MRs through its host bounce pool even when GPUDirect is broken.

Engine discovery is Ray-native: `SchedulerActor`s register stable names carrying the engine's HTTP port, and `get_scheduler_actors` matches on `_port{port}_`/`_tp{rank}_` tokens across namespaces, raising on ambiguous or partial matches. A per-sync `[RDT] sync phase breakdown` log splits stage/load/put/submit/pull_wait.

## Status / requirements

- **Cross-node validated over EFA** via the LIBFABRIC backend (see Performance). Requires the EFA userspace stack.
- **Full-node (TP8) rollouts** use no-double-booking PG reuse (`MILES_RDT_REUSE_PG=1`): sglang reuses miles' already-reserved rollout bundles instead of auto-creating a second placement group. Validated on GLM-4.5-Air-106B TP8 (40 GPU / 5 nodes).
- Requires the sglang-miles branch (sgl-project/sglang#27723): `SchedulerActor.pull_weights`, RayEngine named actors, `enable_engine_info_bootstrap`.
- Requires the sglang (sgl-project/sglang#34621):


miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
```diff
@@ -0,0 +1,427 @@
...
+                # views to NIXL.
+                torch.cuda.synchronize()
+
+                # Re-read post-load in case a weight loader reassigned param.data.
+                tensor_views = [meta.params_dict[name].data for name in transfer_ready_params]
+
+                weights_ref = ray.put(tensor_views, _tensor_transport="nixl")
+                weight_refs.append(weights_ref)
+                for actor in meta.actors:
+                    futures.append(actor.pull_weights.remote([weights_ref], transfer_ready_params))
```
stephanie-wang, line 354: Where is pull_weights defined?

miles/utils/arguments.py
```diff
@@ -84,6 +84,17 @@ def add_cluster_arguments(parser):
                     "Turning this on will also set --offload to true."
                 ),
             )
+            parser.add_argument(
+                "--use-rdt-weight-sync",
+                action="store_true",
+                default=False,
+                help=(
+                    "Deprecated alias for --update-weight-transfer-mode rdt. Enables RDT/NIXL "
```
stephanie-wang: Why do we have both this flag and the --update-weight-transfer-mode option?

miles/utils/arguments.py
```diff
@@ -2269,25 +1996,32 @@ def miles_validate_args(args):
...
+    # both spellings work and the rest of the code can gate on either (single source
+    # of truth = the transfer mode).
+    if args.update_weight_transfer_mode == "rdt":
+        args.use_rdt_weight_sync = True
+    elif getattr(args, "use_rdt_weight_sync", False):
+        args.update_weight_transfer_mode = "rdt"
+
     # always true on offload for colocate at the moment.
-    if args.update_weight_transfer_mode == "p2p":
+    if args.update_weight_transfer_mode in ("p2p", "rdt"):
```
stephanie-wang, line 2915: Also need to check that we're using ray as the backend and not mp?

stephanie-wang:
Also do we have tests?

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
xyuzh: https://github.com/sgl-project/sglang/pull/27723
There is a pr at sglang side

miles/utils/arguments.py
xyuzh: yeah, that shall be enforced in the running scripted

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
```diff
@@ -0,0 +1,398 @@
...
+        model_loader_module.post_load_weights = lambda *args, **kwargs: None
+        try:
+            with ParallelismContext(parallelism_config):
+                model = get_model(
+                    model_config=ModelConfig(model_path),
+                    load_config=load_config,
+                    device_config=DeviceConfig(device="cuda"),
+                )
+        finally:
+            model_loader_module.post_load_weights = original_post_load_weights
```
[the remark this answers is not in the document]
xyuzh: We will enforce sglang version for the rdt backend

miles/utils/arguments.py
xyuzh: removed

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
```diff
@@ -0,0 +1,398 @@
...
+            if first_engine_rank:
+                self._shared_params_dict = params_dict
+                self._shared_param_mapper = ParameterMapper.from_model(model_replica)
+                first_engine_rank = False
+
+            gpu_bucket = torch.empty(
+                self.args.update_weight_buffer_size,
+                dtype=torch.uint8,
+                device=torch.cuda.current_device(),
+            )
```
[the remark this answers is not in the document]
xyuzh: This is not rdt specific, could follow up with a solution

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -428,6 +435,58 @@ def unload_lora_adapter(self, lora_name: str):
...
+        """
+        if self._scheduler_actors:
+            return self._scheduler_actors
+
+        import ray
+
+        tp_size = getattr(self.args, "rollout_num_gpus_per_engine", 1)
+        host = self.server_host.strip("[]")
+        name_prefix = f"sglang_scheduler_node{host}"
+
```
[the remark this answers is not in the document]
xyuzh: we use the sglang-miles branch from sglang

xyuzh:
> Also do we have tests?

I have done testing using the Anyscale job config, would follow up with MILES team for guidance to add test

yueming-yuan:
Can you support weight transfer for the multi-node SGLang engine?

yueming-yuan:
Could you add an e2e CI test, in a similar format to the p2p weight update test (tests/e2e/megatron/test_qwen3_30B_A3B_p2p.py), for the RDT weight update? thanks!

xyuzh:
> Can you support weight transfer for the multi-node SGLang engine?

yes multi-node SGLang engine supported
Qwen3-30B-A3B (MoE) sync to tp=16 sglang engine across two node took:
2.20 s

xyuzh:
> Could you add an e2e CI test, in a similar format to the p2p weight update test (tests/e2e/megatron/test_qwen3_30B_A3B_p2p.py), for the RDT weight update? thanks!

test added, tested CI pass locally

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -244,8 +257,30 @@ def _sanity_check_server_args(actual_server_args, expect_server_args):
...
     def _init_normal(self, server_args_dict):
-        logger.info(f"Launch HttpServerEngineAdapter at: {self.server_host}:{self.server_port}")
-        self.process = launch_server_process(ServerArgs(**server_args_dict))
+        use_rdt = self.args.update_weight_transfer_mode == "rdt"
+        if use_rdt:
+            if self.node_rank != 0:
+                # For a multi-node engine, the node-0 server's RayEngine spawns
+                # the SchedulerActors of ALL ranks (placed cross-node via the
+                # placement group), so non-zero node ranks launch nothing.
+                self.process = None
```
yueming-yuan: This breaks Miles' contract of using the process to control the lifecycle of server engines, which will conflict with the process-based control logics, including cleanup, fault tolerance, etc...

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -514,6 +552,51 @@ def unload_lora_adapter(self, lora_name: str):
             {"lora_name": lora_name},
         )
 
+    def get_scheduler_actors(self) -> list:
```
yueming-yuan, line 579: This function looks a bit of hardcoded

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -132,13 +140,18 @@ def __init__(
         base_gpu_id: int | None = None,
         sglang_overrides: dict | None = None,
         num_gpus_per_engine: int | None = None,
+        placement_group: PlacementGroup | None = None,
```
yueming-yuan: importing ray here and using placement group does not match the files' scope, these are usually controlled under /ray

miles/ray/rollout/server_group.py
```diff
@@ -73,6 +73,14 @@ def start_engines(
 
         pg, reordered_bundle_indices, reordered_gpu_ids = self.pg
 
+        # RDT no-double-booking: hand miles' reserved rollout bundles to sglang's
```
yueming-yuan: nit: (for all files) we might not prefer very long comments; recommend keeping them short and precise

miles/backends/sglang_utils/sglang_engine.py
xyuzh: In non-rdt path, we also return None for the process for non-zero node ranks
https://github.com/xyuzh/miles/blob/xinyu/rdt-weight-sync/miles/backends/sglang_utils/sglang_engine.py#L69-L80

The engine only needs the process on rank 0 node

miles/backends/sglang_utils/sglang_engine.py
xyuzh: SGLangEngine subclasses miles.ray.ray_actor.RayActor
placement_group removed, ray only imported at method level

miles/backends/sglang_utils/sglang_engine.py
xyuzh: addressed hardcoding by adding `get_scheduler_actor_name` as a helper function

miles/ray/placement_group.py
```diff
@@ -126,13 +137,16 @@ def create_placement_groups(args):
...
+    # with a colocated rollout the way the fractional reservation allows.
+    num_gpus_per_actor = 1 if args.update_weight_transfer_mode == "rdt" else 0.4
     train_group_cls = _select_train_group_class()
     return train_group_cls(
         args=args,
         num_nodes=num_nodes,
         num_gpus_per_node=num_gpus_per_node,
         pg=pg,
-        num_gpus_per_actor=0.4,
+        num_gpus_per_actor=num_gpus_per_actor,
```
maocheng23, line 149: Would this be conflict with PPO?
@guapisolo to double confirm

miles/utils/arguments.py
```diff
@@ -819,10 +819,12 @@ def add_rollout_arguments(parser):
             )
             parser.add_argument(
                 "--update-weight-transfer-mode",
-                choices=["broadcast", "p2p", "disk-delta"],
+                choices=["broadcast", "p2p", "disk-delta", "rdt"],
```
maocheng23, line 822: rdt only works for megatron right? Let's add an assert here.

xyuzh: Done

miles/ray/train/actor_factory.py
```diff
@@ -71,11 +82,16 @@ def allocate_gpus_for_actor(
...
     ft = args.use_fault_tolerance
-    TrainRayActor = ray.remote(
-        num_gpus=1,
-        runtime_env={"env_vars": env_vars},
-        **(dict(concurrency_groups={"heartbeat_status": 1, "default": 1, "fault_injector": 1}) if ft else {}),
-    )(_with_ft_concurrency_groups(actor_impl) if ft else actor_impl)
+    remote_kwargs = {"num_gpus": 1, "runtime_env": {"env_vars": env_vars}}
+    if ft:
+        remote_kwargs["concurrency_groups"] = {"heartbeat_status": 1, "default": 1, "fault_injector": 1}
+    elif args.update_weight_transfer_mode == "rdt":
```
guapisolo, line 88: Update: I traced this through Ray 2.56 and retract this P1. This `elif` is intentional: defining any named concurrency group makes Ray create a separate executor for the built-in default group even when its concurrency remains 1 ([manager](https://github.com/ray-project/ray/blob/936f0d7d49d9da8ac1a9f04cc8a89faf2cb3c42a/src/ray/core_worker/task_execution/concurrency_group_manager.cc#L46-L54), [condition](https://github.com/ray-project/ray/blob/936f0d7d49d9da8ac1a9f04cc8a89faf2cb3c42a/src/ray/core_worker/task_execution/thread_pool.h#L35-L40)). The unannotated `update_weights()` in the FT actor therefore already runs off the core-worker main thread; the non-FT path needs `max_concurrency > 1` only to force that same off-main-thread executor.

Also, `"default"` in `concurrency_groups` is only a named group and has no bound method here, so increasing it would not affect `update_weights()`. NIXL is [one-sided](https://github.com/ray-project/ray/blob/936f0d7d49d9da8ac1a9f04cc8a89faf2cb3c42a/python/ray/experimental/rdt/tensor_transport_manager.py#L78-L90), and the required [SGLang PR #27723](https://github.com/sgl-project/sglang/pull/27723) gives the destination `SchedulerActor` `max_concurrency=2` so `pull_weights` can run beside `run_event_loop`. Setting actor-level `max_concurrency=1+tp` for FT would instead allow ordinary trainer RPCs to overlap and potentially race. No behavior change is justified without an FT+RDT E2E hang.

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
```diff
@@ -0,0 +1,397 @@
...
+    def _is_source(self) -> bool:
+        """Whether this training rank sends weights to rollout. Every rank holds a
+        full replica after all-gather, so only the first _rollout_num_gpus are used."""
+        return self.transfer_plan._gathered_dp_rank < self.transfer_plan._rollout_num_gpus
+
+    def connect_rollout_engines(
+        self,
+        rollout_engines: Sequence[ActorHandle],
+        rollout_engine_lock: ActorHandle | None = None,
+        engine_gpu_counts: Sequence[int] | None = None,
```
guapisolo, line 155: [P2] Build RDT targets from the per-engine GPU counts

`--sglang-config` allows each server group to override `num_gpus_per_engine`, and `RolloutManager` passes the resolved `engine_gpu_counts` here, but this method ignores them and uses a `RemoteTransferPlan` built from the global `args.rollout_num_gpus_per_engine`. For example, with four rollout GPUs, a TP=2 CLI default, and groups producing counts `[1, 1, 2]`, the plan creates `(0,1)` and `(1,1)` (out-of-range scheduler ranks) and never targets engine 2, so connection fails at `engine_actors[t.engine_rank]` or leaves that engine stale. Build or update the plan from `engine_gpu_counts`, or explicitly reject heterogeneous RDT layouts.

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
```diff
@@ -0,0 +1,397 @@
...
+        """Re-point each ``params_dict[name].data`` to a contiguous view inside the
+        fixed bucket, returning the views in ``names`` order."""
+        offset = 0
+        views: list[torch.Tensor] = []
+        capacity = self.gpu_bucket.numel()
+        for name in names:
+            shape, dtype, nbytes = self.param_specs[name]
+            itemsize = torch.empty((), dtype=dtype).element_size()
+            offset = ((offset + itemsize - 1) // itemsize) * itemsize
+            assert offset + nbytes <= capacity, (
```
guapisolo: [P2] Validate or grow the RDT bucket before the first sync

The shared bucketizer deliberately lets an oversized first update unit through when the current bucket is empty, but `stage()` requires every complete destination parameter to fit in this fixed allocation. `--update-weight-buffer-size` still defaults to 512 MiB, while a legal TP=1 Qwen3-30B-A3B rollout has a bf16 `embed_tokens` parameter of 622,329,856 bytes (~0.58 GiB); the new E2E also raises the buffer to 1 GiB because of this parameter. Selecting `rdt` with otherwise valid/default arguments therefore reaches this assertion on the first weight update. Once `param_specs` are available in `connect_rollout_engines()`, allocate at least `max(configured_size, max_param_nbytes)` or reject the configuration there with a clear error.

miles/ray/train/actor_factory.py
```diff
@@ -38,6 +38,17 @@ def allocate_gpus_for_actor(
         **args.train_env_vars,
     }
 
+    if args.update_weight_transfer_mode == "rdt":
```
guapisolo, line 41: [P2] Add the new transfer-mode field to fast-test defaults

This unconditional access already makes `pytest -q tests/fast/ray/test_actor_factory.py` fail at this line because its `SimpleNamespace` has no `update_weight_transfer_mode`. The shared `tests/fast/ray/rollout/conftest.py::make_args()` also lacks the field, while `ServerGroup.start_engines()` now reads it directly and raises the same `AttributeError` before actor creation. Add `update_weight_transfer_mode="broadcast"` to both fixtures, matching the parser default, so existing non-RDT fast tests still exercise their intended paths.

miles/ray/placement_group.py
guapisolo: fixed

guapisolo: Confirmed — this deterministically conflicts with Shared Actor/Critic PPO. RDT reserves one full GPU per trainer rank, while PPO reuses the actor placement-group bundles for critic ranks, so critic actor creation cannot be scheduled after actor initialization. I pushed e2f87822a1, which rejects this combination during argument validation and adds parser-level regression coverage. Supporting RDT with PPO would require separate critic GPU bundles.

miles/utils/arguments.py
guapisolo: Confirmed. 048dd487f5 added the Megatron-only assertion, and e2f87822a1 adds parser-level coverage for the supported Megatron path and the FSDP rejection.

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
xyuzh: Added validation: connect_rollout_engines() now raises a ValueError if engine_gpu_counts is heterogeneous (any count != --rollout-num-gpus-per-engine). This fails safely before touching any connection state.

Allowed cases: Uniform, None, and empty counts still pass unchanged.

Deferred work: The ideal fix for heterogeneous layouts is to build the transfer plan per-engine. Since there is no RDT use case for this yet, I'm deferring it until one actually shows up.

miles/backends/megatron_utils/update_weight/update_weight_from_rdt.py
xyuzh: Dynamic bucket sizing: Each engine-rank bucket is now allocated at max(--update-weight-buffer-size, largest destination param nbytes) based on that rank's param_specs. Added a warning when the buffer needs to grow.

Better CI coverage: Removed the hand-tuned 1 GiB override from the 30B e2e test. It now runs at the 512 MiB default, which forces the new growth path to trigger on the 0.58 GiB embed_tokens in every CI run.

miles/ray/train/actor_factory.py
xyuzh: kept the concurrency setup as-is.

miles/ray/train/actor_factory.py
xyuzh: Updated test configs to match the parser default. Set update_weight_transfer_mode="broadcast" in:

- tests/fast/ray/rollout/conftest.py::make_args()

- The SimpleNamespace in test_actor_factory.py

xyuzh:
Two failed CI not related to the PR

yueming-yuan (approved):

guapisolo (approved):

guapisolo (approved):
approve to unblock

xyuzh:
Need to pin sglang version to pass the final CI

guapisolo:
kicked off ci

PR #1314, opened 2026-06-09
[OPD] [3/N] Multi-teacher routing: per-sample teacher selection via --opd-teacher-urls

maocheng23:
…-opd-teacher-urls

Route each sample to a task-specific teacher (e.g. math vs code specialist) during sglang-mode OPD scoring. Each sample is still scored by exactly one teacher, so scoring cost and the loss are unchanged.

- --opd-teacher-urls NAME=URL ...: routing map; reserved name 'default' is the fallback. Unset keeps the single-teacher --rm-url path byte-for-byte.
- --opd-teacher-key: sample metadata key holding the teacher name (default 'opd_teacher'), populated from the dataset metadata column.
- Missing/unknown names without a 'default' entry fail loudly instead of silently distilling from the wrong teacher.
- Validation: sglang mode only; malformed/duplicate entries fail at startup.
- Works with both sampled-token (top-k=0) and Rethinking top-k paths; student-side scoring still goes to the student router.
- Example: run-qwen3-8B-opd-multi-teacher.sh (Qwen3-32B math teacher + Qwen3-Coder-30B-A3B code teacher -> Qwen3-8B student) + docs.


guapisolo (approved):
LGTM. Very clean PR.

PR #1315, opened 2026-06-09
docs: fix deepseek-v4 launcher env var and path-default guidance

yueming-yuan:
## What

**Env var / path-default guidance (deepseek-v4-flash.md + deepseek-v4-pro.md §3.2)**
- The doc claimed there are no `MILES_SCRIPT_*` env vars that preconfigure launcher paths. This is wrong: every `run_deepseek_v4.py` option is bound to `MILES_SCRIPT_<FIELD_NAME_UPPER>` via the bare `@dataclass_cli` decorator (`miles/utils/typer_utils.py` sets `envvar=` on each typer Option), with precedence CLI flag > env var > built-in default. Verifiable via `python scripts/run_deepseek_v4.py train --help`, which prints `[env var: ...]` on each option.
- `--model-local-dir` default corrected from `/root/local_data` to unset → same as `--model-dir` (matches `ScriptArgs.model_local_dir = None` + the `__post_init__` fallback).
- Pro: replaced the "TBD — Pro-specific overrides or env-var notes" placeholder with a one-line pointer to V4-Flash §3.2.

**Disaggregated rollout mode, per #1310 (deepseek-v4-flash.md)**
- §3.1: add a disaggregated launch example (`--num-nodes 16 --rollout-num-nodes 8`).
- New §3.3: `--rollout-num-nodes` semantics — default colocate; N nodes serve rollout, the rest train; verified parallelism recipes are keyed on training nodes.
- §5.1: note that the parallelism table counts training nodes.
- §5.3 (flash + pro): drop `--use-miles-router` from the example SGLANG_ARGS — since #1310 the DSV4 launcher runs R3 with `--use-rollout-routing-replay` only.

### DIFF
--- a/docs/models/deepseek/deepseek-v4-flash.md
+++ b/docs/models/deepseek/deepseek-v4-flash.md
@@ -52,10 +52,10 @@ The Python launcher (`scripts/run_deepseek_v4.py`) takes its path arguments from
 |---|---|---|
 | `--data-dir` | `/root/datasets` | HF datasets (e.g. dapo-math-17k, …) |
 | `--model-dir` | `/root/models` | parent directory holding the HF checkpoint and Megatron `_torch_dist` artifacts as separate sibling sub-directories |
-| `--model-local-dir` | `/root/local_data` | local NVMe path on each node; `prepare-cp` rsyncs the HF checkpoint and `_torch_dist` here so the trainer reads from local disk instead of shared storage |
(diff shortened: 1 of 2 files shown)


docs/models/deepseek/deepseek-v4-flash.md
```diff
@@ -158,7 +169,6 @@ SGLANG_ARGS=(
    --sglang-chunked-prefill-size 8192
    --sglang-mem-fraction-static 0.5  # leave headroom for Megatron during wake_up
```
guapisolo, line 170: could we increase this?

Zhichenzzz (approved):

PR #1317, opened 2026-06-10
fix: load Qwen 3.5 checkpoint with unfused experts

lawrence-harmonic:
For some reason on HuggingFace the fp8 checkpoint has unfused `gate_proj` and `up_proj`:
https://huggingface.co/Qwen/Qwen3.5-35B-A3B-FP8/blob/main/model.safetensors.index.json

which is different from the bf16 checkpoint, which has fused `gate_up_proj`:
https://huggingface.co/Qwen/Qwen3.5-35B-A3B/blob/main/model.safetensors.index.json

Current miles code handles fused `gate_up_proj` only.


guapisolo:
Could you fix the lint first. no rush.

PR #1318, opened 2026-06-11
feat: add FlashQLA backend for Qwen GDN linear-attention layers

Zhichenzzz:
Sync from slime: THUDM/slime#1947

fla vs flashqla kernel bfloat16: max_abs_diff 4.88e-04, mean 1.15e-05, max_rel 7.84e-03
under Qwen3.5-4B RL 20 steps, the max abs_diff is 0.00746 both for fla and flashqla.

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -140,11 +140,15 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
                 help="The qkv layout.",
             )
             parser.add_argument(
-                "--qwen-gdn-backend",
+                "--linear-attention-backend",
                 type=str,
                 choices=["fla", "flashqla"],
                 default="fla",
-                help="GDN implementation backend for Qwen linear-attention layers.",
+                help=(
+                    "Backend for Qwen GDN linear-attention layers. "
+                    "'fla' (flash-linear-attention) is portable and runs on any supported GPU. "
+                    "'flashqla' (FlashQLA) requires NVIDIA SM90 (Hopper) or newer, CUDA 12.8+, and PyTorch 2.8+."
+                ),
             )
             parser.add_argument(
                 "--true-on-policy-mode",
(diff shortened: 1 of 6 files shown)


miles/utils/arguments.py
```diff
@@ -139,6 +139,13 @@ def add_train_arguments(parser):
                 default="thd",
                 help="The qkv layout.",
             )
+            parser.add_argument(
+                "--qwen-gdn-backend",
+                type=str,
+                choices=["fla", "flashqla"],
+                default="fla",
+                help="GDN implementation backend for Qwen linear-attention layers.",
```
guapisolo: Could you add some cmt here for which hardware support which backend?

guapisolo: Also, do you think we should rename the param to `--linear-attention-backend` as instead? Maybe in the future we have more models with linear attention 🤔. but i am not sure whether the naming has conflict with megatron's naming..

miles_plugins/models/qwen_gdn_backend.py
```diff
@@ -0,0 +1,46 @@
...
+    if (major, minor) < (9, 0):
+        raise RuntimeError(f"FlashQLA backend requires NVIDIA SM90 or newer, got sm{major}{minor}.")
+
+    cuda_version = torch.version.cuda
+    if cuda_version is not None and _parse_version(cuda_version) < (12, 8):
+        raise RuntimeError(f"FlashQLA backend requires CUDA 12.8 or newer, got CUDA {cuda_version}.")
+
+
+def get_chunk_gated_delta_rule(backend: str):
+    if backend == "fla":
```
guapisolo, line 8: It looks install flashqla backend is not included in dockerfile? We should avoid user to install it manually.

miles_plugins/models/qwen_gdn_backend.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            from fla.ops.gated_delta_rule import chunk_gated_delta_rule
```
guapisolo: better no lazy-import

guapisolo (approved):
LGTM. but could you provide a repro script and some accuracy result?

PR #1322, opened 2026-06-11
[OPD] [4/N] Teacher ensembles + exact tail-bucket top-k KL + scoring robustness

maocheng23:
# [OPD] [4/N] Teacher ensembles + exact tail-bucket top-k KL + scoring robustness

**Stacked on #1314** (multi-teacher routing), which stacks on #1298 → #994 → #993 (merged).

## Motivation

[slime PR #2033](https://github.com/THUDM/slime/pull/2033) (draft) adds two OPD capabilities miles lacked: combining **multiple teachers into one distillation target** and treating the top-k support as a **proper distribution with an explicit tail bucket**. We reviewed that PR in depth; this PR adopts the two ideas that are genuinely better — the probability-space mixture formula and the k+1-bucket KL — while keeping them on miles' rollout-side architecture, which is structurally cheaper and more general than slime's training-side design:

| | slime 2033 | this PR |
|---|---|---|
| Teacher placement | weight-swapped copies inside the Megatron train actor | external SGLang servers (unchanged) |
| Teacher constraint | **architecture/vocab/parallelism-homogeneous with the student** (a 32B teacher for an 8B student is impossible; their example self-distills) | any size/architecture sharing the tokenizer |
| Multi-teacher cost | N serialized forwards **on training GPUs** + N+1 full weight restores per step + N resident weight snapshots | parallel HTTP fan-out — wall clock ≈ max(teacher latencies), zero training-step cost |
| Ensemble weights | uniform only | per-teacher weights (`URL@W`) |
| Mixture math | `logsumexp(stack) − log N` ✓ (adopted) | same formula, weighted, float64 |
| Tail bucket | `(1 − Σp).clamp(1e-10).log()` in **bf16** → log≈−23 spikes when top-k mass rounds to ≥1 | float64 + `log1p`; student tail ≥1 → exact 0 contribution; floor only as a last-resort guard |
| Composability | replaces `policy_loss` entirely (no PPO clip, no task reward); forked `train_actor` subclass + module-global actor registry | unchanged: per-token penalty folded into advantages, composes with every estimator and PPO clipping |
| Per-sample data shipped to training | O(R·k) indices + logprobs (~6.3 MB at R=8k, k=64) | O(R) floats (~32 KB), unchanged |

Routing (#1314) and ensembling are complementary, not competing: route *across* domains, ensemble *within* a group of same-domain peers. This PR makes each routed name a (possibly weighted) group, so the two compose.

## What's added

| Flag | Behavior |
|---|---|
| `--opd-teacher-urls NAME=URL[@W][,URL[@W]...]` | Extended: a name now maps to a *group*. One URL = routing (3/N behavior, byte-for-byte unchanged). Multiple comma-separated URLs = ensemble: every member scores the sample in parallel and targets combine as a weighted probability-space mixture `log p̄(v) = logsumexp_m(log w_m + log p_m(v)) − log Σw_m`. Weights default to 1.0. |
| `--opd-topk-tail-bucket` | Top-k reward becomes the exact reverse KL over k+1 buckets: selected ids at raw full-softmax probabilities + one tail bucket `(1 − Σp)`. No renormalization — the estimate stays sensitive to mass the student pushes outside the top-k (the structural blind spot of the renormalized estimate). Float64 + `log1p`; requires `--opd-reward-weight-mode student_p` and `--opd-top-k-strategy only_stu`/`intersection` (the partition is only exact when all student logprobs come from one softmax — the other strategies mix the rollout harvest with a separate rescoring pass). |
| `--opd-scoring-timeout-secs` | Dedicated scoring timeout, defaulting to `--sglang-router-request-timeout-secs`. Previously the only knob was the shared router flag, so bounding a slow 32B teacher also changed every generation request's timeout. |

Plus robustness/correctness fixes flagged in the same review:

- **Bounded retry** in `_post_json`: one retry with jitter on timeout / connection error / HTTP 5xx; 4xx and `CancelledError` propagate immediately. Previously any transient teacher hiccup killed the whole rollout loop.
- **CP hole fixed**: `get_rollout_data` now CP-slices `teacher_log_probs` / `opd_reverse_kl` exactly like `rollout_log_probs`. Previously sglang-mode OPD with `--context-parallel-size > 1` crashed on the shape checks in `apply_opd_kl_to_advantages` (slime 2033 designed for CP; we didn't). Identity at cp=1.
- **Heterogeneous-batch gate**: `convert_samples_to_train_data` gated OPD keys on `samples[0]` only — a partially scored batch silently dropped the key or crashed later on a length mismatch. Now: all-or-raise with per-sample diagnostics.
- **Memory**: `sample.metadata["opd_student_top_logprobs"]` (O(R·k) Python lists, 50–150 MB/sample at 8k×64) is freed after the KL is computed instead of riding every Ray transfer forever.
- **Kimi-VL multimodal + CP**: the media-token-expansion re-slice in `mm_data.py` covered `teacher_log_probs` but not `opd_reverse_kl` (stale zigzag layout after expansion → shape mismatch or silent misalignment), and the OPD tensors are now placed on GPU like `rollout_log_probs` so that path's NCCL all-reduce works.
- **Fail-at-startup URL validation**: every parsed group member must be an `http(s)://` URL, so a comma inside a member URL (which would split into phantom members) fails at launch instead of mid-rollout.

## Ensemble semantics

- Mixture of raw probabilities **before** any weighting/renormalization — a mixture of renormalized truncations is not the truncation of the mixture. The result is the log-probability of the actual mixture distribution (arithmetic mean in probability space), not a geometric mean of logprobs.
- Sampled-token path (`top-k=0`): mixture at each sampled token, stored in `sample.teacher_log_probs` as before.
- Top-k path: ensembles require `--opd-top-k-strategy only_stu` (validated at startup) so every member is scored at the *same* student top-k ids per position and the per-id mixture is exact. A missing id in any member's response raises loudly.
- With `--opd-topk-tail-bucket`, the mixture tail equals the weighted average of member tails over the same id set, so the k+1-bucket KL stays exact for ensembles too.
- Everything downstream of `reward_func`/`post_process_rewards` is untouched — the loss never knows the teacher was a mixture.

**Deliberately not ported from slime 2033** (with reasons, from our review): colocated weight-swapped teachers (homogeneity constraint defeats the main OPD use case; serialized forwards inflate every step), the replacement `topk_opd_loss` (unclipped importance ratios, bf16 logprob arithmetic, forked train actor + module-global registry — their own PR body flags the pattern as provisional). A differentiable train-side bucket loss remains a possible 5/N, tracked separately.

## Files changed

- `miles/rollout/on_policy_distillation.py` — group parsing (`NAME=URL[@W][,...]`, last-`@` weight split so userinfo URLs survive), `_teacher_targets_for_sample`, parallel `_post_teacher_group` fan-out, `_mixture_log_probs` / `_mixture_logprob_maps` (float64 logsumexp), `_tail_bucket_reverse_kl`, retry + `_scoring_timeout`, metadata cleanup.
- `miles/utils/arguments.py` — the three flags + fail-fast validation (ensemble⇒`only_stu` when top-k>0, tail-bucket⇒`student_p`+top-k>0, positive timeout, all rejected without `--use-opd`).
- `miles/backends/training_utils/data.py` — CP slicing for OPD tensors.
- `miles/ray/rollout/train_data_conversion.py` — all-or-raise OPD key gate.
- `docs/advanced/on-policy-distillation.md` — "Teacher Ensembles" + "Tail-Mass Bucket" sections, Key Arguments rows.
- `examples/on_policy_distillation/run-qwen3-8B-opd-ensemble.sh` — Qwen3-8B student, Qwen3-32B + Qwen3-30B-A3B teachers mixed 2:1, tail bucket on. README updated.
- `tests/fast/rollout/test_on_policy_distillation.py` — 36 new tests (parsing groups/weights/userinfo-URLs/malformed, group routing, mixture math vs hand-computed values, tail-bucket KL incl. mass≥1 / floor / empty-set edges, ensemble end-to-end + non-`only_stu` rejection, retry classification + retry/no-retry/exhausted, timeout fallback, `reward_func` fan-out, ensemble `post_process_rewards`, metadata freeing, ensemble+tail-bucket with non-uniform weights vs hand-computed mixture KL, top-k `reward_func` fan-out, mixed-source strategy rejection).
- `tests/fast/ray/rollout/test_train_data_conversion.py` — 2 tests for the batch gate.

The diff was adversarially reviewed pre-submission (3 independent lenses + per-finding verification agents); all confirmed findings — tail-bucket strategy restriction, the Kimi-VL CP gaps, URL-scheme validation, timeout-flag guard — are fixed in this revision.

## Tests

80 passed locally (`tests/fast/rollout/test_on_policy_distillation.py` + `tests/fast/ray/rollout/test_train_data_conversion.py`, run standalone due to local env). `ruff check` clean; `bash -n` clean on the example script. Single-teacher and routing paths are covered by the existing 1314 tests, all updated and passing — single-URL behavior is unchanged.

### DIFF
--- a/examples/on_policy_distillation/run-qwen3-8B-opd-ensemble.sh
+++ b/examples/on_policy_distillation/run-qwen3-8B-opd-ensemble.sh
@@ -1,0 +1,219 @@
+#!/bin/bash
+
+# Teacher-ensemble OPD: every sample is scored by a GROUP of teachers whose
+# distributions are combined as a weighted mixture in probability space
+# (logsumexp of weighted logprobs), with the exact tail-bucket top-k KL.
+#
+# Student:    Qwen3-8B (2 training GPUs + 4 rollout GPUs)
+# Teacher A:  Qwen3-32B            (GPU 6) weight 2.0
+# Teacher B:  Qwen3-30B-A3B        (GPU 7) weight 1.0
+#
+# Both teachers score every sample in parallel (asyncio.gather), so scoring
+# wall clock is max(teacher latencies), not the sum, and training-step time is
+# unchanged versus single-teacher OPD. Ensemble only same-domain peers — for
+# task specialists, use per-sample routing instead (see
+# run-qwen3-8B-opd-multi-teacher.sh); the two compose: each routed name can be
+# its own ensemble group. All teachers must share the student's tokenizer
+# (scoring sends input_ids), which holds for the Qwen3 family used here.
+#
+# usage: bash examples/on_policy_distillation/run-qwen3-8B-opd-ensemble.sh
+
+set -ex
+
+TEACHER_A_IP="127.0.0.1"
+TEACHER_A_PORT=13141
+TEACHER_B_IP="127.0.0.1"
+TEACHER_B_PORT=13142
+
+start_teacher() {
+    local gpu=$1 model_path=$2 port=$3
+    local log_file
+    log_file="/tmp/sglang_$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 6).log"
+    CUDA_VISIBLE_DEVICES=$gpu python3 -m sglang.launch_server \
+        --model-path "$model_path" \
+        --host 0.0.0.0 \
+        --port "$port" \
+        --tp 1 \
+        --chunked-prefill-size 4096 \
+        --mem-fraction-static 0.6 \
+        > "$log_file" 2>&1 &
+    echo "$log_file"
+}
+
+wait_teacher() {
+    local ip=$1 port=$2 log_file=$3
+    until curl -sf "http://$ip:$port/health_generate" > /dev/null; do
+        echo "Waiting for teacher server at $ip:$port..."
+        tail -n 10 "$log_file"
+        sleep 5
+    done
+    curl "http://$ip:$port/get_model_info"
+    echo "Teacher server is up at $ip:$port."
+}
+
+A_LOG=$(start_teacher 6 /root/Qwen3-32B "$TEACHER_A_PORT")
+B_LOG=$(start_teacher 7 /root/Qwen3-30B-A3B "$TEACHER_B_PORT")
+wait_teacher "$TEACHER_A_IP" "$TEACHER_A_PORT" "$A_LOG"
+wait_teacher "$TEACHER_B_IP" "$TEACHER_B_PORT" "$B_LOG"
+sleep 10
+
+
+export PYTHONBUFFERED=16
+
+NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
+if [ "$NVLINK_COUNT" -gt 0 ]; then
+    HAS_NVLINK=1
+else
+    HAS_NVLINK=0
+fi
+echo "HAS_NVLINK: $HAS_NVLINK (detected $NVLINK_COUNT NVLink references)"
+
+source "/root/miles/scripts/models/qwen3-8B.sh"
+
+
+CKPT_ARGS=(
+   --hf-checkpoint /root/Qwen3-8B
+   --ref-load /root/Qwen3-8B_torch_dist
+   --load /root/Qwen3-8B_miles/
+   --save /root/Qwen3-8B_miles/
+   --save-interval 20
+)
+
+ROLLOUT_ARGS=(
+   --prompt-data /root/dapo-math-17k/dapo-math-17k.jsonl
+   --input-key prompt
+   --apply-chat-template
+   --rollout-shuffle
+   --num-rollout 300
+   --rollout-batch-size 16
+   --n-samples-per-prompt 4
+   --rollout-max-response-len 16384
+   --rollout-temperature 1
+
+   --global-batch-size 64
+   --balance-data
+)
+
+RM_ARGS=(
+   --custom-rm-path miles.rollout.on_policy_distillation.reward_func
+   --custom-reward-post-process-path miles.rollout.on_policy_distillation.post_process_rewards
+   # One 'default' group of two teachers, mixed 2:1 in probability space.
+   # Every sample is scored by both members in parallel.
+   --opd-teacher-urls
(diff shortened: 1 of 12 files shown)


examples/on_policy_distillation/run-qwen3-8B-opd-multi-teacher.sh
guapisolo, line 18: Do you think python script is easier to maintain for OPD? 🤔

miles/backends/training_utils/data.py
```diff
@@ -93,6 +93,35 @@ def get_rollout_data(args: Namespace, rollout_data_ref: Box) -> RolloutBatch:
...
+    # through here.) At cp_size == 1 this is an identity slice.
+    for opd_key in ("teacher_log_probs", "opd_reverse_kl"):
+        if opd_key in rollout_data:
+            sliced_values = []
+            for i, (value, total_length, response_length) in enumerate(
+                zip(
+                    rollout_data[opd_key],
+                    rollout_data["total_lengths"],
+                    rollout_data["response_lengths"],
+                    strict=False,
```
guapisolo, line 109: Shall we do `strict=True` here 🤔

guapisolo (approved):
Approve with nits.

examples/on_policy_distillation/run-qwen3-8B-opd-ensemble.sh
```diff
@@ -0,0 +1,219 @@
...
+
+TEACHER_A_IP="127.0.0.1"
+TEACHER_A_PORT=13141
+TEACHER_B_IP="127.0.0.1"
+TEACHER_B_PORT=13142
+
+start_teacher() {
+    local gpu=$1 model_path=$2 port=$3
+    local log_file
+    log_file="/tmp/sglang_$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 6).log"
```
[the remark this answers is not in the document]
Shi-Dong: I agree with Gemini and let's fix this.

Shi-Dong (approved):
LGTM.  You might want to run the formatter and address the Gemini comment before merging.

PR #1323, opened 2026-06-12
[fix] stop merging agentic turns at first non-COMPLETED turn

Shi-Dong:
When MILES_EXPERIMENTAL_ROLLOUT_REFACTOR is on, each LLM call within a multi-turn agentic trajectory produces its own Sample, and merge_samples folds them left into one training Sample via _merge_sample_pair. _merge_sample_pair asserts the accumulated turn is COMPLETED before appending the next, encoding the invariant that only the final turn may be non-COMPLETED.

A trajectory whose intermediate turn TRUNCATED (hit --rollout-max-response-len mid-generation) but whose agent harness still produced later turns violated that invariant and crashed the rollout loop with AssertionError: a.status must be COMPLETED, got Status.TRUNCATED.

Fix: in merge_samples, stop folding at the first non-COMPLETED turn so the trajectory simply ends there (merged Sample retains the truncated status). Keeps the assertion in _merge_sample_pair as a real invariant. No behavior change for fully-COMPLETED trajectories.

Repro: DeepSeek-V4-Flash agentic RL on Terminal-Bench-2/terminus-2; crashed right after step 0 on a polyglot task that truncated an intermediate turn.


guapisolo (approved):
LGTM.

PR #1330, opened 2026-06-12
[fix] set deepseek v32 override_hf_native=True

yueming-yuan:
Newest transformers (from 6.10) supported deepseek v32 but sglang's stable version does not. Set override.


guapisolo (approved):
LGTM.

PR #1329, opened 2026-06-12
[refactor] refactor weight update post-process and weight-checker

yueming-yuan:
ci-sglang-pr: #28082

PR #812, opened 2026-03-26
[weight checker] add dequant_mean_err_threshold and enable weight checker in all Megatron CIs

yueming-yuan:
ci-sglang-pr: #21521

## Summary
- Enable `--check-weight-update-equal` flag in all 9 Megatron-based e2e CI test files
- This verifies weight synchronization between training actor and SGLang inference engines after every weight update

## Test plan
- [ ] Run `run-ci-megatron` CI to validate all Megatron tests pass with weight checking enabled

PR #1336, opened 2026-06-12
[weight checker] enable in all CIs + ULP-based quant error tolerance

yueming-yuan:
ci-sglang-pr: #28082

Replaces #812. sglang-side PR: sgl-project/sglang#28082. Stacked on #1329.

## Summary

- Force-enable `--check-weight-update-equal` whenever `--ci-test` is set, so every future CI verifies weight sync between the training actor and SGLang engines after weight update. The now-redundant explicit flags in `test_qwen3_4B_p2p.py` and `test_qwen3_30B_A3B/_common.py` are removed.
- Replace the manual-threshold design from #812 with a boolean `--check-weight-update-allow-quant-error`, plumbed through `train.py`/`train_async.py` → rollout manager/server/group → engine → sglang `/weights_checker`.
- On the sglang side (sgl-project/sglang#28082), quantized weights are compared via their dequantized tensors; with the flag set, each element may differ by up to 1 ULP of the quantized dtype per side (e.g. fp8e4m3 spacing at that element's magnitude, scaled by the block scale). This is principled (each side is one faithful quantization of the same weight), per-element (a mean threshold lets corruption hide in large tensors), and needs no per-model tuning.
- Quantized CI cases get the flag: qwen3-30B fp8/int4, glm5-4layer under `MILES_TEST_USE_FP8_ROLLOUT`, dsv32 fp8 + mxfp8, dsv4-flash (FP8 checkpoint). bf16 checkpoints keep exact-equality comparison.

## Test plan

- [ ] `run-ci-megatron` with weight checking enabled everywhere


Zhichenzzz (approved):
LGTM

PR #1339, opened 2026-06-12
[AMD CI] [1/N] Support Python 3.10 on ROCm base

XinyuJiangCMU:
Make `miles` compatible with Python 3.10 on the ROCm SGLang base image.
The ROCm base uses Python 3.10 while the CUDA base uses Python 3.12. A few call sites relied on Python 3.11+ APIs, which caused test collection to fail on ROCm before any tests could run.
Co-authored with @JessicaJiang-123 

## Changes
- Add a Python 3.10-compatible `StrEnum` fallback via `backports.strenum`.
- Guard `Exception.add_note()` calls so Python 3.10 preserves the original exception instead of raising `AttributeError` while handling it.
- Update requirements for Python 3.10:
  - `backports.strenum ; python_version < "3.11"`
  - `hypothesis >= 5.40`
## Verification

Verified on the Python 3.10 ROCm base, CPU only.

### StrEnum

Before this PR, test collection failed at `tests/conftest.py` because `StrEnum` could not be imported from `enum`:

```
$ pytest --collect-only tests/fast/utils/chat_template_utils/...
ImportError: cannot import name 'StrEnum' from 'enum'
```

After this PR:

```
$ pytest --collect-only tests/fast/utils/chat_template_utils/...
119 tests collected
```

In CI, `stage-a-cpu` now runs the full suite and passes with 2623 tests.

### Exception.add_note()

On Python 3.10, a bare `e.add_note(...)` raises while handling the original exception, masking it:

```
# before
AttributeError: 'ValueError' object has no attribute 'add_note'

# after
ValueError: original error
```

The guard preserves and re-raises the original exception.

### hypothesis

The ROCm base ships `hypothesis==5.35.1`, which predates `HealthCheck.function_scoped_fixture`.

```
$ pytest --collect-only tests/fast/ray/rollout/test_train_data_conversion.py
E   AttributeError: function_scoped_fixture
collected 0 items / 1 error
```
After updating hypothesis:
```
$ pytest --collect-only tests/fast/ray/rollout/test_train_data_conversion.py
26 tests collected
```
No behavior change on Python 3.11/3.12.

## Notes

No behavior change on Python 3.11/3.12. Native `StrEnum` and `Exception.add_note()` are still used when available, and `backports.strenum` is only installed on Python < 3.11.


XinyuJiangCMU:
@HaiShaw @RolaoDenthu could you take a look at this PR when you have a chance?

guapisolo (approved):
LGTM.

Zhichenzzz (approved):

PR #1345, opened 2026-06-15
fix(rollout): score SGLang logprobs at rollout temperature

EazyReal:
Twin of THUDM/slime#2085 (OPD teacher temperature), plus the Miles-only prefill recompute path. Recut onto current `main`.

## Problem

Importance-sampling corrections (TIS / `--use-rollout-logprobs`), train/rollout mismatch metrics, and on-policy distillation compare SGLang-computed log-probs against training-side log-probs. The invariant is consistency: both sides must use the same temperature, not always T=1.

Two re-score paths were still hardcoding `temperature: 0`:

- `--recompute-logprobs-via-prefill`
- the on-policy-distillation teacher scorer

When `rollout_temperature != 1`, those paths score at the wrong temperature.

## Fix

Score both paths with `args.rollout_temperature` instead of hardcoded `0`. Unchanged at the default `rollout_temperature=1.0`.

## Tests

- `tests/fast/rollout/generate_utils/test_prefill_logprobs.py`
- `tests/fast/rollout/test_on_policy_distillation.py`

### DIFF
--- a/miles/rollout/generate_utils/prefill_logprobs.py
+++ b/miles/rollout/generate_utils/prefill_logprobs.py
@@ -38,7 +38,7 @@ def _build_prefill_scoring_payload(
         "sampling_params": {
             **dict(sampling_params),
             "max_new_tokens": 0,
-            "temperature": 0,
+            "temperature": args.rollout_temperature,
             "skip_special_tokens": False,
         },
         "return_logprob": True,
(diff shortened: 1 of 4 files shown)


EazyReal:
@fzyzcjy @yueming-yuan @guapisolo @Shi-Dong may I get a quick review on this temperature mismatch fix?

PR #1351, opened 2026-06-16
Auto-enable router dp_aware routing when DP attention is on

Shi-Dong:
## Summary

When DP attention is enabled (`--sglang-data-parallel-size > 1`, which Miles already forces alongside `--sglang-enable-dp-attention`), the SGLang engine shards its KV cache across DP ranks — each rank keeps an independent KV pool and its own radix prefix tree. The sgl-router, however, defaults to `dp_aware = False`, so it dispatches at engine granularity and lets SGLang scatter requests across DP ranks by load. That fragments the radix prefix cache across ranks and can sharply reduce the KV-cache hit rate (a shared prefix that used to be cached once is recomputed on every rank it lands on).

This PR couples the two: in the sgl-router path, when `sglang_dp_size > 1`, Miles now auto-enables `router_args.dp_aware` so the router routes at DP-rank granularity (preserving prefix-cache locality).

## Behavior

- Triggers only on the **sgl-router** path (`start_router`'s non-`use_miles_router` branch). The miles-router path (`RadixTreeMiddleware`) and externally-managed routers (`--sglang-router-ip` set) are untouched.
- Honors an explicit `--router-dp-aware` (already `True` → left as-is).
- Provides an opt-out via `MILES_DISABLE_AUTO_DP_AWARE=1`. Because `--router-dp-aware` is a `store_true` flag (no native "explicit False"), the env var is the supported way to keep DP attention on while routing per-engine — and it **warns** when used, since it reintroduces the prefix-cache fragmentation.
- Routing policy is left alone (router already defaults to `cache_aware`; `dp_aware` is orthogonal, controlling worker granularity rather than the selection policy).

## Test plan

- [x] Unit test (`tests/fast/router/test_router_dp_aware.py`) covering the four branches: auto-enable when `dp_size > 1`, no-op when `dp_size == 1`, respect an explicit `dp_aware=True`, and opt-out keeps it disabled + warns.
- [x] Helper logic exercised directly across all four cases locally.
- [ ] Smoke: a short DP-attention rollout run (`dp_size >= 2`) confirming the router log shows `dp_aware` active and the KV-cache prefix hit-rate metric recovers vs. a baseline without the change.

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -30,7 +30,8 @@ def _maybe_enable_router_dp_aware(args, router_args) -> None:
     ``MILES_DISABLE_AUTO_DP_AWARE=1`` opt-out, warning loudly in the opt-out case
     since it reintroduces the prefix-cache fragmentation.
     """
-    if getattr(args, "sglang_dp_size", 1) <= 1 or router_args.dp_aware:
+    dp_size = getattr(args, "sglang_dp_size", 1) or 1
+    if dp_size <= 1 or getattr(router_args, "dp_aware", False):
         return
     if os.environ.get("MILES_DISABLE_AUTO_DP_AWARE") == "1":
         logger.warning(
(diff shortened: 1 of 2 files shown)


miles/ray/rollout/router_manager.py
```diff
@@ -16,6 +17,40 @@
 logger = logging.getLogger(__name__)
 
 
+def _maybe_enable_router_dp_aware(args, router_args) -> None:
```
guapisolo, line 20: I suggest put this implicit argument setting in `arguments.py`

lawrence-harmonic:
This may break PD disaggregation if DP aware does not work correctly with PD disaggregation yet.

lawrence-harmonic:
DP aware support for PD router is in: https://github.com/lightseekorg/smg/pull/1522

This needs to be added to https://github.com/radixark/sgl-router-for-miles I think.

PR #1357, opened 2026-06-17
fix: fork session/router servers before wandb.init to avoid event-loop deadlock

Shi-Dong:
## Summary
The miles session server (`--use-session-server`) and the SGLang router are each launched as a `multiprocessing.Process` forked from `RolloutManager.__init__`. On Linux + Python 3.12 the default start method is `fork`, so the child inherits the parent's memory — including the service-backed objects created by `wandb.init()` (run via `init_tracking(..., primary=False)` earlier in `__init__`).

Those wandb objects carry weakref finalizers that call into a background wandb service thread. The thread is **not** recreated in the fork, so when Python GC finalizes an inherited object inside the child's single uvicorn event-loop thread (during HTTP `handle_events`), the finalizer blocks **forever** on a `concurrent.futures` result that will never be set. The event loop freezes, uvicorn stops `accept()`-ing, the `:30000` accept backlog grows, and every agent `chat/completions` request times out → `collect_records` returns empty → no training step.

This was root-caused with `py-spy dump` on the wedged session-server process; the blocking stack is:
```
wait (threading.py:355) <- result (concurrent/futures/_base.py:451)
<- run (wandb/sdk/lib/asyncio_manager.py:137)
<- api_cleanup_request (wandb/.../service_connection.py:201)
<- _cleanup (wandb/apis/public/service_api.py:31)
<- weakref __call__ ... handle_events (uvicorn/.../h11_impl.py:262)
<- run_session_server (miles/rollout/session/session_server.py:111)
```

## Fix
Move `init_tracking(args, primary=False, ...)` to **after** the router/session-server subprocesses are forked. The children then fork from a process where `wandb.init()` has not yet run, so they never inherit the service-backed objects and the finalizer deadlock cannot occur. Nothing between the old and new call sites logs to wandb, so the ordering change is safe.

This reorder is intentionally a minimal, low-risk stop-the-bleeding fix. The deeper root cause is forking a multi-threaded Ray actor at all: the reorder removes the one wandb instance of the hazard, but the next library that holds a thread or lock before the fork point would reintroduce the same deadlock under a new name. The principled fix is to launch these two processes under a `spawn` context (`mp.get_context("spawn").Process(...)`), so the children inherit none of the parent's threads/finalizers — matching how we already start the SGLang engine from inside a Ray actor (`sglang_engine.py` forces `spawn`). That is left as a follow-up because it needs `args`/`router_args` to be picklable and is best validated with a full run.

## Test plan
- [ ] Relaunch the DeepSeek-V4-Flash disagg agentic run and confirm the session server keeps accepting on `:30000` (Recv-Q stays ~0) and a training step lands.

### DIFF
--- a/miles/ray/rollout/rollout_manager.py
+++ b/miles/ray/rollout/rollout_manager.py
@@ -82,9 +82,7 @@ class RolloutManager:
         # child's single uvicorn event-loop thread blocks it forever and the server
         # stops accepting connections. Forking before wandb.init() avoids inheriting
         # that state entirely.
-        # router_addr stays None when the router was not started (e.g. debug_train_only),
-        # otherwise "http://None:None" would slip past the `is not None` guard in
-        # init_wandb_secondary and point metric forwarding at an invalid endpoint.
+        # No router_addr when the router was not started (e.g. debug_train_only).
         router_addr = (
             f"http://{args.sglang_router_ip}:{args.sglang_router_port}"
             if args.sglang_router_ip and args.sglang_router_port
--- a/miles/ray/rollout/rollout_manager.py
+++ b/miles/ray/rollout/rollout_manager.py
@@ -82,7 +82,15 @@ class RolloutManager:
         # child's single uvicorn event-loop thread blocks it forever and the server
         # stops accepting connections. Forking before wandb.init() avoids inheriting
         # that state entirely.
-        init_tracking(args, primary=False, router_addr=f"http://{args.sglang_router_ip}:{args.sglang_router_port}")
+        # router_addr stays None when the router was not started (e.g. debug_train_only),
+        # otherwise "http://None:None" would slip past the `is not None` guard in
+        # init_wandb_secondary and point metric forwarding at an invalid endpoint.
+        router_addr = (
+            f"http://{args.sglang_router_ip}:{args.sglang_router_port}"
+            if args.sglang_router_ip and args.sglang_router_port
+            else None
+        )
+        init_tracking(args, primary=False, router_addr=router_addr)
 
         self.rollout_engine_lock = Lock.options(num_cpus=1, num_gpus=0).remote()
         self.rollout_id = -1
--- a/miles/ray/rollout/rollout_manager.py
+++ b/miles/ray/rollout/rollout_manager.py
@@ -75,13 +75,9 @@ class RolloutManager:
             self.servers = start_rollout_servers(args, pg)
             start_session_server(args)
 
-        # Initialise tracking only AFTER the router and session-server subprocesses
-        # have been forked. wandb.init() spins up service-backed objects whose
-        # weakref finalizers call into a background service thread; a forked child
-        # inherits those objects but not the thread, so finalizing one inside the
-        # child's single uvicorn event-loop thread blocks it forever and the server
-        # stops accepting connections. Forking before wandb.init() avoids inheriting
-        # that state entirely.
+        # Init tracking only after the servers are forked: a child forked after
+        # wandb.init() inherits its finalizers but not its service thread, which
+        # deadlocks the child's uvicorn event loop.
         # No router_addr when the router was not started (e.g. debug_train_only).
         router_addr = (
             f"http://{args.sglang_router_ip}:{args.sglang_router_port}"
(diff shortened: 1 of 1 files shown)


fzyzcjy:
> _Note: this comment is written by an AI agent (Claude), on behalf of a reviewer. Please double-check the claims and citations before acting on them._

Great root-cause work here — the `py-spy` capture and the wandb finalizer / missing-service-thread deadlock explanation are convincing, and reordering `init_tracking()` after the forks is a fine, low-risk **stop-the-bleeding** fix. 👍

That said, I'd push back on the **"`spawn` was considered and rejected"** paragraph. I believe the stated reason is incorrect, and the conclusion is actually backwards: inside a Ray actor, `fork` is the dangerous/unsupported method and `spawn` is the recommended one. Three independent lines of evidence:

### 1. The "re-import `default_worker` as `__main__`" reason is factually wrong

When `multiprocessing` spawn re-imports the parent's main module in the child, it imports it under the name **`__mp_main__`**, *not* `__main__`. Ray's `python/ray/_private/workers/default_worker.py` has:

- **Module top level (runs on import):** only `import` statements + building an `argparse.ArgumentParser`. No raylet connection, no worker registration, no threads.
- **Everything dangerous** — arg parsing, `Node`/`RayParams`, `ray._private.worker.connect()` to the raylet, the worker main loop — is behind `if __name__ == "__main__":`.

Since the spawned child imports it as `__mp_main__`, `__name__ != "__main__"`, so the guarded bootstrap is **skipped**. The child does **not** try to re-register as a Ray worker or connect to the raylet. The re-import is benign.

### 2. We already spawn from inside a Ray actor in this exact codebase — in production

SGLang forces `spawn`, and miles drives it from inside a Ray actor today:

- `python/sglang/srt/entrypoints/engine.py` → `mp.set_start_method("spawn", force=True)` (in `_set_envs_and_config`, called from `Engine._launch_subprocesses`); scheduler/detokenizer/router subprocesses are all created under that spawn context.
- `miles/backends/sglang_utils/sglang_engine.py:63` → inside `ray.remote(SGLangEngine)` (`server_group.py`: `RolloutRayActor = ray.remote(SGLangEngine)`), we explicitly call `multiprocessing.set_start_method("spawn", force=True)` and then `multiprocessing.Process(target=launch_server, ...)`.

So "spawn from inside a Ray actor" is not theoretical here — it's how our SGLang engine starts every run, and those children are *heavier* (full SGLang server + scheduler + detokenizer) than the router / session-server. If spawning from a Ray actor were broken by `default_worker`, the engine wouldn't come up at all.

### 3. Ray itself says: use spawn, never fork

- Ray maintainers recommend `mp.get_context("spawn")` from inside a `@ray.remote` actor ([discuss.ray.io #2165](https://discuss.ray.io/t/best-solution-to-have-multiprocess-working-in-actor/2165)), and Ray docs sanction user-spawned processes ([Lifetimes of a User-Spawn Process](https://docs.ray.io/en/latest/ray-core/user-spawn-processes.html)).
- `fork` from a Ray worker is a known hang: maintainer quote *"Ray is not working with fork because it has some in-memory states that shouldn't be duplicated"* ([ray#28328](https://github.com/ray-project/ray/issues/28328), nested-fork variant [ray#31263](https://github.com/ray-project/ray/issues/31263)).

### Why this matters for the fix

The current PR removes **one** inheritance source (wandb) by making sure the fork happens before `wandb.init()`. But the structural hazard — forking a multi-threaded Ray actor — is untouched. It relies on the fork point staying "clean enough" forever; the next library that initializes a background thread or holds a lock before that fork point (grpc client, CUDA, another tracker, …) reintroduces the same class of deadlock under a new name.

Switching these two `multiprocessing.Process` calls to a spawn context (`mp.get_context("spawn").Process(...)`) would root-cause the whole category at once, with zero inheritance. The usual spawn caveats are all satisfied here:

- **Target must be in an importable module** — `run_router` / `run_session_server` already are (module-level functions, not `__main__`). ✅
- **Args must be picklable** — `router_args` / `args` are config objects; drop any non-picklable members before launch if needed. ✅
- **Placement group is lost in the child** — irrelevant; these are plain HTTP processes that need no Ray scheduling resources. ✅
- **Child lifecycle** — already `daemon=True`. ✅

### Suggestion

Happy to merge this as the immediate fix, but I'd suggest (a) dropping/rewriting the "spawn was rejected" paragraph since the premise doesn't hold, and (b) a follow-up that moves the router/session-server launches to a `spawn` context (matching how we already start the SGLang engine), which eliminates the inheritance-deadlock class rather than just the wandb instance of it.

PR #1275, opened 2026-06-02
Use q lora atomic update groups

yueming-yuan:
## Summary
- replace q_lora conversion cache with static atomic update groups
- keep q_a_proj and kv_a_proj_with_mqa in the same update unit before HF conversion
- keep convert_to_hf stateless for q_lora weights

## Validation
- python -m compileall miles/backends/megatron_utils/megatron_to_hf/__init__.py
- python -m pytest --confcutdir=tests/fast/backends/megatron_utils tests/fast/backends/megatron_utils/test_hf_weight_iterator_direct.py -q
- pre-commit run --all-files

PR #1361, opened 2026-06-17
Pair q_a_proj / kv_a_proj_with_mqa in the bridge weight iterator

nanjiangwill:
## Summary
- When `q_lora_rank` is set (MLA models like Kimi K2.5 / DeepSeek-V3), SGLang's deepseek loader fuses `q_a_proj` and `kv_a_proj_with_mqa` into a single `fused_qkv_a_proj_with_mqa` weight, but only when both halves arrive in the same `load_weights` call.
- `chunk_named_params_by_size` streams the bridge's HF-name output in iterator order, which can place `q_a_proj` and `kv_a_proj_with_mqa` for the same layer in different chunks — leaving the fusion partially applied and corrupting that layer's weights on the rollout engine.
- This PR pairs the two halves on the sender side in `hf_weight_iterator_bridge.py`, mirroring the approach slime adopted in [THUDM/slime#1753](https://github.com/THUDM/slime/pull/1753): hold each half until its partner arrives, then yield both adjacent so the chunker keeps them together. The pairing covers both the base path (`q_a_proj.weight`) and the LoRA path (`q_a_proj.lora_{A,B}.weight`).
- With this change, no SGLang-side workaround is needed (a follow-up reverts the persistent `cached_a_proj` change from sgl-project/sglang#25141 on the `sglang-miles` branch).

## Notes
- The `hf_weight_iterator_direct.py` path already keeps the pair together via the existing `AtomicUpdateGroup` (`_get_q_lora_atomic_update_groups`), so it's only the bridge path that needed the fix.
- The helper is a small generator local to `hf_weight_iterator_bridge.py`; it's gated on `args.q_lora_rank is not None`, so non-MLA models are unaffected.

## Test plan
- [ ] Run K2.5 LoRA training and confirm `train/train_rollout_logprob_abs_diff` stays bounded (~0.03) across steps, without the persistent-cache patch in `deepseek_weight_loader.py`.
- [ ] Existing CI on non-MLA models (Qwen, Llama, etc.) — the new path is a no-op when `q_lora_rank is None`.

### DIFF
--- a/miles/backends/megatron_utils/update_weight/hf_weight_iterator_bridge.py
+++ b/miles/backends/megatron_utils/update_weight/hf_weight_iterator_bridge.py
@@ -4,11 +4,11 @@ import os
 
 from miles.backends.megatron_utils.lora_utils import is_lora_weight_name
 from miles.utils import megatron_bridge_utils
-from miles.utils.iter_utils import chunk_named_params_by_size
 
 from ..megatron_to_hf import postprocess_hf_param
 from ..megatron_to_hf.processors import quantize_params
 from ..misc_utils import strip_param_name_prefix
+from .common import get_atomic_update_groups
 from .hf_weight_iterator_base import HfWeightIteratorBase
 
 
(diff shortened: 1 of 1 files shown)


yueming-yuan:
is it possible to make it unified with atomic weight update group?
https://github.com/radixark/miles/pull/1264
https://github.com/radixark/miles/pull/1275

nanjiangwill:
> is it possible to make it unified with atomic weight update group? #1264 #1275

make sense! Rewired to use the existing \`AtomicUpdateGroup\` infrastructure (\`get_atomic_update_groups\` already declares \`q_lora_a_proj\` for MLA), so this no longer hard-codes the q-lora suffixes and any future model-specific group (e.g. deepseek-v4 once that lands here) will automatically apply on the bridge path the same way it already does for the direct iterator (#1264) and the distributed mixin

yueming-yuan (approved):

PR #1363, opened 2026-06-18
Clamp PPO ratio exponentials to fix NaN logits

ys-2020:
## Summary

This PR adds a small numerical guard around PPO log-ratio exponentials in the training loss utilities.

It clamps the exponent input before computing importance ratios, and applies the same guard to the `low_var_kl` estimator before its exponential. The change is intentionally local to `miles/backends/training_utils/loss_hub/math_utils.py`.

## Motivation

During long-context agentic RL runs, a small number of samples can produce extreme token log-ratios. The current code computes the PPO importance ratio directly with:

```python
ratio = (-ppo_kl).exp()
```

For very large exponent inputs, this can overflow to `inf`. Once that happens, normal PPO arithmetic can produce `NaN`, for example through `inf * 0` when the corresponding advantage is zero. The clipped PPO branch does not fully avoid this because the unclipped branch is still computed.

The same direct exponential pattern also appears in ESS metric computation, and `low_var_kl` has the same overflow risk when computing `log_ratio.exp() - 1 - log_ratio`.

## What Changed

- Added `_safe_clamp_log_ratio`, which centralizes finite-input handling and exponent-input clamping.
- Added `_safe_exp_neg_ppo_kl`, which:
  - casts the log-ratio to fp32 before exponentiation;
  - maps non-finite inputs to bounded values;
  - clamps the exponent input to `[-20, 20]`;
  - returns a finite importance ratio.
- Replaced direct `exp(-ppo_kl)` usage in policy loss and ESS contribution.
- Added the same finite-input guard before the `low_var_kl` exponential.

## Why This Is Safe

`exp(20)` is already far outside the normal PPO clipping range, so the clamp only affects pathological outliers that would otherwise overflow. For normal PPO ratios, behavior is unchanged.

This does not change rollout behavior, rewards, optimizer state, parallelism, checkpoint format, or data loading. It only prevents non-finite arithmetic in the loss helper.

## Testing

```bash
PYTHONDONTWRITEBYTECODE=1 python3 -m py_compile miles/backends/training_utils/loss_hub/math_utils.py
PYTHONPATH=. python3 -m py_compile tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
PYTHONPATH=. python3 tests/ci/run_suite.py --hw cpu --suite stage-a-cpu --list-only
```

### DIFF
--- a/tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
+++ b/tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
@@ -39,5 +39,14 @@ def test_low_var_kl_extreme_log_ratios_stay_finite():
     kl = compute_approx_kl(log_probs, log_probs_base, kl_loss_type="low_var_kl")
 
     assert torch.isfinite(kl).all().item()
-    assert torch.all(kl <= 10).item()
-    assert torch.all(kl >= -10).item()
+
+
+def test_low_var_kl_matches_unclamped_formula_for_normal_log_ratios():
+    log_probs = torch.tensor([-0.1, 0.0, 0.1], dtype=torch.float32)
+    log_probs_base = torch.zeros_like(log_probs)
+
+    kl = compute_approx_kl(log_probs, log_probs_base, kl_loss_type="low_var_kl")
+
+    log_ratio = -(log_probs - log_probs_base)
+    expected_kl = log_ratio.exp() - 1 - log_ratio
+    torch.testing.assert_close(kl, expected_kl)
--- a/tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
+++ b/tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
@@ -1,0 +1,43 @@
+import torch
+
+from miles.backends.training_utils.loss_hub.math_utils import compute_approx_kl, compute_policy_loss
+
+
+def test_policy_loss_extreme_log_ratios_with_zero_advantages_stay_finite():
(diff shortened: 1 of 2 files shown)


Shi-Dong (commented):
The changes are fine, but it'd be better to have extra unit tests to guardrail the changes. Please include the tests in this PR as well.

tests/fast/backends/training_utils/test_ppo_ratio_numerics.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_low_var_kl_extreme_log_ratios_stay_finite():
+    log_probs = torch.tensor([-1000.0, 1000.0, float("nan"), float("inf"), float("-inf")])
+    log_probs_base = torch.zeros_like(log_probs)
+
+    kl = compute_approx_kl(log_probs, log_probs_base, kl_loss_type="low_var_kl")
+
+    assert torch.isfinite(kl).all().item()
+    assert torch.all(kl <= 10).item()
+    assert torch.all(kl >= -10).item()
```
Shi-Dong: Hmm I thought we are only clamping the inputs, not the outputs?  Are you sure that this will pass?

ys-2020: The original code already clamps the outputs in low_var_kl mode, so the previous test ended up covering that as well. But yes, I’ve revised the unit test to focus on the input-clamping behavior.

Shi-Dong (approved):
LGTM!

PR #1367, opened 2026-06-18
spawn router/session-server subprocesses instead of forking

Shi-Dong:
## Summary
Launch the SGLang/miles router and the session server with a `spawn` start method instead of the Linux default `fork`. `RolloutManager` is a Ray actor, and forking a multi-threaded actor lets the child inherit objects whose backing threads/locks do **not** come across the fork. The most recent instance was wandb: a forked session server inherited `wandb.init()` finalizers but not the wandb service thread, so GC of an inherited object deadlocked the child's uvicorn event loop and the run wedged at "Collected 0/N" (root-caused with `py-spy` in #1357).

#1357 fixes that one instance by reordering `init_tracking()` after the fork. This PR removes the underlying hazard for these two processes: a spawned child is a fresh interpreter that inherits none of the parent's threads/finalizers/locks, so the whole class of fork-inheritance deadlocks (wandb today; the next library that holds a thread or lock before the launch point tomorrow) cannot occur. This matches how we already launch the SGLang engine from inside a Ray actor — `sglang_engine.py:58` forces `set_start_method("spawn", force=True)`.

## Why this is safe
The spawn caveats are all satisfied:
- **Targets are importable** — `run_router` / `run_session_server` are module-level functions, not `__main__`.
- **Args are picklable** — `args` is a plain `argparse.Namespace` of config scalars/lists/None; Ray already cloudpickles it into the actor constructor, and a standard `pickle` round-trip succeeds for both the direct-`args` (session server) and `copy.copy(args)` (miles router) paths. If a non-picklable field is ever added, `Process.start()` raises a clear `PicklingError` in the parent.
- **Ray actor + spawn is supported** — spawn re-imports the parent's main module as `__mp_main__` (not `__main__`), so Ray's `default_worker` bootstrap guarded by `if __name__ == "__main__":` is skipped; the child does not re-register as a Ray worker. Ray maintainers recommend `mp.get_context("spawn")` from inside actors and warn that `fork` from a worker is a known hang.
- **Child lifecycle** — unchanged, still `daemon=True`.

## Test plan
- [ ] Relaunch the DeepSeek-V4-Flash disagg agentic run and confirm the router and session server come up and a training step lands.
- [ ] Confirm `:30000` keeps accepting (Recv-Q ~0) across the run.

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -1,17 +1,16 @@
+import copy
 import logging
 import multiprocessing
 import random
 import uuid
 
+from sglang_router.launch_router import RouterArgs
 
-from miles.utils.http_utils import (
-    _wrap_ipv6,
-    find_available_port,
-    get_host_info,
-    is_port_available,
-    wait_for_server_ready,
-)
-
+from miles.rollout.session.session_server import run_session_server
+from miles.router.router import run_router as run_miles_router
+from miles.utils.http_utils import _wrap_ipv6, find_available_port, get_host_info, is_port_available
+from miles.utils.http_utils import run_router as run_sglang_router
+from miles.utils.http_utils import wait_for_server_ready
 
 logger = logging.getLogger(__name__)
 
@@ -34,20 +33,15 @@ def start_router(args, *, has_pd_disaggregation: bool = False, force_new: bool =
             router_port = find_available_port(random.randint(3000, 4000))
 
     if args.use_miles_router:
-        import copy
-
         assert not has_pd_disaggregation, "miles router does not support PD disaggregation."
-        from miles.router.router import run_router
 
+        run_router = run_miles_router
         router_args = copy.copy(args)
         router_args.sglang_router_ip = router_ip
         router_args.sglang_router_port = router_port
 
     else:
-        from sglang_router.launch_router import RouterArgs
-
-        from miles.utils.http_utils import run_router
-
+        run_router = run_sglang_router
         router_args = RouterArgs.from_cli_args(args, use_router_prefix=True)
         router_args.host = router_ip
         router_args.port = router_port
@@ -112,8 +106,6 @@ def start_session_server(args):
         )
 
     router_url = f"http://{args.sglang_router_ip}:{args.sglang_router_port}"
-
-    from miles.rollout.session.session_server import run_session_server
 
     # spawn (not fork): see start_router for rationale.
     process = multiprocessing.get_context("spawn").Process(target=run_session_server, args=(args, router_url))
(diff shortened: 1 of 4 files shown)


guapisolo (approved):
LGTM. My agent said it's a root fix

Shi-Dong:
Will merge after end-to-end testing on a training run.

fzyzcjy (approved):
my agent claims to see issue caused by fork as well today, thus re-read latest pr and lgtm

fzyzcjy:
We independently hit the same fork-inheritance deadlock on `trainer_ft/dev` (#823), but on the **miles router** process rather than the session server — so this PR fixes (at least) a second instance of the class. Sharing the root cause as corroboration.

**Symptom:** `tests/e2e/lora/test_lora_qwen2.5_0.5B.py` hangs the full 1800s on the step-0 eval (the test runs with `--use-miles-router` + `--eval-interval 2`). `Eval gsm8k` is pinned at `0/1319`; the sglang engines go idle right after `load_lora_adapter_from_tensors 200 OK` and never run another batch.

**Root cause (py-spy, identical across 19 dump cycles / ~19 min):** the forked `run_miles_router` child's single uvloop event-loop thread is frozen while forwarding the eval `generate` POST to sglang, because a GC'd wandb weakref finalizer fires on that same thread and does a blocking `concurrent.futures.Future.result()` on the wandb service thread — which didn't survive the fork:

```
proxy         (miles/router/router.py:127)
do_proxy      (miles/router/router.py:149)        # await self.client.request(...) -> sglang
request       (httpx/_client.py:1527)
build_request (httpx/_client.py:367)
__call__      (weakref.py:590)                     # GC finalizer fires on the event-loop thread
_cleanup      (wandb/apis/public/service_api.py:31)
api_cleanup_request (wandb/.../service_connection.py:201)
run           (wandb/sdk/lib/asyncio_manager.py:137)
result        (concurrent/futures/_base.py:451)    # blocking wait on the (gone) wandb service thread
PyThread_acquire_lock_timed                        # never returns
```

Dependency chain, all from live stacks in the same dump cycle:
- driver blocked on `await rollout_manager.eval.remote(0)` (`train.py:138`, `run_until_complete`)
- eval worker (`ray::RolloutManager.eval`) blocked in `.result()` (`miles/utils/async_utils.py:29`)
- **router loop frozen as above** (`run_miles_router` child, forked from the `RolloutManager` actor)
- all 4 `sglang::scheduler` idle in `recv_pyobj` — the request never reached sglang

So the stall is entirely upstream in the forked router; sglang is healthy and just waiting for a request that never arrives.

This is exactly the hazard your PR removes: `start_router`'s `run_miles_router` path forked the router from the `RolloutManager` actor and inherited wandb's finalizers but not its backing thread. Switching to `get_context("spawn")` gives the router a fresh interpreter with no inherited finalizers, so this can't occur. +1 — this fixes our instance too, and it's a stronger fix than reordering `init_tracking()` since it kills the whole class. We'll pull it in once it lands on `main`.

PR #1376, opened 2026-06-19
Support GLM-5.2 744B-A40B

yueming-yuan:
This PR is migrated from ([THUDM/slime#2093](https://github.com/THUDM/slime/pull/2093)).

Thanks @zhuzilin!

### DIFF
--- a/miles_plugins/models/glm5/glm5.py
+++ b/miles_plugins/models/glm5/glm5.py
@@ -31,6 +31,28 @@ from miles.utils.replay_base import indexer_replay_manager
 from .ops.indexer import generate_varlen_mask_params, lighting_indexer
 from .ops.sparse_mla import SparseMLA
 
+# Names of the indexer submodules. On a DSA model with *cross-layer index
+# sharing* these only exist on "computing" layers; "skip" layers drop them.
+_INDEXER_SUBMODULE_NAMES = ("wq_b", "wk", "k_norm", "weights_proj")
+
+
+def is_skip_topk_layer(layer_number: int, skip_topk_offset: int, topk_freq: int) -> bool:
+    """Whether the (1-indexed) Megatron ``layer_number`` reuses a previous layer's top-k.
+
+    Mirrors ``glm-train-prod``'s ``_get_skip_topk_flags``: a layer *computes* its
+    own top-k when ``max(layer_number - offset, 0) % freq == 0``; otherwise it is a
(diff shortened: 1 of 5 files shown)


yushengsu-thu:
cc catches a potential issue:

 Full-model checkpoint conversion will crash on the new assertion

  run_glm5_2_744b_a40b.py → _prepare_megatron_ckpt (the non-pruned branch) keeps the GLM-5 conversion layout verbatim:

  extra_args += (
      "--pipeline-model-parallel-size 4 "
      "--expert-model-parallel-size 32 "
      "--decoder-last-pipeline-num-layers 18 "
  )

  Conversion runs tools/convert_hf_to_torch_dist.py, which builds the model via get_model(get_model_provider_func(args), ...) (line 110) → get_glm5_spec per pipeline stage. get_glm5_spec loads the GLM-5.2 HF config
  (index_topk_freq=4 > 1), so the new assertion is active during conversion, not just training.

  With PP=4 and last=18 over 78 layers, Megatron splits [20, 20, 20, 18] → stages start at 1-indexed layers 1, 21, 41, 61. For freq=4, offset=3:

  - is_skip_topk_layer(21,3,4) → (18 % 4) = 2 ≠ 0 → skip
  - 41 → (38 % 4)=2 → skip · 61 → (58 % 4)=2 → skip

  So pp_ranks 1/2/3 each start on a skip layer and raise AssertionError("DSA index-share pipeline split is invalid…"). The documented full-model prepare step aborts.

  This is almost certainly an oversight rather than intent: the author did account for the constraint everywhere else — the train layout (PP=8), and even the 5-layer convert path, whose comment explicitly says "nproc=1
  also avoids the convert tool's PP auto-bump (which would split onto a skip layer)." The full-model convert split was just inherited from run_glm5_744b_a40b.py (where freq=1, so the assertion never fired).

  Why it slips through CI: the 5-layer smoke test converts via the <= 5 branch (PP=1, single GPU), so the multi-stage split is never exercised.

  Fix: use an index-share-valid conversion split. Either reuse the training layout (PP=8, first=14/last=16), or a tuned PP=4 split such as first=14/last=24 → [14,20,20,24] → starts 1,15,35,55 (all ≡ computing). Verify the
  chosen split against is_skip_topk_layer for each stage start.

yueming-yuan:
@yushengsu-thu good catch! fixed

maocheng23 (approved):

PR #1378, opened 2026-06-20
fix(http): don't retry /generate after the request was sent

### LINKED ISSUE #1334: Mid-flight `/generate` retry can double-generate: `_post` retries any exception and there is no rid to abort the orphaned request
## Summary

`_post` in `miles/utils/http_utils.py` retries **any** `Exception` up to `max_retries=60`:

```python
async def _post(client, url, payload, max_retries=60, action="post", headers=None):
    ...
        except Exception as e:
            retry_count += 1
            ...
            await asyncio.sleep(1)
            continue
```

For `/generate` this is unsafe: the request payload carries no `rid` (`compute_request_payload` builds only `input_ids` / `sampling_params` / multimodal fields; `git grep '"rid"' -- miles/` is empty), and the only abort ever issued is `{"abort_all": True}`.

## Failure mode

1. A `/generate` POST reaches the SGLang server; generation starts.
2. The connection dies mid-flight (reset, `httpx.RemoteProtocolError`, proxy hiccup) — the server keeps generating; the client sees an exception.
3. `_post` retries → a **second** generation of the same prompt starts.
4. The first (orphaned) request cannot be aborted or deduped — there is no rid — so it runs to completion, wasting KV/compute, and stays in flight where it can race a subsequent pause/flush/memory-release transition.

The retry-anything loop is the right call for genuinely connection-level failures (e.g. the engine briefly refusing connections around a weight update — `httpx.ConnectError`, nothing reached the server, retry is free). It is only the **post-send** retry of a non-idempotent `/generate` that double-generates.

## Possible fixes

1. **rid + abort-by-rid (root cause).** Attach a client-generated `rid` to each `/generate` payload (SGLang accepts it), and on a post-send failure abort that rid before retrying. This keeps the full resilience of the current loop while making the retry idempotent. This is the design the sibling slime change uses (THUDM/slime#2069: pooled client, retry once, same rid, pre-connect errors only).
2. **Minimal alternative:** restrict the `/generate` path's retry to pre-connect errors (`httpx.ConnectError`). Cheaper, but loses retry coverage for the case where the server itself died mid-request (there the orphan died too and a blind retry would actually be safe) — which is why rid-based abort is the cleaner fix.

Happy to send a PR for either direction if maintainers have a preference.

EazyReal:
> Draft: scope needs maintainer clarification. Training-step latency is larger than a wasted retry, so a blanket “never retry /generate after send” may be the wrong tradeoff. Tool-call / generate-hub call sites need a separate discussion. Not ready to merge.

## What

`http_utils._post` retries *any* exception up to `max_retries` (~60). For `/generate` that's unsafe: a failure that happens **after** the request reached the server (while generation is already running) gets retried, and since `/generate` carries no request id to abort the orphan, the retry re-issues generation — a double-generate.

## Fix

Thread an `idempotent` flag through `post`/`_post` (and the Ray `do_post` actor). Non-idempotent calls — the three `/generate` call sites — retry **only** pre-send connection-setup errors (`httpx.ConnectError` / `ConnectTimeout`, where the server never saw the request) and re-raise anything raised after the request was sent. Idempotent GET/DELETE/health/scoring calls are unchanged (default `idempotent=True`).

Adds CPU tests for the three cases (non-idempotent: no retry post-send, still retries connect errors; idempotent: retries post-send).

Fixes #1334

### DIFF
--- a/miles/rollout/generate_hub/multi_turn.py
+++ b/miles/rollout/generate_hub/multi_turn.py
@@ -57,7 +57,9 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
             break
 
         gen_t0 = time.time()
-        output = await post(url, payload, headers=compute_routing_headers(args, sample))
+        # /generate is non-idempotent: a retry after the request reached the
+        # server would re-issue generation. Only pre-send errors are retried.
+        output = await post(url, payload, headers=compute_routing_headers(args, sample), idempotent=False)
         sink = None if input.evaluation else TrajectoryLifecycle().sink
         if sink is not None:
             tokens = output.get("meta_info", {}).get("completion_tokens", "")
--- a/miles/rollout/generate_hub/single_turn.py
+++ b/miles/rollout/generate_hub/single_turn.py
@@ -41,7 +41,9 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
         sample.status = halt_status
         return GenerateFnOutput(samples=sample)
 
-    output = await post(url, payload, headers=compute_routing_headers(args, sample))
+    # /generate is non-idempotent: a retry after the request reached the server
+    # would re-issue generation. Only pre-send connection errors are retried.
+    output = await post(url, payload, headers=compute_routing_headers(args, sample), idempotent=False)
     await update_sample_from_response(args, sample, payload=payload, output=output)
 
     return GenerateFnOutput(samples=sample)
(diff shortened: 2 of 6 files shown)


EazyReal:
Good suggestion — applied. The retry loop now bails on any non-`httpx.HTTPError` (e.g. `AssertionError`/`AttributeError`) so programming errors fail fast instead of looping for `max_retries`. Added a test covering it.

EazyReal:
@fzyzcjy @yueming-yuan @guapisolo @maocheng23 @jybsuper @Zhichenzzz @Shi-Dong Could I get a quick review on this HTTP retry safety fix? It avoids retrying `/generate` after the request may already have been sent.

PR #1382, opened 2026-06-22
fix(deepseek_v4): size rope freqs_cis table dynamically for high CP

kakisong:
## Problem

`wrapped_precompute_freqs_cis` in `miles_plugins/models/deepseek_v4/ops/rope.py` hardcodes the rope frequency table to `max_seq_len = 65536`. But `get_freqs_cis_for_cp` slices that table by **global** position:

```python
return freqs_cis[cp_rank * seqlen_local : (cp_rank + 1) * seqlen_local]
```

so the highest CP rank needs entries up to `seqlen_global - 1 = cp_size * seqlen_local - 1`. Whenever `seqlen_global = cp_size * seqlen_local > 65536`, the high ranks get a truncated/empty slice and the forward crashes in `apply_rotary_emb`:

```
RuntimeError: shape '[1, S, 1, 32]' is invalid for input of size 0
```

In packed-THD dynamic-batch SFT, `--max-tokens-per-gpu` is a **per-rank** budget, so `seqlen_global = max_tokens_per_gpu * cp_size` grows with CP. This trips long-context CP training (e.g. CP8 packing past 64K); a typical CP2 / <=64K-packed setup never hits it.

The failing shapes match the arithmetic exactly:
- CP4, `seqlen_local=24576`: rank 2 slices `[49152:65536]` -> 16384 rows -> input `size 524288` (= 16384 x 32).
- CP8, `seqlen_local=27904`: rank 3 start `83712 > 65536` -> empty slice -> input `size 0`.

## Root cause

`65536` is the model's `original_max_position_embeddings` (the pre-YaRN reference length). The frequencies are YaRN-rescaled (`rotary_scaling_factor=16`), so the model supports positions far beyond 65536 — only the **table length** was left at the pre-extension value, while the slicing indexes global positions that CP x packing easily pushes past 64K.

## Fix

1. Derive `max_seq_len` from the run config instead of hardcoding: `max_tokens_per_gpu * cp_size` (the real packed length in THD dynamic-batch mode; `args.seq_length` does not reflect it and defaults to 4096), floored at `original_max_position_embeddings`.
2. Grow the table on demand in `forward` via `ensure_freqs_cis`, because a single packed sample can exceed the `max_tokens_per_gpu * cp_size` budget while still fitting in memory. The table grows monotonically to the actual global length.

Applied to all three `freqs_cis` buffers: main MLA attention, the KV compressor, and the DSA indexer.

The YaRN frequency rescaling is unchanged, so positions <= 65535 stay bit-identical; positions beyond use the already-rescaled (factor 16) frequencies, which is within the model's designed YaRN range. Memory is unaffected for runs that stay within 64K (the table is still 65536) and scales with the actual context otherwise.

## Validation

8xH20 (64 GPU, TP2/PP4/CP8/EP8, packed SFT): the table inits at 65536 and grows to 106496 on demand when an over-budget sample appears; training runs clean across steps — no rope overflow, no OOM.

### DIFF
--- a/miles_plugins/models/deepseek_v4/deepseek_v4.py
+++ b/miles_plugins/models/deepseek_v4/deepseek_v4.py
@@ -34,7 +34,7 @@ from miles_plugins.models.deepseek_v4.ops.cp_utils import (
 )
 from miles_plugins.models.deepseek_v4.ops.kernel.tilelang_sparse_mla import sparse_attn_tilelang
 from miles_plugins.models.deepseek_v4.ops.qat import fp8_simulate_qat
-from miles_plugins.models.deepseek_v4.ops.rope import apply_rotary_emb, ensure_freqs_cis, wrapped_precompute_freqs_cis
+from miles_plugins.models.deepseek_v4.ops.rope import apply_rotary_emb, wrapped_precompute_freqs_cis
 from miles_plugins.models.deepseek_v4.ops.v4_indexer import V4Indexer
 
 
@@ -182,13 +182,6 @@ class DeepSeekV4Attention(MegatronModule):
             else:
                 self.indexer = None
 
-        rope_base = config.dsv4_compress_rope_theta if self.compress_ratio else config.rotary_base
-        yarn_disabled = not self.compress_ratio
-        freqs_cis = wrapped_precompute_freqs_cis(
-            config, rope_head_dim=self.rope_head_dim, base=rope_base, yarn_disabled=yarn_disabled
-        )
-        self.register_buffer("freqs_cis", freqs_cis, persistent=False)
-
     def sharded_state_dict(
         self,
         prefix: str = "",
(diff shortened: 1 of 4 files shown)


kakisong:
cc @yueming-yuan — you authored the DeepSeek-V4 plugin (#1181); would appreciate your eyes here since it touches the rope table in `ops/rope.py`.

A bit more intuition on why this only surfaces at higher CP (slightly counter-intuitive — each CP rank holds *fewer* tokens, yet the high ranks are the ones that overflow):

The table is indexed by **global absolute position**, not per-rank token count. With contiguous CP, rank `r` holds the slice `[r*seqlen_local, (r+1)*seqlen_local)` of the global sequence, so the **last** rank holds the tail at the highest positions and needs entries up to `seqlen_global - 1`. Since `--max-tokens-per-gpu` is a per-rank budget, `seqlen_global = max_tokens_per_gpu * cp_size` grows with CP:

| config (max_tok=16384) | tokens/rank | seqlen_global | table = 65536 |
| --- | --- | --- | --- |
| CP2 | 16384 | 32768 | ok |
| CP4 | 16384 | 65536 | exactly fits |
| CP8 | 16384 | 131072 | overflow |

Concretely at CP8: every rank holds 16384 tokens, but rank 7 holds global positions 114688-131071 — the table only has rows 0-65535, so the tail ranks get a truncated/empty slice. So it is the absolute *positions* of the tail ranks, not the per-rank count, that exceed the table.

kakisong:
Thanks @gemini-code-assist — good catch, adopted in `cdecd78f4`. `self.config` is set by `MegatronModule.__init__(config=config)`, so the extra `self._config` was redundant; now using `self.config` directly, consistent with `DeepSeekV4Compressor` and `V4Indexer`. The change is behavior-equivalent (same config object), so the earlier 8xH20 / CP8 validation still holds.

yueming-yuan:
@kakisong Thanks for the PR! This is a great catch. I did some changes to this PR to make it slightly cleaner. Will merge it after CI passes.

yueming-yuan (approved):

PR #1390, opened 2026-06-22
Add a Clock abstraction with a fake clock for tests

fzyzcjy:
Add a small `Clock` interface (`RealClock` plus a controllable fake clock) so
time-dependent fault-tolerance code (health checks, heartbeats) can be driven
deterministically in tests.

- miles/utils/clock.py and tests.


Shi-Dong (approved):
LGTM!

yueming-yuan (approved):

PR #1395, opened 2026-06-22
Always reconnect rollout engines on weight-update setup

fzyzcjy:
Unconditionally disconnect-then-reconnect the model-update process group when
(re)connecting rollout engines, guarding the destroy against a missing group, so
a reconfigured/healed engine set can rebuild the NCCL group from scratch.

- broadcast.py: drop the "only disconnect if group exists" short-circuit; guard
  `destroy_process_group` against None.


Shi-Dong (commented):
I think the Gemini bot raised a valid concern.  Please address.

yueming-yuan (approved):

yueming-yuan:
> I think the Gemini bot raised a valid concern. Please address.

hmm I think the concern will not happen given the overall design?

PR #1396, opened 2026-06-22
Add a fault-injection RPC to train actors

fzyzcjy:
Expose an `inject_fault` Ray method on TrainRayActor (in its own concurrency
group) that triggers a configured failure mode via the fault injector, so
fault-tolerance tests can crash/hang specific actors on demand.

- train_actor.py: `inject_fault` RPC.


Shi-Dong (approved):
LGTM.

yueming-yuan (approved):

PR #1397, opened 2026-06-22
Delay splitting train data by DP until actor-side processing

fzyzcjy:
Extract the DP split into a witness-aware split_train_data_by_dp_raw
helper (with unit tests; the key list also carries seq_witness_ids) and
use it to split the training data on the actor side when
delay_split_train_data_by_dp is set, deferring the DP split from the
rollout side to actor-side processing. split_train_data_by_dp stays a
thin wrapper that ray.puts each partition.

- miles/ray/rollout/train_data_conversion.py (+ tests), miles/utils/data.py, actor_group.py, rollout_manager.py.

### DIFF
--- a/miles/utils/data.py
+++ b/miles/utils/data.py
@@ -8,6 +8,8 @@ import re
 import numpy as np
 import ray
 
+from miles.ray.rollout.train_data_conversion import split_train_data_by_dp_raw
+
 try:
     import pyarrow.parquet as pq
 except ImportError:
@@ -272,9 +274,19 @@ def get_minimum_num_micro_batch_size(total_lengths, max_tokens_per_gpu):
     return len(batches)
 
 
-def process_rollout_data(args, rollout_data_ref, dp_rank, dp_size):
-    assert len(rollout_data_ref) == dp_size
-    rollout_data = ray.get(rollout_data_ref[dp_rank].inner)
+def process_rollout_data(
+    args,
+    rollout_data_ref,
+    dp_rank,
+    dp_size,
+):
+    if args.delay_split_train_data_by_dp:
+        raw = ray.get(rollout_data_ref.inner)
+        raw = split_train_data_by_dp_raw(args, raw, dp_size=dp_size)
+        rollout_data = raw[dp_rank]
+    else:
+        assert len(rollout_data_ref) == dp_size
+        rollout_data = ray.get(rollout_data_ref[dp_rank].inner)
 
     partition = rollout_data.pop("partition")
     total_lengths = rollout_data["total_lengths"]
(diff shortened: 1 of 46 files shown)


miles/utils/data_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if args.balance_data:
+        partitions = get_seqlen_balanced_partitions(total_lengths, dp_size, equal_size=True)
+    else:
+        partitions = [range(i, len(total_lengths), dp_size) for i in range(dp_size)]
+
+    ans = []
+
+    for i in range(dp_size):
+        rollout_data = {}
```
Shi-Dong: Hmm I could be wrong but it seems that the previous `rollout_data["prompt"] = data["prompt"]` is shadowed by this line?

miles/utils/data.py
```diff
@@ -272,9 +274,19 @@ def get_minimum_num_micro_batch_size(total_lengths, max_tokens_per_gpu):
     return len(batches)
 
 
-def process_rollout_data(args, rollout_data_ref, dp_rank, dp_size):
-    assert len(rollout_data_ref) == dp_size
-    rollout_data = ray.get(rollout_data_ref[dp_rank].inner)
+def process_rollout_data(
+    args,
+    rollout_data_ref,
```
Shi-Dong, line 279: I think it'd be better to add type annotation for `rollout_data_ref`.

miles/utils/data.py
```diff
@@ -8,6 +8,8 @@
 import numpy as np
 import ray
 
+from .data_utils import split_train_data_by_dp
```
Shi-Dong: Let's avoid relative import whenever possible.

Shi-Dong (commented):
A few nits.

miles/utils/data_utils.py
```diff
@@ -0,0 +1,57 @@
+from typing import Any
+
+from miles.utils.seqlen_balancing import get_seqlen_balanced_partitions
+
+
+def split_train_data_by_dp(args, data: dict[str, Any], *, dp_size: int) -> list[dict[str, Any]]:
```
yueming-yuan: this seems to be duplicated with the original `split_train_data_by_dp`? should we unify

miles/utils/data_utils.py
```diff
@@ -0,0 +1,57 @@
...
+            "tokens",
+            "multimodal_train_inputs",
+            "response_lengths",
+            "rewards",
+            "truncated",
+            "loss_masks",
+            "round_number",
+            "sample_indices",
+            "rollout_log_probs",
+            "rollout_routed_experts",
```
yueming-yuan: `rollout_indexer_topk` is missing

PR #1402, opened 2026-06-22
Add event-log snapshot and restore checkpointing

fzyzcjy:
Add snapshot/restore for the structured event log so the event history survives
cell restarts during fault-tolerant training.

- miles/utils/event_logger/checkpoint.py and tests.


Shi-Dong (approved):
LGTM!

yueming-yuan (approved):

PR #1406, opened 2026-06-22
Add event-log checksum-consistency analysis rules

fzyzcjy:
Add the first event-analyzer rules that replay the structured event log and flag
weight-checksum inconsistencies: a `checksum_compare` helper (flatten nested
dicts, diff flat checksum maps) plus two rules built on it — cross-replica weight
checksum consistency and inference-engine weight checksum consistency — with unit
tests.

- miles/utils/event_analyzer/rules/{checksum_compare,cross_replica_weight_checksum,inference_engine_weight_checksum_consistency}.py and tests.


Shi-Dong (approved):
LGTM!

miles/utils/event_analyzer/rules/cross_replica_weight_checksum.py
yueming-yuan, line 1: qq: Does this check always work in non-deterministic mode? e.g. are all optimizers deterministic, if not the final weight will be different?
(i don't know, just curious...)

if no and this kind checker is only applied to deterministic mode, should we add explicit assert

fzyzcjy: my naive thought is that it should always apply, b/c otherwise diff ranks are training diff models

PR #1414, opened 2026-06-22
Add an in-memory (non-persistent) checkpoint manager

fzyzcjy:
Add an InMemoryCheckpointManager that keeps checkpoints in memory and thread a
`checkpointing_context` / `non_persistent_ckpt` through the model save path so a
non-persistent checkpoint can be requested.

- miles/backends/megatron_utils/in_memory_checkpoint.py, model.py and tests.


Shi-Dong (approved):
LGTM!

yueming-yuan (approved):

PR #1416, opened 2026-06-22
Extract actor construction into a shared allocate_gpus_for_actor factory

fzyzcjy:
Move the inline env-var / backend-selection / ray.remote actor construction out of
RayTrainGroup into a module-level allocate_gpus_for_actor factory, with no behaviour
change. (Pure mechanical extraction; FT context is threaded in a follow-up.)

- miles/ray/train/actor_factory.py, miles/ray/actor_group.py.


Shi-Dong (approved):
LGTM!

yueming-yuan (approved):

PR #1417, opened 2026-06-22
Thread independent-DP / role / cell-index / rollout-manager context through the train group

fzyzcjy:
Add the FT context to the actor factory and train actors: indep_dp_store_addr / role /
cell_index params and per-process-identity logging in TrainRayActor; pass a rollout_manager
into RayTrainGroup (set_rollout_manager, pause rollout health monitors before weight
updates); freeze the v1 group and wire context through placement_group / entrypoints.


Shi-Dong (approved):
LGTM

yueming-yuan (approved):

PR #1424, opened 2026-06-22
Make the tensor dumper fault-tolerance aware

fzyzcjy:
Scope tensor dumps per rollout and make them safe under independent-DP cells:
dumps go into per-rollout subdirectories, only effective-DP rank 0 writes output
files (other ranks still join dumper collectives), weights/grads are dumped once
per rollout pinned to step 0 with a distributed-optimizer grad all-gather, and
dump-dir cleanup is resilient to crashed peers and barriers across cells.

- dumper_utils.py: per-rollout exp_name, rank-gated output, `_build_full_grad_getter`,
  resilient `_cleanup_dump_dir` + `_barrier_after_dump_dir_cleanup`.
- model.py / actor.py: thread `rollout_id` into `forward_only` / `compute_log_prob`
  and the DumperMegatronUtil construction.


miles/utils/dumper_utils.py
```diff
@@ -111,25 +127,96 @@ def _extract_model(model: Sequence[torch.nn.Module]) -> torch.nn.Module:
...
+        phase: DumperPhase,
+        rollout_id: int,
+        overrides: dict[str, Any] | None = None,
+    ) -> bool:
         if overrides is None:
             overrides = _get_phase_override_configs(args, phase)
         if not overrides.get("enable"):
             return False
 
+        exp_name = f"{phase.value}/rollout_{rollout_id}"
```
yueming-yuan: maybe add `store_prefix` (actor/critic/teacher...) here to avoid conflict

yueming-yuan (approved):

PR #1436, opened 2026-06-22
Tolerate per-cell failures by marking dead cells errored and skipping them

fzyzcjy:
When an independent-DP cell's actors raise during a broadcast, isolate that
cell instead of failing the whole group: the execute layer now gathers with
return_exceptions, marks the failing cell as errored, and filters subsequent
broadcasts to the alive cells. `train` checks that at least one surviving cell
produced valid gradients and otherwise raises to signal a retry.

- group.py: `_execute_all_alive_and_catch` (catch + log + alive filter) and
  `_execute_first_alive`; `train` wraps the dispatch in a per-attempt closure
  and calls `_check_train_one_attempt` / `_compute_attempt_outcomes` to bucket
  cells into errored / normal.
- cell.py: `_mark_as_errored`, `is_errored`, and `execute(mark_errored_on_failure=)`
  so a failing dispatch transitions the cell into the errored state.
- cell_state.py: `StateAllocatedErrored`.
- tests: per-cell error isolation, errored-cell skipping, first-alive fallback,
  and `_check_train_one_attempt` outcome bucketing.


yueming-yuan (approved):

miles/ray/train/group.py
```diff
@@ -172,20 +196,23 @@ async def update_weights(self, rollout_id: int | None = None):
         # ranks observe a consistent engine set; the actor releases the lock itself.
         info = await self._rollout_manager.get_updatable_engines_and_lock.remote()
         await self._rollout_manager.health_monitoring_pause.remote()
-        await self._execute_first("update_weights", info=info)
+        await self._execute_first_alive("update_weights", info=info)
```
yueming-yuan, line 199: here we get the lock first and then `_execute_first_alive`, will this cause a deadlock if all cells are dead?

fzyzcjy: If all cells are dead, then we have no way to recover in current codebase (given it transfer weights from one dp group into another dp group). but yes in the future if we somehow want to handle this case we should be more careful

PR #1437, opened 2026-06-22
Retry failed training, save, and weight-update attempts

fzyzcjy:
Wraps RayTrainGroup.train, save_model, and update_weights in retry(...): when an attempt raises (a cell died mid-step, or _check_train_one_attempt signals a should-discard / all-failed step), the whole step is retried, picking up the now-reduced alive cell set (failed cells auto-marked errored). Passes the attempt index through so witness ids are reallocated per attempt. Includes the train-retry and first-alive-fallback tests. The retry helper lives in retry_utils, added separately.


yueming-yuan:
Should we add a `max_attempts` limit to the retry?

PR #1446, opened 2026-06-22
Inject witness ids into the Megatron forward and train step

fzyzcjy:
Feeds witness ids through the Megatron forward path so per-sample witness tracing can run: includes witness_ids in the get_batch key lists, splats them into the model forward (forward_only and train_one_step), guards combined-1f1b against witness, and dumps/clears stale witness state after a NORMAL train step via witness_dump_and_clear_stale. The witness_info parameter threading and the witness allocator/module are added with their own features.


yueming-yuan:
qq: will you add the witness operator in `radixark/Megatron-LM`'s model forward?

fzyzcjy:
yes, need a PR there

PR #1449, opened 2026-06-22
Add CI rollout-data injection with recorded-data metadata round-trip

fzyzcjy:
Adds RolloutDataInjectionUtil (debug_data.py) so CI can replace generated rollout data with a recorded baseline at/after a start rollout id, asserting the freshly generated responses still match the recording above a configurable token-match ratio (catching weight divergence beyond ulp drift). Threads recording metadata through save/load (save_debug_rollout_data metadata= / load returns (data, metadata) / _load_rollout_data_file) and wires injection into RolloutManager._get_rollout_data. Includes the debug_data unit tests for the metadata round-trip and the injection match logic.


yueming-yuan:
qq, why can we do this check? If we have a ULP-level drift at the beginning of the trajectory, will it easily break?

fzyzcjy:
not remember details clearly, but iirc it works in the tests, and I guess it is b/c on average it matches a lot (but anyway this is not a strong test but more like a sanity check)

PR #1464, opened 2026-06-22
Enable observe training entropy without computing entropy loss

zyzshishui:
## Motivation
Algorithm bros want to observe entropy from the training forward pass even when the entropy is not included in calculating loss. The previous `--entropy-coef 0.0` path skipped training entropy entirely, leaving only rollout-time entropy available for dashboards.

## Usage
```
--entropy-coef 0.0 --observe-training-entropy
```
This logs entropy_loss from the training logits as a detached metric. It does not add the entropy to the loss and does not attach the observed entropy to backward.

### DIFF
--- a/miles/backends/training_utils/loss_hub/math_utils.py
+++ b/miles/backends/training_utils/loss_hub/math_utils.py
@@ -859,6 +859,7 @@ def calculate_log_probs_and_entropy(
     tokens,
     tp_group,
     with_entropy: bool = False,
+    entropy_requires_grad: bool = True,
     chunk_size: int = -1,
     true_on_policy: bool = False,
     vocab_size: int | None = None,
@@ -869,6 +870,7 @@ def calculate_log_probs_and_entropy(
             tokens,
             tp_group,
             with_entropy=with_entropy,
+            entropy_requires_grad=entropy_requires_grad,
             vocab_size=vocab_size,
         )
 
@@ -877,6 +879,13 @@ def calculate_log_probs_and_entropy(
     # Without the clone, the backward will trigger inplace edit error.
     # It seems that the function with tp will modify the logits inplace.
     entropy = None
+
+    def compute_entropy(logits_chunk: torch.Tensor) -> torch.Tensor:
+        if entropy_requires_grad:
+            return compute_entropy_from_logits(logits_chunk.clone(), tp_group)
+        with torch.no_grad():
+            return compute_entropy_from_logits(logits_chunk.detach().clone(), tp_group)
+
     if logits.size(0) != 0:
         if chunk_size > 0:
             num_chunks = (logits.size(0) - 1) // chunk_size + 1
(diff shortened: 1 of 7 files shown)


guapisolo (approved):
LGTM. need code owner review.

guapisolo:
CC said the entropy calc definitely introduce some overhead.. so it should be never calculated under `--entropy-coef 0.00`, and can be re-opened by the new `observe-training-entropy` flag.

@zyzshishui Do you think algorithm people happy to workaround with `--entropy-coef 1e-15` (extremely tiny won't degrade training, but can still observe the entropy metrics?)

zyzshishui:
> CC said the entropy calc definitely introduce some overhead.. so it should be never calculated under `--entropy-coef 0.00`, and can be re-opened by the new `observe-training-entropy` flag.
> 
> @zyzshishui Do you think algorithm people happy to workaround with `--entropy-coef 1e-15` (extremely tiny won't degrade training, but can still observe the entropy metrics?)

I dont think they will be happy...

guapisolo:
> > CC said the entropy calc definitely introduce some overhead.. so it should be never calculated under `--entropy-coef 0.00`, and can be re-opened by the new `observe-training-entropy` flag.
> > @zyzshishui Do you think algorithm people happy to workaround with `--entropy-coef 1e-15` (extremely tiny won't degrade training, but can still observe the entropy metrics?)
> 
> I dont think they will be happy...

ok fine.

miles/backends/training_utils/loss_hub/math_utils.py
```diff
@@ -941,7 +953,13 @@ def _calculate_log_probs_and_entropy_true_on_policy(
 
     entropy = None
     if with_entropy:
-        probs = log_probs_full.exp()
-        entropy = -(probs * log_probs_full).sum(dim=-1)
+        entropy_log_probs = log_probs_full if entropy_requires_grad else log_probs_full.detach()
```
yueming-yuan, line 956: qq: curious why don't we directly replace `with_entropy` with `entropy_requires_grad`? like
```
if entropy_requires_grad:
        probs = log_probs_full.exp()
        entropy = -(probs * log_probs_full).sum(dim=-1)
```

zyzshishui: `with_entropy` decides whether entropy is computed, and `entropy_requires_grad` decides whether that entropy participates in autograd. Our algo bro wants observe-only case ——`with_entropy=True` and `entropy_requires_grad=False`

zyzshishui: Adding entropy into computing graph is memory consuming, and leads to OOM which could be avoided by this patch

yueming-yuan: oh sorry i was sb...
just one more question: it seems if we already detach the `entropy_log_probs`, we don't need the  `with torch.no_grad():` branch? 
in the way below i think it's also correct (and shorter)
```
if with_entropy:
        entropy_log_probs = log_probs_full if entropy_requires_grad else log_probs_full.detach()
        probs = entropy_log_probs.exp()
        entropy = -(probs * entropy_log_probs).sum(dim=-1)
```

zyzshishui: I was sb... fixed

yueming-yuan (approved):

PR #1469, opened 2026-06-23
[fsdp] fsdp backend fix

Zhichenzzz:


miles/backends/experimental/fsdp_utils/adaptations/precision.py
```diff
@@ -0,0 +1,47 @@
...
+from dataclasses import dataclass
+
+import torch
+
+
+@dataclass
+class PrecisionPolicy:
+    param_dtype: torch.dtype  # FSDP MixedPrecisionPolicy compute dtype
+    reduce_dtype: torch.dtype  # gradient all-reduce dtype
+    keep_fp32_master: bool = False  # keep an fp32 master copy; downcast to on-disk dtype at weight sync
```
guapisolo: `keep_fp32_master` should be True as default?

miles/backends/experimental/fsdp_utils/adaptations/precision.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# model_types (substring-matched) whose FSDP2 bf16 reshard needs an fp32 master; registered in specs.
+_FP32_MASTER_TYPES: set[str] = set()
```
guapisolo: I don’t think the model dtype and FP32 master-weight setting should be hardcoded together. The master-weight dtype does not affect forward accuracy. 

CI and partial-model tests may only care about forward accuracy, while often operating under tight memory constraints. Hardcoding this relationship unnecessarily increases their memory usage.

I suggest adding `--enable-fp32-master` in `arguments.py`, and True as default.

miles/backends/experimental/fsdp_utils/adaptations/class_patches.py
```diff
@@ -0,0 +1,135 @@
...
+
+import inspect
+import logging
+import textwrap
+
+logger = logging.getLogger(__name__)
+
+
+def apply_flash_attn_saux_guard() -> bool:
+    """Guard ``s_aux`` (attention sink) against None in transformers 5.6.0 flash_attention_forward, which
```
guapisolo: My cc said Transformer 5.8.1 has already fix this. And current sglang 0.5.15 already has transformers 5.12.1. Seems we can remove it.

guapisolo:
The "require_flash_for_attention_sink()" function in #1529 does not exist in this PR. need double check.

miles/backends/experimental/fsdp_utils/adaptations/weight_bridge.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+# qwen3_moe: split transformers>=5.6 batched experts into the per-expert names SGLang expects.
+def _qwen3_moe_matches(name: str, param) -> bool:
+    return getattr(param, "dim", lambda: 0)() == 3 and (
+        name.endswith(".experts.gate_up_proj") or name.endswith(".experts.down_proj")
+    )
+
+
+def _qwen3_moe_expand(name: str, full: torch.Tensor) -> Iterable[tuple[str, torch.Tensor]]:
```
guapisolo: This hand-written split duplicates the conversion contract transformers itself owns (row order, dim-1 halving, per-expert naming are all "HF happens to do it this way today" hardcoded) — if the upstream fusion layout ever shifts, this silently streams wrong weights to rollout.

Now that the stack is on transformers 5.12.1, the revert machinery works at single-tensor granularity, so this can call `revert_weight_conversion` (the exact reverse path `save_pretrained` uses) instead — the streamed names/values match the on-disk dialect SGLang loads by construction.

Fix in **#1734** (commit 41c2040e7c7b83cff819275a982246b701bd9d57, based on this branch): verified byte-identical to this slicing on tiny qwen3_moe + glm4_moe_lite, with dialect-pin tests that turn CI red on any upstream conversion-family drift.

guapisolo (changes requested):
Requesting one change on the WeightBridge expert split; fix is ready in #1734 (based on this branch).

guapisolo:
Stacked on this PR: #1737 adds a warn-only `VERIFIED_MODEL_TYPES` hook in `class_patches.py` — model_types not yet validated on the FSDP backend now get an explicit startup warning instead of loading silently through the generic HF path. Complements the DSA train/infer divergence warn hook here.

miles/backends/experimental/fsdp_utils/adaptations/precision.py
Zhichenzzz: agree

miles/backends/experimental/fsdp_utils/adaptations/precision.py
Zhichenzzz: yeah this should be unified with `keep_fp32_master` as the same argue

miles/backends/experimental/fsdp_utils/adaptations/class_patches.py
Zhichenzzz: could you retest it? at that moment, seems the version is wrong

Zhichenzzz:
> The "require_flash_for_attention_sink()" function in #1529 does not exist in this PR. need double check.

i remembered the gpt-oss model needs changing the flashattn backend to ensure the correctness of swa,

miles/backends/experimental/fsdp_utils/adaptations/weight_bridge.py
Zhichenzzz: thanks, agreed!

miles/backends/experimental/fsdp_utils/adaptations/class_patches.py
guapisolo: Ok, let me do it

guapisolo: @Zhichenzzz Retested with the current `transformers==5.12.1`: `flash_attention_forward` already guards `s_aux`, and the local compatibility hook is a no-op. I removed the obsolete patch and its registry-test expectations in `b61c4db78100dee318367604b3bbc983d24f1b01`; the revert is in https://github.com/radixark/miles/pull/1765.

miles/backends/experimental/fsdp_utils/models/qwen3_moe_hf.py
```diff
@@ -1,11 +1,21 @@
...
 import torch.nn.functional as F
 
+logger = logging.getLogger(__name__)
+
 
 def apply_fsdp_moe_patch():
 
     from transformers.models.qwen3_moe import modeling_qwen3_moe
 
+    # transformers >= 5.6 batches experts (Qwen3MoeExperts/TopKRouter): the legacy graph-forcing patch
```
guapisolo: Does 5.12.1 need this? Maybe we can remove this.

miles/backends/experimental/fsdp_utils/adaptations/__init__.py
guapisolo, line 1: why split specs and models, we can merge them to make it simpler

miles/backends/experimental/fsdp_utils/adaptations/post_load_fixups.py
```diff
@@ -0,0 +1,96 @@
...
+    """Run every registered fixup whose arch-predicate matches; return the names that fired."""
+    fired = []
+    for f in _FIXUPS:
+        if f.applies_to(hf_config) and f.apply(model, ckpt_path):
+            fired.append(f.name)
+    return fired
+
+
+# NemotronH / Mamba2-hybrid: _init_weights re-inits mixer params after loading; re-assert disk values.
+def _is_mamba_hybrid(hf_config) -> bool:
```
guapisolo: This is model-specific.

miles/backends/experimental/fsdp_utils/adaptations/weight_bridge.py
```diff
@@ -0,0 +1,50 @@
...
+def get_param_transform(name: str, param, model_type: str):
+    """Return the ``expand`` fn for the transform matching this param, or None (passthrough)."""
+    for transform in _REGISTRY.get(model_type, ()):
+        if transform.matches(name, param):
+            return transform.expand
+    return None
+
+
+# batched-expert archs (qwen3_moe, glm4_moe_lite, ...): unfuse into the per-expert names SGLang expects.
+def _batched_experts_matches(name: str, param) -> bool:
```
guapisolo, line 39: Is the unfold behavior in transformers only happens for moe? If so we can leave as-is, otherwise we should  use other abstraction.

guapisolo:
Also, we should re-enable fsdp tests in this PR.

miles/backends/experimental/fsdp_utils/adaptations/__init__.py
guapisolo: After checking the import path, I think keeping them split is useful. `specs/` contains lightweight declarations/registrations and is imported eagerly, while `models/` contains concrete implementations that are loaded only when the corresponding hook applies. In particular, `models/qwen3_moe.py` pulls in Triton/SGLang fused-MoE dependencies that are not available in every build, so merging it into the eagerly imported spec would make unrelated FSDP models load those dependencies. The naming/layout could be clearer, but I would keep this declaration/implementation boundary.

miles/backends/experimental/fsdp_utils/actor.py
```diff
@@ -111,11 +105,20 @@ def init(
                 self.tokenizer = load_tokenizer(
                     self.args.hf_checkpoint, chat_template_path=self.args.chat_template_path, trust_remote_code=True
                 )
-                # Vision models have `vision_config` in the config
                 if hasattr(self.hf_config, "vision_config"):
                     self.processor = load_processor(self.args.hf_checkpoint, trust_remote_code=True)
             dist.barrier(group=get_gloo_group())
 
+        # FSDP trains stock HF modeling: HF-compat patches + config-lifetime packing, before construction.
```
guapisolo, line 118: Here we shouldn't lazy import?

miles/backends/experimental/fsdp_utils/actor.py
```diff
@@ -125,12 +128,31 @@ def init(
                 attn_implementation=self.args.attn_implementation,
             )
 
+        from .adaptations.precision import apply_fp32_master, resolve_precision_policy
```
guapisolo: Also here is a lazy import

miles/backends/training_utils/loss_hub/math_utils.py
```diff
@@ -900,9 +900,9 @@ def calculate_log_probs_and_entropy(
 
     def compute_entropy(logits_chunk: torch.Tensor) -> torch.Tensor:
         if entropy_requires_grad:
-            return compute_entropy_from_logits(logits_chunk.clone(), tp_group)
+            return compute_entropy_from_logits(logits_chunk.to(torch.float32, copy=True), tp_group)
```
guapisolo, line 902: why copy here?

miles/backends/experimental/fsdp_utils/models/nemotron_h.py
```diff
@@ -0,0 +1,161 @@
...
+        b, q, _ = hidden_states.size()
+        Q = self.q_proj(hidden_states).view(b, q, self.num_heads, self.head_dim).transpose(1, 2)
+        K = self.k_proj(hidden_states).view(b, q, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+        V = self.v_proj(hidden_states).view(b, q, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+        kk = repeat_kv(K, self.num_key_value_groups)
+        vv = repeat_kv(V, self.num_key_value_groups)
+        qf = Q.transpose(1, 2).reshape(q, self.num_heads, self.head_dim).contiguous()
+        kf = kk.transpose(1, 2).reshape(q, self.num_heads, self.head_dim).contiguous()
+        vf = vv.transpose(1, 2).reshape(q, self.num_heads, self.head_dim).contiguous()
+        ml = int((cu[1:] - cu[:-1]).max())
```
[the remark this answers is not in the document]
guapisolo: Fixed in commit 90eccc1ae. `_patch_attn_forward` now reuses the `max_seqlen` already computed for the packed sequence instead of reducing `cu_seqlens` in every attention layer. I also added a regression test whose `cu_seqlens` cannot be indexed, which proves this path no longer recomputes the maximum; the full HF-compat test file passes (13 tests).

miles/backends/experimental/fsdp_utils/models/nemotron_h.py
```diff
@@ -0,0 +1,161 @@
...
+        except TypeError:
+            position_ids = kwargs.get("position_ids")
+        ctx = packed_seq_context(position_ids)
+        cu = ctx.cu_seqlens if ctx is not None else None
+        si = ctx.seq_idx if ctx is not None else None
+        for mod in self.modules():
+            if isinstance(mod, mixer_cls):
+                mod._packing_seq_idx = si
+            elif attn_cls is not None and isinstance(mod, attn_cls):
+                mod._packing_cu_seqlens = cu
```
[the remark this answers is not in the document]
guapisolo: Fixed in commit 90eccc1ae. `_patch_causallm_forward` now stashes `ctx.max_seqlen` on each attention module alongside `cu_seqlens`, so every layer receives the one precomputed value. I intentionally omitted the fallback: both attributes have the same owner and are written together, so a missing max should expose an internal contract violation rather than silently restore the synchronization path. The regression test and all 13 HF-compat tests pass.

miles/backends/experimental/fsdp_utils/adaptations/post_load_fixups.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    try:
+        from safetensors import safe_open
+    except Exception:  # pragma: no cover
+        return 0
+    files = sorted(glob.glob(os.path.join(ckpt_path, "*.safetensors")))
+    if not files:
+        return 0
+    index = os.path.join(ckpt_path, "model.safetensors.index.json")
+    shard_of = json.load(open(index))["weight_map"] if os.path.exists(index) else {}
```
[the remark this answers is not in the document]
guapisolo: Fixed in commit a40f7d7a8. The optional safetensors index is now opened inside a context manager, giving the post-load fixup a deterministic file-handle lifetime while preserving the existing JSON and I/O error behavior. The focused Mamba gating/registry tests pass (2 tests), and the commit hooks pass.

miles/backends/experimental/fsdp_utils/models/qwen3_moe_hf.py
guapisolo: No. The repo is now pinned to `transformers==5.12.1`, where `Qwen3MoeExperts`/`Qwen3MoeTopKRouter` is always present, so this legacy forward patch could only hit its early-return path. I removed it in `55ad1383d` while keeping the lazy true-on-policy patch. I also removed the stale test and rebased the ROCm patch so it no longer references `qwen3_moe_hf`. The HF-compat tests pass (`12 passed`), and the ROCm patch applies cleanly in a disposable checkout.

miles/backends/experimental/fsdp_utils/adaptations/post_load_fixups.py
guapisolo: Good catch. Addressed locally: `post_load_fixups.py` now keeps only the generic registry/dispatcher. The exact `nemotron_h` predicate and registration live in `specs/nemotron_h.py`, while the concrete reload lives in `models/nemotron_h.py` and is limited to the 42 proven `mixer.dt_bias` / `mixer.out_proj.weight` parameters. Verified with the full HF-compat test file (12 passed) and a real Nano-4B load (263/263 parameters match disk after the fixup). The changes have not been committed or pushed yet.

guapisolo: Agreed. I moved the model-specific predicate to `adaptations/specs/nemotron_h.py` and the checkpoint reload implementation to `models/nemotron_h.py` in `891a6860c`. `post_load_fixups.py` now only contains the generic registry and dispatcher. The spec uses a lazy wrapper, so importing specs does not eagerly load the model implementation. The existing registration name, gating, threshold, and reload behavior are preserved. The focused tests pass (`14 passed`).

miles/backends/experimental/fsdp_utils/adaptations/weight_bridge.py
guapisolo: No—the Transformers conversion framework itself is not MoE-only. In 5.12.1, `revert_weight_conversion()` reverses arbitrary `WeightConverter` mappings, including dense MLP and attention conversions.

However, `ParamTransform` is already the generic abstraction here: each model type opts in only when its HF training layout differs from the SGLang rollout-loader contract. The current helper is specifically the batched-expert unfuse needed by `qwen3_moe` and `glm4_moe_lite`; `qwen3_5_moe` intentionally passes through because its loader consumes the batched layout directly.

So I would leave the code as-is. A future non-MoE mismatch can register a different transform without changing the sync loop.

miles/backends/experimental/fsdp_utils/adaptations/post_load_fixups.py
guapisolo: Follow-up: after validating against the real Nano-4B checkpoint, I tightened the local follow-up beyond `891a6860c`. The post-load registration now matches only exact `model_type == "nemotron_h"`, and the model implementation only reloads the 42 proven `mixer.dt_bias` / `mixer.out_proj.weight` parameters; generic Mamba/hybrid configs and unrelated differing weights no longer match. The full HF-compat file passes (13 tests), and the real Nano-4B check is 263/263 equal to disk after the fixup with zero loading/final mismatches. This follow-up is local and has not been committed or pushed yet.

miles/backends/training_utils/loss_hub/math_utils.py
guapisolo: `copy=True` preserves the previous `clone()` isolation while upcasting bf16 chunks. TP cross-entropy mutates its input in forward, and `_VocabParallelEntropy.backward()` temporarily mutates its saved logits. A plain `.to(torch.float32)` would alias already-fp32 inputs and can trigger an autograd inplace-modification error. I clarified this invariant in a5f6f383d.

miles/backends/training_utils/log_utils.py
```diff
@@ -222,7 +222,7 @@ def log_rollout_data(rollout_id: int, args: Namespace, rollout_data: RolloutBatc
             if "rollout/entropy" in reduced_log_dict:
                 assert 0 < reduced_log_dict["rollout/entropy"] < 0.7
 
-        if args.ci_test and args.true_on_policy_mode:
+        if args.ci_test and args.true_on_policy_mode and not args.ci_disable_logprobs_checker:
```
guapisolo, line 225: why change here?

guapisolo: This is the CI log-probability checker, not a logits computation. It compares the rollout engine recorded `rollout_log_probs` with the FSDP actor recomputed `log_probs` for the same tokens. The adjacent tolerance-based checks already honor `--ci-disable-logprobs-checker`, but this separate true-on-policy exact equality assertion did not, so the opt-out could still fail CI. Commit `5be6a4ad5` makes this exact assertion honor the same flag. It only skips the CI assertion when the opt-out is explicitly set; it does not change logits, loss, gradients, or training behavior.

docker/amd_patch/latest/miles.patch
```diff
@@ -1,39 +1,18 @@
 diff --git a/miles/backends/experimental/fsdp_utils/actor.py b/miles/backends/experimental/fsdp_utils/actor.py
-index 47c2540f9..3c5f50ee8 100644
+index 9d258e733..55c4647fd 100644
```
guapisolo, line 2: This part already reviewed by AMD folks.

miles/backends/training_utils/loss_hub/math_utils.py
guapisolo: why didn't we use the old clone logic?

guapisolo: `clone()` only makes a copy; it does not change the dtype. Previously the logits were already FP32, so that was enough. FSDP can now pass BF16 logits here, so we need both an FP32 conversion and a separate copy for the mutating downstream kernels. `to(torch.float32, copy=True)` does both.

guapisolo (approved):
Fixed details. LGTM now.

Zhichenzzz:
> Fixed details. LGTM now.

Thanks!! we will wait till all of the ci pass

yueming-yuan (approved):

PR #1472, opened 2026-06-24
Opt-in multi-process session server (thin router + IPC → headless workers)

guapisolo:
# Opt-in multi-process session server (thin router + IPC → headless workers)

## Motivation
The session server's dominant cost under multi-turn agent loops with R3 (`routed_experts`) is `json.loads`/validation of large (100+ MiB) response bodies, which runs under a single process's GIL. The previous bounded thread-pool offload could not raise aggregate parse throughput (GIL-bound) and worsened tail latency under concurrency (see `docs/developer/session-server-overhead.md`). This PR escapes the GIL by sharding sessions across OS processes.

## What changed
Adds an **opt-in** multi-process mode via `--session-server-workers N`:
- **`--session-server-workers=1` (default)** keeps today's single-process path (a thin FastAPI adapter over a transport-neutral `SessionCore`); fully backward compatible.
- **`>1`** starts a **thin router process** (the sole client-facing HTTP listener) that shards each request by `blake2b(session_id) % N` over a **framed UNIX-socket IPC channel** (not a second HTTP hop) to **N headless worker processes**. Each worker runs the existing session logic (`SessionRegistry`/`LinearTrajectory`/in-flight gate) on its own asyncio loop and makes the upstream inference HTTP call itself. Sessions are standalone, so sticky-by-hash routing means IPC carries only the per-turn request/response, never accumulated state.

New modules under `miles/rollout/session/`: `session_core.py` (transport-neutral core), `session_ipc.py` (framed multiplexed RPC), `session_worker.py` (headless worker + lightweight `ProxyBackend`), `session_router.py` (thin router), `session_supervisor.py` (lifecycle), `routing.py` (stable hashing). Plus `router_manager.py`/`arguments.py` wiring and a decoupled `sessions.py` adapter. Baseline: branched from `9cf2a0384`.

## Results (production-shaped: 32 sessions × 50 turns × r3_scale=1000, mock backend isolating session-server overhead)
| | OLD (w1) | NEW (w16) | speedup |
|---|---|---|---|
| Wall time | 405.3s | 60.7s | **6.7×** |
| Throughput | 3.95 t/s | 26.4 t/s | **6.7×** |
| Reply p50 | 7763 ms | 750 ms | **10.3×** |
| Reply p95 | 16224 ms | 1819 ms | **8.9×** |
| Reply p99 | 18477 ms | 2262 ms | **8.2×** |
| Peak RSS | 43.6 GiB | 44.9 GiB | ~flat |

**w16 is the sweet spot** for 32 sessions; w32 over-subscribes (slower + heavier). Caveat: synthetic mock backend isolates session-server CPU/IPC overhead, not end-to-end model latency; the gain scales with concurrency × body size.

## One intended client-visible behavior change
`routed_experts`/`indexer_topk` (R3) are stripped from the **client-facing chat response** (uniformly, both modes) — the agent never reads them there; they're kept in the worker's stored record and served via `GET /sessions/{id}` for training-sample reconstruction.

## Testing
`tests/fast/router/` → **144 passing**, no orphan processes. Covers: the single-process behavioral suite (in-flight 409, closing-404>409, slot-release on all error paths, non-200 passthrough unrecorded, invalid-200→502, TITO invariants, R3 pruning); IPC framing/multiplexing/no-HOL/EOF; **workers=N equivalence** of the error/gate/passthrough paths that cross the spawn boundary; client-disconnect stability (no worker death / no fail-fast / no cross-session 503); router backpressure; stable-hash cross-process determinism. Plus a ported end-to-end overhead benchmark (`tests/benchmark/`).

The benchmark (`tests/benchmark/bench_session_server_overhead.py`) additionally gained, for realistic / higher-concurrency studies: `--inference-interval` (mock backend generation delay) and `--tool-interval` (per-session inter-turn idle) to model real rollout pacing instead of back-to-back; and a multi-process load generator (`--bench-driver-procs`, `_bench_load_generator.py`) that shards the client load across processes so a single asyncio event loop's connection-churn ceiling doesn't cap the achievable session count. Per-session driving is robust — failures are classified (server non-2xx vs client transport reset) and tallied rather than aborting the run.

## Design decisions
- Re-established the per-session in-flight gate (409) + strict upstream validation (502) inline (no thread pool).
- Worker death → whole session server fail-fast (surfaced to the rollout path via the supervisor's `.check()`); `PR_SET_PDEATHSIG` + process-group teardown, no orphans.
- Router never `json.loads` the response body; `asyncio.shield` so a client disconnect doesn't cancel the worker; per-worker 503 backpressure.

## Deferred (out of scope)
- Consistent-hashing / rebalanceable routing (v1 uses simple modulo over a stable hash).
- Delta/incremental-R3 protocol (needs SGLang/trainer coordination).

## Notes
Includes fixes for all findings from an adversarial Codex code review (fail-fast→rollout propagation, SIGTERM handling, disconnect backpressure, IPC round-robin/bounded-queue, all-choices R3 strip, off-loop GET serialization).


gongshaotian:
image
Do you have a detailed plan regarding this matter？

PR #1479, opened 2026-06-25
Remove deprecated --use-miles-router from recipe docs

Shi-Dong:
## Summary
The `--use-miles-router` flag is deprecated. This removes its remaining references from the model recipe pages, keeping the still-valid `--use-rollout-routing-replay` (R3) intact:

- `models/nemotron/nemotron-3-nano-moe.md` — dropped the flag from the `SGLANG_ARGS` block and reworded the prose so R3 alone is credited with keeping train/rollout logprobs aligned.
- `models/nemotron/nemotron-3-super.md` — same treatment.
- `models/kimi/kimi-k2.5.md` — dropped the "uses the miles router (`--use-miles-router`)" clause, kept the `--update-weight-buffer-size` detail.

## Notes / scope
- Docs-only. The flag is still defined in code (`miles/utils/arguments.py`) and several `scripts/run-*.sh` launchers still pass it; those are intentionally out of scope for this PR.
- The `/advanced/miles-router` page slug (titled "Rollout Routing Replay (R3)") and the developer/architecture descriptions of the router component are also left untouched here — separate follow-ups if desired.

## Test plan
- [ ] Mintlify preview: the three recipe pages render with R3 guidance intact and no dangling references.

### DIFF
--- a/docs/models/kimi/kimi-k2.5.md
+++ b/docs/models/kimi/kimi-k2.5.md
@@ -87,7 +87,7 @@ The launcher groups its flags into the arrays that are passed to `train.py`. The
 - **`GRPO_ARGS`** and **`OPTIMIZER_ARGS`** set the algorithm and CPU-offloaded Adam (§5.2, §5.4).
 - **`SGLANG_ARGS`** configures the colocated rollout engine (§5.3).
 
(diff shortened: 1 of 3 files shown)


guapisolo (approved):
Good job. LGTM.

PR #1483, opened 2026-06-25
fix(megatron): propagate training-infra args to bridge model config

guapisolo:
## Summary
Bridge mode silently drops args-derived training config (recompute, offload, fp8, ...)

## Symptom & Reproduction
- **Symptom:** under `--megatron-to-hf-mode bridge`, args-derived knobs (`recompute_granularity`, `cpu_offloading`, `fp8`, `attention_backend`) silently keep their dataclass defaults.
- **Symptom:** Qwen3.6-27B VLM at 32k OOMs on the train step (~136 GiB/GPU) at TP4 because activation checkpointing never runs.
- **Reproduction:** run a bridge-mode job with `--recompute-granularity full`; the bridge `transformer_block` reads `self.config.recompute_granularity` as `None`.

## Root Cause
1. `get_model_provider_func` bridge branch builds the provider via `bridge.to_megatron_provider`, bypassing `core_transformer_config_from_args`.
2. It then hand-wires only ~9 fields, so any other args-derived knob keeps its `TransformerConfig` dataclass default.
3. `Qwen3VLTransformerBlock` gates checkpointing on `recompute_granularity == "full"`, so the `None` default makes `_checkpointed_forward` never run.
4. Every other such knob (`cpu_offloading`, `fp8`, `deterministic_mode`) is the same latent gap, surfacing only when it breaks.

## Fix
Give the bridge provider one place that re-injects the args-owned runtime config it would otherwise lose: `_apply_bridge_runtime_config(provider, args)`, so new training-infra flags land here instead of being rediscovered one OOM at a time. It is a curated include-list, not a blind copy of args: in bridge mode the HF checkpoint and the per-model provider own architecture, dtype, and model-specific config, and the bridge weight mapping only warns and continues, so copying those back would silently corrupt the model rather than fail.

## Verification
- **End-to-end:** `recompute_granularity` (the live casualty) reaches `will_checkpoint=True` on Qwen3.6-27B VLM at 32k; the train step completes with no OOM.
- **Field presence:** every propagated field was confirmed present at runtime before assignment, not assumed.
- **Selection audit:** the propagated set was checked against the full args-to-`TransformerConfig` overlap to confirm no runtime/infra knob is missing.

## Review Focus
- Scrutinize `_apply_bridge_runtime_config` membership: every assignment must be a runtime/infra knob, never an architecture or provider-set field (e.g. `cp_comm_type`, `moe_grouped_gemm`, `apply_rope_fusion` stay provider-owned).
- Confirm `bf16`/`params_dtype` stay HF-derived and are not propagated.
- Verify the guard on `cpu_offloading`, the only conditionally-set field among the propagated knobs.

### DIFF
--- a/miles/backends/megatron_utils/model_provider.py
+++ b/miles/backends/megatron_utils/model_provider.py
@@ -23,41 +23,73 @@ from miles.utils.replay_base import routing_replay_manager
 logger = logging.getLogger(__name__)
 
 
-# Args-owned knobs to re-inject onto a bridge-built provider, which skips
-# core_transformer_config_from_args. NOT a blind args->provider copy -- architecture, dtype, and
-# provider-set fields stay HF-checkpoint-authoritative; overriding them would silently corrupt the model.
-_BRIDGE_RUNTIME_OVERRIDE_FIELDS = (
+def _apply_bridge_runtime_config(provider, args: argparse.Namespace) -> None:
+    """Re-inject args-owned runtime config onto a bridge-built provider.
+
+    Bridge mode builds the model from the HF checkpoint and skips
+    core_transformer_config_from_args, so CLI args never reach the provider. We
+    hand-pick fields here instead of blind-copying every arg: on this path the
+    provider is authoritative for what the HF checkpoint defines, and
+    architecture / dtype / provider-deliberate fields sit at mismatched defaults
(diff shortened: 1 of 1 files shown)


Zhichenzzz (approved):
LGTM!

PR #1492, opened 2026-06-27
fix(weights): handle empty colocated tensor buckets

EazyReal:
Twin of [THUDM/slime#2134](https://github.com/THUDM/slime/pull/2134), adapted to the current `radixark/miles` `main` and the shared Megatron/FSDP weight-sync paths.

## Problem

Colocated raw weight sync assumed every TP rank produced the same number of serialized dtype buckets. A rank with no tensors could either construct `FlattenedTensorBucket([])` before participating in the Gloo gather, or leave the gather source indexing a missing column when ranks had ragged bucket counts. Uneven PP/EP/MoE layouts can expose both cases.

## Fix

Empty local chunks now participate in `gather_object` with an empty list. After the gather, per-rank lists are aligned to the widest non-empty rank, and missing entries receive a lazily created, reusable zero-length serialized bucket. All-empty gathers remain no-ops. The alignment and placeholder construction live in one shared helper used by both Megatron and FSDP.

The SGLang receiver selects the serialized entry for its own TP rank and synchronizes once per update call. Padding therefore preserves equal call counts while giving ranks without tensors a valid no-op payload.

## Validation

- Changed-file pre-commit checks passed.
- Focused backend suite: 33 passed.
- Real CUDA IPC against the repository-pinned SGLang receiver: two spawned consumers each deserialized and reconstructed the same placeholder twice; every payload had zero elements and reconstructed to zero tensors.
- Full regular `stage-a-cpu` suite: 5,794 passed, 32 skipped.

GitHub-hosted exact-head workflows are awaiting maintainer approval; applying the `run-ci-weight-update` label will approve and run them.

### DIFF
--- a/miles/backends/fsdp_utils/update_weight_utils.py
+++ b/miles/backends/fsdp_utils/update_weight_utils.py
@@ -17,6 +17,10 @@ except ImportError:
 
 from sglang.srt.utils import MultiprocessingSerializer
 
+from miles.backends.training_utils.serialized_buckets import (
+    align_serialized_bucket_columns,
+    empty_flattened_tensor_data,
+)
 from miles.utils.distributed_utils import get_gloo_group, init_process_group
 
 try:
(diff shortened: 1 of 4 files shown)


EazyReal:
@fzyzcjy @yueming-yuan @maocheng23 @yushengsu-thu @Zhichenzzz Could I get a quick review on this empty colocated weight-bucket fix? It avoids failing a weight update when a rank owns no tensors.

EazyReal:
Rebased this onto current `radixark/miles` `main` and completed an exact-head review and verification pass on `df23a25e`.

The final regression coverage exercises all-empty gathers, short/ragged rank lists, lazy placeholder reuse, and both the Megatron and FSDP call sites. Results:

- changed-file pre-commit checks: passed
- focused backend suite: 33 passed
- real CUDA IPC against the repository-pinned SGLang receiver: two spawned consumers each deserialized and reconstructed the same zero-length `uint8` placeholder twice; all reconstructions produced zero tensors
- full regular `stage-a-cpu` suite: 5,794 passed, 32 skipped

I also checked the receiver-side contract: each TP rank selects its own serialized entry and synchronizes once per update call. Max-width alignment with no-op padding is therefore required to keep call counts equal, while all-empty gathers correctly remain no-ops.

The exact-head hosted workflows are held for approval rather than failing. A maintainer can apply `run-ci-weight-update` to approve and run them.

PR #1493, opened 2026-06-27
fix(ppo): preserve raw KL metric tensor

EazyReal:
Port of THUDM/slime#2114 for miles' refactored advantage helper. Recut onto current `main`.

Summary:
- build PPO token-level rewards out of place instead of mutating the `kl` tensor list
- preserve the raw approximate KL tensor used by rollout/kl logging
- add a CPU test directly on `compute_advantages`

Validation:
- test fails on the old `k *= kl_coef` in-place loop
- test passes after the out-of-place multiply

### DIFF
--- a/miles/backends/training_utils/loss_hub/advantages.py
+++ b/miles/backends/training_utils/loss_hub/advantages.py
@@ -60,9 +60,8 @@ def compute_advantages(
         terminal_rewards = rewards
         token_rewards = []
         kl_coef = -args.kl_coef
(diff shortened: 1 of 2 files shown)


EazyReal:
@fzyzcjy @yueming-yuan @maocheng23 @Zhichenzzz @Shi-Dong Could I get a quick review on this raw KL metric fix? It keeps rollout/KL logging raw while training can still use the shaped reward.

PR #1494, opened 2026-06-27
fix(train): support eval-only mode (--num-rollout 0)

EazyReal:
Port of THUDM/slime#2109 for miles.

Summary:
- compute estimated train iterations once
- clamp the scheduler-visible `train_iters` to at least 1 so Megatron LR decay steps stay valid
- keep the training loop semantics controlled by `args.num_rollout`; eval-only still performs zero training rollouts
- reject positive-rollout configs with zero training iterations instead of silently papering over an inconsistent train setup

Validation:
- `uv run --with pytest --with torch pytest --confcutdir=tests/fast/backends/megatron_utils tests/fast/backends/megatron_utils/test_eval_only_optimizer_scheduler.py -q` -> 3 passed

### DIFF
--- a/miles/backends/fsdp_utils/lr_scheduler.py
+++ b/miles/backends/fsdp_utils/lr_scheduler.py
@@ -9,6 +9,8 @@ import torch
 from torch.optim.lr_scheduler import LRScheduler
 from typing_extensions import override
 
+from miles.backends.training_utils.train_iters import compute_train_iters
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


EazyReal:
@fzyzcjy @yueming-yuan @maocheng23 @yushengsu-thu @Zhichenzzz Could I get a quick review on this eval-only fix? It lets `num_rollout=0` run evaluation without breaking Megatron scheduler setup.

EazyReal:
Closing in favor of #2685.

`train_iters = 1` was only a dummy so Megatron/FSDP scheduler constructors would accept eval-only. Eval-only does not train, so it should not build the optimizer or LR scheduler. #2685 skips that stack and still loads actor weights for `rollout_manager.eval`. Twin of THUDM/slime#2296.

PR #1505, opened 2026-06-29
[fix] DeepSeek-V4 blockwise FP8: keep linear_weights_proj to bf16

xiuhu17:
## Changes

**`scripts/run_deepseek_v4.py` — keep the DSA indexer `weights_proj` in BF16 on the trainer**
- Add `_DSV4_TE_PRECISION_CONFIG`, a TE per-module precision override that matches
  `*.self_attention.indexer.linear_weights_proj` and forces it to BF16.
- Inject it via `--te-precision-config-file` in the `fp8_training` branch (skipped if the
  user already passes their own `--te-precision-config-file` through `--extra-args`).
  Blockwise FP8 on `weights_proj` is numerically unstable, so the trainer keeps it BF16.

**`tools/convert_hf_to_fp8.py` — keep more weights BF16 in the rollout checkpoint**
- Extend the skip list so `head.`, `wo_a`, `ffn.gate.`, and `compressor.` are left in BF16
  instead of being cast to FP8, matching the trainer's BF16 set. (In run_deepseek_v4, we do not need convert_hf_to_fp8 since the checkpoint already in fp8 format. Keep the change to avoid future bug.)

**`miles_plugins/models/deepseek_v4/ops/qat.py` — fix FP8 simulate scale dtype**
- Pass `scale_dtype=torch.float32` to `act_quant`. The auto dtype picks `float8_e8m0fnu`
  on Blackwell+DeepGEMM, but the downstream `per_token_cast_back` only accepts int32/fp32
  scales. `scale_fmt="ue8m0"` still rounds scale values to powers of two, so numerics are
  unchanged.

## Notes
- Scope is **blockwise FP8 only**
- The `--te-precision-config-file` injection is guarded so an explicit user-provided
  config in `--extra-args` takes precedence.

Screenshot 2026-06-29 at 11 08 14 AM

### DIFF
--- a/miles_plugins/models/deepseek_v4/ops/qat.py
+++ b/miles_plugins/models/deepseek_v4/ops/qat.py
@@ -13,7 +13,9 @@ def fp8_simulate(x: torch.Tensor, block_size: int):
     the rest of the DeepSeek stack.
     """
     x_c = x.contiguous()
-    y, scale = act_quant(x_c, block_size, "ue8m0")
+    # Force fp32 scale storage: act_quant's auto dtype picks float8_e8m0fnu on
+    # Blackwell+DeepGEMM, but per_token_cast_back below only accepts int32/fp32 scales
+    y, scale = act_quant(x_c, block_size, "ue8m0", scale_dtype=torch.float32)
 
     N = x_c.size(-1)
     y_flat = y.view(-1, N)
--- a/miles_plugins/models/deepseek_v4/ops/qat.py
+++ b/miles_plugins/models/deepseek_v4/ops/qat.py
@@ -13,9 +13,7 @@ def fp8_simulate(x: torch.Tensor, block_size: int):
     the rest of the DeepSeek stack.
     """
     x_c = x.contiguous()
-    # Force fp32 scale storage: act_quant's auto dtype picks float8_e8m0fnu on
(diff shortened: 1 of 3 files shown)


miles_plugins/models/deepseek_v4/ops/qat.py
```diff
@@ -13,7 +13,9 @@ def fp8_simulate(x: torch.Tensor, block_size: int):
     the rest of the DeepSeek stack.
     """
     x_c = x.contiguous()
-    y, scale = act_quant(x_c, block_size, "ue8m0")
+    # Force fp32 scale storage: act_quant's auto dtype picks float8_e8m0fnu on
+    # Blackwell+DeepGEMM, but per_token_cast_back below only accepts int32/fp32 scales
+    y, scale = act_quant(x_c, block_size, "ue8m0", scale_dtype=torch.float32)
```
yueming-yuan: This has already been forced in today's code

yueming-yuan (approved):

miles_plugins/models/deepseek_v4/ops/qat.py
xiuhu17: Thanks for pointing out, fixed.

PR #1510, opened 2026-06-29
(2/7) refactor: extract session core with direct HTTP responses

guapisolo:
> Stacked on #1569 (the session_ prefix rename). Base = `refactor/session-pkg-rename`.

## Summary
Extract session route behavior into `SessionCore`. Pure extraction; behavior preserved. The R3 client-strip stacks on this as its own PR (#1563).

## Motivation
`sessions.py` mixed FastAPI request adaptation with session/TITO state transitions. This PR makes the route layer thin while keeping Starlette `Response` as the HTTP return contract.

## Before / After
| Area | Before | After |
|---|---|---|
| Route layer | `sessions.py` owned FastAPI adaptation and session flow. | `sessions.py` reads `Request` data and calls `SessionCore`. |
| Core layer | Session behavior lived in route handlers. | `SessionCore` owns session CRUD, chat/TITO state transitions, and proxy response rendering. |
| Backend proxy | `do_proxy` accepted FastAPI `Request`. | `do_proxy` accepts `ProxyRequest`. |

## Behavior Preservation
Functional behavior is preserved: HTTP status codes, parsed response content, request/response header filtering, the hardcoded upstream flags (`logprobs`/`return_meta_info`/`no_stop_trim`/conditional R3 flags/`input_ids`), the three chat lock phases, every upstream-response validation check and its order, the `num_assistant`-mismatch handling (warn + skip state update + return the upstream 200), the `SessionRecord` fields, and non-200 passthrough. Existing session route and race-condition tests pass.

Two intended **non-functional** changes:
1. **Body-read timing / lock scope.** The FastAPI route now reads `await request.body()` before entering `SessionCore.chat_completions` (previously the body was read inside `session.lock`, after the session lookup). JSON parsing, TITO mutation, and record writes remain under `session.lock`; backend proxying remains outside it. Consequences: a chat to a missing/closing session consumes the full request body before returning the error; in a DELETE-vs-chat race during body upload the chat is more likely to be rejected (404). The race outcome was already non-deterministic and the race-condition tests pass.
2. **Debug request-logging removed.** The per-request `debug_request_logger` middleware (the `[session-server] REQUEST ARRIVED/DONE …` INFO lines and the `_inflight_chat` counter) and the two DELETE lock `debug` lines are removed. Logging-only; no contract impact.

## Verification
- `python -m pytest tests/fast/router/test_sessions.py -q` → 7 passed.
- `python -m pytest tests/fast/router/test_session_race_conditions.py -q` → 9 passed.
- `python -m pytest tests/fast/router/` → 73 passed.
- `python -m black --check miles/rollout/session/core.py miles/rollout/session/sessions.py miles/rollout/session/server.py` passed.

## Review Focus
- `SessionCore.chat_completions`: TITO record updates and accepted body-read timing.
- `proxy_result_to_response`: backend response rendering.
- `SessionServer.do_proxy`: request forwarding with `ProxyRequest`.

### DIFF
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -185,7 +185,7 @@ class SessionCore:
             raise UpstreamResponseError(
                 "meta_info and output_token_logprobs must be in choice (requires logprobs=True)"
             )
-        assistant_message = choice.get("message", {})
+        assistant_message = choice.get("message") or {}
         if assistant_message.get("content") is None:
             raise UpstreamResponseError(
                 "assistant message content is None, when tool call parser failed SGLang should still return "
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -16,7 +16,12 @@ from dataclasses import dataclass
 
 from starlette.responses import Response
 
-from miles.rollout.session.errors import SessionNotFoundError, TokenizationError, UpstreamResponseError
+from miles.rollout.session.errors import (
+    MessageValidationError,
+    SessionNotFoundError,
+    TokenizationError,
+    UpstreamResponseError,
+)
 from miles.rollout.session.linear_trajectory import SessionRegistry
 from miles.rollout.session.types import GetSessionResponse, SessionRecord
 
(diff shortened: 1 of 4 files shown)


Shi-Dong (approved):
LGTM!

PR #1512, opened 2026-06-29
Megatron e2e: weight-check skip-list, Qwen3.5 MTP cases

guapisolo:
ci-sglang-pr: #27750

> Stacked on #1513 (CaseConfig refactor); merge that first.

## What

1. **`--check-weight-update-selector` / `--check-weight-update-skip-list`** for the post-update weight-equality check (previously all-or-nothing):
   - `selector`: `all` (target + draft/MTP model) or `target` (target model only; skips the draft, e.g. when MTP training is off).
   - `skip_list`: substrings matched against tensor names; matches become non-fatal.
   Threaded `train.py -> rollout_manager -> rollout_server -> server_group -> sglang_engine` into the SGLang `/weights_checker` request. Requires the SGLang-side change (see `ci-sglang-pr`).

2. **Qwen3.5-35B-A3B MTP + spec-v2 e2e cases** (`test_qwen3_5_35B_A3B_mtp/`, cp2/ep8 on 8 GPUs):
   - `test_mtp1_spec_v2_r3`: MTP on + R3 on; selector `all` (draft is synced, so it is checked).
   - `test_mtp0_spec_v2`: MTP off + R3 off; selector `target` (the un-synced draft is not checked).
   Both pass `check_weight_update_skip_list=("visual",)`: miles has no VLM/vision implementation on the training side, so Qwen3.5's vision weights are never synced.

## Open items
- Runtime (8×H100) verification of the qwen3.5 cases has not been run; on the first run, confirm `selector=target` + the `visual` skip let `mtp0` pass (extend the per-case skip list if other un-synced weights surface).

## Test
- `python -m py_compile` on all touched dirs.
- Weight-check skip/selector logic unit self-check; qwen3.5 `build_train_args` smoke check (selector / skip-list / mtp / r3 flags per case).


miles/utils/arguments.py
```diff
@@ -1532,6 +1532,23 @@ def add_debug_arguments(parser):
                 default="torch",
             )
             parser.add_argument("--check-weight-update-equal", action="store_true")
+            parser.add_argument(
+                "--check-weight-update-selector",
+                type=str,
+                default="all",
```
yueming-yuan, line 1538: nti: maybe directly add choices, ["target", "draft", "all"]

train.py
yueming-yuan, line 1: should we also update `train_async.py`?

yueming-yuan (approved):

train.py
guapisolo: Oh you're right, I forgot that..

PR #1525, opened 2026-06-30
[fsdp] Add PrecisionPolicy dataclass + fp32-master registry + resolver

Zhichenzzz:
Adds a PrecisionPolicy that resolves the FSDP mixed precision dtypes and whether a model keeps an fp32 master, plus a small registry for archs to opt into an fp32 master. Nothing consumes it yet; the actor wiring lands later in the chain.

### DIFF
--- a/miles/backends/experimental/fsdp_utils/adaptations/precision.py
+++ b/miles/backends/experimental/fsdp_utils/adaptations/precision.py
@@ -1,0 +1,28 @@
+"""Precision policy for the FSDP backend: resolve MixedPrecisionPolicy dtypes and whether to keep an fp32 master."""
+
+from dataclasses import dataclass
+
+import torch
+
+
+@dataclass
+class PrecisionPolicy:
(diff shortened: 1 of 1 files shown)


miles/backends/experimental/fsdp_utils/adaptations/precision.py
```diff
@@ -0,0 +1,28 @@
...
+from dataclasses import dataclass
+
+import torch
+
+
+@dataclass
+class PrecisionPolicy:
+    param_dtype: torch.dtype  # FSDP MixedPrecisionPolicy compute dtype
+    reduce_dtype: torch.dtype  # gradient all-reduce dtype
+    keep_fp32_master: bool = False  # keep an fp32 master copy; downcast to on-disk dtype at weight sync
```
guapisolo, line 12: Highly suggest we set `keep_fp32_master` to default True. To make sure the best accuracy under default mode.

PR #1526, opened 2026-06-30
[fsdp] Add apply_fp32_master in-place fp32 master conversion

Zhichenzzz:
Adds `apply_fp32_master`, which converts a model to a uniform fp32 master in place after recording each param on disk dtype so the weight sync can downcast correctly. Needed for mixed dtype checkpoints whose routing or mixer params are stored in fp32.


guapisolo (approved):
LGTM.

PR #1539, opened 2026-07-01
[fsdp] Skip legacy qwen3_moe MoE-block patch on transformers>=5.6

Zhichenzzz:
Makes the legacy qwen3_moe MoE block patch a no op on transformers 5.6 and later, where experts are already batched. This prevents double patching the new expert layout.

### DIFF
--- a/miles/backends/experimental/fsdp_utils/models/qwen3_moe_hf.py
+++ b/miles/backends/experimental/fsdp_utils/models/qwen3_moe_hf.py
@@ -1,10 +1,19 @@
+import logging
+
 import torch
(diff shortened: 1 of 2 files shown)


miles/backends/experimental/fsdp_utils/models/qwen3_moe_hf.py
yueming-yuan, line 1: maybe just remove this patch? (it is ugly) we can assert transformer version >= 56.

yueming-yuan (changes requested):

PR #1559, opened 2026-07-01
Add GLM-5/5.1/5.2 (744B MoE) LoRA RL support 

yushengsu-thu:
## Results glm5.2 lora RL training:

- miles:
  - https://github.com/radixark/miles/pull/1559
- SGLang (miles)
  - https://github.com/sgl-project/sglang/pull/29874
- megatron-bridge: 
  - https://github.com/radixark/Megatron-Bridge/pull/15
  which is split into:
1. https://github.com/radixark/Megatron-Bridge/pull/16
2. https://github.com/radixark/Megatron-Bridge/pull/17

Screenshot 2026-07-02 at 11 19 27 AM

Screenshot 2026-07-02 at 11 18 49 AM

To-do and limitation:

0. To reduce the review effort, I will submit another lora weight checker PR (since it cannot use currently base model weight checker and require adding APIs on sglang side.) 

1. currently, we only use up_proj but did not include  down_proj (moe expert down projection weight) since grad_nore will aggregate to there and cause larger logprobdiff and effect the reward. We are fixing this in issue #

2. memory optimization - currently the mem take too much

3. fp8 rollout and improve perf.

## Below is CC’s summary of the dev journal and unit test correctness. You can ignore it.

**Summary.** End-to-end GLM-5 / 5.1 / 5.2 (744B-A40B MoE + MLA + DSA) LoRA RL on the Megatron-Bridge path (`--megatron-to-hf-mode bridge`): selectable DSA kernel backend, MoE-expert LoRA in two layouts, MLA LoRA targets with HF↔Megatron name mapping, per-model runners + registries + CI combination-matrix tests, and several colocate-rollout correctness fixes.

**Fixes**
- Rollout served the base model: the refactor rollout path never tagged `lora_path` (`c4d9d49cb`); `lora_B=0` init masked it at step 0.
- MLA up-proj adapters silently dropped: stale `q_b/kv_b` exclusion emptied (`_SGLANG_UNSUPPORTED_HF_TARGETS = {}`).
- `--lora-base-cpu-backup` defaults ON in the runners — OFF silently gives fake on-policy (KL ~1.0 vs ~1e-4).
- DSA-indexer RoPE style (`INDEXER_ROPE_NEOX_STYLE=0`) and `--sglang-max-lora-rank` / `--sglang-lora-backend triton` pinned by the runners (engine-init crashes otherwise).
- `NVSHMEM_DISABLE_NCCL=1` defaulted (second NCCL communicator hung SGLang CUDA-graph replay).
- Weight-update checker works under frozen-base LoRA (base re-ship forced when the checker is on; GLM CI passes the engine-side stacked params via `--check-weight-update-skip-list`).

**Features**
- `--dsa-attention-backend {megatron,tilelang}` (default `tilelang`): `tilelang` = fused TileLang kernels (thd, matches slime rollout numerics), `megatron` = portable unfused megatron-core kernels (bshd). Backend is a `GLM5ModelProvider` dataclass field on the MB side; runners derive `--qkv-format` automatically.
- MoE-expert LoRA layouts: **per-expert by default**; `--experts-shared-outer-loras` (same field name end-to-end, incl. the serve-side auto-pair) opts into shared-outer. Serving uses sglang virtual-experts by default (`--no-sglang-lora-use-virtual-experts` selects the alignment path).
- `--target-modules all-linear` adds the MLA projections only when the HF config is MLA (dense models stay dense); the DSA indexer (`wq_b/wk/weights_proj`) is opt-in by explicit list.
- Runners `scripts/run_glm5_{1,2}_744b_a40b_lora.py` + `scripts/models/*_lora.sh` registries; CI: GLM-5.2 / GLM-5.1 tests run the combination matrix {shared-outer+virtual-experts, per-expert+no-virtual-experts} × {tilelang, megatron} (all must pass), plus a new gpt-oss-20b MoE-LoRA CI with both serving combinations.

**Validation (8×H200)**
- CI combination matrix: **10/10 green** (GLM-5.2 4/4, GLM-5.1 4/4, gpt-oss 2/2); Qwen2.5 dense LoRA CI (all-linear) green.
- Full 744B GLM-5.2, 64 GPU: rollout→train→save to 50+ steps; post-tagging-fix abs_diff ~0.0096 flat, `rollout/raw_reward` 0.45→0.8-0.99.
- **down_proj ablation** (full 744B, only difference = expert `down_proj` in targets): with it, abs_diff drifts to 0.058@43 (KL 0.02); without it, flat 0.0044-0.0077 (KL ~2e-4) at **no reward cost** — the down-proj delta lands on the residual stream and amplifies the bf16 kernel-numerics gap (matches Kimi team's finding). Recommendation: exclude expert down-proj for clean parity; `use_tis` / rollout-logprobs are the mitigations if its capacity is needed. wandb: https://wandb.ai/glm-zero/miles-glm5_att_unfused_moe (group `no-down-proj-260702-0948`).

**Naming note.** `--dsa-attention-backend` values are now `megatron` / `tilelang` (earlier reviews may say `megatron-bridge[-native]` / `slime` / `glm-native`); pairs with Megatron-Bridge PRs #16/#17. Runner-split parity verification: https://gist.github.com/yushengsu-thu/a23d258e7b211a86e97d63631cc09970


miles/utils/replay_base.py
```diff
@@ -194,10 +194,39 @@ def check_replay_result(self, old_topk_fn, scores, topk, top_indices, *args, **k
...
+            mm = is_mismatch.nonzero(as_tuple=False).squeeze(1)
+            if mm.numel() > 0:
+                smm = sf[mm]
+                srt, _ = smm.sort(dim=-1, descending=True)
+                k = min(topk, srt.shape[1] - 1)
+                top1, rk, rk1 = srt[:, 0], srt[:, k - 1], srt[:, k]
+                r2k = srt[:, min(2 * k - 1, srt.shape[1] - 1)]
+                rep = replay_flat[mm]
+                repv = rep != -1
+                rep_sc = smm.gather(1, rep.clamp(min=0).long())
```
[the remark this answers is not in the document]
yushengsu-thu: False positive — `rep` is 2D here, not 1D.

`replay_flat` is explicitly reshaped to 2D a few lines up (`top_indices.view(-1, top_indices.shape[-1])` → `[n_tokens, topk]`), and `mm` is a 1D tensor of row indices (`nonzero().squeeze(1)`), so `rep = replay_flat[mm]` selects rows → `[n_mismatch, topk]`. The gather is therefore 2D-input/2D-index with matching `size(0)`, which is valid, and returns the scores of **all** replayed experts per mismatched token — `rep_sc[repv]` then masks the `-1` padding with the same-shaped boolean mask. That per-pick semantics is intentional (the diagnostic averages over every replayed expert, not one per token).

Verified on torch 2.12.1: the current code runs and produces `[n_mismatch, topk]`; the suggested `unsqueeze(1)` makes the index 3D against the 2D input and raises exactly `RuntimeError: Index tensor must have the same number of dimensions as input tensor` — i.e. the suggestion would introduce the failure it aims to fix. Resolving.

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -0,0 +1,545 @@
...
+    # (q/k/v/o + MLA q_a/kv_a/q_b/kv_b) until the sglang probe is patched.
+    _tm = args.target_modules
+    # MoE/MLP LoRA on gate_proj/up_proj/down_proj is KEPT BY DEFAULT now (regular per-expert MoE LoRA;
+    # the bridge layout is selected by --experts-shared-outer-loras via lora_utils.create_lora_instance).
+    # Set KEEP_MOE_LORA=0 to restore the attention-only drop -- needed ONLY when serving via the sglang
+    # colocate rollout, which cannot yet serve MoE-expert LoRA (the dense gate_up "12288 vs 24576"
+    # scheduler crash). Training / train-only / dumped-rollout do not need the drop.
+    _keep_moe_lora = os.environ.get("KEEP_MOE_LORA", "1") != "0"
+    if _is_full and not _keep_moe_lora:
+        _tm = ",".join(m for m in _tm.split(",") if m.strip() not in ("gate_proj", "up_proj", "down_proj"))
```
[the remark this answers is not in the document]
yushengsu-thu: Note on anchoring first: this review predates the split — it was made against the pre-split `scripts/run_glm5_lora.py` (original commit `815b691e`), and GitHub's rename detection migrated the thread onto `scripts/run_glm5_2_744b_a40b_lora.py`. The block does carry over verbatim, so assessing it against the new file anyway:

The identified gap is real but narrow: it needs `KEEP_MOE_LORA=0` (non-default escape hatch) **and** `--target-modules all-linear` (non-default) together, and the failure mode is the loud scheduler crash at serving init rather than silent corruption.

We are not taking the suggested patch as-is: it expands `all-linear` **unconditionally**, which hijacks the default (`KEEP_MOE_LORA=1`) path — `all-linear` is a magic value resolved downstream by `lora_utils.convert_target_modules_to_megatron`, and its expansion is `lora_type`-dependent (`_STANDARD_LORA_ALL_MODULES` vs `_CANONICAL_LORA_ALL_MODULES`). A hand-rolled HF list in the runner would override that canonical resolution for CanonicalLoRA users and duplicates module-set knowledge that would silently drift. If we close the gap later, the right shape is confined to the `KEEP_MOE_LORA=0` branch (fail loudly on `all-linear`, or reuse the `lora_utils` expansion) — deferring that as a known edge case for now. Resolving.

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -0,0 +1,545 @@
...
+    # and the CI smoke tests previously fell back to csgmv (only run_glm5_lora_multinode.sh forced
+    # triton via --extra-args). Emitted as --sglang-lora-backend; override per-run with
+    # --sglang-lora-backend csgmv (or, via the .sh wrappers, SGLANG_LORA_BACKEND=...).
+    sglang_lora_backend: str = "triton"
+    # fp8 rollout: serve sglang from a pre-converted _fp8 ckpt (tools/convert_hf_to_fp8.py). fp8
+    # halves rollout weight mem so the 744B model fits engine=8 (1 node); megatron TRAIN stays bf16
+    # (it dequantizes the fp8 HF base via the bridge). Point --hf-checkpoint at the _fp8 dir.
+    fp8_rollout: bool = False
+
+    enable_wandb: bool = True
```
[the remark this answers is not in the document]
yushengsu-thu: Done in 44293dbbb (note this thread was rename-migrated from the pre-split `run_glm5_lora.py`; applied to both split runners). `ScriptArgs.experts_shared_outer_loras: bool = True` — default keeps the current (GPU-validated) shared-outer behavior; `False` selects the regular per-expert MoE-LoRA layout. Wired into the emission rather than just adding the field: the flag only takes effect when MoE-expert LoRA is active (`KEEP_MOE_LORA=1` with expert projections in `--target-modules`), and the serve side needs no extra plumbing — `arguments.py` auto-pairs `--sglang-experts-shared-outer-loras` from the train flag. Verified both values via a mocked `execute_train` capture (flag present with True, absent with False).

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -0,0 +1,545 @@
...
+    #       / down lora_B shared across experts, expert_dim=1; arguments.py also auto-sets the serve-
+    #       side --sglang-experts-shared-outer-loras so sglang knows the layout -- SGLang PR #21466).
+    #   --sglang-lora-use-virtual-experts (SERVE side: sglang serves the expert LoRA via the
+    #       virtual-experts path).
+    # Turn on only one and serving breaks (e.g. expert gate_up LoRA-B dim 768 vs 24576 = 32x ep ->
+    # "scheduler died during init"), so both go on together here. KEEP_MOE_LORA=0 turns both off
+    # (attention-only LoRA: q/k/v/o + MLA q_a/kv_a/q_b/kv_b), the previous bring-up default.
+    lora_args = f'--lora-rank {args.lora_rank} --lora-alpha {args.lora_alpha} --lora-dropout {args.lora_dropout} --target-modules "{_tm}" '
+    if _keep_moe_lora:
+        lora_args += "--experts-shared-outer-loras "
```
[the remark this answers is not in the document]
yushengsu-thu: Already done in 44293dbbb together with the companion field suggestion (r3509359709) — the emission is now exactly `if _keep_moe_lora and args.experts_shared_outer_loras:` in both split runners. Resolving.

miles/backends/megatron_utils/lora_utils.py
```diff
@@ -78,12 +78,24 @@
     "kv_a_proj_with_mqa": "linear_kv_down_proj",
     "q_b_proj": "linear_q_up_proj",
     "kv_b_proj": "linear_kv_up_proj",
+    # DSA indexer (GLM-5.1 / DeepSeek-V3.2): HF/SGLang leaf names wq_b/wk/weights_proj
+    # vs Megatron-Bridge names linear_wq_b/linear_wk/linear_weights_proj. Same HF<->Megatron
+    # mismatch handling as MLA above, so one --target-modules name resolves to the Megatron
+    # name for training and back to the HF name for SGLang rollout.
+    "wq_b": "linear_wq_b",
+    "wk": "linear_wk",
+    "weights_proj": "linear_weights_proj",
```
yushengsu-thu, line 83: If model has no these modules, will it occurred the error?

miles/backends/megatron_utils/lora_utils.py
```diff
@@ -315,13 +327,36 @@ def create_lora_instance(args: Namespace):
         lora_A_init_method=getattr(args, "lora_A_init_method", "xavier"),
         lora_B_init_method=getattr(args, "lora_B_init_method", "zero"),
     )
-    # Opt-in to SGLang PR #21466's shared-outer grouped-expert LoRA. Only the
-    # standard ``LoRA`` class supports the flag today.
-    if lora_cls is LoRA and getattr(args, "experts_shared_outer_loras", False):
-        lora_kwargs["experts_shared_outer_loras"] = True
```
yushengsu-thu, line 321: If we choose the moe layer lora, we expect it use experts_shared_outer_loras as the default but it's okay to sswitch - thus, it will read from arg.

miles/utils/arguments.py
```diff
@@ -220,6 +220,19 @@ def add_train_arguments(parser):
...
+                default="slime",
+                help=(
+                    "DSA sparse-MLA kernel backend for GLM (glm_moe_dsa) under --megatron-to-hf-mode bridge. "
+                    "'slime' (default) uses the fused TileLang kernels (SparseMLA + lighting_indexer) for "
+                    "rollout<->train numerical parity; 'megatron-bridge' uses the portable unfused megatron-core "
+                    "kernels. 'slime' requires --qkv-format thd and the optional tilelang dep, and is "
+                    "training/forward-only (no KV cache, cannot serve inference). Both support GLM-5.1 and "
+                    "GLM-5.2, full or LoRA. No effect on non-DSA models or the 'raw' path."
+                ),
+            )
```
yushengsu-thu, line 235: This argument is specific to Megatron-Bridge. If you use Megatron-Bridge (not m-bridge) as your training backend, you can choose either the Megatron-Bridge-native or GLM-native DSA implementation.

You should rename this argument throughout the codebase, including the .py and .sh scripts. You also need to modify the corresponding class/function names in the megatron-bridge repo.

Also, is this the right place to configure it? Or should we set it so that, when Megatron-Bridge is used, the option is exposed through dsa-attention-backend?

miles/utils/replay_base.py
```diff
@@ -194,10 +194,39 @@ def check_replay_result(self, old_topk_fn, scores, topk, top_indices, *args, **k
...
+                    flush=True,
+                )
+
         threshold = float(os.environ.get("MILES_TEST_R3_THRESHOLD", self.replay_check_max_mismatch_fraction))
         mismatch_threshold = threshold * orig_flat.shape[0]
         mismatch_indices = is_mismatch.nonzero(as_tuple=False).squeeze(1)
-        for idx in mismatch_indices:
+        for idx in mismatch_indices[:5]:
             i = idx.item()
             lines = []
```
yushengsu-thu: Is this correct? review and check again.

miles/backends/megatron_utils/lora_utils.py
yushengsu-thu: No — this mapping table itself can never error: `convert_target_modules_to_megatron` is a pure string rename (HF ↔ Megatron names), it never inspects the model. What happens downstream when the model lacks these modules:

- **Train (Megatron-Bridge)**: tolerant — `ModuleMatcher` logs a warning for target entries that matched no module (`peft/module_matcher.py:209`) and simply creates no adapter for them; training proceeds.
- **Serve (sglang)**: fail-fast at engine startup — LoRA buffers are sized per module NAME from the model config (`lora/utils.py get_hidden_dim`), so `kv_b_proj`/`q_b_proj` on a non-MLA model hits a missing `config.kv_lora_rank`/`q_lora_rank` (AttributeError), the DSA indexer names (`wq_b`/`wk`/`weights_proj`) hit the `get_dsa_index_*` arch asserts (the exact thing 808e6b552c relaxed for DeepSeek-V4), and unknown names raise `NotImplementedError` (`lora/utils.py:224`).

So: harmless rename here, warning-and-skip on the train side, loud startup crash on the rollout side. In practice the GLM runners only use these names for GLM (MLA+DSA) checkpoints, so the crash path is only reachable by pointing this target list at a non-MLA/non-DSA model — and it fails loudly at init rather than silently serving wrong adapters.

miles/backends/megatron_utils/lora_utils.py
yushengsu-thu: Confirmed — the current code already behaves as expected; the removed 4-line block was replaced (not dropped) by a stronger version right below (`lora_utils.py:330+`):

```python
_shared_outer = bool(getattr(args, "experts_shared_outer_loras", False))
lora_kwargs["share_expert_adapters"] = _shared_outer
if lora_cls is LoRA:
    lora_kwargs["experts_shared_outer_loras"] = _shared_outer
```

- **Reads from arg / switchable**: yes — both layouts are supported and selected by `--experts-shared-outer-loras`, which also drives `share_expert_adapters` consistently and auto-pairs the serve-side `--sglang-experts-shared-outer-loras` (arguments.py).
- **Shared-outer as the default when MoE LoRA is chosen**: yes at the runner level — since 44293dbbb, `run_glm5_{2,1}_744b_a40b_lora.py` default `experts_shared_outer_loras=True` and emit the flag; `--no-experts-shared-outer-loras` switches to per-expert.

One deliberate nuance: the **raw miles CLI** default (bare `train.py` without the flag) is per-expert, not shared-outer — the two layouts produce incompatible adapter checkpoints, so the CLI keeps shared-outer as an explicit opt-in and logs the chosen layout loudly whenever expert projections are targeted. The launchers' full-scale validation runs use that per-expert default. Keeping it this way avoids silently flipping checkpoint compatibility for non-runner entrypoints. Resolving.

miles/utils/arguments.py
yushengsu-thu: Done — renamed throughout both repos and validated on GLM-5.2_5layer:

- **miles `697cd8aac`**: `--dsa-attention-backend` choices are now `megatron-bridge-native` (portable unfused megatron-core kernels) and `glm-native` (fused TileLang kernels, vendored from slime); default stays the fused backend (now `glm-native`). Renamed across arguments.py, the bridge helpers, both split runners (Literal/qkv-format coupling/docstrings), and the model registries.
- **Megatron-Bridge `a758d75a`**: identifiers follow the backend name — `fused/slime_mla.py` → `fused/glm_native_mla.py`, `SlimeMLASelfAttention` → `GlmNativeMLASelfAttention`, `_slime_*` → `_glm_native_*`, plus the dispatch/default values and the layout assert message. Provenance references to the slime project (vendored kernel origin, reference implementations) are intentionally kept.
- On the placement question: kept as-is — the option stays exposed through `--dsa-attention-backend`, which is only parsed under `--megatron-to-hf-mode bridge`; with the value rename there is no longer a name collision with the training backend itself.

Validated: TOY RL smoke on `megatron-bridge-native` (abs_diff 0.0105 / kl 1.17e-4 at steps 0-1) and an attention-only toy run on `glm-native` (thd) through rollout→train step 1, both clean. PR descriptions of both PRs updated to the new names (sglang #29874 had no backend-name mentions). Resolving.

miles/utils/replay_base.py
yushengsu-thu: Reviewed and tested — the block is correct as written; no changes needed.

**Verification (executed the real `check_replay_result` from this file, torch 2.12.1):** hand-computed vs printed diag values match exactly in both overlap modes — legacy `min_overlap_ratio=0.0` (mismatch only on zero overlap, per the docstring): 1 mismatch, `gap=0.7 / orig_top2=0.85 / replay=0.015 / ratio=0.0176`; strict `ratio=1.0`: 2 mismatches incl. a near-tie, `gap=0.3505 / ratio=0.3809`. Also covered: the threshold `AssertionError` path, `topk == E` (the `k = min(topk, srt.shape[1]-1)` clamp), partial/full `-1` padding (excluded via `repv` / `is_padding`), and the silent early return when nothing mismatches. The earlier gemini concern about `smm.gather(1, rep...)` is a false positive — `rep = replay_flat[mm]` is 2D `[n_mismatch, topk]`, and the suggested `unsqueeze(1)` would itself raise the very RuntimeError it warns about (verified empirically).

**Why the block exists:** the base checker only counts mismatches; it cannot distinguish a benign near-tie (recompute & replay pick different but ~equally-scored experts → same MoE output → logprobs still align) from a real divergence (replay picked genuinely lower-scored experts). The two printed discriminants — `gap(rank k − rank k+1)` (≈0 ⇒ ill-conditioned tie) and `replay_experts_score / orig_topk_score` (≈1 benign, ≪1 real) — separate the two; in the test the pure-divergence case gives ratio 0.018 vs 0.38 once a near-tie is mixed in.

Two usage notes for future readers: `MILES_R3_DIAG` is an opt-in debug env var (this file is its only consumer; unset ⇒ the block is skipped entirely and nothing depends on it), and the `mismatch_indices[:5]` cap below is the deliberate log-flood guard from the introducing commit ("cap mismatch warnings") — the 5 per-token dumps are illustrative samples, while the counts/diag statistics are computed over ALL mismatched tokens. Resolving.

yushengsu-thu: Note. This is for debugging. If review dislike this, I can remove this.

miles/utils/arguments.py
```diff
@@ -1083,9 +1096,18 @@ def add_algo_arguments(parser):
...
+                help="R3 rollout routing replay (https://arxiv.org/abs/2510.11370): replay the rollout's "
+                "MoE top-8 in training for rollout<->train on-policy parity. Cheap (sglang routed-experts "
+                "capturer ~0.5 GB/rank, no host buffer) and recommended for MoE RL -- the GLM-5 launchers "
+                "pass it explicitly. DEFAULT OFF: it is MoE-only (a dense/non-MoE rollout emits no "
+                "routed_experts, so enabling it raises 'rollout_routed_experts is required' at step 0) and "
+                "R3 replay modules are only registered for models that opt in, so a default-on would "
+                "silently change training numerics / crash for non-GLM-MoE and dense runs. Enable with "
+                "--use-rollout-routing-replay. NOTE: distinct from the debug-only "
+                "--use-rollout-indexer-replay (stays default-off; it triggers the ~78-128 GB/rank "
+                "IndexerTopkCapturer host buffer that OOMs the colocate pod).",
```
yushengsu-thu: could you revert this part modification? still keep action="store_true",

yushengsu-thu: Done in 46092b2cb — reverted to `action="store_true"` / `default=False` with the short help text. The GLM-5 LoRA runners were the only callers of the `--no-use-rollout-routing-replay` form; their `use_r3=False` branch now just omits the flag (same behavior, the default is already off). Verified no other caller in the repo uses the `--no-` form.

miles/utils/arguments.py
```diff
@@ -220,6 +220,19 @@ def add_train_arguments(parser):
                 default="raw",
                 help="The method to convert megatron weights to hugging face weights for SGLang.",
             )
+            parser.add_argument(
+                "--dsa-attention-backend",
+                choices=["megatron-bridge-native", "glm-native"],
```
yueming-yuan: will "megatron" "tilelang" be a clearer naming here?

miles/utils/arguments.py
```diff
@@ -2303,7 +2352,17 @@ def miles_validate_args(args):
     )
 
     if args.ci_test and not args.debug_rollout_only and not args.debug_train_only:
+        from miles.backends.megatron_utils.lora_utils import is_lora_enabled
+
         args.check_weight_update_equal = True
+        if is_lora_enabled(args):
```
yueming-yuan: is lora weight checker supported in this PR? if no it would be better not to add this for now

yushengsu-thu:
1. remove check weight - if not support 
2. remove the comments in the core code

miles/utils/arguments.py
yushengsu-thu: Done in c1d163c4e — choices renamed to `megatron` / `tilelang` (default `tilelang`), applied across the arg definition, GLM-5 LoRA runners, bridge helpers, and the Megatron-Bridge side (package dirs `models/glm5/{megatron,tilelang}/`, `TileLangMLASelfAttention`, commit 73a0d3fc on bridge-dev-glm-merging).

miles/utils/arguments.py
yushengsu-thu: Good point — done in 1dc86b423. To clarify: this block wasn't a LoRA weight checker; it kept the *existing* checker (auto-enabled by `--ci-test`, line above) from false-failing under frozen-base LoRA — sglang stacks q_a+kv_a into `fused_qkv_a_proj_with_mqa` at load (and the DSA indexer weights likewise), so a LoRA run has no 1:1 trainer export to re-ship them and the checker's reset→verify protocol reports them as mismatched. That knowledge is GLM/MLA-specific though, so the LoRA-specific block is now removed from `arguments.py` entirely; the two GLM LoRA CI tests pass the skips explicitly via the existing `--check-weight-update-skip-list` flag instead.

miles/backends/megatron_utils/bridge_lora_helpers.py
```diff
@@ -102,6 +102,10 @@ def _setup_lora_model_via_bridge(args: Namespace) -> list:
         provider.num_layers_in_first_pipeline_stage = args.decoder_first_pipeline_num_layers
     if getattr(args, "decoder_last_pipeline_num_layers", None) is not None:
         provider.num_layers_in_last_pipeline_stage = args.decoder_last_pipeline_num_layers
+    # GLM DSA kernel backend: override the provider's "megatron" default from the
+    # miles arg. hasattr-guarded so non-DSA providers are untouched.
```
yushengsu-thu: remove the comments

yushengsu-thu: Done in 7834d7223 — removed, along with the twin comment in model_provider.py (comment-only change, stripped ASTs verified identical).

yushengsu-thu: done

miles/backends/megatron_utils/bridge_lora_helpers.py
```diff
@@ -128,5 +130,28 @@ def apply_lora_hook(model_chunks):
...
+                    _cfg.dsa_attention_backend = _backend
+                    _n += 1
+                if _m.__class__.__name__ == "TileLangMLASelfAttention" and getattr(_m, "config", None) is not None:
+                    _mla += 1
+                    _m.config.dsa_attention_backend = _backend
+        print(
+            f"[dsa-fix:lora] backend={_backend!r}; forced on {_n} module configs; "
+            f"{_mla} TileLangMLASelfAttention modules",
+            flush=True,
+        )
```
yushengsu-thu: Should I have this? Could I simplify this or move this part to megatron-bridge

yushengsu-thu: Simplified per your suggestion — the whole post-build force loop is deleted (miles 33cd2df44) and the knob moved to Megatron-Bridge as a declared `GLM5ModelProvider` dataclass field (MB f98108b9): since `provide()` hands the provider object itself to the model as its config, the field set before `provide()` reaches every module's config with no caller-side propagation. The loop was redundant belt-and-braces from debugging (its own log — 'forced on 196 module configs' — showed the configs already carried the value). Re-validated e2e on GLM-5.2_5layer `--ci-test` (tilelang default; a wrong/missing flag would crash loudly on the thd input, and the run is clean with step-1 KL ~1e-4). MB PRs #16/#17 rebuilt with the same change.

miles/utils/external_utils/command_utils.py
yueming-yuan, line 1: this should not be changed

miles/utils/arguments.py
```diff
@@ -2204,7 +2219,39 @@ def miles_validate_args(args):
         assert args.target_modules is not None, "'--target-modules' is required when LoRA is enabled."
 
         if args.target_modules == "all-linear":
-            modules = ["q_proj", "k_proj", "v_proj", "o_proj", "gate_proj", "up_proj", "down_proj"]
+            # "all-linear" = dense attention/MLP + (on MLA models only) the MLA up/down
```
yueming-yuan: Need to implement in a cleaner way...

miles/utils/arguments.py
```diff
@@ -2219,6 +2266,8 @@ def miles_validate_args(args):
             modules = [m for m in modules if m not in exclude_set]
 
         args.target_modules = modules
+        # Log the resolved set so a changed adapter shape on an existing recipe is visible.
```
yueming-yuan: maybe remove debug print?

miles/utils/replay_base.py
```diff
@@ -194,10 +194,36 @@ def check_replay_result(self, old_topk_fn, scores, topk, top_indices, *args, **k
         if mismatch_count == 0:
             return
 
+        # MILES_R3_DIAG: classify a replay/recompute mismatch as a benign near-tie (replayed
```
yueming-yuan: I would prefer not to add this to keep the code clean.. We can slightly tune up the R3 checker threshold if needed for each model; adding this means the previous thresholds are all invalid.

yueming-yuan:
maybe try to remove the multi-line comments in core files. try prompt below to your agent

> learn the codestyle in the core files in miles codebase. Remove all comments for code that can self-explain. only keep needed ones and keep them clean

miles/backends/megatron_utils/lora_utils.py
```diff
@@ -315,10 +319,13 @@ def create_lora_instance(args: Namespace):
...
+    # MoE-expert LoRA layout: per-expert by default; --experts-shared-outer-loras selects the
+    # shared-outer layout (validated and logged in arguments.py). Only the standard ``LoRA``
+    # class supports the shared-outer flag.
+    shared_outer = bool(getattr(args, "experts_shared_outer_loras", False))
+    lora_kwargs["share_expert_adapters"] = shared_outer
+    if lora_cls is LoRA:
+        lora_kwargs["experts_shared_outer_loras"] = shared_outer
```
yushengsu-thu: Should I do this? It will be assigned in arguemnt.py. isn't it?

yushengsu-thu: Yes, keep this — it's not a duplicate assignment. `arguments.py` only operates on `args` (flag definition, train/serve pairing, layout log, virtual-experts auto-enable); the Megatron-Bridge `LoRA` object never reads `args`, so its constructor kwargs (`share_expert_adapters`, and `experts_shared_outer_loras` for the standard `LoRA` class only) must be translated somewhere at build time — this is that single consumption point, already reduced to the minimal 4-line translation. Which class accepts which kwarg is MB-class knowledge, so it belongs next to `lora_cls(**lora_kwargs)` rather than in `arguments.py`.

miles/utils/external_utils/command_utils.py
yushengsu-thu: Done in 6aedd89d7 — command_utils.py is restored to main's exact form (the env-gated --no-wait/--submission-id option removed; nothing in the repo set those env vars — they only served out-of-repo dev launchers, which no longer export them either).

miles/utils/arguments.py
yushengsu-thu: Cleaned up in 18538c08e — the hand-rolled config.json read (try/except + pseudo-config fallback) is replaced with miles' existing `load_hf_config` helper and plain `getattr` gating (16 lines → 6).

miles/utils/arguments.py
yushengsu-thu: Removed in 2299d88d5 — the resolved list is visible via the emitted train command and the saved adapter config anyway.

miles/utils/replay_base.py
yushengsu-thu: Removed in the latest push — the MILES_R3_DIAG block is out of the tree (kept as a local debug patch on my side). One clarification: it never changed the checker semantics — it was env-gated pure logging, so existing thresholds were unaffected either way; per-model tuning via `replay_check_max_mismatch_fraction` remains the mechanism.

miles/utils/replay_base.py
```diff
@@ -197,7 +197,7 @@ def check_replay_result(self, old_topk_fn, scores, topk, top_indices, *args, **k
         threshold = float(os.environ.get("MILES_TEST_R3_THRESHOLD", self.replay_check_max_mismatch_fraction))
         mismatch_threshold = threshold * orig_flat.shape[0]
         mismatch_indices = is_mismatch.nonzero(as_tuple=False).squeeze(1)
-        for idx in mismatch_indices:
+        for idx in mismatch_indices[:10]:
```
yushengsu-thu, line 200: I still modify here.
Because if r3 fail, it will stay here too longer to show the mismatch_indices. So, i just let it extract final 10 to quickly move forward and the users can also get the error msg.

miles/ray/placement_group.py
```diff
@@ -191,7 +191,12 @@ def create_rollout_manager(args, pg):
 
     if args.check_weight_update_equal:
         ray.get(rollout_manager.check_weights.remote(action="snapshot"))
-        ray.get(rollout_manager.check_weights.remote(action="reset_tensors"))
+        # Honor the skip list on the reset too: params excluded from the compare (e.g. the
```
guapisolo: good fix. the logic is straight forward so you can remove this cmt?

yueming-yuan (approved):

yushengsu-thu:
Lora related CI run in local:

**Environment**: 1 node × 8 H200 (tests use 4 GPUs, TP=EP=4); Megatron-Bridge
`bridge-dev-glm-merging` @ `d5b93af1`; sglang @ `7c9dad97cb` (#29874 content); each combo is the
full rollout → train → weight-update-check → save loop; each test file executed verbatim
(`python3 tests/e2e/megatron/<test>.py`), exit code 0.

| Test | Combination | Result | step-1 abs_diff / KL |
|---|---|---|---|
| `test_glm5_2_744b_a40b_5layer_lora_ci.py` | tilelang + shared-outer + virtual-experts | ✅ | 0.0101 / 1.11e-4 |
| | megatron + shared-outer + virtual-experts | ✅ | 0.0098 / 1.06e-4 |
| | tilelang + per-expert + no-virtual-experts | ✅ | 0.0100 / 1.09e-4 |
| | megatron + per-expert + no-virtual-experts | ✅ | ~0.010 / ~1.1e-4 |
| `test_glm5_1_744b_a40b_6layer_lora_ci.py` | same 4 combinations | ✅ 4/4 | ~0.0104 / ~1.2e-4 |
| `test_gpt_oss_20b_moe_lora_ci.py` | shared-outer + virtual-experts | ✅ | — |
| | per-expert + no-virtual-experts | ✅ | 0.0111 / 4.9e-4 |

- Weight-update checker ON for all GLM combos (`--ci-test` +
  `--check-weight-update-skip-list fused_qkv_a_proj_with_mqa indexer.`):
  **0 × `check tensor equality failed`** across all runs.
- gpt-oss uses expert-only LoRA targets (gate/up/down, grouped-GEMM experts, TP=4, bridge mode,
  `lmsys/gpt-oss-20b-bf16`).

These should turn green on the hosted CI once the runners pick up the rebuilt v0.5.14 image
(its Dockerfile checks out the `sglang-miles` branch tip, which now contains the fixes).

kailash109:
Hello, I've added a PR to sglang that enables GLM 5.2 lora fp8 rollouts (there was a bug in Lora mem pool miscalculating tp dims): https://github.com/sgl-project/sglang/pull/31066

When enabling fp8 rollouts in miles for GLM 5.2 lora, logprob diff hovers around 0.05, config available [here](https://github.com/modal-labs/multinode-training-guide/pull/91/changes#diff-5c0fd24f2da16c87e70443790238cf547bb4a57b69228cb296a3898347ff5505) (apologies it's through modal so different syntax)

kailash109:
Hello, updated to do a full run with fp8 rollouts using GLM5.2 lora, config is [here ](https://github.com/modal-labs/multinode-training-guide/pull/91/changes#diff-c465de9c8c2d7f9baa3061e4cac721f62c5dd0d8ce077013adae02d99a10f89f)

Relevant patches to sglang: 
- miles/glm5_tilelang_safe_indices.patch — TileLang sparse-MLA backward produced NaN gradients: out-of-bounds loads/atomics through padded indices, plus the TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE miscompile. Added index guards  + disabled merge pass. 

- miles/sglang_fp8_lora_fix.py — SGLang's LoRA-B buffers were incorrectly sized on quantized column-parallel layers (input-axis sharding probe is poisoned by Fp8LinearMethod, which stamps input_size_per_partition == input_size), crashing engine init at LoRA attach. Instead I size buffers from the base module's output_size_per_partition instead. 

- miles/sglang_tp1_shared_expert_fix.py — with SGLANG_SHARED_EXPERT_TP1=1 (required: 128×128 quant blocks can't shard the 2048-row shared expert across TP32), the replicated shared-expert output was added after the all reduce, but FlashInfer AllReduce Fusion and dp-attention's reduce-scatterv defer/replace that all-reduce — so the shared output was summed once per TP rank (32×), which corrupts every Moe layer. Instead folded it into the pre-reduction add scaled by 1/tp_size

validation run with dapo-math 17k: logprob diff ~ 0.04-0.05
image

reward goes up to ~0.7-0.8: 
image

PR #1522, opened 2026-06-30
feat(utils): allow custom loss mask module paths

waitong94:
## Summary

- Allow --loss-mask-type to accept existing built-in names or a dotted module path.
- Validate dotted module paths during argument parsing so malformed values fail early.

## Why

This lets users plug in custom loss mask implementations without adding each implementation name to the argument parser.

## Validation

- python -m py_compile miles/utils/arguments.py tests/fast/utils/test_arguments.py
- ruff check miles/utils/arguments.py tests/fast/utils/test_arguments.py
- git diff --check
- isolated argparse validation for accepted built-ins, accepted module paths, and rejected malformed values

## Local notes

- pytest tests/fast/utils/test_arguments.py -q was blocked locally during test setup by a dependency mismatch in the local environment.
- pre-commit run --files miles/utils/arguments.py tests/fast/utils/test_arguments.py passed through ruff, autoflake, and isort; the Black hook did not complete locally.

PR #1561, opened 2026-07-02
refactor(utils): make loss mask generation strategy-based

TSunny007:
This refactors the loss mask path from a hardcoded tokenizer_type if/else into an interface/contract-based design, making it easier to add custom loss masks without modifying core code.

Changes:
- Introduce `LossMaskStrategy` ABC and a `register_loss_mask` decorator / `LOSS_MASK_REGISTRY`.
- Convert the existing Qwen family masks (`qwen`, `qwen3`, `distill_qwen`) into registered strategy classes.
- Make `MultiTurnLossMaskGenerator` a dispatcher that resolves the strategy by registered name or fully-qualified class path.
- Update `--loss-mask-type` in `arguments.py` to accept any registered name or class path.
- Update `tests/fast/utils/test_mask_utils.py` for the new contract and add registry/class-path tests.
- Document the new hook in `docs/user-guide/customization.md` and `docs/user-guide/argument-groups.md`.

Custom usage:

```python
from miles.utils.mask_utils import LossMaskStrategy

class MyMask(LossMaskStrategy):
    def get_loss_mask(self, messages, tools=None):
        ...
        return token_ids, loss_mask
```

Then pass `--loss-mask-type my_pkg.module.MyMask`, or register it with `@register_loss_mask("my_mask")` and use `--loss-mask-type my_mask`.

Related to the discussion in #1522 about a more extensible loss mask implementation.

### DIFF
--- a/miles/utils/mask_utils.py
+++ b/miles/utils/mask_utils.py
@@ -1,8 +1,17 @@
-import importlib
+"""Loss-mask strategies for SFT training.
+
+Follows the same ABC + registry + dispatcher pattern as
+``miles/utils/tracking_utils/base.py`` (TrackingBackend / BACKEND_REGISTRY /
+TrackingManager).  Strategies may hold per-run cached state (e.g.
+system_message_length) computed once at construction and reused across samples.
+"""
+
 from abc import ABC, abstractmethod
 from typing import Any
 
 from transformers import AutoTokenizer
+
+from miles.utils.misc import load_function
 
 
 def get_response_lengths(loss_masks: list[list[int]]) -> list[int]:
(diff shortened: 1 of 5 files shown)


miles/utils/mask_utils.py
```diff
@@ -6,24 +10,65 @@ def get_response_lengths(loss_masks: list[list[int]]) -> list[int]:
...
+        # protect from cases of module reload
+        if name in LOSS_MASK_REGISTRY and LOSS_MASK_REGISTRY[name] is not cls:
+            raise ValueError(f"Loss mask strategy {name!r} is already registered by a different class")
+        LOSS_MASK_REGISTRY[name] = cls
+        return cls
+
+    return decorator
+
+
+class LossMaskStrategy(ABC):
```
Rockdu, line 48: Seems like this strategy only exposes a member function "get_loss_mask". In this case, a new strategy class is heavy. Curious if more method can potentially to be added to the class in the future, if not maybe we should consider turn it to a customized function.

TSunny007: From the public interface alone, a single `get_loss_mask` does look like it could be a plain function. Two reasons I'd like to keep the class shape here:

1. **It isn't really a single-method abstraction.** `BasePerMessageLossMaskStrategy` shares `_get_system_message_length`, `_find_all_sublist_indices`, and `_apply_step_loss_mask`, plus two cached lengths (`system_message_length`, `gen_token_length`) between the `qwen` and `qwen3` strategies. Those lengths are computed once per run (a few `apply_chat_template` calls) and reused across thousands of samples. Just as important, `distill_qwen` extends `LossMaskStrategy` directly so it *skips* that precompute entirely — previously we ran it unconditionally for every tokenizer type, including distill, where the sublist unpack could raise. So the base/subclass split is doing real work (shared machinery + selective, safe precompute), not just wrapping one function.

2. **It matches the existing pattern.** This mirrors `miles/utils/tracking_utils/base.py` (`TrackingBackend` ABC + `BACKEND_REGISTRY` + `TrackingManager` dispatcher). Keeping loss-mask strategies in the same pattern keeps the two extension points consistent for maintainers.

A function-registry rewrite is feasible, but it would still need a context object to hold the cached per-run state, plus an opt-in flag for which functions precompute, and it'd lose the `issubclass(LossMaskStrategy)` check that both `register_loss_mask` and the fully-qualified-path loader rely on (otherwise any importable callable resolves). Net, I think the class earns its keep. Happy to add a short module docstring pointing at `tracking_utils` to make the shared-machinery intent obvious if that'd help.

PR #1563, opened 2026-07-02
(3/7) feat(session): strip R3 replay payloads from client chat responses

guapisolo:
## Summary
Strip the R3 replay payloads (`routed_experts` / `indexer_topk`) from client-facing chat responses. A deliberate, explicitly-owned **client-contract change**, isolated in its own PR.

> Stacked on #1569 → #1510 (`refactor/session-core-extract`). The multi-process data plane (#1518) and control plane (#1519) stack on this.

## Why
R3 payloads are replay-only training inputs: the rollout data path consumes them from the **session records** via `GET /sessions/{id}` (`_compute_sample_from_openai_record` reads `record.response`), never from the live chat reply — so shipping them to the agent client is dead weight, 100+ MiB per turn at the production R3 shape.

The concrete motivation is the multi-process session server stacked on this branch: without the strip, every R3 blob is relayed by the single router process, which measures at **100% CPU and caps 16-worker scaling at ~4.6×**; with it the router drops to ~5% CPU and scaling reaches **7.2–7.6×** (workers pegged — the intended bottleneck).

## Mechanics
`_chat_client_response` renders the client 200 from the already-parsed upstream response via `_strip_replay_payloads` — copy-on-write (only the dicts on the path to `meta_info` are copied), so the recorded response object is never mutated and records keep full R3 for the training path. All three 200-return points in `chat_completions` route through it (happy path + the two race-skip paths); non-200 passthrough and the generic `proxy()` are unchanged.

## What the client sees
`choices[].meta_info` without the two R3 keys; everything else (status, other meta_info keys, headers minus upstream framing) unchanged. `GET /sessions/{id}` records still carry the full response.

## Verification
- New test `test_chat_response_strips_replay_payloads_but_record_keeps_them` — the mock backend injects both R3 keys; asserts the live chat reply lacks them while the stored record retains them (which also pins the copy-on-write invariant end-to-end).
- `python -m pytest tests/fast/router/` → 74 passed at this tip. black / isort / ruff clean.

### DIFF
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -52,13 +52,6 @@ _CLIENT_STRIPPED_META_KEYS = ("routed_experts", "indexer_topk")
 
 
 def _strip_replay_payloads(response: dict) -> dict:
-    """Copy-on-write removal of the R3 replay payloads from a parsed chat response.
-
-    The client-facing chat response omits ``routed_experts`` / ``indexer_topk``;
-    the session record keeps the full response (served by ``GET /sessions``).
-    Only the dicts on the path to ``meta_info`` are copied, so the recorded
-    response object is never mutated.
-    """
     stripped_choices = []
     for choice in response.get("choices", []):
         meta = choice.get("meta_info")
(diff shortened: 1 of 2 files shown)


Shi-Dong (approved):
LGTM!

PR #1572, opened 2026-07-03
[optim]--rematerialize-param-from-master-weight: save the bf16 weight backup in colocate

yueming-yuan:
## TLDR

On GLM 5.2, 64 * GB300, before -> after optimization:
CPU memory 858.2 GB -> 729.2 GB (-129 GB)
weight update time 76.3 s -> 73.8 s (-2.5 s)
wake_up time (include rematerialize) 4.6 s -> 4.8 s (+0.2 s)

## Motivation

In colocate, the weights_backuper keeps a pinned CPU copy of the bf16 weights (2 bytes/param/rank) to (a) restore params after the TMS rollout pause and (b) feed update_weights while the actor is paused. On GLM-5.2 744B that copy is 26-37.5 GB/rank — 104-150 GB per node sitting inside a rollout-peak budget that measured 858/898 GiB.

Both jobs can be done without it:

- **update_weights** runs with the param buffer still resident (phase 1 of the sleep pauses grads/main/mv only) and reads live GPU weights; the buffer is dropped without backup right after the update. Engine pools are sized on an empty GPU at startup, so engine + 2P coexist only during the short update window — the KV pool is untouched.
- **restore at wake** replays exactly what produced the bf16 weights at step end: `_copy_main_params_to_model_params` (fp32 main -> bf16 cast) + param all-gather. Bit-identical by construction.
- Only tensors not derivable from fp32 mains keep a KB-scale pinned backup: expert_bias buffers and fp32-dtype params (their optimizer "main" aliases the param itself). update_weights reads those from the backup since their TMS region is paused.

## Guarantees

- `--check-rematerialize-param-from-master-weight` (opt-in, off by default): per-tensor SHA256 recorded at backup, re-verified after restore, for the first 2 cycles. 2 cycles also catches param-gather state-machine damage, which surfaces in the following step's weights.
- Init-time coverage assert: every named parameter must be in the DDP param buffers (restorable via cast+AG) or in the extras backup — fails loudly on e.g. frozen params. (DDP buffer membership is the correct local criterion: all distributed-optimizer structures only cover this rank's owned shard under DP>1.)
- Config asserts exclude non-megatron backends, LoRA, `--debug-disable-optimizer`, `--indep-dp`, the old-actor tag, TE-managed precision-aware optimizers (without `--optimizer-cpu-offload`; with it, the HybridDeviceOptimizer holds standalone masters and restore replays its copy-back hooks), overlap-param-gather, and --disable-compute-advantages-and-returns (the per-cycle restore runs in that block). `--indep-dp` is out because the param-buffer pause lives in update_weights, which `RayTrainGroup` runs on the first alive cell only while sleep runs on all of them — every other cell would keep the bf16 shard resident for the whole run; the assert can be lifted once group.py's "use all cells to update weights" TODO lands. critic-only warmup steps are fenced (they rerun update_weights without an intervening actor wake_up; #1588 removes the redundant updates and lifts the fence). --debug-train-only silently disables the flag and the critic actor keeps the baseline pause path - neither ever runs update_weights, where the param buffer is paused.
- ref/teacher tags are supported: they have no optimizer masters, so they keep full pinned copies via a delegated normal backuper; only the actor copy is dropped.
- update_weights' weight source asserts every tensor is either in the extras backup or in the DDP param buffers - a new buffer admitted into the sync source fails loudly instead of handing out a pointer into the paused region.

## Measured

| | before | after |
|---|---|---|
| host rollout peak, max node (744B, 16x GB300) | 858.2 GB | **729.2 GB (-129)** |
| engine KV pool | 1.64M tokens | unchanged |
| steady step time | ~1000 s | 972-989 s (restore cast+AG ~ +1.5 s, per-step backup D2H removed) |
| weight bit-exactness | pinned-copy restore | SHA256-identical x3 cycles @744B TP8xPP4xDP2xEP16, x4 @Qwen3-4B TP2xDP2 |

### Qwen3-30B-A3B and Qwen3.5-35B-A3B, 8x H200

3 rollouts, off vs on with otherwise identical args; colocate + `--optimizer-cpu-offload --overlap-cpu-optimizer-d2h-h2d --use-precision-aware-optimizer` (the HybridDeviceOptimizer path) + `--use-kl-loss` (ref tag), i.e. both paths this PR adds.

**Qwen3-30B-A3B (TP4 / EP8)**

| | off | on |
|---|---|---|
| host memory, after_offload_train (mean of 3) | 943.6 GB | **863.1 GB (-80.5, -8.5%)** |
| GPU used, update_weights window | 22.8 GB | 30.7 GB (+7.9 = the per-rank bf16 shard) |
| GPU used, global peak (training fwd/bwd) | 47.2 GB | 49.3 GB (+2.1) |
| cycle time | - | +1.4 s (~1.5% of train; per-step backup D2H removed, restore cast+AG added) |
| bit-exactness | - | `--check` run: SHA256 identical, both checked cycles; **0 extras** - every tensor rebuilt via HDO copy-back replay + all-gather |

**Qwen3.5-35B-A3B (TP1 / CP2 / EP8, MTP)**

| | off | on |
|---|---|---|
| host memory, after_offload_train (mean of 3) | 1013.0 GB | **910.6 GB (-102.4, -10.1%)** |
| GPU used, global peak | 57.8 GB | 57.9 GB (+0.05) |
| update_weights | 8.0 s | 7.1 s |
| extras backup | - | 30 fp32 `linear_attn.A_log` tensors, 0.0 MiB |
| bit-exactness | - | CI `test_mtp1_spec_v2_r3` (same combo + `--check`) passed |

The +8-11 GB the resident param buffer adds sits inside the update window, whose baseline is far below the training fwd/bwd peak - the jobs' actual GPU ceilings moved by +2.1 GB (30B) and +0.05 GB (35B). Per-rollout timing on 35B is noisy (random rollout lengths); no regression observed.


miles/backends/megatron_utils/actor.py
```diff
@@ -518,6 +530,8 @@ def update_weights(self, info: "EnginesAndLock") -> None:
                 else:
                     self.weights_backuper.backup("old_actor")
 
+        if self.args.rematerialize_param_from_master_weight:
+            torch_memory_saver.pause(tag="param_buffer")
```
yushengsu-thu, line 774: Not exception-safe: if anything in the update block above raises (engine death, the ci_test weight-version check), this pause is skipped and nothing ever pauses param_buffer afterwards — the full bf16 shard stays resident. Also this same epilogue is copy-pasted in the debug_skip early return; a try/finally with a single owner would cover both.

miles/backends/megatron_utils/actor.py
```diff
@@ -215,8 +220,13 @@ def sleep(self) -> None:
         print_memory("before offload model")
         destroy_process_groups()
 
-        tag = "default" if is_lora_enabled(self.args) else None
-        torch_memory_saver.pause(tag=tag)
+        if self.args.rematerialize_param_from_master_weight and self.role == "actor":
+            # Keep params resident for update_weights, which pauses them afterwards.
+            torch_memory_saver.pause(tag="grad_buffer")
```
yushengsu-thu, line 306: Keeping the param buffer resident from here through update_weights raises the weight-sync GPU peak by the per-rank bf16 shard, right when engine weights + conversion buffers are also live. Fine on GB300, but mem-fraction values tuned for the old flow can OOM on 80GB parts — worth a line in the help text.

miles/utils/arguments.py
```diff
@@ -2035,6 +2050,35 @@ def _resolve_eval_datasets(args) -> list[EvalDatasetConfig]:
...
 
 
+def _validate_rematerialize_param_from_master_weight(args):
+    if not args.rematerialize_param_from_master_weight:
+        return
+    if args.debug_train_only:
+        # update_weights never runs, so the param buffer would never be paused.
+        args.rematerialize_param_from_master_weight = False
+        return
+    assert args.colocate and args.offload_train
```
yushengsu-thu, line 2608: These asserts read megatron-only args — with --train-backend fsdp this AttributeErrors (or silently enables). Guard on the backend first. Two more cheap rejections that belong here: --debug-disable-optimizer (optimizer=None → restore crashes with AttributeError after full init) and the LoRA check that currently sits in build_main_cast_context and only fires after the whole cluster spins up.

miles/utils/tensor_backper.py
```diff
@@ -74,6 +86,91 @@ def restore(self, tag: str) -> None:
...
+        expected = self._expected_hashes
+        assert expected is not None
+        assert actual.keys() == expected.keys(), (
+            f"main-cast restore changed the tensor set: "
+            f"missing={sorted(expected.keys() - actual.keys())[:5]} "
+            f"extra={sorted(actual.keys() - expected.keys())[:5]}"
+        )
+        mismatches = [name for name in expected if actual[name] != expected[name]]
+        if mismatches:
+            raise RuntimeError(
```
yushengsu-thu, line 173: Per-rank raise: under DP>1 only the rank with the bad shard dies, the rest march into the next collective and hang until the NCCL watchdog — which buries exactly the mismatch this flag is meant to surface. all_reduce a failure flag and abort together?

miles/utils/tensor_backper.py
```diff
@@ -74,6 +86,91 @@ def restore(self, tag: str) -> None:
...
+            if backup is None:
+                assert (
+                    id(tensor) in self._ctx.rematerializable_ids
+                ), f"{name} is neither in the DDP param buffers nor in the extras backup"
+                backup = tensor.detach()
+            out[name] = backup
+        return out
+
+    def _compute_hashes(self) -> dict[str, str]:
+        from miles.backends.megatron_utils.ci_utils import _hash_tensor_sha256
```
yushengsu-thu: utils importing a backend-private helper — ci_utils pulls in megatron.core at import, so check mode breaks in a megatron-less env (the new tests already need importorskip for this). _compute_hash_dict in this file does the same job.

yushengsu-thu (commented):
Nice memory win. Left a few inline notes — the exception-safety one and the fsdp/validation gaps are the ones that can be fixed before merge as well.

miles/backends/megatron_utils/actor.py
Zhichenzzz: agreed!

miles/utils/tensor_backper.py
Zhichenzzz: yes, i think we should consider distributed scenarios

yueming-yuan: In this case the driver will crash immediately without fault tolerance; For FT it's a real issue, but this PR is now not compatible with FT (require a TODO in FT's group.py). I added a `indep_dp` assert now.

miles/backends/megatron_utils/actor.py
yueming-yuan: same to another one, this is not reachable by try/expect in current design; but will cause issue with fault tolerance. For simplicity, assert indep-dp == 0 now and will fix it after FT.

miles/utils/tensor_backper.py
```diff
@@ -102,6 +198,13 @@ def restore(self, tag: str) -> None:
         torch.cuda.synchronize()
 
 
+def _hash_tensor_sha256(x: torch.Tensor) -> str:
```
Zhichenzzz, line 208: could we move this checker to miles/utils? since local_weight_checksum also use it to check equal

miles/utils/tensor_backper.py
```diff
@@ -74,6 +87,89 @@ def restore(self, tag: str) -> None:
         torch.cuda.synchronize()
 
 
+class _TensorBackuperMainCast(TensorBackuper):
+    """Rematerialize weights from the optimizer's fp32 master weights (same cast +
+    param all-gather as the step end, so bit-identical) instead of a pinned CPU copy;
+    only `extras_getter` tensors keep a small pinned backup. With `check`, the first
+    `_check_num_cycles` restores are SHA256-verified against backup time."""
+
+    _check_num_cycles = 2
```
Zhichenzzz, line 100: why here we need 2 cycles

yueming-yuan: This is just a manually picked value, in case there's some stale bug that will not appear in the first cycle but will break the following steps

Zhichenzzz (approved):

PR #1573, opened 2026-07-04
Add Search-R1 group-relative length penalty

ys-2020:
## Summary

Add an opt-in, group-relative length penalty for the Search-R1 example: among the responses in a prompt group that answer correctly, responses that are much longer than their sibling successes get a small reward deduction, so the policy is rewarded for being right *and* concise. Enabled in the example via `--custom-reward-post-process-path length_reward.post_process_rewards`.

## Motivation

GRPO compares rewards only within a prompt group, so response length is free: a policy can inflate trajectories without losing advantage. On Search-R1-style multi-turn search tasks this shows up as steadily growing responses. A group-relative penalty (rather than a global length cap) adapts the allowance to each prompt's difficulty, using the group's own successful responses as the reference.

## How it works

For each prompt group with at least 2 successes (reward >= `success_threshold`; by default the sample must also have COMPLETED status):

- reference length = the `ref_quantile` quantile of the successes' effective lengths (`sum(loss_mask)` when available, else `response_length`)
- allowed length = `ref_len * (1 + relative_slack) + absolute_slack`
- penalty = `min(cap, beta * excess / ref_len)`, applied only to successes over the allowance
- shaped reward = `max(min(success_floor, reward), reward - penalty)` — a penalized success never drops below the floor, so it still ranks above failures

Failed samples and groups with fewer than 2 successes are untouched. Because a custom reward post-process replaces the trainer's built-in normalization, the module re-applies GRPO/GSPO group mean-centering (and optional std normalization) to the shaped rewards; raw rewards are returned unchanged for logging and pass@k.

## Configuration

All knobs are env vars prefixed `SEARCH_R1_LENGTH_PENALTY_`, with defaults: `SUCCESS_THRESHOLD=1.0`, `BETA=0.10`, `CAP=0.20`, `REF_QUANTILE=0.25`, `REL_SLACK=0.10`, `ABS_SLACK=16.0`, `SUCCESS_FLOOR=0.80`, `REQUIRE_COMPLETED=true`, `LOG_STATS=true`. Setting `BETA` or `CAP` to 0 disables shaping.

## Observability

Each sample gets `length_*` metadata (penalty, applied flag, base/shaped reward, delta, effective length, success and group-eligibility flags), and `train_data_conversion.py` passes these keys through to train data for downstream logging/analysis.

## Scope

Only the Search-R1 example enables this. The penalty logic itself is not search-specific — any GRPO/GSPO task with threshold-style rewards can point `--custom-reward-post-process-path` at this module — but it ships as an example, not a core feature. The only core change is the metadata passthrough.

## Tests

- `tests/fast/examples/test_search_r1_length_reward.py`: penalty within groups, disabled config, interleaved groups, single-sample groups (no NaNs)
- `tests/fast/ray/rollout/test_train_data_conversion.py`: metadata passthrough and DP split


ys-2020:
@copilot+gemini-3.1-pro-preview Please review the PR again

PR #1585, opened 2026-07-06
[sglang] fix: actively drain pending requests in flush_cache to prevent colocate offload timeout

HJSang:
## What

Under `--colocate`, the SGLang engine must flush and free its KV memory at **every** rollout→train offload (`SGLangEngine.release_memory_occupation` → `flush_cache`). The sglang scheduler only performs `/flush_cache` when there are **zero** queued and **zero** running requests; otherwise it returns HTTP 400 (`Scheduler.flush_cache`: *"Cache not flushed because there are pending requests"*).

`sglang_rollout.abort()` is issued before offload, but aborting a **running** request only sets `to_abort=True` and relies on **one more decode forward pass** to reap it (sglang `Scheduler.abort_request`, "Abort method 3"). A single request that never reaches that decode step (wedged engine, torn-down batch) keeps `#running-req` at 1, and the old passive retry loop in `flush_cache` then raises:

```
TimeoutError: Timeout while flushing cache.
  File ".../miles/backends/sglang_utils/sglang_engine.py", in release_memory_occupation
ray::SGLangEngine.release_memory_occupation() → TimeoutError: Timeout while flushing cache.
ray::RolloutManager.offload() → TimeoutError: Timeout while flushing cache.
```

This crashes colocate RL runs. Observed twice on Qwen3-8B DAPO-Math colocate runs: the scheduler logged `#queue-req: 0, #running-req: 1` — i.e. the queue was empty and exactly **one** running request would not drain — and `flush_cache` timed out after its 60×1s budget.

## Why this approach is correct

The old loop *passively* polls `/flush_cache` and waits, but a running request marked `to_abort=True` only gets reaped on its next decode step. If that step never comes, waiting cannot help. The fix re-issues `/abort_request {abort_all: True}` on **each** flush retry, so the scheduler keeps re-marking the lingering request every poll until the batch drains and `/flush_cache` returns 200 — turning a passive wait into an active drain. The abort payload matches what `sglang_rollout.abort()` already sends.

Behavior when nothing is pending is unchanged (first `/flush_cache` returns 200, no abort sent). The 60-retry budget and the final `TimeoutError` are preserved as a backstop for a genuinely unrecoverable engine.

## Tests

`miles/backends/sglang_utils/test_sglang_engine.py`:
- re-issues `abort_all` once per pending poll, stops when `/flush_cache` returns 200
- immediate 200 → no abort issued
- never drains → still raises `TimeoutError` after the budget
- non-primary node (`node_rank != 0`) → no-op

## Risk

Scoped to `flush_cache`; only adds an abort POST on the non-200 retry path that already sleeps and loops. Disaggregated runs (which never offload mid-run) are unaffected — they don't call `flush_cache` at a barrier.

### DIFF
--- a/miles/backends/sglang_utils/sglang_engine.py
+++ b/miles/backends/sglang_utils/sglang_engine.py
@@ -399,13 +399,34 @@ class SGLangEngine(RayActor):
         )
 
     def flush_cache(self):
-        """Flush the cache of the server."""
+        """Flush the cache of the server, actively draining pending requests first.
+
+        The sglang scheduler only performs ``/flush_cache`` when there are zero
+        queued and zero running requests; otherwise it returns HTTP 400 (see
+        ``Scheduler.flush_cache``). Under ``--colocate`` the engine must flush and
+        free its KV memory at every rollout->train offload, so any request that
+        has not drained blocks the flush.
+
+        ``abort`` (``sglang_rollout.abort``) is issued before offload, but aborting
+        a *running* request only sets ``to_abort=True`` and relies on one more
+        decode forward pass to reap it. A single request that never reaches that
+        decode step (wedged engine, torn-down batch) keeps ``#running-req`` at 1
+        and the old passive retry loop then times out with
+        ``TimeoutError: Timeout while flushing cache`` — killing colocate runs
+        (see the ``godly-eddy`` / ``brave-altar`` failures: ``#queue-req: 0,
+        #running-req: 1``).
+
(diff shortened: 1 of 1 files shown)


HJSang:
Note: the accompanying unit test (`test_sglang_engine.py`) was dropped from this branch at the requester's direction to keep the change fix-only. Happy to re-add the drain/immediate-success/timeout/non-primary tests if reviewers want them before merge.

PR #1590, opened 2026-07-07
feat(dpsk-v4): add {tool,user} role support for text-protocol RL

Shi-Dong:
## Summary

Add proper `user`-role support for the DeepSeek-V4 TITO tokenizer so text-protocol harnesses (e.g. OpenEnv) that append environment output as plain `user` turns don't NaN on the first backward.

Registers a `{tool, user}` `FixedTemplateRow` for `DeepSeekV4TITOTokenizer` with `extra_kwargs={"drop_thinking": False}`.

## Why (root cause)

DeepSeek-V4's encoder forces `effective_drop_thinking=False` **only when a message carries `tools`**. Native tool-calling harnesses (terminus-2) always pass `tools=`, so thinking is never stripped and TITO's append-only invariant holds — that surface stays HF-native (`{tool}` row, no kwargs).

Text-protocol scaffolds append env output as plain `role=user` turns and pass **no `tools`**. With the default `drop_thinking=True`, once a new `user` turn advances `last_user_index` the encoder strips every prior assistant's `reasoning_content` — a **non-append-only history mutation** that corrupts the cached pretokenized prefix and produces **NaN gradients on the first backward**.

Pinning `drop_thinking=False` on the `{tool, user}` surface restores append-only rendering. This mirrors how V3.2 registers only `{tool}` for the same reason.

## Change set

- `miles/utils/chat_template_utils/tito_tokenizer.py` — add the `{tool, user}` row + docstring explaining the invariant.
- `tests/fast/utils/chat_template_utils/test_fixed_templates.py` — pin both dsv4 rows (`{tool}` → HF-native/no kwargs; `{tool, user}` → `drop_thinking=False`).

## Caveat — needs a pod smoke run

The **sglang serving side must also render with `drop_thinking=False`** so generation and re-tokenization are byte-identical. This PR pins the training-side tokenizer; a short DeepSeek-V4-Flash smoke run is still needed to confirm the first backward is NaN-free with the serving config in use.

## Test plan

- [ ] CI fast tests (`tests/fast/utils/chat_template_utils/test_fixed_templates.py`) green.
- [ ] DeepSeek-V4-Flash text-protocol smoke run reaches step >1 with finite gradients.


guapisolo (commented):
running e2e tests.

guapisolo (dismissed):
test passed. lgtm

PR #1598, opened 2026-07-08
Add nvidia-resiliency-ext to requirements

fzyzcjy:
## What

Add `nvidia-resiliency-ext` (Linux-only) to `requirements.txt`.

## Why

The fault-tolerance trainer auto-sets `non_persistent_ckpt_type="local"` whenever `train` is in `--ft-components` (`miles/utils/arguments.py`), and Megatron requires the `nvidia_resiliency_ext` module for local checkpointing. Without it, any FT training run crashes at init with:

```
RuntimeError: nvidia_resiliency_ext is required for local checkpointing
```

The dependency is already declared on the trainer-FT branch, but it needs to be on `main` first so the `radixark/miles:dev` CI image (which bakes deps at build time via `pip install -r requirements.txt`, built from `main`) ships the module. Landing this standalone unblocks the FT e2e tests: once merged and the `dev` image is rebuilt, the FT suite can run and pass in CI.

This is a pure dependency addition — no code changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ httpx[http2]
 hypothesis>=5.40
 mcp[cli]
 memray  # needed for debugging (but is lightweight), we can put it to dev mode when using pyproject.toml
+nvidia-resiliency-ext; platform_system == "Linux"
 omegaconf
 pillow
 polars
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ httpx[http2]
 hypothesis>=5.40
 mcp[cli]
 memray  # needed for debugging (but is lightweight), we can put it to dev mode when using pyproject.toml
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Good call — pinned to `nvidia-resiliency-ext~=0.6.0; platform_system == "Linux"` (i.e. `>=0.6.0,<0.7.0`).

Rationale: `~=0.6.0` caps at the 0.6.x line so a future major release can't silently break CI/training, while still allowing patch updates. `0.6.0` is the current PyPI latest (what an unpinned install resolves to today), and it exposes the `LocalCheckpointManager` API that Megatron's local-checkpointing path requires. The FT e2e suite will validate this exact version once this lands and the `dev` image is rebuilt.

yushengsu-thu (approved):

PR #1599, opened 2026-07-08
fix(dpsk-v4): real-history TITO merge and parse-kwargs alignment for the {tool,user} surface

guapisolo:
> **Depends on:** #1600

## Summary
DeepSeek-V4 `{tool, user}` TITO session support for text-protocol RL scaffolds.

## Motivation
Text-protocol scaffolds (OpenEnv) feed environment output back as plain `user` turns. V4 folds contiguous `tool`/`user` turns into ONE `<｜User｜>` block, so the base per-segment TITO machinery fabricates `<｜User｜>` markers. The encoder's default `drop_thinking=True` mutates history on append, breaking the pretokenized-prefix invariant. sglang's `deepseek-v4` reasoning parser separates reasoning only when the request carries `thinking`.

## Usage
`--tito-model deepseekv4 --tito-allowed-append-roles tool user` on the session server (`MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1`); full recipe in `tests/e2e/sglang/test_session_server_multi_role/test_deepseekv4.py`.

## Design Notes
- **Key choices:** the `{tool, user}` row pins `drop_thinking=False`, keeping V4 renders append-only.
- **Key choices:** `tokenize_additional_non_assistant` diffs the real-history render; dummy contexts cannot know whether the `<｜User｜>` block is still open.
- **Key choices:** the tokenizer derives the effective `thinking` mode; the session server forwards it with client keys winning.
- **Key choices:** opener handling lives in #1600's `add_generation_prompt`; this layer carries no opener strips.

## Verification
- **Tests added:** session e2e `test_deepseekv4.py` — `tito_session_mismatch_rate` 0.0 on all four components, 64/64 rollback/append_user coverage.
- **Tests added:** Stage-2 verifier `{tool, user}` 48/48 both thinking modes; `test_via_tito_pass_on_deepseek_v4` fast test.

## Review Focus
- Scrutinize `tokenize_additional_non_assistant` in `tito_tokenizer.py`: the append-only assertion is the only guard on prefix drift.
- Scrutinize the `chat_template_kwargs` merge in `session/core.py`: client keys must win over the server render config.


miles/utils/chat_template_utils/tito_tokenizer.py
```diff
@@ -793,9 +820,18 @@ class DeepSeekV4TITOTokenizer(TITOTokenizer):
...
+            template=None,
+            extra_kwargs={"drop_thinking": False},
+        ),
     )
 
     _DEFAULT_ASSISTANT_START = "<｜Assistant｜>"
+    # After a user-tail render the encoder always emits assistant-start plus a
+    # thinking bracket; these two-token suffixes are what merge must strip when
+    # the appended messages keep extending the same user block.
+    _OPENER_STRS = ("<｜Assistant｜><think>", "<｜Assistant｜></think>")
```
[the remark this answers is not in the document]
guapisolo: The premise doesn't hold for the dsv4 encoder: a thinking bracket is always appended after `<｜Assistant｜>` in both modes — `thinking_mode="thinking"` renders a user tail as `...<｜Assistant｜><think>`, and `thinking_mode="chat"` as `...<｜Assistant｜></think>` (verified directly against sglang's `encoding_dsv4.encode_messages`; its render path is commented "Normal generation: append Assistant + thinking token"). `_OPENER_STRS`'s second entry is exactly the chat-mode opener, so both modes are already covered. A bare `<｜Assistant｜>` fallback would be dead code — the longer bracket entries always match first — and would silently absorb a genuine encoder behavior change instead of surfacing it as a verifier mismatch.

miles/utils/chat_template_utils/tito_tokenizer.py
```diff
@@ -814,6 +850,73 @@ def __init__(
...
+        # messages.  Gated on the old tail role so tokens the model actually
+        # generated (assistant tail) are never touched.
+        if (
+            old_messages
+            and old_messages[-1].get("role") != "assistant"
+            and len(prefix) >= 2
+            and prefix[-2] == self._assistant_id
+            and prefix[-1] in self._think_bracket_ids
+        ):
+            prefix = prefix[:-2]
```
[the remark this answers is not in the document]
guapisolo: Same premise as the other `_OPENER_STRS` thread: the dsv4 encoder never emits a bare `<｜Assistant｜>` tail — `thinking_mode="chat"` appends `<｜Assistant｜></think>` — so the 1-token strip variant is unreachable. The strict role-gated 2-token strip is intentional: if the encoder's opener behavior ever changes, we want a loud verifier mismatch, not silent absorption.

miles/utils/chat_template_utils/tito_tokenizer.py
```diff
@@ -780,9 +781,32 @@ class DeepSeekV4TITOTokenizer(TITOTokenizer):
...
+    ``tokenize_additional_non_assistant`` to diff the real-history render
+    instead (see below).
+
+    Two append surfaces are registered:
+
+    - ``{tool}`` (native tool-calling harnesses, e.g. terminus-2): the encoder
+      forces ``drop_thinking=False`` whenever any message carries ``tools``
+      (``encoding_dsv4.encode_messages``), so prior assistants keep their
+      thinking and history never mutates on append.
+    - ``{tool, user}`` (text-protocol scaffolds like OpenEnv, which feed env
```
Shi-Dong: Let's refrain from mentioning the "OpenEnv" harness since it is a private harness (it is not shipped with the OpenEnv repo).  I think you should also mention in the PR description a contract for agent harness in order for `{tool, user}` to work.

Shi-Dong (approved):
Approved since the changes are not intrusive.

miles/rollout/session/core.py
```diff
@@ -207,6 +207,14 @@ async def chat_completions(
...
             # token IDs still come from logprobs below.
             request_body["no_stop_trim"] = False
+            # Chat template kwargs should also be forwarded to sglang to make sure
+            # parsers work correctly.
+            server_ctk = self.registry.tito_tokenizer.chat_template_kwargs
+            if server_ctk:
+                request_body["chat_template_kwargs"] = {
+                    **server_ctk,
+                    **(request_body.get("chat_template_kwargs") or {}),
+                }
```
Shi-Dong, line 217: This change seems unrelated to `{tool, user}` support?  If yes, ideally it should be split into a separate PR.  You don't have to do it here, though. I'm curious whether this might have caused issues previously if `chat_template_kwargs` was not forwarded to sglang?

guapisolo: I think so, will give a more concrete detail

guapisolo: This is not specific to `{tool, user}`, but the DSV4 session path requires it. 

SGLang needs `chat_template_kwargs["thinking"]` to configure its `deepseek-v4` response parser. Without it, reasoning is returned inside `content` instead of `reasoning_content`, corrupting subsequent session turns. This was also a latent issue for the existing DSV4 `{tool}` path.

But other model's reasoning parsers don't care about the input chat template kwargs. So there is no bug.

miles/utils/chat_template_utils/tito_tokenizer.py
guapisolo: good point, let me remove the cmt

PR #1600, opened 2026-07-08
refactor(chat-template): merge the DeepSeek V3.2/V4 bridges into one dispatch module

guapisolo:
## Summary
One `deepseek.py` bridge replaces the V3.2/V4 duplicates; `add_generation_prompt` is honored.

## Motivation
The two family bridges were byte-identical except the encoder module, so every shared fix had to be pasted twice. The bridge also silently dropped `add_generation_prompt`, forcing TITO callers to hand-strip the encoder's auto opener. Default render bytes stay parity-locked against sglang.

## Before / After
- **Before / After:** two family modules plus two copy-pasted `template.py` branches become one `model_type()` guard plus `apply_chat_template` dispatch.
- **What moved where:** each family is a `DeepSeekFamily` instance (`V32` / `V4`) wrapping its `encoding_dsv*` module.
- **What moved where:** the known-kwargs whitelist is gone; the encoder validates its own kwargs.
- **What moved where:** `render_messages` strips the user-tail auto opener when `add_generation_prompt=False`.

## Behavior Preservation
- **How we know:** per-family parity tests lock default-path renders byte-for-byte against sglang `encode_messages`.
- **How we know:** two deliberate deltas are test-pinned instead: `add_generation_prompt=False` strips the opener (previously ignored); unknown kwargs surface encoder errors, not a miles-side whitelist.

## Verification
- **Existing test suite:** `tests/fast/utils/chat_template_utils` — 1448 passed.
- Stage-2 verifier deepseekv4 `{tool}`: 26/26, both thinking modes.

## Review Focus
- Scrutinize `DeepSeekFamily._build_encode_config`: alias precedence `thinking_mode` > `enable_thinking` > `thinking` must consume losing aliases.
- Scrutinize the `_GENERATION_OPENERS` strip in `render_messages` against the encoders' user-tail openers.


yueming-yuan (approved):

Shi-Dong (approved):
LGTM!

PR #1602, opened 2026-07-09
(4/N) fix(rollout): drop the per-session /health probe that stalls session creation under load

guapisolo:
## What happened

A production sync rollout froze at 72/128 for 30+ minutes. The training log filled with a `/health` 503 storm — **7,808** × `Server error '503 Service Unavailable' for url 'http://<rollout_ip>:30002/health'`, **7,681** × `retrying...`, and **128** × `Failed to get session server health` (one per pending session) — while the inference engine sat nearly idle (`#running-req: 4-5`).

The verified chain:

1. At a trajectory-completion wave, `GET /sessions/{id}` renders GiB-scale full-records bodies (accumulated R3 replay payloads) as one synchronous `json.dumps` on the owning worker's single event loop, blocking it well past 5s (measured 5–23s per session at 30k–120k-token shapes).
2. The router's `/health` pings every worker under `asyncio.wait_for(..., _HEALTH_TIMEOUT=5.0)` and treats a timeout the same as worker death, so `/health` flips to 503 while every worker is alive — all-or-nothing across N workers.
3. Every new session's tracer probed `/health` before `POST /sessions`, through `post()` which blind-retries any error 60×1s. Each failed attempt also burns the router-side 5s hold, so one probe ≈ 60 × 6s ≈ **6 minutes**. A 128-session over-sampling wave hit this simultaneously (first probe 09:31:21, retries exhausted 09:37:30 — the arithmetic matches exactly), and the rollout never recovered until SIGTERM.

## Why the probe can simply be deleted

The probe existed only to read back `session_server_instance_id`. That id is minted **driver-side** in `start_session_server` (`router_manager.py`) on the very args Namespace the generate path later receives — the server merely echoes it on `/health`, and the probe then wrote the fetched value back onto the same args (a no-op). It is a provably redundant round-trip on the hot path, and it was already fail-open (create proceeds with `None` once retries exhaust).

## Change

- `OpenAIEndpointTracer.create` reads `getattr(args, "session_server_instance_id", None)`; the rollout path no longer issues any `/health` request.
- The `with_session_server` test fixture now mirrors production minting: the id is set on the caller's args. (Previously it minted the id only on the server's private namespace, so the identity-forwarding test passed only via the `/health` echo — a fixture/production wiring divergence.)
- Unit tests pin the new contract: `create()` issues only `POST /sessions`; id sourced from args; `None` when absent.

Deliberately unchanged: `/health` itself (its semantics and the supervisor's startup readiness polling), worker-death fail-fast, and the GiB records-GET path (its cost is tracked separately as the in-worker sample-assembly proposal).

## Verification

- **A/B under a reproduced jam** (session-server bench: 16 sessions × 50 turns, GiB full-records GETs, 8 workers): the old probe path took **359.22s** per create with 60 health retries — reproducing the incident's 6-minute arithmetic exactly; the fixed path took **0.01s** with zero `/health` requests, even while two workers were pinned at 100% CPU writing GiB reply frames.
- The incident's log signatures become structurally impossible from the rollout path: no `GET /health` is issued (pinned by unit test), and the `Failed to get session server health` log line no longer exists in the code.
- `tests/fast/rollout/generate_utils/test_openai_endpoint_utils.py`: 23 passed on this branch. The full `test_multi_turn` + `tests/fast/router` suites (144 passed, 12 skipped) were verified with this same change in the session-server branch environment; on this base my local env has a pre-existing incompatibility in the multi_turn fixtures (fails identically with and without this change), so CI is the gate for those here.

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -129,17 +129,22 @@ def start_session_server(args):
     # Spawn all children before waiting on any: each child pays the ~10s
     # transformers import, so N servers start in ~one import of wall-time.
     # spawn (not fork): see start_router for rationale.
+    instance_ids: dict[int, str] = {}
     processes = []
     for port in ports:
         child_args = copy.copy(args)
         child_args.session_server_port = port
         child_args.session_server_instance_id = uuid.uuid4().hex
+        instance_ids[port] = child_args.session_server_instance_id
         process = multiprocessing.get_context("spawn").Process(
             target=run_session_server, args=(child_args, router_url)
         )
         process.daemon = True
         process.start()
         processes.append((port, process))
+    # The per-port map OpenAIEndpointTracer.create reads instance ids from,
+    # replacing the per-session /health probe.
+    args.session_server_instance_ids = instance_ids
     for port, process in processes:
         wait_for_server_ready(ip, port, process, timeout=30)
     logger.info(f"Session servers launched at {ip}, ports {ports} ({len(ports)} instances)")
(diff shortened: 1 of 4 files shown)


Shi-Dong (approved):
LGTM!

PR #1604, opened 2026-07-09
[AMD] Fix HF to torch_dist conversion segfault during checkpoint save

JessicaJiang-123:
Co-authored-with: @XinyuJiangCMU

## Problem

On ROCm, `convert_hf_to_torch_dist.py` segfaults during checkpoint save. Miles already has a workaround (`ROCmFileSystemWriterAsync`, which forces `non_blocking=False`), but it was only patched into the `filesystem_async` namespace, while the writer is actually instantiated in `strategies/torch.py`, which holds its own `from .filesystem_async import FileSystemWriterAsync` binding.

## Fix

Patch the `strategies/torch` namespace too, so the writer it instantiates uses `ROCmFileSystemWriterAsync`.


guapisolo (approved):
LGTM.

PR #1610, opened 2026-07-10
Rocm/Disabling fused MLA + aiter MoE for GLM4.7

sreerohi:
Disabling fused MLA + aiter MoE for GLM4.7 on rocm .

Also changing the reasoning parser mismatch threshold to 0.4 

Supersedes #1240


guapisolo:
Could you just use os.env["xxx"] to set env for rocm

sreerohi:
> Could you just use os.env["xxx"] to set env for rocm

Verified flags correctly reach  runtime-env-json and ray workers with this mechanism. Updated the files.

PR #1618, opened 2026-07-10
rocm: fix hipBLASLt BGRADB wgrad failure on gfx950

sreerohi:
The bridge provider defaults gradient_accumulation_fusion to True (via can_enable_gradient_accumulation_fusion) and ignores --no-gradient-accumulation-fusion from the CLI. The fix propagates the flag from Megatron args to the bridge provider.

Context:
hipBLASLt on gfx950 (MI350/MI355) has no algorithm for the triple combination of bf16 output with fp32 accumulate + HIPBLASLT_EPILOGUE_BGRADB epilogue + accumulate=True. This fires during TE's LayerNormLinear backward when gradient_accumulation_fusion=True and the layer has bias. So we don't use gradient_accumulation_fusion and need this change for that to take effect.

Supersedes #1153


guapisolo (approved):
LGTM.

PR #1628, opened 2026-07-11
feat(tito): add DeepSeek V3.2 speculative session verification

guapisolo:
## Summary
DeepSeek V3.2 TITO session verification plus a vendored encoder that keeps historical thinking.

## Motivation
Add a manual TP8/EP8 session test and expose the fixed EAGLE serving recipe through ModelConfig. Keep SGLang TP/DP/PP/EP namespace fields on Miles' short-name contract, and forward the effective V3.2 thinking mode so SGLang separates reasoning before TITO comparison.

Upstream `encoding_dsv32` gates every thinking block on `last_user_idx`: a new user turn strips historical think and rewrites the rendered prefix, which is why V3.2 registered only the `{tool}` surface. Vendoring the encoder lets Miles fix this at the render level without forking sglang.

## Usage
```bash
DEEPSEEK_V32_MODEL=/cluster-storage/models/deepseek-ai/DeepSeek-V3.2 python -m pytest tests/manual/tito/deepseek_v32.py -q -s
```

The test launches one session verifier with `tp=8`, `ep=8`, and EAGLE steps/top-k/draft-tokens set to `2/1/3`.

## Design Notes
- **Key choices:** `encoding_dsv32` is vendored verbatim under `miles/utils/chat_template_utils/templates/` (provenance header only, formatter-excluded); V4 stays on sglang's `encoding_dsv4`.
- **Key choices:** the `protocol.Tool` seam stays on sglang so tool schemas keep serializing with the server's defaults.
- **Key choices:** the vendored dsv32 mirrors dsv4's gates — `drop_thinking=False` renders every thinking block position-independently while `drop_thinking=True` stays byte-identical to upstream.
- **Key choices:** every V3.2 surface pins `drop_thinking=False`, which legalizes the `{tool, user}` surface.
- **Key choices:** `DeepSeekV32TITOTokenizer` forwards the resolved thinking mode as `chat_template_kwargs["thinking"]`; `add_sglang_arguments` maps parallel-size aliases to Miles' short Namespace fields.
- **Alternatives considered:** preserve-think via bridge-side message preprocessing — rejected because the cut happens in the encoder's render gates, not in message content.

## Verification
- **Tests added:** upstream-parity corpus plus raise parity pin every `drop_thinking=True` path to the installed sglang encoder.
- **Tests added:** append-only byte tests cover user appends, the tool-tail flip shape, plus tool-loop drop-mode equivalence.
- **Tests added:** `test_deepseek_v32_forwards_effective_thinking_mode`, `test_sglang_parallel_sizes_use_short_namespace_fields`, `test_namespace_to_train_args_enables_eagle_speculative_decoding`.
- **Existing test suite:** 2566 fast tests pass on the rebased branch.
- **Manual E2E:** V3.2 without speculative decoding plus V4 with EAGLE both completed 64/64 samples with zero TITO mismatches.
- **Known blocker:** V3.2 EAGLE stops in SGLang's NextN loader because `DeepseekV3ForCausalLMNextN` lacks `fuse_qkv_a_proj`.

## Review Focus
- Scrutinize the vendored dsv32 gate rewrites against upstream: header, four gate sites, `encode_messages` pass-through.
- Scrutinize `DeepSeekV32TITOTokenizer` effective-mode precedence.
- Scrutinize `namespace_to_train_args` fixed EAGLE option serialization.


tests/fast/utils/chat_template_utils/test_deepseek_v32.py
```diff
@@ -159,6 +159,120 @@ def test_thinking_mode_changes_output():
...
+    "developer": [{"role": "developer", "content": "do the thing"}],
+    "thinking_history": _THINKING_HISTORY,
+    "tool_tail": _TOOL_TAIL_HISTORY,
+}
+
+
+@pytest.mark.parametrize("scenario", list(_UPSTREAM_PARITY_SCENARIOS), ids=list(_UPSTREAM_PARITY_SCENARIOS))
+@pytest.mark.parametrize("thinking", [False, True], ids=["chat", "thinking"])
+def test_drop_thinking_true_matches_upstream_sglang(scenario, thinking):
+    from sglang.srt.entrypoints.openai import encoding_dsv32 as upstream
```
Shi-Dong: please hoist this import to the beginning of the script

tests/fast/utils/chat_template_utils/test_deepseek_v32.py
```diff
@@ -159,6 +159,120 @@ def test_thinking_mode_changes_output():
...
+    thinking_mode = "thinking" if thinking else "chat"
+    assert _reference_encode(messages, thinking=thinking, drop_thinking=True) == upstream.encode_messages(
+        messages, thinking_mode=thinking_mode, drop_thinking=True
+    )
+
+
+def test_drop_thinking_true_raise_parity_with_upstream():
+    # thinking mode + a post-last-user assistant without reasoning_content or
+    # tool_calls raises upstream; the vendored copy keeps that contract.
+    from sglang.srt.entrypoints.openai import encoding_dsv32 as upstream
```
Shi-Dong: please hoist this import to the beginning of the script

tests/fast/utils/chat_template_utils/test_deepseek_v32.py
```diff
@@ -159,6 +159,120 @@ def test_thinking_mode_changes_output():
...
+
+    from miles.utils.chat_template_utils.templates import encoding_dsv32 as vendored
```
Shi-Dong: please hoist this import to the beginning of the script

miles/utils/test_utils/session_verify_runner.py
```diff
@@ -166,8 +167,17 @@ def namespace_to_train_args(ns: argparse.Namespace) -> str:
...
+        parts.append(f"--sglang-expert-parallel-size {ns.sglang_ep_size}")
+    if ns.enable_spec:
+        parts.extend(
+            [
+                "--sglang-speculative-algorithm EAGLE",
+                "--sglang-speculative-num-steps 2",
+                "--sglang-speculative-eagle-topk 1",
+                "--sglang-speculative-num-draft-tokens 3",
+            ]
+        )
```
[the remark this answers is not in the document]
guapisolo: This runner intentionally exposes one fixed EAGLE recipe behind `enable_spec`. Its serializer requires a fully-shaped Namespace and deliberately avoids serializer-level defaults. Adding `getattr` fallbacks or per-model knobs without a current second recipe would broaden the harness contract, so we’re keeping the fixed 2/1/3 values.

miles/utils/chat_template_utils/templates/encoding_dsv4.py
Shi-Dong, line 1: Do we have to keep a chat template for DeepSeek v4 in Miles as well? My understanding is that the cut-thinking behavior only exists in the v3.2 chat template, doesn't it?

miles/utils/chat_template_utils/templates/encoding_dsv32.py
Shi-Dong, line 1: The PR description says that this file is vendored "verbatim" but it doesn't seem so. Perhaps you'd want to update the PR description.

tests/manual/tito/deepseek_v32.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+from tests.e2e.sglang.test_session_server_multi_role._common import ModelConfig, run_one
+
+CONFIG = ModelConfig(
+    model_name=os.environ.get("DEEPSEEK_V32_MODEL", "deepseek-ai/DeepSeek-V3.2"),
+    reasoning_parser="deepseek-v3",
+    tool_call_parser="deepseekv32",
+    tito_model="deepseekv32",
+    allowed_append_roles=("tool",),
```
Shi-Dong, line 10: Aren't we supposed to test `("tool", "user")` here?

miles/utils/chat_template_utils/templates/encoding_dsv4.py
guapisolo: Yes.. I also import v4 for potential usage. 🤔

guapisolo: okay let me remove v4. cuz people actually dont need to change v4

tests/manual/tito/deepseek_v32.py
guapisolo: this is modified in #1777

miles/utils/chat_template_utils/templates/encoding_dsv32.py
guapisolo: no in this pr it's the same to keep consistancy behavior with existing code. template change happens in #1778

tests/manual/tito/deepseek_v32.py
```diff
@@ -0,0 +1,23 @@
...
+    num_gpus=8,
+    tp_size=8,
+    ep_size=8,
+    enable_spec=True,
```
Shi-Dong, line 14: In the PR description it says that EAGLE fails with Dpsk v3.2?  Why is speculative decoding enabled in this test?

Shi-Dong (approved):
Overall LGTM. Added one comment and the PR description might need updates.

tests/manual/tito/deepseek_v32.py
guapisolo: Enable spec to speed up the test, it's rollout-only and can speed up a lot.
There is actually some issue for dsv32 + mtp. So I naively skip it.

PR #1632, opened 2026-07-11
fix(ci): separate image and nightly FT tag scopes

guapisolo:
## Summary
Apply distinct FT tag scopes to image, nightly, and all CI.

## Symptom & Reproduction
- **Symptom:** Once PR #1622 enables `ft-short`, every broad CI trigger still expands to the same unrestricted test-label scope.
- **Reproduction:** Run `.venv/bin/python -m pytest -q --confcutdir=tests/ci tests/ci/test/test_run_suite.py -k BroadScopes` against the added scope cases.

## Root Cause
1. `.github/workflows/pr-test.yml` maps every broad trigger to `--match-all-labels`.
2. `filter_tests` cannot exclude scope-specific labels after broad inclusion.

## Fix
Add `--exclude-labels` as a final filter. Image passes `ft-short ft-long`, nightly passes `ft-long`, and all passes no exclusions; the workflow applies `run-ci-all` > nightly > `run-ci-image` when labels overlap.

## Verification
- **New / updated test:** `test_image_scope_excludes_all_ft_labels` covers image scope.
- **New / updated test:** `test_nightly_scope_includes_ft_short_and_excludes_ft_long` covers nightly scope.
- **New / updated test:** `test_all_scope_includes_every_label` covers all scope.
- **Workflow lint:** `actionlint` accepts all seven updated stage expressions.

## Review Focus
- Scrutinize the exclusion ordering in `filter_tests`.
- Scrutinize scope precedence in `.github/workflows/pr-test.yml`.


tests/ci/run_suite.py
```diff
@@ -84,23 +84,28 @@ def filter_tests(
...
     * `match_all_labels=False` (default): include only tests where
       `not test.labels or (set(test.labels) & labels)`. `labels` here is
       the already-stripped domain-label set produced by
       `strip_run_ci_prefix`. A test registered with `labels=[]` (or
       omitted) is treated as always-run: it survives an empty PR-label
       set; a test with non-empty `labels` survives only when its labels
       intersect the PR-supplied set.
+    * `exclude_labels`: after either inclusion mode, remove tests carrying any
+      excluded label. This lets broad scopes omit a specific expensive class;
+      an exclusion always wins over inclusion.
```
yushengsu-thu, line 108: "An exclusion always wins over inclusion" has one surprising consequence: a PR carrying `run-ci-image` **and** `run-ci-ft-short` — someone explicitly asking for FT coverage on an image bump — gets ft-short silently stripped (match-all mode discards the `--labels` set, then the exclusion removes it). Consider exempting labels that were explicitly present in the stripped `--labels` set from exclusion, or at least printing a warning when an excluded label was also explicitly requested, and documenting the interaction in `01-label.md`.

tests/ci/run_suite.py
```diff
@@ -113,6 +118,9 @@ def filter_tests(
         label_set: set[str] = labels or set()
         ci_tests = [t for t in ci_tests if not t.labels or (set(t.labels) & label_set)]
 
+    excluded_label_set = exclude_labels or set()
+    ci_tests = [t for t in ci_tests if not (set(t.labels) & excluded_label_set)]
```
yushengsu-thu, line 122: Two small things here:

1. `exclude_labels` values are never validated against `KNOWN_LABELS`. A typo in the workflow YAML (`ft_long`, `ft-lonng`) excludes nothing, silently — the failure mode is expensive soak tests quietly running on every image bump (or the reverse). `strip_run_ci_prefix` already warns on malformed `--labels` input; a one-line unknown-label check here would match that defensive style.
2. (nit) Guarding with `if excluded_label_set:` skips the per-test `set()` construction on the majority of runs that have no exclusions, and — worth a docstring sentence either way — the filter also removes matching **disabled** tests from `skipped_tests`, so excluded-and-disabled tests vanish from the skip report entirely.

Also: the three new tests all use `match_all_labels=True`; the docstring promises the exclusion applies "after either inclusion mode", but the `match_all_labels=False` + exclusion path is untested.

.github/workflows/pr-test.yml
```diff
@@ -97,6 +100,7 @@ jobs:
         --auto-partition-id ${{ matrix.partition_id }} --auto-partition-size 4
         --labels ${{ join(github.event.pull_request.labels.*.name, ' ') }}
         ${{ (github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'run-ci-image') || contains(github.event.pull_request.labels.*.name, 'run-ci-all') || (github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'nightly'))) && '--match-all-labels' || '' }}
+        ${{ ((github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'nightly')) && !contains(github.event.pull_request.labels.*.name, 'run-ci-all') && '--exclude-labels ft-long') || (contains(github.event.pull_request.labels.*.name, 'run-ci-image') && !contains(github.event.pull_request.labels.*.name, 'run-ci-all') && github.event_name != 'schedule' && !contains(github.event.pull_request.labels.*.name, 'nightly') && '--exclude-labels ft-short ft-long') || '' }}
```
yushengsu-thu, line 103: This ~470-char expression is now copy-pasted 7 times — this PR is itself evidence of the cost (7 near-identical line edits), and a one-stage drift would be invisible until a nightly quietly runs the wrong test set. The precedence logic here is also the riskiest artifact of the PR, and it's only covered by actionlint (syntax, not behavior) — the new unit tests verify `filter_tests` given already-correct flags, not that the workflow emits those flags per trigger.

Suggested follow-up: compute a `scope-flags` output once in a preliminary job (`resolve-ci-image` is a natural home) and reference it per stage — or better, pass the raw event name + labels into `run_suite.py` and compute the scope in Python where it's unit-testable.

(nit: clause 2's `github.event_name != 'schedule' && !contains(..., 'nightly')` guards are unreachable — any case where they'd matter is already consumed by clause 1 or killed by `!contains(run-ci-all)` — so they could be dropped to lighten the line.)

yushengsu-thu (approved):
Traced the new scope expression through every trigger/label combination (schedule, workflow_dispatch, and PRs carrying nightly / run-ci-image / run-ci-all in all overlaps) — the claimed precedence run-ci-all > nightly > run-ci-image holds, and the Python side is safe (labels=None is normalized to [] by ci_register, so the unconditional exclusion filter can't crash). Docs and tests are updated consistently.

Approving — nothing blocking. Three non-blocking issues inline: (1) an exclusion silently overrides an explicitly requested run-ci-ft-short label, (2) --exclude-labels values aren't validated against KNOWN_LABELS so a YAML typo fails silently, (3) the 7x-duplicated scope expression is a maintenance trap and its precedence logic is only covered by actionlint, not by the unit tests.

tests/ci/run_suite.py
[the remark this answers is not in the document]
guapisolo: Adopted in the follow-up PR #1644 (`perf(ci): short-circuit the exclusion filter when nothing is excluded`): the filter now uses `any(lbl in exclude_labels for lbl in t.labels)` and is skipped entirely when no exclusions resolve. Note the surrounding code changed shape there — `--exclude-labels` moved off the CLI into `resolve_scope`.

PR #1635, opened 2026-07-12
fix(cp): inherit requires_grad in empty-response redistribute fill

EazyReal:
## Summary

Make empty-response context-parallel fill tensors inherit requires_grad from the redistributed input.

## Why

The detached fill tensor can break backward when a context-parallel shard contains no response tokens. This matches the Slime fix from THUDM/slime#2123.

## Testing

`pytest -q tests/fast/backends/training_utils/test_ulysses_cp_utils.py`

### DIFF
--- a/miles/backends/training_utils/cp_utils.py
+++ b/miles/backends/training_utils/cp_utils.py
@@ -367,7 +367,7 @@ def allgather_cp_redistribute(
                     response_length,
                     dtype=value.dtype,
                     device=value.device,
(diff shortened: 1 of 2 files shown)


EazyReal:
Closing. The empty-response CP fill already uses `requires_grad=True` in `allgather_cp_redistribute` (GLM-5, Feb 2026). Inherit-from-input vs True isn't worth keeping this open.

PR #1637, opened 2026-07-12
feat(rm): allow per-sample custom_rm_path override

EazyReal:
## Summary

Allow each sample metadata record to override the configured custom reward function path, including batched reward dispatch.

## Why

Mixed-task batches need to route samples to different reward functions without splitting rollout jobs. Override samples use the single-sample reward contract; unoverridden samples remain batched through the configured global function. Samples without an override keep the existing behavior. This ports the Slime configuration behavior from THUDM/slime#2081.

## Testing

`pytest -q tests/fast/rollout/rm_hub/test_rm_hub.py`

### DIFF
--- a/miles/rollout/generate_utils/sample_utils.py
+++ b/miles/rollout/generate_utils/sample_utils.py
@@ -126,6 +126,7 @@ def _merge_sample_pair(a: Sample, b: Sample, tokenizer) -> Sample:
             status=b.status,
             metadata=_merge_metadata(),
             generate_function_path=_merge_equal_value("generate_function_path"),
+            custom_rm_path=_merge_equal_value("custom_rm_path"),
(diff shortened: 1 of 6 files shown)


EazyReal:
Closing. Per-sample RM dispatch landed as `Sample.reward_spec.custom_rm_path` in #1746 (multi-LoRA). That's the right type; this PR's metadata override isn't the current contract.

This had been open since Jul 12. A pointer that #1746 covered the per-sample RM path would have helped me drop the metadata version instead of leaving it conflicted. If mixed-task batches still need something beyond RewardSpec, that's a new PR on that type — happy to write it if it's wanted.

A one-line "landed in #N / still wanted / not in scope" on overlapping community PRs is enough to stay motivated to keep contributing.

PR #1638, opened 2026-07-12
[lora] feat: multi lora async (full PR, see stacked diff PRs instead)

mathewjhan:
# Summary

Allow disagg fully async training on multiple loras (all-linear, excluding expert) per step in Miles using megatron-bridge.

Two LoRAs trained together (DAPO + GSM8K)

image

image

# Feature

- Train multiple LoRAs in a single training step
- Use as a long running service, with `in_place` pause generation and LoRA upsert updates
- Use as normal training job, stopping when no more LoRAs are left to train
- FastAPI control plane to register and deregister adapter runs

# Dependent PRs

## SGLang

- `rid` prefix abort: https://github.com/sgl-project/sglang/pull/30912
- upsert LoRA loading: https://github.com/sgl-project/sglang/pull/30913

Unified above: https://github.com/sgl-project/sglang/pull/31253 (for testing - do not need merge)

## Megatron-Bridge (radixark)

- multi lora support: https://github.com/radixark/Megatron-Bridge/pull/20

# Dev setup

Tested image:  `radixark/miles:dev-202607090055`

You need to install the 3 forks:

## SGLang

This fork has two small changes to SGLang (has both PRs):

- Upsert semantics for loading LoRAs with the same lora name, to avoid having to unload; currently, unloading requires there to be zero requests for that LoRA to be loaded.  When running pause generation before unload, we cannot complete more requests, when there are still running requests on unload, it is impossible to unload the adapter and miles will hang
- Abort by rid prefix; this is useful to target specific requests under a "namespace", such as when a LoRA run has been deregistered from miles and we want to abort all of the requests for that LoRA adapter. The alternative is to track all requests for that LoRA adapter individually and send in individual abort requests, since SGLang currently doesn't support batch abort

```
git clone https://github.com/mathewjhan/sglang
cd sglang
git checkout sglang-miles-mathew-dev-2
uv pip install --system --break-system-packages --no-deps -e python
```

## Megatron-Bridge 

This fork adds a multi-lora support to Megatron-Bridge along with some helper methods/user api
 
```
git clone https://github.com/mathewjhan/Megatron-Bridge
cd Megatron-Bridge
git checkout radixark/bridge/multilora-2
uv pip install --system --break-system-packages --no-deps -e .
```

## Miles

```
git clone https://github.com/mathewjhan/miles
cd miles
git checkout feat/multi-lora-async
uv pip install --system --break-system-packages --no-deps -e .
```

# Running

see: `examples/multi_lora`

For normal training (not as a service):

1. run `examples/multi_lora/provision.sh`
2. configure W&B credentials and settings in `examples/multi_lora/run_job.sh`
3. run `examples/multi_lora/run_job.sh |& tee run.log`

For multi-lora training as a long running service:

1. run `examples/multi_lora/provision.sh`
2. configure W&B credentials and settings in `examples/multi_lora/run_service.sh`
3. run `examples/multi_lora/run_service.sh |& tee run.log` in one shell
4. wait for the service to be ready ("No adapters; sleeping for 5.0s...")
5. optionally run the smoke test in another shell

```
python3 examples/multi_lora/service_smoke.py \
      --api-url http://127.0.0.1:8068 \
      --data /root/gsm8k/train.parquet \
      --input-key messages \
      --label-key label \
      --rm-type math \
      --steps 5
```

# Minor features added to existing code (backwards compatible)

- Add `AdapterRef` and `RewardSpec` to `Sample` type so individual samples can access their own reward functions and adapter names during rollout

# Missing features for future

- Colocated + update_weight_from_tensor currently unsupported, only update_weight_from_distributed
- Per adapter batch size currently unsupported, all adapters stepped at once (will be added later when decoupling the optimizer)
- Currently doesn't checkpoint optimizer + scheduler state yet, but can be added later as future PR
- Dataset checkpoint loading per adapter
- MultiLoRA not applied to experts as of now due to more complex bookkeeping required (need to keep track of the [adapter index, routed experts] together for grouped gemm)
- Doesn't support pure LoRA load from \*.bin/\*.safetensors yet, can be added later as a future PR, only resume from a megatron checkpoint or train from scratch
- Support add_samples to recycle stale/aborted samples


miles/backends/megatron_utils/actor.py
```diff
@@ -513,12 +518,61 @@ def train_actor(
...
+            dist.broadcast_object_list(broadcast_buffer, src=0, group=get_gloo_group())
+        snapshot = broadcast_buffer[0]
+        should_be_loaded = {**snapshot["active"], **snapshot["pending"], **snapshot["retiring"]}
+        cleanup_names = set(snapshot["cleanup"])
+
+        loaded_names = set(self.loaded_adapters)
+        adapters_to_load = [adapter for name, adapter in should_be_loaded.items() if name not in loaded_names]
+        adapters_to_clean_up = [
+            self.loaded_adapters[n] for n in loaded_names if n in cleanup_names or n not in should_be_loaded
+        ]
```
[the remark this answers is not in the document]
mathewjhan: resolved

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,244 @@
...
+    data = sorted(data, key=lambda g: first_sample(g).index)
+
+    batch_adapters = sorted({first_sample(g).adapter.name for g in data if g and first_sample(g).adapter})
+    if batch_adapters:
+        await get_multi_lora_controller().record_batch_adapters.remote(rollout_id, batch_adapters)
+
+    if (x := args.rollout_sample_filter_path) is not None:
+        load_function(x)(args, data)
+
+    await recompute_samples_rollout_logprobs_via_prefill(
```
yushengsu-thu, line 541: **Logprob recompute is broken in multi-LoRA mode: it always sends the single-adapter name.** `_build_prefill_scoring_payload` (miles/rollout/generate_utils/prefill_logprobs.py:40-41) keys LoRA solely on `is_lora_enabled(args)` — which multi-LoRA forces true — and unconditionally sets `lora_path=LORA_ADAPTER_NAME` (`"miles_lora"`). In multi-LoRA mode only `__miles_slot_{N}` names are ever registered on the engines, so this call fails every step with `The following requested LoRA adapters are not loaded: miles_lora`. And even if the name resolved, the batched variant applies `payloads[0]`'s single `lora_path` to samples from *different* adapters — silently wrong logprobs / importance weights for a mixed-adapter batch. The payload builder needs to take the per-sample adapter (`sample.adapter.slot → __miles_slot_{N}`), and the batched path must group by adapter.

miles/utils/multi_lora.py
```diff
@@ -0,0 +1,409 @@
...
+        ):
+            try:
+                resp = await self.client.get(f"{self.router_url}{endpoint}")
+                if resp.status_code == 200:
+                    return extract(resp.json())
+            except Exception:
+                continue
+        return []
+
+    async def abort_adapter_requests(self, adapter_name: str) -> None:
```
yushengsu-thu: **One-shot abort + drain-free upsert + slot reuse = a retired adapter's surviving requests decode under the NEXT adapter's weights.** The abort here fires exactly once, at the RETIRING→CLEANUP flip, warn-only on failure, and is never repeated or verified. Two ways requests survive it: (a) a multi-turn group that is between turns POSTs its next turn (fresh rid `A::uuid`) right after the single abort round; (b) sglang-side, a request admitted to the tokenizer but not yet forwarded to the scheduler passes the abort unmatched (filed on sgl-project#30912). The same reconcile then frees slot N; when a *different* adapter B registers into slot N, the next update_weights upserts B's rank/alpha/weights into the **same** sglang lora_id (`__miles_slot_N`) with no request drain (sglang upsert doesn't wait for in-flight requests — filed on sgl-project#30913). A's orphan requests then continue decoding under B's weights for up to max_new_tokens — no error anywhere, just corrupted rollout groups. Cheap hardening on the miles side: re-run the prefix abort at CLEANUP time (second round in `_deregister_adapter`), reject rollout submissions for non-sampleable adapters at the worker, and/or verify zero in-flight rids for the namespace before `free_slot()`.

yushengsu-thu (commented):
Ran a deep verification pass over this branch together with its two sglang dependency PRs (sgl-project#30912/#30913); every finding below was adversarially re-verified against the code. Two issues to flag on the miles side — one is a hard crash/correctness bug in the logprob-recompute path, the other is a lifecycle race that silently corrupts rollout data across adapter slot reuse. (Side note: the deeper sglang-side findings are filed on the two dependency PRs directly.)

miles/rollout/multi_lora/async_rollout.py
yushengsu-thu: Fixed in mathewjhan/miles@8bdd7d49fb57a94407f127d38f35083b82218b31: `_build_prefill_scoring_payload` now resolves the LoRA name per sample — adapter-tagged samples score under their own `__miles_slot_{N}` (via a new shared `slot_lora_name()` helper, also used by the rollout payload and the weight-push mixin), single-adapter LoRA keeps `miles_lora`, base models send none. The batched path groups by `(logprob_start_len, lora_path)` and `_build_batch_prefill_scoring_payload` rejects mixed-adapter payloads outright.

Tests added in `tests/fast/rollout/generate_utils/test_prefill_logprobs.py`: per-sample slot resolution, per-adapter batch grouping, and mixed-batch rejection (all passing).

miles/utils/multi_lora.py
yushengsu-thu: Hardened in mathewjhan/miles@1e93ea40930be334f136641241baa5d3a75d44fd:

1. `MultiLoRABackend.free_slot()` re-runs the prefix abort right before the slot is released for reuse (a second round, a full train step after the RETIRING→CLEANUP one), and the controller now routes `free_slot` through the backend;
2. `sglang_rollout.generate` refuses to POST for an adapter that is no longer sampleable and aborts the sample locally — closing escape (a), the between-turns next-turn POST;
3. escape (b) (the engine tokenizer window) is fixed engine-side in sgl-project/sglang#30912.

I skipped the "verify zero in-flight rids before free_slot" idea: sglang exposes no cheap way to enumerate live rids across workers. As @mathewjhan noted, the batch-time collection filters (`adapter_name not in current_adapters` + slot-version staleness) remain the final backstop keeping any survivor out of training data — these changes shrink the window and stop wasting decode on orphans rather than replace that backstop.

Tests added for the free_slot second abort round in `tests/fast/utils/test_controller_backend.py` (passing).

yushengsu-thu:
@mathewjhan `examples/multi_lora/tests/test_multi_lora_async_rollout.py::test_process_group_recycles_aborted` fails on this branch (pre-existing at e371b2e0, unrelated to the M1/M2 fixes): the test expects aborted groups to be re-queued, but the code path documents that re-queuing is not wired up. Which direction do you want — wire up re-queuing for aborted groups, or adjust the test to pin the current drop-and-count behavior? I'm happy to pick it up once the direction is set.

yushengsu-thu:
Thanks for confirming — marked `test_process_group_recycles_aborted` as `xfail(strict=True)` in 50f6ff14 so the suite is green on this PR while still pinning the intended end-state behavior (wiring up re-queuing will XPASS and force the marker off). Two things worth bundling into that future PR: (1) the re-queuing itself (per-adapter source needs a write path), and (2) aborted/stale-dropped rows currently still count toward `num_row`, so adapters quietly train on fewer samples than configured — probably the more impactful half. Also worth skipping re-queue for RETIRING adapters (their slot is being freed; nothing can serve the regenerated group).

yushengsu-thu:
Merged `radixark/main` into this branch to resolve the conflict (the branch was 36 commits behind). Single conflict in `miles/rollout/sglang_rollout.py` — an import-line collision between the multi-LoRA imports and main's new `call_agent_abort_hook`; resolved as the union. `compileall` clean; PR is now MERGEABLE.

Updated with [Claude Code](https://claude.com/claude-code).

yushengsu-thu:
Note on Muon support (currently rejected at launch for multi-LoRA): it would be a small change (~50 LOC). Megatron's Muon path mirrors `build_multi_lora_optimizer` exactly — same freeze-others param-group trick, same bf16/LayerWise delayed-master-weight dance, same `use_distributed_optimizer=False` + no-fp16 constraints — so per-slot Muon children slot straight into the existing LayerWise assembly.

What the ~50 LOC covers (the three Adam-bound spots):

1. `zero_optimizer_state_for_adapter` only scrubs `exp_avg`/`exp_avg_sq` and the Adam step clocks. Muon state is a momentum buffer with different keys, so a retired slot's momentum would silently leak into the next tenant — this is the reason for the hard launch gate rather than a soft warning.
2. `_adam_init_state_fn` eagerly initializes Adam moments; Muon needs its own init fn.
3. Construction: per-slot children currently come from `get_megatron_optimizer` with adam config; Muon goes through the `dist_muon` naming (mcore folded it into `get_megatron_optimizer`, layer-wise via the `dist_` prefix) and produces a different children layout (matrix params on Muon, the rest on Adam), so the slot-to-children mapping needs re-verification.

Everything else is already optimizer-agnostic: `step_adapter_slots` uses base-class `prepare_grads`/`clip`/`step_with_ready_grads`, and the 1/batch_size scaling holds for Muon too.

Caveats: LoRA `lora_B` names don't match the `linear_qkv.weight` check so `muon_split_qkv` wouldn't apply, and Muon's benefit on r×d low-rank matrices is unproven. Deferred until there's an experimental need. Note the current "adam" path is effectively AdamW (`decoupled_weight_decay` defaults to true in mcore).

yushengsu-thu:
until here is good and the following is refactor

examples/multi_lora/README.md
```diff
@@ -0,0 +1,133 @@
+# Multi-LoRA Training Example (fully-async)
```
yushengsu-thu, line 1: All files in in example need to be chekced again after mofiied the codes.

miles/backends/megatron_utils/model.py
```diff
@@ -280,6 +290,7 @@ def forward_step(
                 "response_lengths",
                 "max_seq_lens",
                 "witness_ids",
+                "adapter_slots",
```
yushengsu-thu: will make the code hard to read or it has too strong invasion of add to here. If there's other elegant way?

miles/backends/megatron_utils/model.py
```diff
@@ -291,6 +302,11 @@ def forward_step(
         total_lengths = batch["total_lengths"]
         response_lengths = batch["response_lengths"]
 
+        if "adapter_token_counts" in batch:
```
yushengsu-thu, line 304: do you have this key "adapter_token_counts" in batch. I think you never use this?

miles/backends/megatron_utils/model.py
```diff
@@ -390,12 +417,16 @@ def train_one_step(
...
-    if not disable_optimizer:
-        optimizer.zero_grad()
+    if multi_lora:
+        from miles.backends.megatron_utils.multi_lora_optimizer import reset_grad_metadata_keep_grads
+
+        # Retain accumulated per-adapter gradients; reset only the per-iteration
+        # DDP bookkeeping. Slot grads are zeroed selectively at step time.
+        reset_grad_metadata_keep_grads(model)
+    else:
+        _zero_grads(model, optimizer, disable_optimizer)
```
yushengsu-thu, line 429: Why it doesn't need _zero_grads, but it require now?

yushengsu-thu (approved):

miles/backends/megatron_utils/model.py
yushengsu-thu: Reworked in a1ce9e4d: the callers never actually read batch["adapter_slots"] — the only consumer is get_batch's own token-count computation — so get_batch now fetches the key itself (same pattern as the existing dynamic_global_batch_size self-fetch) and it's gone from both key lists. Non-multi-LoRA runs get None from get_next and everything downstream no-ops.

miles/backends/megatron_utils/model.py
yushengsu-thu: It is used — the producer is get_batch itself: data.py sets batch["adapter_token_counts"] whenever the batch carries adapter_slots (every multi-LoRA micro-batch), and this hook feeds it to set_tokens_per_adapter_slot so the multi-LoRA layers know which token range routes to which slot's A/B. It's needed on both forward paths: here (logprob recompute) and the training forward — without it a mixed-adapter batch would score under the wrong adapters. Normal runs never have the key, so the branch is skipped.

miles/backends/megatron_utils/model.py
yushengsu-thu: Because multi-LoRA accumulates gradients across train calls: a driver batch carries only part of an adapter's global batch, so grads must stay in the DDP buffer until that adapter's batch completes — then step_adapter_slots steps and zeroes only that slot's grad region. Zeroing everything at the top (the normal path) would wipe other adapters' partial accumulation. reset_grad_metadata_keep_grads only resets DDP's per-iteration bookkeeping so the next backward can run; it doesn't touch grad data. The else branch is unchanged from main — _zero_grads is just the old inline 4 lines extracted because they now appear twice.

miles/backends/megatron_utils/model.py
mathewjhan: @yushengsu-thu i was thinking the same thing a while ago too, i think # slots might be available from args as well, but not sure if passing in args is good either for this function

examples/multi_lora/README.md
yushengsu-thu: Reworked the example in 323a9918 to match the repo's current .py-launcher convention (modeled on run_glm5_2_744b_a40b_lora.py): provision.sh / run_job.sh / run_service.sh are replaced by a single run_multi_lora.py with prepare / train / full-train / serve commands via command_utils. Dataset/model paths follow the shared /root/datasets + /root/models layout, the adapter yamls and service_smoke usage are updated to match, and every remaining file in the example dir (README, yamls, smoke client) has been re-checked against the current flags and paths.

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
```diff
@@ -157,6 +157,52 @@ def _update_lora_weight_implementation(self, named_tensors: list[tuple[str, torc
...
+        """Multi-LoRA variant of ``_update_lora_weight_implementation`` (kept
+        separate so the single-adapter path stays untouched; the transport is
+        the same metadata-over-Ray + NCCL broadcast sequence).
+
+        Differences from the single-adapter path: ``lora_name`` is the
+        adapter's slot name (``__miles_slot_{N}``), ``lora_config`` carries the
+        adapter's own ``r`` / ``lora_alpha``, and the RPC always upserts —
+        insert-or-overwrite in place, no unload/register — which is both the
+        first load and the refresh path for the fixed multi-LoRA pool.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
```diff
@@ -157,6 +157,52 @@ def _update_lora_weight_implementation(self, named_tensors: list[tuple[str, torc
...
+                shapes=shapes,
+                group_name=self._group_name,
+                upsert=True,
+            )
+            for engine in self.rollout_engines
+        ]
+        # NCCL broadcast requires contiguous buffers, but slice_lora_to_rank yields
+        # a strided (non-contiguous) view for lora_B (column slice). Materialize
+        # contiguous copies (no-op when already contiguous) and hold the list so
+        # the buffers stay alive until the async broadcasts complete.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
```diff
@@ -263,6 +263,51 @@ def _update_lora_weights(self) -> None:
         self._update_lora_weight_implementation(accumulated_named_tensors)
         self._lora_loaded = True
 
+    def _update_multi_lora_weights(self) -> None:
+        """Push the selected adapters (upsert, never unload). The push set is
+        chosen by the actor before each call and is identical on every rank, so
+        per-adapter TP collectives line up. Version bumps are recorded by the
+        actor after the push succeeds."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
```diff
@@ -263,6 +263,51 @@ def _update_lora_weights(self) -> None:
...
+                )
+
+        if not self._is_lora_source:
+            return
+
+        if not accumulated_named_tensors:
+            raise RuntimeError(
+                f"Multi-LoRA weight sync failed for adapter {adapter.name!r}: the weight iterator "
+                "produced no LoRA weights (no lora_A/lora_B names found). This usually means the "
+                "Megatron-Bridge or SGLang version is incompatible."
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
```diff
@@ -300,32 +345,44 @@ def update_weights(self) -> None:
...
+            is_multi_lora = is_lora and is_multi_lora_enabled(self.args)
+
+            # Base weight sync:
+            #   Full-param RL: base weights change every step -> always sync.
+            #   LoRA RL: the base is frozen and the (remote) rollout engines already
+            #     hold it (they load ``hf_checkpoint`` at init), so there is nothing
+            #     new to push. Skip it -- this mirrors UpdateWeightFromTensor's
+            #     ``skip_base_sync`` for the distributed case, and avoids routing frozen
+            #     base weights (e.g. a VLM vision tower, unsupported by the direct
+            #     convert_to_hf) through the base path.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/actor.py
```diff
@@ -235,6 +236,15 @@ def init(
             is_lora=is_lora_enabled(args),
         )
 
+        # Adapters currently loaded into Megatron slots on this rank.
+        self.loaded_adapters: dict[str, object] = {}
+        # Adapters whose engine-side weights are stale: newly loaded at
+        # reconcile, or stepped by the last train call. Consumed (pushed) by
+        # the next update_weights. Derivable identically on every rank because
+        # the loaded set only mutates at reconcile and the stepped set ships
+        # with the train data.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/actor.py
```diff
@@ -513,12 +523,72 @@ def train_actor(
...
     @with_logs
+    @timer
+    def reconcile_adapters(self) -> None:
+        """Load what the controller wants served, tear down what it retired.
+        The snapshot is read once on the main rank and broadcast.
+
+        Deregistered (RETIRING) adapters are retired here, at the sync point
+        before the next generate — their untrained tail (buffered groups and
+        any partially accumulated gradients) is discarded.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+- optimizer state is sharded across DP ranks at whole-parameter granularity;
+- gradients flow through plain DDP all-reduce (``use_distributed_optimizer``
+  must be OFF), which is also what makes cross-batch gradient retention
+  idempotent: the retained, already-reduced portion of the buffer is identical
+  on every rank, so re-reducing it is a no-op.
+
+Selective stepping (``step_adapter_slots``) steps only the slots whose
+adapter batch completed, zeroes only their gradients, and leaves every
+other slot's accumulated gradients and optimizer state untouched.
+"""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+    model_chunks: Sequence,
+) -> MegatronOptimizer:
+    """Build one Adam per adapter slot under a LayerWiseDistributedOptimizer.
+
+    The returned optimizer is a ``ChainedOptimizer`` whose ``chained_optimizers``
+    hold one Float16-wrapped Adam per slot, in slot order; each child's param
+    groups are tagged with ``miles_multi_lora_slot``. Param groups are narrowed
+    to this rank's whole-parameter shard by LayerWise, so Adam state exists only
+    for owned params.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+    assert (config.optimizer or "").lower() == "adam", (
+        "multi-LoRA per-slot optimizers only implement Adam semantics (state init, "
+        f"slot retirement cleanup, step clocks); got optimizer={config.optimizer!r}"
+    )
+
+    pg_collection = ProcessGroupCollection.use_mpu_process_groups()
+
+    # Delay master-weight creation into LayerWise (post-sharding), so fp32
+    # masters exist only for owned params. LayerWise unwraps FP32Optimizer
+    # children and re-wraps with Float16OptimizerWithFloat16Params itself.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+                base_optimizers.append(child)
+                init_fns.append(_adam_init_state_fn)
+    finally:
+        config.bf16 = reset_bf16
+
+    optimizer = LayerWiseDistributedOptimizer(base_optimizers, config, pg_collection, init_state_fn_list=init_fns)
+
+    # LayerWise aggregates grad stats globally (params are scattered across DP
+    # ranks at whole-parameter granularity), so per-child norm/clip reductions
+    # must span the world too, not just the model-parallel group.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def reset_grad_metadata_keep_grads(model_chunks) -> None:
+    """Between-batch replacement for ``DistributedDataParallel.zero_grad_buffer``.
+
+    Resets the per-iteration bookkeeping (``grad_added_to_main_grad`` flags and
+    bucket-group sync metadata) WITHOUT zeroing the grad buffers, so per-adapter
+    gradient accumulation survives across train batches. Slot gradients are
+    zeroed selectively at step time instead.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+            for param in model_chunk.params_with_grad:
+                param.grad_added_to_main_grad = False
+        for bucket_group in model_chunk.bucket_groups + model_chunk.expert_parallel_bucket_groups:
+            bucket_group.reset()
+
+
+def zero_adapter_slot_grads(model, slot: int) -> None:
+    """Zero one slot's gradients everywhere they live: the DDP ``main_grad``
+    buffer views (every rank holds the full buffer under plain DDP) and any
+    lingering ``grad``/``main_param.grad`` references."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+    scale by 1/batch_size, clip per slot, run Adam on the owned shard, copy
+    masters back to model params, and zero the slot's gradient state. One param
+    all-gather at the end propagates updated weights (unchanged slots gather
+    identical bytes).
+
+    Returns the grad norm per stepped slot. Non-finite gradients get the same
+    treatment as baseline bf16 training: they poison that adapter's weights
+    (visible as a NaN grad norm in the logs) — co-tenants are unaffected since
+    slots' params, gradients, and clipping are disjoint.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+    grad_norms: dict[int, float] = {}
+
+    for slot, batch_size in step_batch_sizes.items():
+        children = _slot_children(optimizer, slot)
+        # Copy model main_grads -> owned masters' grads (bf16 has no loss
+        # scaler, so prepare_grads performs no found-inf handling here), then
+        # normalize the accumulated sum into the adapter-batch mean.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+        for child in children:
+            child.prepare_grads()
+            for main_param in child.get_parameters():
+                if main_param.grad is not None:
+                    main_param.grad.mul_(1.0 / batch_size)
+
+        # Per-slot norm over the union of the slot's children, reduced across
+        # the world (params are scattered whole across DP ranks; every param is
+        # counted exactly once on its owner rank).
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_scheduler.py
```diff
@@ -0,0 +1,100 @@
...
+is per adapter, advancing with the adapter's own trained samples. This gives
+every adapter its own warmup and decay trajectory (a late-registered adapter
+no longer starts on an already-decayed shared schedule), and resume is
+deterministic: the position is rebuilt from the adapter's committed step
+count, with no stored or replayed scheduler state.
+
+Adapters without a known ``num_step`` at load time (indefinite service-mode
+runs, or num_epoch-derived stopping that resolves later on the rollout side)
+have no horizon to decay over: they warm up, then hold ``--lr`` constant.
+"""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_utils.py
```diff
@@ -0,0 +1,459 @@
...
+
+    chained = getattr(optimizer, "chained_optimizers", [optimizer])
+    for chained_optimizer in chained:
+        inner = getattr(chained_optimizer, "optimizer", chained_optimizer)
+        if inner is None:
+            continue
+        # TE/apex FusedAdam keeps the Adam clock per param GROUP, not per
+        # param: reset the retired slot's groups so the next tenant gets
+        # fresh bias correction (the per-param reset below only covers
+        # torch's AdamW fallback, whose clock lives in state["step"]).
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_utils.py
```diff
@@ -0,0 +1,459 @@
...
+                for hf_name, weight, _megatron_name in bridge.export_adapter_weights(
+                    model,
+                    cpu=True,
+                    show_progress=False,
+                ):
+                    # The model allocates every slot at --lora-rank; slice the
+                    # export down to this adapter's real rank so the tensors
+                    # match the r written to adapter_config.json (PEFT refuses
+                    # the checkpoint otherwise). The weight-sync push path does
+                    # the same. clone(): safetensors can't save aliased views.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_utils.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+def step_stepped_adapter_slots(args, model, optimizer, rollout_data, rollout_id: int, step_id: int) -> float:
+    """Optimizer-step the slots whose adapter batch completes with this train
+    batch, and advance their per-adapter LR/WD schedules. Returns the max grad
+    norm across stepped slots (0.0 when none stepped).
+
+    The shared opt_param_scheduler is never stepped under multi-LoRA: schedule
+    parameters inherit the args, only the position is per adapter.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/megatron_utils/multi_lora_utils.py
```diff
@@ -0,0 +1,459 @@
...
+    """Pick the adapters whose engine-side weights are stale: newly loaded +
+    stepped since the last push (tracked identically on every rank, so
+    per-adapter TP collectives line up). New engines need every loaded adapter.
+
+    Returns (adapters to push keyed by name, names getting a version bump).
+    Version bumps drive the staleness filter, so only adapters whose weights
+    actually changed get one: a new-engine full resync pushes every loaded
+    adapter, but re-sending unchanged weights must not age the unchanged
+    adapters' buffered groups toward the drop limit.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -341,14 +342,22 @@ def load_lora_adapter_from_tensors(
...
         added_tokens_config: dict | None = None,
+        upsert: bool = False,
     ):
-        """Load a LoRA adapter. ``serialized_named_tensors[tp_rank]`` is bytes for TP rank N."""
+        """Load a LoRA adapter. ``serialized_named_tensors[tp_rank]`` is bytes for TP rank N.
+
+        When ``upsert`` is set, the adapter named ``lora_name`` must already be
+        loaded and is overwritten in place (no unload/register); used by the
+        multi-LoRA in-place update path.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -369,13 +378,19 @@ def load_lora_adapter_from_distributed(
...
 
         Mirrors ``update_weights_from_distributed``: only metadata is sent here;
         the tensors arrive via NCCL broadcast (src=0), so no CUDA IPC is used and
         this works across nodes. ``init_weights_update_group`` must have created
         ``group_name`` already.
+
+        When ``upsert`` is set, the adapter named ``lora_name`` must
+        already be loaded on the engines; its weights are overwritten in place
+        (no unload, no register, no wait_for_unload). This is the in-place update
+        path for the fixed multi-LoRA pool.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/ray/multi_lora/backend.py
```diff
@@ -0,0 +1,196 @@
...
+        """Free the adapter's slot, after one final abort round.
+
+        The abort in ``retire_adapters`` fires once at the RETIRING->CLEANUP
+        flip, but requests can survive it: a multi-turn group between turns
+        submits its next turn only after that round, and a request still inside
+        the engine's tokenizer adapter batch can be missed by the scheduler-side
+        matching. Aborting again here — right before the slot becomes reusable —
+        closes those escapes, so a later tenant of the slot cannot serve a
+        retired adapter's orphaned requests.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/ray/multi_lora/registry.py
```diff
@@ -0,0 +1,256 @@
...
+        while len(self.batch_records) > MAX_BATCH_RECORDS:
+            self.batch_records.pop(next(iter(self.batch_records)))
+
+    def mark_batch_trained(self, rollout_id: int) -> list[str]:
+        """A train call over this batch succeeded: bank each adapter's groups, fire steps.
+
+        This is the only place accumulation/step state advances, so a failed or
+        retried train call leaves the registry untouched. Returns the adapters
+        that stepped.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/ray/rollout/train_data_conversion.py
```diff
@@ -85,6 +88,27 @@ def convert_samples_to_train_data(
...
         train_data["teacher_log_probs"] = [sample.teacher_log_probs for sample in samples]
 
+    if any(sample.adapter is not None for sample in samples):
+        assert all(sample.adapter is not None for sample in samples), "Cannot mix adapter and adapter-less samples"
+        train_data["adapter_slots"] = [sample.adapter.slot for sample in samples]
+        # Adapters whose adapter batch completes with this batch (the
+        # collection loop's step decision, lifted from sample metadata in postprocess),
+        # with their adapter batch sizes: the trainer scales each stepping slot's
+        # accumulated gradient by 1/adapter batch and advances the LR schedule by the
+        # summed adapter batches.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/generate_utils/prefill_logprobs.py
```diff
@@ -6,10 +6,25 @@
...
 from miles.utils.types import Sample
 
 
+def _lora_path_for_sample(args: Any, sample: Sample) -> str | None:
+    """The adapter name this sample's scoring request must be served with.
+
+    Multi-LoRA samples carry an adapter ref and must score under their own
+    slot's adapter — the engines only know ``__miles_slot_{N}`` names there.
+    Single-adapter LoRA uses the fixed name; base models use none.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/generate_utils/prefill_logprobs.py
```diff
@@ -63,14 +78,21 @@ def _build_batch_prefill_scoring_payload(
     if any(payload["logprob_start_len"] != logprob_start_len for payload in payloads):
         raise ValueError("Batched SGLang prefill scoring requires a shared logprob_start_len")
 
+    lora_paths = {payload.get("lora_path") for payload in payloads}
+    if len(lora_paths) > 1:
+        # A batch payload carries a single lora_path; scoring a mixed-adapter
+        # batch under one adapter would silently corrupt the other samples'
+        # logprobs. Callers must group by adapter first.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/generate_utils/prefill_logprobs.py
```diff
@@ -137,12 +159,15 @@ async def recompute_samples_rollout_logprobs_via_prefill(
     flush_url = url.rsplit("/", 1)[0] + "/flush_cache"
 
     if _can_batch_prefill_score(args, samples_to_score):
-        samples_by_logprob_start_len: dict[int, list[Sample]] = defaultdict(list)
+        # A batch shares one logprob_start_len and one lora_path, so group by
+        # both: mixed-adapter batches (multi-LoRA) must not score every sample
+        # under whichever adapter came first.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+A background producer generates continuously into per-adapter buffers. Each
+train batch is collected by popping groups from the buffers round-robin, in
+multiples of the adapter's ``min_groups_per_dp_split`` capped at its remaining
+batch, so:
+
+- any batch splits evenly across data-parallel ranks;
+- an adapter's batch (``rollout_batch_size`` prompt groups, i.e.
+  ``adapter_global_batch_size`` samples) is never overshot;
+- adapters whose batch completes here are stamped as stepping.
+"""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+# Safety valve, same convention as fully_async's queue.Queue(maxsize=1000):
+# never hit in practice, just bounds memory if training stalls entirely.
+MAX_BUFFERED_GROUPS = 1000
+EMPTY_BATCH_TIMEOUT_S = 30.0
+
+
+class GroupBuffer:
+    """One adapter's completed prompt groups: a FIFO queue you can also
+    len(), and sweep for staleness. Bounded; the oldest group is dropped
+    when a put exceeds the cap."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        return None
+    return result
+
+
+class MultiLoRAWorkerMetrics:
+    """The worker's cross-batch metric state, kept out of its buffer
+    machinery. Two cadences: dynamic-filter drops and staleness drops drain
+    every batch, per-adapter sample stats flush when the adapter's optimizer
+    step completes. Has its own lock — the producer thread records drops
+    while the trainer thread drains."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+
+    def __init__(self) -> None:
+        self.lock = threading.Lock()
+        self.dynamic_filter_drop_counts: dict[str, int] = defaultdict(int)
+        # Staleness of dropped groups per adapter, drained every batch.
+        self.staleness_values: dict[str, list[int]] = defaultdict(list)
+        # Shipped-sample values, accumulated per adapter across train batches
+        # and flushed as step statistics when the adapter steps. Bounded by
+        # adapter_global_batch_size values per adapter.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        self.step_rewards: dict[str, list[float]] = defaultdict(list)
+        self.step_response_lens: dict[str, list[float]] = defaultdict(list)
+        # Per-sample mean engine log prob (rough per-adapter entropy trend).
+        self.step_log_prob_means: dict[str, list[float]] = defaultdict(list)
+        # Group outcomes for zero-std rates: total groups shipped, and the
+        # common reward of each uniform-reward (zero advantage) group.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+    ) -> dict[str, dict[str, float]]:
+        """Accumulate the shipped batch's rewards and response lengths per
+        adapter; for adapters stepping with this batch, flush statistics over
+        their whole adapter batch (accumulated across shipped batches, so the
+        stats cover all ``adapter_global_batch_size`` samples of the step, not
+        just this batch's slice). Returns {adapter name: flushed metrics}.
+
+        Counted at ship time, not train commit: a failed train call aborts the
+        run anyway, so the distinction has no practical effect.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+                    logger.warning(
+                        f"Adapter '{name}' stepped with {len(rewards)} shipped samples, expected "
+                        f"adapter_global_batch_size={expected}; batch accounting drifted"
+                    )
+                # Keys are single-segment ("raw_reward_mean", not
+                # "rollout/raw_reward/mean") so that, prefixed with
+                # "{name}/", they sit one level under the "{name}/*" glob —
+                # the same shape as "train/loss" under "train/*". Deeper keys
+                # never get their axis: glob expansion only matches one
+                # segment on the server.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        self.dynamic_filter = (
+            load_function(args.dynamic_sampling_filter_path) if args.dynamic_sampling_filter_path else None
+        )
+        # Guards the buffers: the producer thread puts completed groups while
+        # get_groups (trainer side) pops them.
+        self.buffer_lock = threading.Lock()
+        self.buffers: dict[str, GroupBuffer] = defaultdict(GroupBuffer)
+        # Fairness cursor: the adapter whose buffer get_groups visits first.
+        # Advances past every visited adapter, persisting across calls and
+        # batches, so adapters are served round-robin.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        self.generate_fn = generate_fn
+        self.concurrency = concurrency or args.rollout_batch_size
+        self.running = True
+        self.worker_thread: threading.Thread | None = None
+        self.state = GenerateState(args)
+        self.dynamic_filter = (
+            load_function(args.dynamic_sampling_filter_path) if args.dynamic_sampling_filter_path else None
+        )
+        # Guards the buffers: the producer thread puts completed groups while
+        # get_groups (trainer side) pops them.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        self.buffer_lock = threading.Lock()
+        self.buffers: dict[str, GroupBuffer] = defaultdict(GroupBuffer)
+        # Fairness cursor: the adapter whose buffer get_groups visits first.
+        # Advances past every visited adapter, persisting across calls and
+        # batches, so adapters are served round-robin.
+        self.rotation: deque[str] = deque()
+        self.metrics = MultiLoRAWorkerMetrics()
+        # Last seen registration id per adapter name; a change means the name
+        # was re-registered and inherited buffer/metric state must be dropped.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        self.registrations: dict[str, str] = {}
+        # Set when run_loop dies; collect_batch surfaces it instead of letting
+        # the batch collection time out with a misleading empty-batch error.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        ``group_counts`` (adapter name -> groups in the batch); passing the
+        counts back on each fetch is what keeps a batch from overshooting an
+        adapter's remaining groups.
+
+        Pops round-robin from the cursor, one ``min_groups_per_dp_split`` at a
+        time, until ``num_samples`` is covered (the final multiple may overshoot
+        it) or no adapter can contribute. An adapter can't contribute when its
+        buffer holds less than a whole multiple, or the batch already holds all
+        its remaining groups.
+        """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        dp_size = self.args.multi_lora_dp_size
+        max_staleness = getattr(self.args, "max_weight_staleness", None)
+        group_counts = dict(group_counts)  # updated copy; the argument is not modified
+        popped: list[Group] = []
+        popped_samples = 0
+
+        with self.buffer_lock:
+            # Adapters retired at the last reconcile sync point: their buffered
+            # tail is discarded (base deregistration semantics), along with any
+            # partially accumulated reward stats.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+            for name in list(self.buffers):
+                if name not in adapters:
+                    self.buffers.pop(name)
+                    self.metrics.discard_adapter(name)
+                    self.registrations.pop(name, None)
+
+            # A re-registered name is a new tenant. The sweep above never ran
+            # when no generate happened between retirement and re-registration
+            # (e.g. the last adapter retired and the driver idled), so drop any
+            # buffered groups and partial stats inherited from the old tenant.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+    fully_async's generate_rollout_async): keep popping group multiples until
+    the batch reaches ``--global-batch-size`` samples, or it is non-empty and
+    made no progress for ``--multi-lora-max-coalesce-wait-s`` (the target can
+    be permanently unreachable when the live adapters' remaining batches are
+    smaller than the target, so ship what there is).
+
+    The remaining-groups math relies on the sequential trainer loop: the
+    previous batch's ``mark_batch_trained`` has landed before this generate
+    call, so the snapshot's ``accumulated_groups`` is current.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+    assert args.rollout_global_dataset
+
+    state = GenerateState(args)
+    worker = AsyncMultiLoRAWorker.get_or_create(args, data_source, generate_fn)
+    start_time = time.time()
+    queue_sizes = worker.queue_sizes()
+
+    # Driver contract: generate is only called with live adapters, and the
+    # sequential loop retires adapters and commits accumulated_groups only
+    # between generate calls — so one snapshot serves the whole collection.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+        args,
+        [s for g in data for s in iter_group_samples(g)],
+        url=get_model_url(args, "default"),
+        sampling_params=state.sampling_params,
+    )
+
+    # Adapter metrics live on the adapter's own optimizer-step axis
+    # ({name}/step), not rollout/step: one point per completed step, means
+    # over exactly the samples that step trained on. adapters[name].step is
+    # the committed count at snapshot time; this batch completes step + 1.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/multi_lora/async_rollout.py
```diff
@@ -0,0 +1,650 @@
...
+
+    stale_drops = worker.metrics.pop_stale_drops()
+    all_staleness = [staleness for values in stale_drops.values() for staleness in values]
+    metrics = {
+        **worker.metrics.pop_metrics(),
+        "perf/fully_async/queue_length": sum(queue_sizes.values()),
+        "perf/fully_async/stale_dropped": len(all_staleness),
+        # Per-adapter cycle metrics: {name}/perf/* rides rollout/step. Two
+        # segments under {name}/ keeps these off the step axis (glob expansion
+        # only reaches one segment); the {name}/perf/* glob catches them.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/rm_hub/__init__.py
```diff
@@ -28,15 +29,29 @@ async def remote_rm(args, sample: Sample):
             return await resp.json()
 
 
+def _resolve_reward_config(args, sample: Sample) -> tuple[str | None, str]:
+    # Spec fields win when set; unset fields fall through to the sample
+    # metadata and the process-wide args. A reward_spec with empty fields (an
+    # adapter config without its own rm_type) must not disable the fallbacks.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/rollout/sglang_rollout.py
```diff
@@ -172,7 +173,24 @@ async def generate(args: Namespace, sample: Sample, sampling_params: dict[str, A
...
 
-    if is_lora_enabled(args):
+    if sample.adapter is not None:
+        from miles.ray.multi_lora.controller import AdaptersCache
+
+        if (adapter := await AdaptersCache().get(sample.adapter.name)) is None:
+            # Adapter no longer sampleable (deregistered / cleaned up). Don't
+            # POST a request the retire-time abort round can no longer see:
+            # once the slot is reused, such an orphan would keep decoding under
+            # the next tenant's weights and silently pollute its group.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/tracking_utils/base.py
```diff
@@ -33,11 +33,21 @@ def log(self, metrics: dict[str, Any], step: int | None = None, **kwargs) -> Non
     @abstractmethod
     def finish(self) -> None: ...
 
+    def define_step_key_metric_group(self, prefix: str, step_key: str) -> None:
+        """Declare a runtime metric group (``{prefix}/*``) plotted against its
+        own step key. Only meaningful for backends where the chart axis is
+        configuration rather than per-call data; the default is a no-op —
+        tensorboard/mlflow take the step numerically on every log call."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/tracking_utils/base.py
```diff
@@ -51,6 +61,18 @@ def log(self, metrics: dict[str, Any], step: int | None = None, **kwargs) -> Non
 
         wandb.log(metrics)
 
+    def define_step_key_metric_group(self, prefix: str, step_key: str) -> None:
+        # Runtime analog of the static groups in _init_wandb_common. Call from
+        # the primary tracking process: definitions asserted by secondary
+        # shared-mode writers are lost nondeterministically.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/tracking_utils/tracking.py
```diff
@@ -28,6 +28,18 @@ def init_tracking(args, primary: bool = True, **kwargs):
...
 
+def define_step_key_metric_group(prefix: str, step_key: str) -> None:
+    """Declare a runtime metric group plotted against its own step key
+    (e.g. a multi-LoRA adapter's ``{name}/*`` against ``{name}/step``).
+
+    Fans out to the active backends; only wandb does anything (chart axes are
+    configuration there), and it deduplicates internally. Must be called from
+    the primary tracking process (the driver): definitions asserted by
+    secondary shared-mode writers are lost nondeterministically.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/multi_lora.py
```diff
@@ -0,0 +1,201 @@
...
+    Glob expansion only reaches one path segment, so {name}/perf/* keys are
+    out of {name}/*'s reach despite the shared prefix — each key group must
+    stay exactly one segment under its glob.
+
+    Must run in the the primary tracking writer, whose wandb
+    definitions are the only ones that reliably persist — and before the
+    adapter's first metrics, which is guaranteed at snapshot time: an adapter
+    can't ship step metrics until it has been promoted and trained a full
+    adapter batch.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/multi_lora.py
```diff
@@ -0,0 +1,201 @@
...
+    """Set ``args.multi_lora``, then validate and default the multi-LoRA arg
+    surface. Called from ``miles_validate_args``; a no-op for normal runs."""
+    args.multi_lora = getattr(args, "multi_lora_n_adapters", 0) > 0
+    if not args.multi_lora:
+        return
+
+    # Multi-LoRA ships its own rollout fn and data source (per-adapter
+    # buffers, controller-driven sampling); the standard defaults don't
+    # understand adapters. Swap them in unless the user pointed the flag
+    # at something else (e.g. a subclass).
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/multi_lora.py
```diff
@@ -0,0 +1,201 @@
...
+    if args.multi_lora_max_adapter_global_batch_size is None and getattr(args, "global_batch_size", None) is not None:
+        args.multi_lora_max_adapter_global_batch_size = 4 * args.global_batch_size
+    if args.multi_lora_max_adapter_global_batch_size is not None:
+        assert (
+            args.multi_lora_max_adapter_global_batch_size > 0
+        ), "--multi-lora-max-adapter-global-batch-size must be positive"
+
+    # Effective data-parallel size of the trainer; adapter batch shapes are
+    # validated against it at registration (min_groups_per_dp_split). Guarded for
+    # harnesses that validate miles args without the megatron arg set.
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/types.py
```diff
@@ -6,6 +6,30 @@
 import torch
 
 
+@dataclass(frozen=True)
+class AdapterRef:
+    """Per-sample handle identifying which LoRA adapter this sample is bound to.
+
+    Set by the multi-LoRA data source and consumed at training (slot routing)
+    and inference (per-request lora_path) sites. ``None`` means no adapter.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

miles/utils/types.py
```diff
@@ -6,6 +6,30 @@
...
+    slot: int
+
+
+@dataclass(frozen=True)
+class RewardSpec:
+    """Per-sample handle describing how this sample's response is scored.
+
+    Decoupled from ``AdapterRef`` because reward dispatch is a separate concern
+    from adapter routing — single-adapter or non-LoRA flows can use this too.
+    """
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/megatron_utils/test_multi_lora_scheduler.py
```diff
@@ -0,0 +1,116 @@
+"""Per-adapter LR schedules: parameters inherit the global args, position is
+per adapter. Pins the two defects this replaces: a late/loaded adapter no
+longer inherits the shared schedule's decayed position, and resume rebuilds
+the position deterministically from the adapter's committed step count."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/megatron_utils/test_multi_lora_slot_cleanup.py
```diff
@@ -0,0 +1,96 @@
+"""zero_optimizer_state_for_adapter must reset the retired slot's Adam state
+on both clock layouts: TE/apex FusedAdam keeps the step per param GROUP, while
+torch's AdamW fallback keeps it per param in state["step"]. Either way the
+slot's next tenant must start from zero moments and a fresh bias-correction
+clock, and co-tenant slots must be untouched."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/megatron_utils/test_multi_lora_slot_cleanup.py
```diff
@@ -0,0 +1,96 @@
...
+class FakeMultiLoRALinear:
+    def __init__(self, adapters):
+        self.adapters = adapters
+
+
+@pytest.fixture()
+def rig(monkeypatch):
+    # zero_optimizer_state_for_adapter imports the bridge module lazily; stub
+    # it in sys.modules so the test does not depend on a bridge build that
+    # ships multi-LoRA (and can shape the module graph freely).
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/megatron_utils/test_slice_lora_to_rank.py
```diff
@@ -0,0 +1,50 @@
+"""slice_lora_to_rank trims max-rank-padded LoRA tensors to an adapter's real
+rank. Both consumers depend on it: the weight-sync push and the HF PEFT
+checkpoint export (whose adapter_config.json declares the real rank — padded
+tensors would make PEFT refuse the checkpoint)."""
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/training_utils/test_get_batch_multi_lora_cp.py
```diff
@@ -0,0 +1,154 @@
...
+  summed per slot, with the rank-local stream padding attributed to the last
+  slot, and sums to the rank-local padded token count;
+- zigzag slicing gives every CP rank the same per-sample lengths, so counts
+  must be identical across ranks;
+- unsorted ``adapter_slots`` in a micro-batch is rejected;
+- (strict xfail) bshd tokens must match a single zigzag slice — data.py
+  currently slices bshd tokens twice under CP>1.
+
+CUDA is stubbed out so the tests run on CPU-only workers.
+"""
```
yushengsu-thu: summarize comments - reduce to one or two lines

tests/fast/backends/training_utils/test_get_batch_multi_lora_cp.py
```diff
@@ -0,0 +1,154 @@
...
+    batch["adapter_slots"] = [2, 0, 1, 0, 2]
+    with pytest.raises(AssertionError, match="not sorted"):
+        get_batch(_FakeIterator(batch, N_ADAPTERS), KEYS, pad_multiplier=8, qkv_format="thd")
+
+
+def test_bshd_cp2_tokens_are_single_sliced(monkeypatch):
+    # Regression test: the bshd path used to slice tokens twice under CP>1
+    # (pre-slice at the top of the branch, then again in the non-allgather
+    # arm), silently replacing the back half of each rank's tokens with pad
+    # zeros while keeping the shape unchanged.
```
yushengsu-thu: summarize comments - reduce to one or two lines

yushengsu-thu (commented):
Summarize the comments into one or two lines. If a comment doesn't make sense, feel free to delete or remove it.

yushengsu-thu:
## Splitting this PR into 7 stacked PRs

To reduce review load, this PR is being split into 7 module-level PRs, each an independently testable unit. **This PR stays open as the integration reference** (its E2E validation — 2×H200 smoke, full lifecycle — covers the whole stack). The 7 branches are pushed to `mathewjhan/miles` and stack on top of each other; **they must be merged in order 1/7 → 7/7**.

Every prefix of the stack keeps `main` green: the whole feature is gated behind `--multi-lora-n-adapters > 0` and the driver lands last, so merging 1..k leaves everything dark until 7/7. Branch `multi-lora/07-example`'s tree is byte-identical to this PR's head (`47c3ec09`) — the split is exact, no code differences.

| # | Branch | Scope | Diff | CI tests carried in the PR |
|---|--------|-------|------|---------------------------|
| 1/7 | `multi-lora/01-utils` | Utils foundation: `types.py` (AdapterRef/RewardSpec), `adapter_config.py`, `utils/multi_lora.py` (helpers + validation), CLI flags, tracking metric axes | +496/−2 | `test_arguments.py` multi-LoRA validation tests |
| 2/7 | `multi-lora/02-controller` | Adapter controller: `ray/multi_lora/` package — registry state machine, backend, control-plane HTTP API, named Ray actor | +1,322 (half tests) | `test_controller_backend` + `test_controller_http` (631 test lines) |
| 3/7 | `multi-lora/03-trainer` | Trainer core: `multi_lora_{optimizer,scheduler,utils}.py`, `model.py`, `bridge_lora_helpers.py`, `data.py`/`loss.py` | +1,255/−30 | scheduler, slot_cleanup, slice_to_rank, get_batch CP=2 suites (387 test lines) |
| 4/7 | `multi-lora/04-weight-sync` | Weight sync + actor integration: `update_weight/` mixin+broadcast, `sglang_engine.py` upsert, `actor.py` reconcile/train/save/update hooks | +207/−28 | ⚠ no new dedicated tests — covered by main's existing weight-sync validation suite + the 7/7 E2E; the one that most needs human eyes |
| 5/7 | `multi-lora/05-rollout-routing` | Rollout request routing: `sglang_rollout.py` (per-sample lora_path/rid), `prefill_logprobs.py`, `rm_hub` per-sample rewards | +197/−15 | `test_prefill_logprobs`, `test_reward_config_resolution` |
| 6/7 | `multi-lora/06-batching` | Async batch collection + data conversion: `rollout/multi_lora/` (async_rollout + data_source), `ray/rollout/` conversions | +1,267/−2 | batch_collection, process_group, train_data suites (452 test lines) |
| 7/7 | `multi-lora/07-example` | Driver + example: `train_multi_lora_async.py`, `examples/multi_lora/` (typer launcher, adapter yamls, smoke client) | +640 | no fast tests; acceptance = the E2E smoke already validated on this PR |

Merge order = import dependency order: everything imports 1; the trainer (3) and rollout (5, 6) talk to the controller (2); weight sync (4) uses trainer exports; the driver (7) wires it all.

**Verification done on each branch**: import-graph completeness per layer (no branch references a module that lands later), and each layer's fast suites pass on its own tree (layer 1: 2456 passed; layer 2: 2488; layers 3–4: 266 each incl. all existing megatron_utils tests; layer 5: 400; layer 6: full `tests/fast/ray/ + rollout/`).

**Mechanics note**: since these merge with squash, after each merge GitHub will still show the previous layers' changes in the next PR's diff until it is rebased — after merging k/7, rebase branches k+1..7 onto the new `main` (mechanical, no conflicts by construction).

https://github.com/radixark/miles/pull/1742
https://github.com/radixark/miles/pull/1743
https://github.com/radixark/miles/pull/1744
https://github.com/radixark/miles/pull/1745
https://github.com/radixark/miles/pull/1746
https://github.com/radixark/miles/pull/1747
https://github.com/radixark/miles/pull/1748

yushengsu-thu (approved):
All content landed via the 7 stacked PRs (#1742-#1748); this umbrella merge is a no-op content-wise and closes the loop.

PR #1639, opened 2026-07-12
rollout: decouple oversampling-abort teardown into a pluggable agent hook

Shi-Dong:
## Problem

With oversampling on, once the trainer has collected enough samples, the rollout aborts the in-flight **SGLang** generations via `/abort_request {"abort_all": true}`. But the external agent loops driving the agentic rollouts are never told.

Each aborted generation just looks like one finished LLM call to the agent, so the agent immediately issues the *next* completion request back to SGLang. The loops keep hammering SGLang and holding their sandbox containers until they independently hit their own `max_seq_len` / trial timeout — wasting inference during the drain and pinning containers.

## Why not a hard-coded flush in core

Miles' core rollout is agent-host-agnostic: `agentic_tool_call.generate` delegates *all* agent-host knowledge to a plugin loaded from `--custom-agent-function-path`. Baking a Harbor-specific `/flush` call into the core rollout would wrongly assume every deployment uses a Harbor agent server that speaks `/flush`, which isn't always true.

## Fix (pluggable hook, both sync and async paths)

- **Core stays agnostic.** New `miles.utils.misc.call_agent_abort_hook(args)` looks for a sibling `abort` callable in the *same module* as the configured `--custom-agent-function-path` and awaits it if present. No hook → drain as before. Zero Harbor strings, no new core args. Non-agentic and non-Harbor runs are unaffected.
- **Wired into BOTH oversampling-abort implementations:**
  - `miles/rollout/sglang_rollout.py::abort` — the legacy/sync path (default when `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR` is unset).
  - `miles/rollout/inference_rollout/inference_rollout_train.py::abort` — the experimental-refactor path used by **async** training (`train_async.py` + `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1` → `InferenceRolloutFn`). This is the path Shi's agentic async runs actually take, so covering it is the point.
- **Harbor teardown lives in the plugin.** The Harbor `/flush` logic is in `examples/experimental/swe-agent-v2/swe_agent_function.py` as its `abort(args)` hook — next to the `AGENT_SERVER_URL` + `/run` code it already owns. It flushes the agent server's in-flight `/run` trials for this run's `session_server_instance_id` (sending `HARBOR_ADMIN_SECRET` as a bearer token when present). No-op unless the URL + instance id are available.

No Harbor server change is needed — `/flush` already exists in `harbor-miles-v0.13.1` and returns HTTP 200 on cancel.

## Test plan

- [ ] Sync agentic run (legacy path): on oversampling abort, plugin logs `Flushed agent server ...` and Harbor `/run` tasks terminate instead of continuing to call SGLang.
- [ ] Async agentic run (`train_async.py` + refactor): same behavior via `inference_rollout_train.abort`.
- [ ] Agentic run whose plugin defines no `abort`: `call_agent_abort_hook` is a silent no-op; drain unchanged.
- [ ] Non-agentic (rule-based RM) run: `custom_agent_function_path` is None → hook lookup short-circuits; abort behavior unchanged.


miles/rollout/inference_rollout/inference_rollout_train.py
```diff
@@ -29,6 +29,10 @@ async def abort(state: GenerateState, pendings: set, rollout_id: int) -> list[li
     logger.info(f"Abort request for {urls}")
     await asyncio.gather(*[post(f"{url}/abort_request", {"abort_all": True}) for url in urls])
 
+    # Let the agent integration tear down its in-flight trials so they stop hitting
+    # SGLang, instead of running on until their own max_seq_len / timeout.
+    await call_agent_abort_hook(args)
```
maocheng23, line 34: Would this be triggered even not under SWE job?

miles/rollout/sglang_rollout.py
```diff
@@ -361,6 +361,10 @@ async def abort(args: Namespace, rollout_id: int) -> list[list[Sample]]:
         if isinstance(result, Exception):
             logger.warning(f"Failed to abort worker at {url}: {result}")
 
+    # Let the agent integration tear down its in-flight trials so they stop hitting
+    # SGLang, instead of running on until their own max_seq_len / timeout.
+    await call_agent_abort_hook(args)
```
maocheng23, line 366: Same as above

miles/rollout/inference_rollout/inference_rollout_train.py
Shi-Dong: I don't think so: if `custom_agent_function_path` is not set, then `call_agent_abort_hook` is just a no-op.

maocheng23 (approved):

maocheng23 (approved):

yueming-yuan (approved):

PR #1643, opened 2026-07-12
fix(weight-update): track disk-delta engine freshness

guapisolo:
## Summary
Keep disk-delta weight updates connected across rollout-engine lifecycle changes.

## Symptom & Reproduction
- **Symptom:** `tests/e2e/megatron/test_qwen3_4B_disk_delta.py` fails before weight sync with `AttributeError: 'UpdateWeightFromDiskDelta' object has no attribute 'is_rollout_engines_fresh'`.
- **Reproduction:** Run `python3 tests/e2e/megatron/test_qwen3_4B_disk_delta.py` on its registered `stage-c-8-gpu-h100` environment.

## Root Cause
1. `MegatronTrainRayActor.update_weights` requires each updater to report connection freshness.
2. `UpdateWeightFromDiskDelta` omitted the freshness methods and connection state.

## Fix
Track rollout-engine handles and stale state inside `UpdateWeightFromDiskDelta`, then clear stale state when engines reconnect. This completes the existing updater lifecycle contract without a caller-side fallback.

## Verification
- **Regression e2e:** `tests/e2e/megatron/test_qwen3_4B_disk_delta.py` passed on its registered `stage-c-8-gpu-h100` job in 334s before the review-only removal of the standalone unit test.
- **Review follow-up:** Removing the standalone test changes no production code; `git diff --check` and `compileall` pass.
- **Interface completeness:** Every Megatron weight updater defines the freshness contract.

## Review Focus
- Scrutinize `UpdateWeightFromDiskDelta` freshness transitions around `connect_rollout_engines`.


yueming-yuan (approved):
approved with nit: we might not need a seperate test for this straightforward logic

PR #1645, opened 2026-07-13
Add Qwen3 / Qwen3-Coder support to swe-agent-v2 example (ROCm-safe)

lizamd:
- run.py: parameterize SGLang/TITO parsers (GLM-4.7-Flash defaults unchanged); set RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES only under torch.version.hip so it is a no-op on NVIDIA (CUDA device management unchanged there)
- run-qwen3-swe.py: thin Qwen3/Qwen3-Coder launcher reusing run.py; --coder uses Qwen3-Coder-30B-A3B-Instruct (rope_theta=1e7 via MODEL_ARGS_ROTARY_BASE)
- server.py: add the missing task-agnostic Harbor /run server (README references it; closely related to the codecontests server in #1603) plus turns/tool_calls metrics so agent/turns_mean populates
- README: Qwen3/Coder usage and ROCm prerequisites


lizamd:
**CI infra flake: CPU test stages fail at `conftest` import when HuggingFace can't serve `Qwen/Qwen3-0.6B`**

The `stage-a-cpu` / `stage-b-cpu` failures on this PR are **unrelated to the diff** (which only touches `examples/experimental/swe-agent-v2/`; `Run pre-commit` is green). Root cause is a pre-existing test-collection issue.

**Symptom** — pytest aborts during collection with exit code 4:
```
ImportError while loading conftest '.../tests/conftest.py'.
...
miles/utils/test_utils/mock_tools.py:144: in <module>
    _TOKENIZER = load_tokenizer("Qwen/Qwen3-0.6B", trust_remote_code=True)
...
E   OSError: Can't load tokenizer for 'Qwen/Qwen3-0.6B'.
```

**Root cause** — `miles/utils/test_utils/mock_tools.py:144` loads the `Qwen/Qwen3-0.6B` tokenizer from HuggingFace **at module-import time**. It's pulled into every CPU test run via `tests/conftest.py` → `tests/fast/fixtures/generation_fixtures.py:21` (`from miles.utils.test_utils import mock_tools`). Because the fetch happens at import, a single failed HF download aborts **all** collection (exit 4), not just tests that use the tokenizer.

**Intermittent, not deterministic (→ HF rate-limit / transient fetch)** — within the same commit `33a95a1`, `stage-a-cpu (2)` passed while `stage-a-cpu (0)/(1)/(3)` and `stage-b-cpu` failed with the tokenizer error; all CPU shards passed earlier on `7d887f4`. Consistent with anonymous HF rate-limiting when shards fetch concurrently.

**Suggested fixes** (maintainer side):
1. Add an `HF_TOKEN` secret to the CPU test workflow (authenticated pulls avoid the anonymous rate limit), and/or
2. `actions/cache` on `~/.cache/huggingface` so shards don't each re-download, and/or
3. Vendor the small `Qwen/Qwen3-0.6B` tokenizer as a test asset + `HF_HUB_OFFLINE=1`.
   Optionally make `mock_tools._TOKENIZER` lazy so a plain import no longer performs a network call (limits blast radius to tests that actually need it).

yushengsu-thu (changes requested):
create an examples/experimental/swe-agent-v2-amd and put your files into in

lizamd:
> create an examples/experimental/swe-agent-v2-amd and put your files into in

@yushengsu-thu done — moved the AMD/Qwen3 work into `examples/experimental/swe-agent-v2-amd/` and reverted `examples/experimental/swe-agent-v2/` to its base state (no shared-file edits now).

The new dir is self-contained: it has its own `run.py` (the generalized parser config + ROCm `HIP_VISIBLE_DEVICES` handling), plus `run-qwen3-swe.py`, `server.py`, and a standalone `README.md`. It also carries copies of the runtime deps `generate.py` / `swe_agent_function.py` so it runs without touching the shared example — happy to reference the shared modules instead if you'd prefer no duplication.

(FYI the earlier Gemini findings were already addressed in an earlier commit. The remaining CI red is the pre-existing `Qwen/Qwen3-0.6B` tokenizer-fetch flake I noted above, unrelated to this diff.) PTAL 🙏

yushengsu-thu (approved):

PR #1649, opened 2026-07-13
fix(metrics): move pass@k from trainer to rollout manager

HJSang:
## Summary

- Move pass@k computation from the Megatron trainer (`log_passrate` in `log_utils.py`) to the rollout manager (`log_rollout_data` in `metrics.py`)
- The trainer-side `log_passrate` assumed `raw_reward` had contiguous groups of `n_samples_per_prompt`, but custom converters can filter individual samples from within groups, breaking alignment and crashing the reshape
- The rollout side runs **before** `convert_samples_to_train_data`, so all samples are present and group boundaries are intact

## Motivation

Custom `convert_samples_to_train_data` implementations (e.g. for agent RL) filter out individual failed/zero-reward samples per-prompt to save forward-pass compute. This shrinks the batch below `rollout_batch_size * n_samples_per_prompt` and can leave partial groups (e.g., 78 samples with group_size=8). The trainer-side `log_passrate` then crashes on:

```python
assert len(flat_rewards) == num_groups * group_size  # 78 != 16*8
```

Moving pass@k to the rollout side avoids the issue entirely — the data is complete and aligned at that point. The eval path (`log_eval_rollout_data`) already computed pass@k on the rollout side and is unchanged.

## Changes

| File | Change |
|------|--------|
| `miles/ray/rollout/metrics.py` | Add `_compute_passrate_from_samples`, call it from `log_rollout_data` |
| `miles/backends/training_utils/log_utils.py` | Remove `log_passrate` function and call site |
| `tests/fast/ray/rollout/test_metrics.py` | Add `TestComputePassrateFromSamples` |

## Metric namespace change

Before: `passrate/pass@k` (logged via `gather_log_data("passrate", ...)` on trainer)
After: `rollout/pass@k` (logged inline with other rollout metrics)

The `log_passrate` config flag still controls whether the metric is computed.

## Test plan

- [x] Unit tests for `_compute_passrate_from_samples` (perfect/zero/mixed/non-divisible)
- [ ] Verify W&B metrics appear under `rollout/pass@k` on a real run
- [ ] Confirm no crash with custom converters that filter samples


Rockdu:
LGTM, thanks for this fix for pass@k calculation timing! pass@k better fits rollout_log and thus should be in rollout metrics. 

But the TestComputePassrateFromSamples claimed in the PR description still seems to be missing. Could you add this unit test to the PR? Then I think we can merge it!

HJSang:
> LGTM, thanks for this fix for pass@k calculation timing! pass@k better fits rollout_log and thus should be in rollout metrics.
> 
> But the TestComputePassrateFromSamples claimed in the PR description still seems to be missing. Could you add this unit test to the PR? Then I think we can merge it!

Done. Unit test added. I am wondering what is the principle of miles about unit test? I have this question because coding agent is often aggressively adding unit tests, which i don't like though.

Rockdu:
> > LGTM, thanks for this fix for pass@k calculation timing! pass@k better fits rollout_log and thus should be in rollout metrics.
> > But the TestComputePassrateFromSamples claimed in the PR description still seems to be missing. Could you add this unit test to the PR? Then I think we can merge it!
> 
> Done. Unit test added. I am wondering what is the principle of miles about unit test? I have this question because coding agent is often aggressively adding unit tests, which i don't like though.

Oh, no worries! I asked simply because of the mismatch between the description and the actual changes.

Regarding the unit test standard, I checked our current PR checklist (https://github.com/radixark/miles/blob/main/docs/developer/contributor-guide.md#pr-checklist), which generally encourages every PR to include a unit test. However, after discussing this with a few teammates, we realized that the expectations around unit tests should probably be more nuanced and fine-grained.

Zhichenzzz (approved):
LGTM, thanks!

yueming-yuan (approved):

PR #1654, opened 2026-07-13
feat: miles dashboard

yueming-yuan:
Observability dashboard for miles RL post-training: an **efficiency view** (per-GPU wall-clock timeline: phase bands, GPU util + per-process memory breakdown, sglang engine overlay, rank carpet, config-tuning advisory), a **training-dynamics view** (step metrics grouped by rollout/train with tooltips → per-sample table with GRPO group navigation → token-level entropy/logprob drill-down, wandb link), and a **trajectory lifecycle view** (per-sample generate/tool/queue swimlanes with segment-level weight-version staleness and a conversation view from the trajectory dump — the fully-async batch anatomy).

## Design principles

- **Offline-first, file-mediated.** The collector is the only writer; everything lands under `{dump-details}/dashboard/` as append-only JSONL (high-rate streams hourly-partitioned). The server is a stateless reader — a finished (or crashed) run reopens identically from files alone.
- **Observability must never kill training.** All hooks are fire-and-forget with rate-limited warnings; sinks batch (64 events / 2 s); no sink attached = no-op.
- **Minimal core footprint.** Outside `miles/dashboard/` + tests the whole feature is a thin seam across the rollout/train path (registry entry, CLI args, Timer event sinks, engine topology accessor, transport-free dump-helper extractions, and the trajectory lifecycle probes in `generate_and_rm` / `multi_turn` / openai record assembly / session core — gated behind a stdlib-only seam in `miles/utils/lifecycle.py`, same layer as `Timer.event_sinks`).
- **Scales by construction:** columnar in-memory store (polars, ~16 B/row), lazy partition reads (open parses nothing; queries O(window)), hard 4 h viewport with lazy pan, rank carpet + manual lane selection (`rank:/node:/gpu:/engine:/role:` grammar, URL-shareable), 32-lane render budget, binary heatmap endpoint.
- **Brand-consistent UI.** Chrome, categorical palettes, and typography pulled from the docs' Mintlify config (`docs/docs.json`) rather than ad hoc colors — colorblind-validated via the dataviz palette checker.

## What's new since the initial draft

- **wandb link** at the top of the Metrics page, derived from the run's tracking args.
- **GRPO group navigation** — jump between samples in the same group from the token view.
- **Readable timeline x-axis** — auto-scaled tick spacing with an absolute-time toggle.
- **Metrics page reclassification + tooltips** — rollout/log-probs metrics moved into the `train` group where they're computed; every wandb metric gets an explanatory tooltip.
- **GPU per-process memory breakdown** in the efficiency view.
- **Databuffer length status**, surfaced from the rollout manager.
- **sglang config-tuning advisory** — heuristic call-outs (e.g. low prefix-cache hit rate, low concurrency vs. `--sglang-max-running-requests`) computed from the engine metric series.
- **Brand UI redesign** — chrome, categorical/sequential/diverging palettes, and headings/body fonts now come from the docs' brand tokens; the "miles dashboard" wordmark became the real logo; the run-status header explains live vs. static mode.
- **Batch anatomy readability** — a hairline gap between trajectory rows plus a hover-highlight band, so a single row is a reliable click target even when hundreds of samples are compressed into a few px each.
- **Sentence-case text pass** across panel titles, buttons, captions, and advisory messages (previously an inconsistent mix of ALL CAPS panel headers and ad hoc lowercase).
- **Merged current `main`** (41 commits ahead) — reconciled with `main`'s independent `debug_data.py`/`train_data_conversion.py`/tracking-utils refactors, and fixed the two tests those changes broke.
- **Real fix, not a workaround, for a `flush_cache` bug**: the retry loop wasn't backing off on the normal "pending requests" 400 response, burning through all 60 attempts in a fraction of a second instead of giving in-flight generation time to drain.

## Usage

```
# training: add to train args (train.py and train_async.py both wired)
--dump-details /path/to/dump --use-miles-dashboard

# view (live: --follow; post-hoc: without)
python -m miles.dashboard.serve --dump-details /path/to/dump --follow --port 7788
```

## Validation

- 3680 passed, 0 failed, 37 skipped on the full `tests/fast/` suite (post-merge, on an H200 devbox).
- Validated live against real runs: qwen3-30B-A3B colocate (8×H200, wait-ratio bubble discovery), 4 train + 4 rollout fully async (role-split lanes, engine overlay, staleness), and a full qwen3-4B fully-async GRPO run (`--pause-generation-mode in_place`, 15/15 rollouts, clean weight updates every boundary, Ray job `SUCC`) with the dashboard served live throughout.

## Review guide

41 stacked single-purpose commits, each self-contained with its tests — commit-by-commit is the intended review path. Bottom-up: store → dump reader → server/SPA → timeline → scraper/sampler/collector → core wiring → carpet/selection → scaling (columnar/partitions/viewport) → trajectory lifecycle (probes → sink → columns → views) → gap-closing features (wandb link, GRPO nav, x-axis, metrics reclass, GPU per-process memory, databuffer status, config advisory) → brand UI redesign → main merge → text/UX polish.


yueming-yuan:
Pushed a batch of fixes from live dogfooding on real runs (colocate 8×H200 3-step, 4+4 fully-async, and a 744B disaggregated agentic run):

- **Live phase bands (open intervals)**: `Timer.start` now emits an OPEN marker; in-progress phases render as growing bands instead of appearing only when they end — the first agentic rollout (40 min+) is visible from second one.
- **External-engine overlay**: pure sglang engines (no miles actor) are synthesized into the topology from scraper worker addrs; the frontend attaches their series by node when GPU placement is unknown.
- **Global lane numbering**: `g{index}` labels everywhere + `g:0-31` selection grammar; per-node relative ids demoted to secondary text.
- **Spaced default selection**: opening the timeline seeds 8 evenly-spaced lanes as a real, editable selection (chips/URL) instead of first-N.
- **Data freshness badge**: `live · data → HH:MM:SS` from the newest server-side timestamp (a stalled collector reads as stale).
- **Batch anatomy**: zoom/pan; per-sample lifecycle lane embedded on the token (L2) page; single-turn/aborted gen spans close at attempt end (no more phantom uniform end times); session-server agent gaps emitted as tool-wait spans, bounded exactly by a new optional `SessionRecord.request_timestamp` (+3 lines, wire-compatible).
- **Robustness**: dumps without train `log_probs` degrade to null columns instead of HTTP 404; `Cache-Control: no-cache` on the SPA so upgrades never serve a stale frontend (existing browsers need one hard refresh); trainer log now prints the serve command.

Same 23-commit stack shape; all commits amended in place. 149 dashboard tests green.

yueming-yuan:
**Batch: metrics L0 redesign (wandb/grafana parity)** — force-pushed `d34d87b04`.

- **Category page replaces the pinned-chart L0** (PR-6): left sidebar with one button per wandb category (`rollout` / `perf` / `train` / `eval` — names verbatim from the `log()` fan-out), single-select, every key of the active category renders as a line chart. Filter box narrows within the category.
- **`sglang` category** (PR-8): appears when scraped engine metrics exist (`/api/meta.engine_metric_keys`, derived from the data, not a hardcoded whitelist). Each metric renders as a multi-line chart, one thin line per engine, hover carries engine identity. Verified against the qwen30b-async3s2 run: 8 metrics incl. disagg queues.
- **No perpetual repaint** (PR-6/8): follow mode still polls every 5s, but each chart keeps a data signature and repaints only when its series actually changed — a quiet run leaves the page visually still.
- **Chart click-through removed; `steps` is top-level nav** (PR-6): charts are hover-only; the per-step data view (batch anatomy, samples, groups) is reached via a `steps` link next to `metrics · timeline`, landing on the newest train step.
- **`dump/*` gated server-side** (PR-5): dump-derived keys are advertised only for dump-only dirs (no telemetry stream); with telemetry the L0 never reads dumps. `mean_abs_lp_diff`'s telemetry counterpart is `train_inference_abs_diff`; composition stats moved to the step page.
- **`mixed-version frac` stat box** (PR-22) added to the step-page statgrid next to zero-std groups.
- Fix: render crash from an interim bad merge (`crumbs` referenced `meta` before fetch).

Verification: local tip 151 passed; devbox `tests/fast/dashboard` 152+3 passed (one failure was a stale local Ray cluster on the box — cleared), `tests/fast/rollout` 388 passed. Devbox serve on :7788 updated to this tip.

yueming-yuan:
sglang page polish (follow-up, force-pushed `d32fd8c3e`): per-engine line colors from a fixed 10-color categorical palette assigned by sorted-addr index (stable across refreshes, cycles past 10) — PR-8; metrics grid now two charts per row and the category sidebar narrowed to 160px — PR-6. Verified on the qwen30b-async3s2 run.

yueming-yuan:
Metrics interaction follow-up (force-pushed `5f11f2747`): all L0 charts now support **drag-to-zoom** on the x-range (translucent selection band, double-click resets, zoom survives live data refreshes since it lives on the canvas, per-chart independent) — implemented once as `installDragZoom` in charts.js and shared by `drawChart` (PR-6) and `drawMultiLine` (PR-8). Sub-4px presses still count as clicks, so the step-page scatter's open-sample behavior is unchanged. Multi-line colors are assigned by full-list index so an engine keeps its color when zoom drops other series; x labels stay run-relative under zoom. Layout: sidebar 200px, chart columns capped at 500px.

yueming-yuan:
Nav + step-jump follow-up (force-pushed `874548a1b`): top bar is now dark with bold 15px nav blocks at equal spacing — **Metrics · Compute Utilization · Data / Trajectories** — whole-block gray hover, accent underline marks the active section (replaces the `› efficiency` breadcrumb). Step view header gained a number input: type a step and press Enter to jump straight to it (PR-6/8). Also fixed at its origin: PR-6's `crumbs()` was called before the meta fetch (the TDZ crash previously patched at PR-8 actually originated here).

yueming-yuan:
Drag-zoom fix (force-pushed `49aa28397`, PR-6): real drags route mousemove/mouseup to whatever element is under the pointer, so the previous canvas-local handlers lost any drag that strayed outside the 220px chart band (or crossed into the next chart) and the zoom never completed — synthetic-event tests had masked this. `installDragZoom` now anchors on mousedown, listens for move/up on `window` until release, and `preventDefault()`s the native text selection. Verified with genuine mouse input against the live run on :7788.

yueming-yuan:
sglang legend (force-pushed `8839f1b18`, PR-8): the sglang page gains an **ENGINES** panel on the right — one checkbox per scraped engine (color swatch matches its line), all checked by default. Unchecking hides that engine's line in every chart on the page and drops it from the y-scale. Colors are pinned to the page-wide sorted engine order (`opts.colorIndex` in `drawMultiLine`), so a hidden or zoom-clipped engine never shifts the others' colors; toggles repaint from each chart's cached series without refetching. Verified on the live run (2 engines).

yueming-yuan:
Zoom rework + slow-load notice (force-pushed `c7e322a77`):

- **Why most drags "did nothing"**: the old x-zoom silently reset whenever the selected window held <2 data points — on 3-point step charts (rollout/perf/train of a 3-step run) nearly every reasonable drag hit that fallback; only wide sweeps (e.g. out to step 0) survived. The fallback is gone: zoom domains are now explicit, and a window with few or zero points just renders that window (PR-6/8).
- **Both axes**: drag now selects a box — `canvas._zoom = {x, y}`. An axis participates only when dragged ≥10px, so a horizontal-ish sweep zooms x and lets y re-autoscale to the window instead of pinning y to hand wobble; the live overlay shows a full-height band until the drag gains y-extent. Marks are canvas-clipped to the plot area so out-of-domain data can't paint over margins. Double-click still resets.
- **Detokenize notice** (PR-6): the sample page now says "loading sample… the first open of a step detokenizes its whole dump and can take several minutes" instead of a bare "loading…", so a minutes-long first open reads as work, not a hang.

Verified with real mouse input on the live run: box drag on gen_throughput zooms x to +6:40–+23:20 and y to the selected band; narrow drags on 3-point charts now zoom reliably.

yueming-yuan:
Weight-version hue ring (force-pushed `ae2b9ee36`, PR-23): the batch-anatomy view now colors each weight version from a cyclic spectral ring — `hsl(55 + (v·137.508) % 320, 70%, 46%)`: golden-angle steps keep consecutive versions far apart on the wheel for any version count, and the ring starts past the tool-wait amber band so version colors never read as span colors (gen green / tool amber / queue gray). Detailed rows color the `v#` label (mixed rows: each endpoint in its own color with a stale dash between); compact rows get a colored left tick per lane, two ticks when versions are mixed. Legend text updated. Verified on the live run: v1 lanes tick in the ring's 192.5° cyan.

yueming-yuan:
Weight-version coloring, corrected semantics (force-pushed `222a80f41`, PR-23): **generation spans themselves** are now colored by the weight version that produced them, using a sequential cyclic ring — `hsl(55 + (105 + 25v) % 320, 69%, 37%)`. v0 is the familiar gen green, each +1 steps 25° (green → teal → blue → … → back around every ~13 versions), so neighbouring versions read as neighbouring colors and every version bump is a visible new shade. The ring skips the tool-wait amber band; tool/queue/consume and the Compute-Utilization view are untouched. Version ticks/labels reuse the same ring. Also fixed at the read side: coarse gen spans open before any version exists, so `trajectory_lanes` now backfills the span's version from the closing `attempt_end` event — old dumps get correct per-span versions without re-running. Verified on the live run: gen bars render `versionColor(1)` teal, matching the dump's weight_version=1 for all three steps.

yueming-yuan:
Follow-up (force-pushed `7547be767`, PR-23): removed the separate weight-version label column / left ticks from batch anatomy — the generation spans themselves carry the version color now, so the extra column was redundant. Detailed-mode left margin shrinks 150→96px; legend cleaned up. Note on the uniform color in the current run: the dump records weight_version=1 for all samples of all three steps (core-side accounting), so a single hue is the faithful rendering.

yueming-yuan:
avg staleness stat (force-pushed `53b2f65ec`, PR-22): the step page statgrid gains **avg staleness** = mean over samples of (trainer version at consume − sample's oldest generation version), with trainer-at-step-N = v(N+1) (one startup push + one per train step). Train view only; renders — when version data is absent. Verified on the live run: step 0 → 0, step 2 → 2 (trainer at v3 consuming v1 data).

yueming-yuan:
Per-token metrics chart (force-pushed `edf549dd`→tip, PR-6): the L2 "imp_ratio per token position" chart is now a general **metric vs token position** panel — a button row selects the metric (imp_ratio / lp_diff / advantages / returns / train / rollout / ref log-probs, whatever the dump carries), the chart spans the whole response (fetched once, independent of the token-strip window), and drag-zoom selects a token range with double-click reset. `drawChart` gained generic auto-bucketing: above 700 in-domain points, adjacent tokens collapse into mean buckets drawn as a polyline (hover shows `pos a–b · mean (n pts)`); zooming in re-buckets until raw per-token points with dots reappear. Switching metric keeps the x-range and re-autoscales y. Verified on the live run against an 8337-token sample: full-view bucketed line, drag → tokens 510–1804, chip switch preserves range.

yueming-yuan:
**PR-24 + PR-25 end-to-end verified on a real multi-turn run** (tip `fde52b9f0`).

Run recipe (simplest trajectory trigger, zero new code): the qwen30b async recipe with `--custom-generate-function-path miles.rollout.generate_hub.multi_turn.generate` + retool_v2's `code_interpreter` sandbox tools, on 8 tool-forcing micro-prompts. Result: **64/64 samples recorded conversations with real tool calls**; the sample page renders USER → ASSISTANT (`<think>` + `<tool_call>{"code": "print(2 ** 20)"}`) → TOOL·CODE_INTERPRETER (`Output: 1048576`) → ASSISTANT (`Answer: \boxed{1048576}`), coexisting with the token strip and per-token metrics.

Field notes from getting there, each worth knowing:
- `--generate-*` args only register under `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1` (the arg-hook is gated; the test harness sets it, real launches must too);
- tools reach the chat template only when `sample.prompt` stays a message list — `--apply-chat-template` pre-renders to a string and silently drops tool advertisement for the engine multi-turn path (retool_v2 sidesteps this via its SFT model);
- base Qwen3-30B on dapo prompts never volunteers tool calls even with tools advertised — instruction-level forcing was needed;
- `examples/fully_async/fully_async_rollout.py:288,324` crash on list prompts (`prompt + response` debug prints) — two-line upstream fix worth submitting separately.

Also in this push (PR-18): `time_range()` no longer leaks the open-interval `t1=-1.0` sentinel through partition `edge_stamps` — the epoch-scale `+29734410:00` axis-label bug; `line_stamps` now reuses each record's own `timestamps()`.

yueming-yuan:
L2 split into tabs (force-pushed, PR-25): the sample page now opens on a **conversation | tokens** tab bar when a trajectory sidecar exists — conversation (sidecar read, instant) is the default; the token machinery (probe → full dump load → detokenize, the several-minutes first-open cost) fires only when the tokens tab is first clicked (verified: zero /tokens requests before the click). Runs without a sidecar keep today's behavior: no tab bar, tokens load directly.

yueming-yuan:
Raw-assistant display parsing (force-pushed, PR-25): the engine multi_turn path records the model's verbatim emission, so the conversation renderer now does a display-level split when the structured fields are absent — `<think>…</think>` becomes the italic thinking block, `<tool_call>…</tool_call>` becomes a name chip with pretty-printed JSON arguments, and trailing special tokens (`<|im_end|>` etc.) are stripped; unknown markup stays visible verbatim. Session-path rows with structured `reasoning_content`/`tool_calls` render as before; the dump itself stays raw (display-only transform). Verified on the live retool run: think block + `code_interpreter` chip + clean final answer.

Zhichenzzz (approved):
LGTM, thx!

fzyzcjy (approved):

PR #1660, opened 2026-07-14
(3/N) test(session): session-server overhead benchmark on the multi-instance topology

guapisolo:
> **Depends on:** #1659

## Summary
Manual HTTP benchmark of session-server per-turn overhead on the multi-instance topology.

## Motivation
The overhead benchmark from #1564 gated the multi-process stack; the multi-instance design needs the same instrument. It stands up the real production server processes against a mock R3 backend, so the server pays its real JSON / TITO / record-store cost while model generation stays off the clock.

## Usage
```
python tests/manual/session/bench_session_server_overhead.py \
    --sessions 32 --turns 50 --r3-scale 1000 --session-server-instances 16
```
Measured at this shape on 192 cores, 16 instances vs 1: 8.2x wall-time and throughput (58.1 to 476.5 turns/s), mean chat latency 542ms to 41ms, p99 1190ms to 114ms, 1600/1600 turns clean in both runs, peak server-group RSS 12.5 to 34.4 GiB.

## Design Notes
- **Key choices:**
  - One `run_session_server` process per consecutive port, all spawned before any readiness wait — the production launch pattern.
  - Each session picks one instance URL at create time and stays on it, mirroring `OpenAIEndpointTracer.create`.
  - The mock backend replays pre-built response bytes, SO_REUSEPORT-sharded so it never caps the measurement.
- **Alternatives considered:** none — the first three commits are #1564 cherry-picked unchanged; only the last commit retargets the topology.

## Verification
- **Tests added:** manual benchmark, not CI; smoke-verified at 2 instances x 4 sessions, plus the production shape above, with zero chat / GET errors.

## Review Focus
- Scrutinize only the last commit (the retarget); the first three are verbatim #1564.
- Scrutinize the per-session `random.choice(base_urls)` in `lg_drive_all` — the bench's session affinity must match the production client.


Shi-Dong (approved):
LGTM!

PR #1673, opened 2026-07-14
[feat] support `sample` rollout submission granularity to keep fully async concurrency

ys-2020:
## Motivation

In group-based RL (GRPO-style, `n_samples_per_prompt` samples per prompt) with long-horizon agentic rollouts, the async schedulers currently backfill at **group** granularity: a replacement group is only submitted when an entire group task returns. When individual trials run 20-120 minutes (e.g. SWE-agent tasks at 64K context), every group is gated by its slowest sibling, and engine capacity idles while stragglers finish. In our production runs (n=8, 32 concurrent groups on SWE tasks) this straggler gating was a primary rollout-throughput limiter.

## What this PR does

- Adds two submission schedulers (`miles/rollout/submission_scheduler.py`), shared by the sync rollout driver (`generate_rollout_async`) and the fully-async worker loop: `GroupLevelSubmission` (the pre-existing pacing, verbatim) and `SampleBackfillSubmission` (counter + wakeup event), selected by `make_submission_scheduler(args, default=...)`.
- Adds **sample-completion backfill** as a submission policy: each finished sample fires a completion credit via `task.add_done_callback`, and once `n_samples_per_prompt` credits accumulate — from whichever groups they came — one replacement prompt group is submitted immediately, instead of waiting for whole groups to return.
- Adds `--rollout-submission-granularity {group,sample}`, default unset. Each driver resolves its own default when unset (see below); passing the flag overrides both.
- Wires an optional `sample_done_callback` through `generate_and_rm_group` and both inference-rollout drivers, plus the fully-async example worker loop (refactored around a `submit_one_group()` helper).

## Defaults, and why they differ per driver

The two drivers pay very different prices for overshooting a step's batch, so they default differently:

| Driver | Default | Why |
|---|---|---|
| `FullyAsyncRolloutFn` (`--fully-async`) | `sample` | Groups completed beyond a step's batch stay in the output queue and are consumed by later steps, so backfilling past a group's straggler costs nothing. |
| `generate_rollout_async` (sync / one-step async) | `group` | The step ends as soon as `rollout_batch_size` groups complete and `abort()` drops everything still pending — without `--partial-rollout` those prompts are consumed from the data source and discarded, so backfill trades wasted generation for lower step latency. Opt in with `--rollout-submission-granularity sample`. |

Under `sample` granularity the in-flight bound is on unfinished *samples*, not on live group tasks: a group releases all its credits at `asyncio.gather` while its task is still pending. When one sibling per group straggles, replacements compound to roughly `rollout_batch_size × n_samples_per_prompt` submitted groups. That is the intended trade on the fully-async path and the reason the sync path does not take it by default.

Scope note: only the class-based rollout path is affected (`MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1` → `InferenceRolloutFn` / `FullyAsyncRolloutFn`). The default `--rollout-function-path`, `miles.rollout.sglang_rollout.generate_rollout`, has its own `generate_and_rm_group` and is untouched, so the new flag is a no-op there.

## Testing

- New CPU CI tests (`tests/fast/rollout/inference_rollout/test_sample_completion_backfill.py`):
  - `generate_rollout_async` defaults to `group` granularity and wires no sample callback;
  - with `sample` granularity, a replacement group is scheduled after `n_samples_per_prompt` sample completions, before any group task returns;
  - no replacement is scheduled one sample short of a full group;
  - scheduler unit tests: factory resolution, capacity accounting under both granularities, orphaned-credit resync, and `wait_for_progress` returning on either a group completion or a single sample completion.
- `tests/fast/rollout/test_fully_async_rollout.py`: the worker defaults to `sample` granularity; `--rollout-submission-granularity group` opts it back out; existing stubs updated for the new `sample_done_callback` parameter.
- Full PR CI green on `acba7df07`: **24 passed, 0 failed** (stage-a/b CPU shards plus the label-gated `stage-b/c` H100/H200 GPU suites).

https://claude.ai/code/session_013vyY58s2TFpCzqEP4KaSxC


miles/rollout/inference_rollout/inference_rollout_train.py
```diff
@@ -138,26 +195,114 @@ async def generate_rollout_async(
...
     )
+    submit_groups(target_data_size)
 
-    # reset the global state to prevent effects on the next rollout or eval.
-    state.reset()
+    while len(data) < target_data_size:
+        if not pendings:
+            # Defensive fallback for group-level task exceptions. Normal flow keeps
+            # pending sample slots replenished from sample completion credits.
+            submit_groups(max(1, target_data_size - len(data)))
```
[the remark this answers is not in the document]
ys-2020: Good catch — fixed in f814b44: `submit_groups` returning 0 with nothing in flight now breaks out of the loop instead of blocking on the sample-completion queue.

miles/rollout/inference_rollout/inference_rollout_common.py
```diff
@@ -132,9 +137,10 @@ async def generate_and_rm_group(
         current_sampling_params = sampling_params.copy()
         if getattr(args, "sglang_enable_deterministic_inference", False):
             current_sampling_params["sampling_seed"] = args.rollout_seed + idx
-        tasks.append(
-            asyncio.create_task(generate_and_rm(state, sample, current_sampling_params, evaluation=evaluation))
-        )
+        task = asyncio.create_task(generate_and_rm(state, sample, current_sampling_params, evaluation=evaluation))
+        if sample_done_callback is not None:
+            task.add_done_callback(lambda _task: sample_done_callback())
```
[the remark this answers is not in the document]
ys-2020: This is deliberate, and conditioning on success would introduce a subtler bug: the callback fires exactly once per submitted sample task (success, exception, or cancellation), so completion credits *conserve in-flight sample concurrency exactly* — every n-credit batch corresponds to exactly one freed group slot. If failed samples were skipped, credits would leak and concurrency would decay monotonically with every failure (unless paired with group-level replenish, which then over-submits on partial group failures — see the reply on the other thread). A failed sample does free engine capacity, so counting it is the semantically correct choice for a throughput scheduler. Documented the invariant at this call site in f814b44.

miles/rollout/sglang_rollout.py
```diff
@@ -325,9 +329,10 @@ async def generate_and_rm_group(
         if getattr(args, "sglang_enable_deterministic_inference", False):
             seed = state.group_sampling_seeds[idx]
             current_sampling_params["sampling_seed"] = seed
-        tasks.append(
-            asyncio.create_task(generate_and_rm(args, sample, current_sampling_params, evaluation=evaluation))
-        )
+        task = asyncio.create_task(generate_and_rm(args, sample, current_sampling_params, evaluation=evaluation))
+        if sample_done_callback is not None:
+            task.add_done_callback(lambda _task: sample_done_callback())
```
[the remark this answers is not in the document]
ys-2020: Same rationale as in inference_rollout_common.py: credits deliberately fire for every sample outcome so that in-flight concurrency is conserved exactly; conditioning on success would leak credits. Comment added in f814b44.

miles/rollout/inference_rollout/inference_rollout_train.py
```diff
@@ -138,26 +195,114 @@ async def generate_rollout_async(
...
-        args,
-        [sample for group in data for sample in group],
-        url=f"http://{args.sglang_router_ip}:{args.sglang_router_port}/generate",
-        sampling_params=state.sampling_params,
+        for task in group_done:
+            try:
+                group: list[Sample] = task.result()
+            except Exception as e:
+                logger.error(f"[rollout] Task raised exception: {e!r}", exc_info=True)
+                continue
```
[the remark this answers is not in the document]
ys-2020: The premise doesn't hold for the current implementation: when a group task raises, its per-sample done callbacks still fire (they're attached to the individual sample tasks, which run to completion independently of the group-level gather), so the failed group's n credits still arrive and trigger exactly one replacement group — the slot is not lost. Adding `submit_groups(1)` here on top of that would double-replenish: e.g. a group where 1 of 8 samples raises would emit 7-8 credits *and* an immediate extra group, over-submitting by up to ~2x per failed group. Keeping replenishment purely credit-driven preserves the exact conservation invariant.

examples/fully_async/fully_async_rollout.py
```diff
@@ -96,6 +96,54 @@ async def continuous_worker_loop(self):
...
+                    generate_kwargs["sample_done_callback"] = on_sample_done
+
+                # Create new async task
+                task = asyncio.create_task(generate_and_rm_group(**generate_kwargs))
+
+                # Add completion callback
+                def make_callback(gid):
+                    def task_done_callback(done_task):
+                        result = done_task.result()
+                        self.output_queue.put((gid, result))
```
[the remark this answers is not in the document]
ys-2020: This pattern is pre-existing on main (the same `output_queue.put` in the task callback, this PR only moved it into `submit_one_group()`), and the full-queue condition is unreachable in practice: in-flight groups are capped at `rollout_batch_size` (typically 32-64) while the queue's `maxsize` is 1000 and the consumer drains continuously. Happy to switch to `put_nowait` with an executor fallback in a follow-up if maintainers prefer, but keeping this PR's surface minimal.

miles/utils/arguments.py
```diff
@@ -458,6 +458,17 @@ def add_rollout_arguments(parser):
                     "You could use `miles.rollout.filter_hub.dynamic_sampling_filters.check_reward_nonzero_std` as an example."
                 ),
             )
+            parser.add_argument(
+                "--rollout-sample-completion-backfill",
+                action="store_true",
+                default=False,
```
nblintao: I think this is strictly better in most cases. Are we going to make the default to be true or even deprecate the legacy behavior in the future?

yueming-yuan:
Rebased onto #1717 (fully-async rollout rewritten as a class-based `FullyAsyncRolloutFn`) and reworked the implementation to fit it — force-pushed, the previous 3 commits are replaced by one. Original authorship preserved on the commit. Summary of what changed and why:

**The driver is no longer forked.** The old version duplicated the whole rollout loop into `_generate_rollout_group_level_async` + `_generate_rollout_sample_completion_backfill_async` (+181 lines, with the filter/print/append block copied verbatim into both — a permanent divergence hazard). The two policies actually differ in only two decisions: *when may another group be submitted* and *what counts as progress*. Both are now owned by one small `SubmissionScheduler` in `inference_rollout_common`:

```python
while len(data) < target_data_size:
    scheduler.arm()
    while scheduler.has_capacity(pending_groups=len(pendings), group_budget=target_data_size - len(data)):
        samples = data_source(args.over_sampling_batch_size)
        scheduler.on_submit(samples)
        pendings.update(submit_generate_tasks(state, samples, scheduler.sample_done_callback))
    done, pendings = await scheduler.wait_for_progress(pendings)
```

Disabled, `has_capacity` counts groups and `wait_for_progress` is plain `asyncio.wait` — bit-identical to today's default path, and no callback is even wired. Enabled, slots are freed per sample and the wait also returns on a single sample completion. Net: the sync driver is +14 lines instead of +181, with zero duplicated logic.

**The same scheduler serves the fully-async worker** (`FullyAsyncRolloutFn._worker_loop`), so the feature works in both drivers from one implementation. The old example-worker code path this PR used to patch no longer exists.

**Credit accounting is gone.** `asyncio.Queue[int]` + drain-loop + `sample_done_credit` + `accept_sample_done` collapse into an `Event` plus a `samples_in_flight` counter that the submit predicate reads directly. `arm()` clears stale notifications before the capacity check; since callers never await between `arm()` and `wait_for_progress()`, no completion can be lost.

**Dropped, as dead code:**
- the `sglang_rollout.py` hunk — after #1717 nothing passes `sample_done_callback` to the legacy stack (its driver never supported backfill);
- the data-source-exhaustion guard — `RolloutDataSource.get_samples` always returns the requested count (it rolls to the next epoch), so that branch was unreachable.

**Budget semantics.** Backfill now keeps `(target - collected) * n` samples in flight, i.e. the exact sample-level analogue of the existing group rule, so oversampling shrinks as the batch fills instead of staying pinned at full width.

**Tests** now assert behavior rather than which internal function got called (those functions are gone): a replacement group goes out after n sample completions *while the groups are still pending*; nothing is submitted one sample short of a group; default-off wires no callback at all; plus direct `SubmissionScheduler` unit tests and a fully-async worker case. 16 pass locally.

Note the base is now `yueming/fully-async-class-api` (#1717, which stacks on #1716). Happy to split the sync-driver half back onto main if you'd rather it land before that stack — it's an independent cherry-pick.

miles/utils/arguments.py
yueming-yuan: agree

yueming-yuan (approved):

guapisolo:
I am curious why we didn't directly impl `SampleLevelSubmission` but a `SampleBackfill` as instead? @yueming-yuan  Is it hard to control staleness or starve, or easy to impl, or other reason?

guapisolo:
Why I ask this question is b/c, I think in very large scale training on extermely hard tasks. People will set grpo group to 64 or even 128 with 256 prompts. But I am worried 64 or 128 are still not fine-grained enough and will create big bubble?  (batch size per sglang engine maybe <= 32 for agentic tasks).

miles/rollout/submission_scheduler.py
```diff
@@ -0,0 +1,63 @@
...
+class GroupLevelSubmission:
+    """A submission slot frees only when the whole group task returns."""
+
+    sample_done_callback = None
+
+    def has_capacity(self, *, pending_groups: int, group_budget: int) -> bool:
+        return pending_groups < group_budget
+
+    def on_submit(self, groups: list[list[Sample]]) -> None:
+        pass
```
guapisolo, line 25: Better add cmt: pending_groups counter is hold in external variable.

miles/rollout/submission_scheduler.py
```diff
@@ -0,0 +1,63 @@
...
+        self._sample_done = asyncio.Event()
+
+    def sample_done_callback(self) -> None:
+        self.samples_in_flight -= 1
+        self._sample_done.set()
+
+    def has_capacity(self, *, pending_groups: int, group_budget: int) -> bool:
+        """A False return arms the sample wakeup: only later completions wake ``wait_for_progress``."""
+        if pending_groups == 0:
+            # nothing in flight: drop orphaned credits from groups that never spawned sample tasks
```
guapisolo: Is it used for boundary cases? Why add this. I am not sure whether this should be an assert

guapisolo:
## Codex question

Could the following edge case occur?

```text
Submit a group; count += 4
    ↓
One sample raises an exception
    ↓
asyncio.gather causes the parent group task to exit early
    ↓
The other 3 sibling sample tasks are still running
    ↓
pendings becomes empty
    ↓
has_capacity() resets samples_in_flight to 0
    ↓
Submit a replacement group; count = 4 again
    ↓
Callbacks from the old siblings arrive
    ↓
They incorrectly decrement the accounting for the replacement group
```

If this edge case is possible, the late callbacks would be charged against the replacement group, making the capacity accounting inaccurate.

guapisolo (approved):

miles/rollout/submission_scheduler.py
yueming-yuan: add at the top of the file

miles/rollout/submission_scheduler.py
yueming-yuan: added a logging

PR #972, opened 2026-04-11
[fix] [test] attention_output_gate TP slice when num_kv_heads < TP

Zhichenzzz:
When num_query_groups < tensor_parallel_size (e.g. num_kv_heads=2, TP=4), Megatron slices query heads per TP rank but forgets to slice the gate tensor in SelfAttention.get_query_key_value_tensors, causing gate.view(*x.shape) to fail in _apply_output_gate. This affects Qwen3.5 and Qwen3-Next models.

Fix: after gate reshape, apply the same TP slice as query when num_query_groups < world_size.

Verified with Qwen3.5-35B-A3B TP=4 CP=2 EP=8 (4 training steps, all passed, logprob_abs_diff stable at 0.01-0.02).

PR #1675, opened 2026-07-14
openenv/tbench2: Daytona sandbox mode — one cloud sandbox per episode, as its own agent function

nblintao:
Final PR of the stack (#1710 sandbox recipe → #1711 orphan TTL + ownership labels → this): the adapter-side Daytona sandbox mode. The existing shared-server path is behaviorally unchanged (asserted by a dispatch unit test).

## What

A second agent function for the OpenEnv TB2 adapter: point `--custom-agent-function-path` at **`openenv_daytona_agent_function.run`** (the launcher does this automatically when `OPENENV_TB2_TASKS_DIR` is set) and **every episode runs in its own Daytona cloud sandbox**, built from the task's official image (`task.toml` `docker_image`) plus a tbench2_env server layer, deleted when the episode ends.

Same per-task image fidelity as docker mode, with zero shared infrastructure: no Docker socket, no env server to size or babysit (no `MAX_CONCURRENT_ENVS` / `ulimit` tuning), no cross-episode state. Cost profile: the first episode of a task builds its image (~10 min, layer-cached by definition hash), repeats start in ~1 min; no named snapshots, so no org snapshot quota. Sandboxes carry ownership labels and a TTL dead-man's switch (#1711), and creation is throttled process-wide with jittered backoff (Daytona rate-limits creates).

## Architecture (shaped by review)

Two agent-function modules with the identical public shape — **`run()`** for miles (session-server policy wiring + training failure semantics: timeout → reward 0, no verdict → drop sample) and **`run_episode()`** for callers that bring their own policy client and failure semantics (`eval_tbench2_via_api`):

- `openenv_agent_function` — the shared env server leg, and home of the shared agent loop + training wrapper.
- `openenv_daytona_agent_function` — the Daytona sandbox mode; holds all Daytona-specific machinery (throttled creates, cancel-reap, per-episode env lifecycle).

Everything that differs between the legs enters the shared loop as three keyword parameters filled in only by each module's `run_episode`: `run_body` (how an env comes into being), `native_evaluate` (server contract, below), `post_episode` (shared-server trial-dir hygiene). No backend objects, no env-var-sniffing dispatch. **Behavior change**: setting `OPENENV_TB2_TASKS_DIR` while pointing at `openenv_agent_function.run` no longer switches modes implicitly — mode selection is explicit plugin selection.

## Why the two legs score differently (`native_evaluate`)

Each leg matches what its env server actually provides:

- **Shared-server leg** (unchanged): adapter-driven canonical exec + reward-marker parse, compensating for an UNMODIFIED older upstream tbench2_env server.
- **Daytona sandbox mode** (new): the standard `evaluate` action. The recipe (#1710) bakes the installed `tbench2_env` into each task image; upstream main (≥ huggingface/OpenEnv#965 + #972) carries the server-side contract this leg scores through — canonical `tests/test.sh` run natively, task WORKDIR resolved server-side (WORKDIRs are not uniformly `/app`), and verifier assets withheld outside the verify window (`TB2_WITHHOLD_TESTS`, the reward-hacking fix). The launcher preflights the installed server source for exactly this contract: an older install wouldn't fail per-episode, it would silently mis-score every episode.

Once the shared-server deployment upgrades to current upstream, a follow-up can flip `native_evaluate` there and retire the compensation machinery.

## Daytona key supply (review follow-up)

The key value never rides ray's logged channels (the submit command is echoed into driver logs; runtime_env persists in job metadata). Workers resolve `DAYTONA_API_KEY` from their own node-local environment (platform-injected, or single-host inheritance) first, else read a key file (`DAYTONA_API_KEY_FILE`, default `~/.config/daytona/api_key` — a dotfile, K8s Secret mount, or shared-FS path). The launcher forwards only the file PATH and fails fast with provisioning guidance when neither supply is available. `requirements.txt` now also declares `openai` (already load-bearing in miles core; previously satisfied only transitively via sglang).

## Tooling (ships with the mode; produced the validation evidence)

- `scan_golden.py` — infra baseline with no LLM: replays each task's OFFICIAL `solution/solve.sh` (oracle-faithful, chunk-uploaded from the local checkout since the image withholds verifier assets) through the same sandbox + `evaluate` scoring; `--logs` captures solve/test log tails for anything below 1.0.
- `eval_tbench2_via_api.py` — the exact training agent-env loop with any OpenAI-compatible API as the policy (no GPU, no Ray/Megatron), via each module's `run_episode`.
- `tests/` — 20 offline unit tests (`pytest examples/experimental/openenv/tests/ -q`): episode dispatch on both legs through the public `run_episode` entries, create-throttling retry/backoff/give-up + cancel-reap, typed throttle-error classification, key-resolution chain, recipe hygiene, labels, TTL keepalive lifecycle.

### Post-refactor re-validation (4×H200 devbox, full stack green)

Re-ran the whole pipeline against the **final two-module form** to confirm the split (own agent function + `native_evaluate` params + no backend objects) broke nothing. `tbench2_env` installed from **upstream OpenEnv main** (`39c91bf`, the #965/#972 merge the Daytona leg scores through):

- **Offline** — `20 passed` (`pytest examples/experimental/openenv/tests/ -q`).
- **Daytona leg, no LLM (scan_golden.py)** — golden **3/3** (`fix-git`, `chess-best-move`, `circuit-fibsqrt` all 1.0) through the real per-task-image sandbox + native `evaluate`.
- **Daytona leg, live API policy (eval_tbench2_via_api.py)** — `fix-git` solved by DeepSeek-V4-Flash through the Daytona sandbox (dispatch confirmed `env=daytona sandboxes`).
- **Shared-server leg, real Docker fidelity** — ran the shared leg (public `run_episode`, `OPENENV_TB2_TASKS_DIR` unset → dispatch selects `openenv_agent_function`, not the Daytona module) end-to-end against a live **`TB2_MODE=docker` `tbench2_env` server** (per-task official-image containers), with DeepSeek-V4-Flash as the policy: **1/2 tasks solved, 0 errored**, `reward` correctly `1.0` (real `tests/test.sh` passed in the real container) vs `0.0` (real fail) — the `native_evaluate=False` canonical-exec scoring branch (byte-identical to pre-split, just relocated) produces correct rewards against a real Docker-mode server. Also covered by the dispatch unit test.
- **Full GRPO loop closure on 4×H200** (GLM-4.7-Flash colocated, **TP=2 EP=2 DP=2** — the 4-GPU analog of the 8-GPU recipe; `torch_dist` checkpoint reshards automatically from its saved `(TP1,PP8)` layout): `rollout 0` (2 `fix-git` episodes via Daytona sandboxes, `raw_reward` 0.5) → `train_step outcome=NORMAL` → weight update (461 tensors broadcast to the SGLang engines) → `rollout 1` with updated weights (`raw_reward` 1.0) → `train_step outcome=NORMAL`. On-policy fidelity: rollout vs. ref logprobs `-0.535`/`-0.543`, KL ≈ 0.0115 (token-faithful TITO recording). Robustness confirmed under a saturated org: while the Daytona org CPU quota was fully consumed by an unrelated resident pool, the leg's `DaytonaValidationError` (`Total CPU limit exceeded`) was caught per-episode and the sample dropped cleanly across hundreds of failed creates — no crash, no leaked sandbox.

### Prior evidence (larger scale, pre-split)

- **scan_golden.py** — golden sweep **82/89** of the TB2 suite (0 infra errors; the 7 remaining have upstream-broken solutions, each with solve.log evidence).
- **eval_tbench2_via_api.py** — full 89-task sweep with DeepSeek-V4-Flash (single sample, 30-turn cap, up to 38 concurrent episodes) — **33/89 solved, 0 systematic infra failures**.
- **GRPO smoke on 8×H200** (TP=4 EP=2, 8-task subset): rollout 16/16 episodes through Daytona sandboxes, train step `outcome=NORMAL`, weight update, next rollout — loop closed; prefix-cache hit rate 0.96; sandboxes carried the TTL/ownership contract in production.


examples/experimental/openenv/openenv_launch_common.py
```diff
@@ -147,8 +149,12 @@ def base_env_vars(args: LaunchArgs, script_dir: str, megatron_path: str, miles_r
...
 
 def apply_optional_env_vars(env: dict[str, str], args: LaunchArgs) -> None:
-    """Add host-rewrite env vars when the args request them."""
+    """Add host-rewrite / per-task-Daytona env vars when the args request them."""
     if args.miles_host_ip:
         env["MILES_HOST_IP"] = args.miles_host_ip
     if args.router_external_host:
         env["MILES_ROUTER_EXTERNAL_HOST"] = args.router_external_host
+    if args.openenv_tb2_tasks_dir:
+        assert args.daytona_api_key, "DAYTONA_API_KEY required in per-task Daytona mode"
```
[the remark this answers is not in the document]
nblintao: Adopted in b604dc9 — switched to `raise ValueError`. Also makes this file consistent with the adapter, which already raises ValueError for the missing-task_id case.

examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -179,6 +199,117 @@ def _load_tbench2() -> dict[str, Any]:
...
+
+
+def _per_task_mode() -> bool:
+    """True when episodes run in per-task Daytona sandboxes instead of OPENENV_ENV_URL."""
+    return bool(os.getenv("OPENENV_TB2_TASKS_DIR", "").strip())
+
+
+def _is_throttle_error(exc: BaseException) -> bool:
+    s = str(exc).lower()
+    return "throttler" in s or "too many requests" in s or "429" in s
```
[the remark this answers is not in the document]
nblintao: Deliberate, and now documented in the function docstring (b604dc9): the daytona SDK is a lazy, per-task-mode-only dependency — shared-server users don't install it — so an isinstance check on its exception classes isn't available at module scope. Rate limits also surface in several shapes across layers (SDK ThrottlerException, HTTP 429 text, wrapped errors), which one class check wouldn't cover. Same approach as `_is_retryable_env_error` above, which matches websockets close exceptions by name for the same no-hard-dependency reason.

examples/experimental/openenv/scan_golden.py
```diff
@@ -0,0 +1,126 @@
...
+                try:
+                    import tomllib
+
+                    cfg = tomllib.loads(
+                        (Path(os.environ["OPENENV_TB2_TASKS_DIR"]) / task_id / "task.toml").read_text()
+                    )
+                    for k, v in (cfg.get("solution", {}).get("env", {}) or {}).items():
+                        sol_env += f" {k}={v!r}"
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
nblintao: Adopted with a stronger fix in b604dc9: this tool only exists on the per-task backend at all, so `main()` now fails fast with a clear message when `OPENENV_TB2_TASKS_DIR` is unset. That makes `golden_one`'s `os.environ[...]` read unconditionally valid, and the try/except is left tolerating only its real target — a task.toml without `[solution].env`.

examples/experimental/openenv/openenv_agent_function.py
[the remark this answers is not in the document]
nblintao: Update (1f9ed50e2): partially adopting after all — the module-scope constraint stands, but by the time a create has FAILED the daytona SDK has necessarily been imported, so a lazy in-function import can do a typed check. `_is_throttle_error` now checks `isinstance(exc, DaytonaRateLimitError)` first (the SDK normalizes HTTP 429 to it) and keeps the text match only as a fallback for older SDKs / text-only server messages. Covered by a new unit test asserting the typed path matches even when the message carries no throttle keywords, and that sibling error classes do not.

examples/experimental/openenv/openenv_launch_common.py
```diff
@@ -147,8 +150,48 @@ def base_env_vars(args: LaunchArgs, script_dir: str, megatron_path: str, miles_r
...
+        src_text = server_src.read_text(encoding="utf-8") if server_src.is_file() else ""
+        if "TB2_WITHHOLD_TESTS" not in src_text:
+            raise RuntimeError(
+                "the installed tbench2_env server lacks the per-task sandbox "
+                "contract (canonical test.sh scoring / TB2_WITHHOLD_TESTS): "
+                "install the pinned checkout from this directory's README, "
+                "not upstream main"
+            )
+        env["OPENENV_TB2_TASKS_DIR"] = args.openenv_tb2_tasks_dir
+        env["DAYTONA_API_KEY"] = args.daytona_api_key
```
Shi-Dong: Hmm if we save `DAYTONA_API_KEY` here, the ray job might print it out in the logs, which means that the key might be leaked.

examples/experimental/openenv/eval_tbench2_via_api.py
```diff
@@ -0,0 +1,141 @@
...
+  python eval_tbench2_via_api.py --data /root/tbench2_train.jsonl
+"""
+
+import argparse
+import asyncio
+import json
+import os
+import sys
+
+from openai import AsyncOpenAI
```
Shi-Dong, line 45: Probably we should declare `openai` in Miles requirements?

examples/experimental/openenv/openenv_agent_function.py
Shi-Dong, line 1: I'd suggest leaving `openenv_agent_function.py` as-is, and create another script for the Daytona code path. Daytona is only one of the many sandbox providers that users can choose from, and mixing it with the generic sandbox-agnostic code path feels a little bit weird 😂

examples/experimental/openenv/eval_tbench2_via_api.py
nblintao: Added. `openai` was already in `openenv_agent_function.py` before but we were probably using the dependency from sglang. It's nicer to make it explicit.

examples/experimental/openenv/openenv_launch_common.py
nblintao: Good catch! Fixed. I changed it to read from server env or file directly, instead of passing through ray env.:
```
  DAYTONA_API_KEY              the Daytona API key, authenticating every
                     sandbox create/delete. Read from the worker's own
                     node-local environment; nothing forwards it. Supply it
                     via platform-injected pod env, or by exporting it in
                     the shell that starts ray on a single host.
  DAYTONA_API_KEY_FILE         fallback when DAYTONA_API_KEY is unset: path
                     of a file holding the key (default
                     ~/.config/daytona/api_key). Launchers forward this path
                     instead of the key itself, because ray runtime_env is
                     logged in plaintext. Point it at a file every node can
                     read: a dotfile, K8s Secret mount, or shared-FS path.
```
I noticed that wandb keys may have similar issue. Let me create a separate PR to fix that.

examples/experimental/openenv/openenv_agent_function.py
nblintao: Makes sense. Fixed. I also made some more refactors to make is more clear. 

I think we can also deprecate the adaptor path since the fixes are already in openenv. But we can do it in a separate PR.

examples/experimental/openenv/README.md
```diff
@@ -50,6 +50,55 @@ concurrency. Per-task containers are heavy on disk — if you'd rather not coloc
 them with the GPU workload, run the env server on a separate Docker host and point
 the launcher at it via `--openenv-env-url http://<env-host>:8003`.
 
+### 2b. Alternative: Daytona cloud sandboxes (no Docker host)
```
nblintao, line 53: I'll have another round of polish after some more factorings in a separate PR

nblintao:
I've re-ran the validation. See the updated summary.

examples/experimental/openenv/openenv_daytona_agent_function.py
```diff
@@ -0,0 +1,249 @@
...
+    The daytona SDK is a lazy dependency of this module only (shared-server
+    users don't install it), so its exception classes cannot be imported at
+    module scope -- but by the time a create has FAILED, daytona has
+    necessarily been imported, so the typed check happens here. The SDK
+    normalizes HTTP 429 to DaytonaRateLimitError; keep the text match as a
+    fallback for older SDKs and server messages that only surface as text
+    (e.g. "ThrottlerException: Too Many Requests").
+    """
+    try:
+        from daytona.common.errors import DaytonaRateLimitError
```
Shi-Dong, line 94: please hoist this import to the beginning of the script

examples/experimental/openenv/openenv_daytona_agent_function.py
```diff
@@ -0,0 +1,249 @@
...
+
+def _get_create_sem() -> asyncio.Semaphore:
+    global _create_sem
+    if _create_sem is None:
+        _create_sem = asyncio.Semaphore(_CREATE_CONCURRENCY)
+    return _create_sem
+
+
+def _start_declarative(task_id: str, tasks_dir: str) -> tuple[Any, str]:
+    import tb2_sandbox_daytona
```
Shi-Dong: please hoist this import to the beginning of the script

examples/experimental/openenv/scan_golden.py
```diff
@@ -0,0 +1,162 @@
...
+        async def run() -> dict:
+            async with odaf._episode_env(classes["env"], {"task_id": task_id}) as env:
+                await env.reset(task_id=task_id)
+                t = time.monotonic()
+                # Official oracle convention (harbor OracleAgent): solution dir
+                # staged at /solution, DEBIAN_FRONTEND=noninteractive, task.toml
+                # [solution].env exported, cwd = task workdir.
+                sol_env = "DEBIAN_FRONTEND=noninteractive"
+                try:
+                    import tomllib
```
Shi-Dong: please hoist this import to the beginning of the script

examples/experimental/openenv/openenv_agent_function.py
Shi-Dong: Sounds good

examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -301,43 +400,29 @@ async def body(env: Any) -> tuple[float | None, int, list[float], list[float], f
...
-            )
-        except Exception:
-            pass
+        if native_evaluate:
+            # The server runs the canonical test.sh natively inside the
+            # standard evaluate action; there is no adapter-side test.sh
+            # exit-code marker to parse.
+            eval_result = await env.step(action_cls(action_type="evaluate"))
+            eval_time = time.monotonic() - t0
+            reward = float(getattr(eval_result, "reward", 0.0) or 0.0)
```
Shi-Dong: If there is an error in task eval, probably we should just leave `reward = None`?  In the PR description it says "no verdict → drop sample", but it seems here that no verdict returns zero reward.

Shi-Dong (approved):
LGTM with minor comment.

examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -301,43 +400,29 @@ async def body(env: Any) -> tuple[float | None, int, list[float], list[float], f
...
+            testsh_rc = None
+        else:
+            # Older server: adapter-driven canonical exec + marker parse.
+            eval_result = await env.step(action_cls(action_type="exec", command=_CANONICAL_EVAL_CMD))
+            eval_time = time.monotonic() - t0
+            eval_output = _obs_field(eval_result, "output")
+            reward = _parse_reward_marker(eval_output)
+            testsh_rc = _parse_testsh_rc(eval_output)
```
Shi-Dong, line 428: I think we can probably remove this branch since the changes have already landed in OpenEnv?

nblintao: Yes. I‘ll remove that in a separate PR. Note the non-daytona path (the one you created) is still using this for now. In openenv, part of the fixes s only done on local mode. I need to fix docker too.

examples/experimental/openenv/openenv_agent_function.py
nblintao: Ahh, good catch. Fixed.

examples/experimental/openenv/openenv_daytona_agent_function.py
nblintao: Done.

examples/experimental/openenv/openenv_daytona_agent_function.py
nblintao: I think it's better to keep this one in function because of SDK compatibility issue. I've updated the details in the docstring.

examples/experimental/openenv/scan_golden.py
nblintao: Done.

PR #1677, opened 2026-07-15
Add --async-max-concurrent-samples to decouple fully-async generation concurrency from batch size

yueming-yuan:
## Problem

In fully async mode the generation concurrency is hard-wired to the training batch: `AsyncRolloutWorker` keeps at most `rollout_batch_size` groups in flight, i.e. exactly one training batch worth of trajectories. There is no way to run the generation side ahead of consumption (e.g. to keep a larger engine fleet or a larger sandbox pool saturated) without also inflating the training batch.

## Change

- New `--async-max-concurrent-samples` (default `None`): maximum number of concurrently generating **trajectories** in fully async mode. The worker converts it to in-flight groups (`value // n_samples_per_prompt`). Unset keeps the legacy bound (`rollout_batch_size` groups), so existing runs are unaffected.
- Validated in `miles_validate_args`: at least `--n-samples-per-prompt` (the worker submits whole groups, so one group already puts that many trajectories in flight). Values beyond the fleet-wide client cap (`sglang_server_concurrency x num_engines`) just queue on the client semaphore, so the worker logs a warning instead of failing.
- Dropped the dead `AsyncRolloutWorker(concurrency=...)` constructor parameter — it was stored and never read; the real bound is the one above.

Consumption is intentionally untouched: each train step still collects `rollout_batch_size` groups. The new knob only controls how far generation runs ahead, trading higher engine/sandbox utilization for higher average weight staleness (combine with `--max-weight-staleness` to bound the tail).

## Validation

GLM-5.2 744B fully-async disaggregated run (8 train + 8 inference nodes, agentic tbench2 episodes): `--rollout-batch-size 8 --n-samples-per-prompt 8 --async-max-concurrent-samples 128` doubles in-flight trajectories from 64 to 128 (16 groups); unset falls back to the previous behavior.

### DIFF
--- a/examples/fully_async/fully_async_rollout.py
+++ b/examples/fully_async/fully_async_rollout.py
@@ -82,10 +82,14 @@ class AsyncRolloutWorker:
 
     def __init__(self, args, data_buffer: DataSource):
         if args.async_max_concurrent_samples is not None:
-            assert args.async_max_concurrent_samples <= args.sglang_server_concurrency, (
-                f"--async-max-concurrent-samples ({args.async_max_concurrent_samples}) must not exceed "
-                f"--sglang-server-concurrency ({args.sglang_server_concurrency})"
+            client_capacity = (
+                args.sglang_server_concurrency * args.rollout_num_gpus // args.rollout_num_gpus_per_engine
             )
+            if args.async_max_concurrent_samples > client_capacity:
+                print(
+                    f"--async-max-concurrent-samples ({args.async_max_concurrent_samples}) exceeds the "
+                    f"client concurrency cap ({client_capacity}); the excess queues on the semaphore"
+                )
         self.args = args
         self.data_buffer = data_buffer  # Directly save data_buffer reference
         self.running = True
--- a/examples/fully_async/fully_async_rollout.py
+++ b/examples/fully_async/fully_async_rollout.py
@@ -94,7 +94,12 @@ class AsyncRolloutWorker:
         print("Continuous async rollout worker started")
 
         active_tasks = set()
-        max_concurrent_tasks = self.args.rollout_batch_size
+        # Concurrency is counted in trajectories; each in-flight group holds
+        # n_samples_per_prompt of them.
+        if self.args.async_max_concurrent_tasks is not None:
+            max_concurrent_tasks = max(1, self.args.async_max_concurrent_tasks // self.args.n_samples_per_prompt)
+        else:
+            max_concurrent_tasks = self.args.rollout_batch_size
         group_id_counter = 0
 
         while self.running:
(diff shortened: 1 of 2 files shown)


examples/fully_async/fully_async_rollout.py
```diff
@@ -80,10 +80,14 @@ class AsyncRolloutWorker:
     Supports continuous running, independent of rollout function lifecycle
     """
 
-    def __init__(self, args, data_buffer: DataSource, concurrency=10):
+    def __init__(self, args, data_buffer: DataSource):
+        if args.async_max_concurrent_samples is not None:
+            assert args.async_max_concurrent_samples <= args.sglang_server_concurrency, (
+                f"--async-max-concurrent-samples ({args.async_max_concurrent_samples}) must not exceed "
+                f"--sglang-server-concurrency ({args.sglang_server_concurrency})"
+            )
```
Shi-Dong, line 87: I'd suggest replacing this assertion with a warning: i.e. if `async_max_concurrent_samples` is too large, then we replace it with `min(async_max_concurrent_samples, sglang_server_concurrency)`, and log a warning.  Or perhaps it's fine to remove this constraint altogether?

The reason is that we want `async_max_concurrent_samples` to be large (otherwise it'll just be redundant), but this assertion makes a training run prone to being tripped by an upper bound.

miles/utils/arguments.py
```diff
@@ -490,6 +490,17 @@ def add_rollout_arguments(parser):
...
             )
+            parser.add_argument(
+                "--async-max-concurrent-samples",
+                type=int,
+                default=None,
+                help=(
+                    "Maximum number of concurrently generating trajectories in fully async mode, "
+                    "decoupling generation concurrency from the training batch size. None (default) "
+                    "keeps the legacy bound of one training batch worth of trajectories "
+                    "(rollout_batch_size groups, i.e. rollout_batch_size * n_samples_per_prompt)."
```
Shi-Dong, line 501: It'd be great if we could add a guideline here saying that ideally it should be set larger than global batch size.

Zhichenzzz:
Pushed 62499421e reworking the concurrency validation, verified e2e on 8xH200 (Qwen3-4B, 4 actor + 4 rollout TP1, 3 rollouts each):

- baseline (knob unset): per-engine running-req avg 12.6 — unchanged legacy behavior
- `--async-max-concurrent-samples 128`: avg 27.0 (~2x in-flight, the feature works as described)
- `--async-max-concurrent-samples 1024`: previously killed by the constructor assert even though the real client cap here is `512 x 4 engines = 2048` — the assert compared against the per-engine constant. The client semaphore (`GenerateState`) is `sglang_server_concurrency x rollout_num_gpus / rollout_num_gpus_per_engine`, and exceeding it is harmless (excess tasks queue on the semaphore), so the assert is now a warning against the fleet-wide cap; same 1024 config trains green after the change.

Also moved the real invariant into `miles_validate_args`: the value must be >= `n_samples_per_prompt`, since the worker submits whole groups (this also rejects 0/negative values, which previously became one group silently).

Zhichenzzz (approved):

PR #1678, opened 2026-07-15
feat: lora GLM-5.2 FP8 rollout support

yushengsu-thu:
## Summary

Development branch for GLM-5.2 FP8 support in miles (rollout + training integration). Opened as a draft to track the work; implementation commits will land on this branch.

Based on `main` @ 81d2f02.

Logprob diff:
image

Reward:
image

## Companion PRs

- sglang: https://github.com/sgl-project/sglang/pull/31251

## Test plan

- [ ] GLM-5.2 FP8 rollout smoke test
- [ ] Training E2E on H200 cluster

### DIFF
--- a/scripts/run_glm5_2_744b_a40b_lora.py
+++ b/scripts/run_glm5_2_744b_a40b_lora.py
@@ -323,11 +323,11 @@ def _train(args: ScriptArgs):
             "INDEXER_ROPE_NEOX_STYLE": "0",
             "SGLANG_NSA_FORCE_MLA": "1",
             # PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True breaks torch_memory_saver
-            # fp8 + LoRA: the quantized load path builds mlp.shared_experts TP-sharded while
-            # the LoRA B buffer is sized from the global gate_up dim, so set_lora_info fails
-            # at engine init (e.g. B 4096 vs partition prefix 2048). TP1 shared experts match
-            # the bf16 layout the LoRA buffers assume.
-            **({"SGLANG_SHARED_EXPERT_TP1": "1"} if args.fp8_rollout else {}),
+            # Do NOT set SGLANG_SHARED_EXPERT_TP1 here: TP1-replicated shared experts get
+            # double-counted under dp-attention/EP topologies that skip the in-layer
+            # all-reduce (garbage rollouts). The fp8 LoRA-init crash it papered over is
+            # fixed in sglang lora/mem_pool.py (column-parallel output-axis shard probe);
+            # fp8_rollout requires an sglang build with that fix.
         },
         megatron_path=args.megatron_path,
     )
--- a/scripts/run_glm5_2_744b_a40b_lora.py
+++ b/scripts/run_glm5_2_744b_a40b_lora.py
@@ -252,12 +252,8 @@ def _train(args: ScriptArgs):
 
     if _is_full:
         # mirrors run_glm5_744b_a40b.py; bf16 ~1488GB needs >=~22 GPUs/engine while fp8
-        # fits engine=min(8, ngpu) on one node.
-        # The big-engine fp8 profile is FULL-MODEL ONLY: on the 5-layer toy the ep8/dp8
-        # engine gives a large rollout<->train deviation (abs_diff 2.19 / kl 1.64 vs
-        # 0.27 / 0.06 at ep2/dp2; bisected: not the kv dtype, not the decode backend,
-        # not the LoRA layout), so the toy keeps the small-engine bf16-style serving
-        # profile and only swaps the weights to fp8.
+        # fits engine=min(8, ngpu) on one node. Big-engine fp8 profile is full-model only:
+        # the toy at ep8/dp8 shows a large rollout<->train deviation (abs_diff 2.19 vs 0.27).
         _fp8_full = args.fp8_rollout and args.model_name == "GLM-5.2"
         _eng = min(8, args.num_gpus_per_node) if _fp8_full else args.rollout_num_gpus_per_engine
         _decode = "flashmla_kv" if _fp8_full else "flashmla_sparse"
--- a/scripts/run_glm5_2_744b_a40b_lora.py
+++ b/scripts/run_glm5_2_744b_a40b_lora.py
@@ -280,12 +280,8 @@ def _train(args: ScriptArgs):
         sglang_args = f"--rollout-num-gpus-per-engine {args.rollout_num_gpus_per_engine} --sglang-mem-fraction-static {args.sglang_mem_fraction_static} --sglang-cuda-graph-max-bs 64 --sglang-moe-runner-backend triton --sglang-disable-shared-experts-fusion --sglang-lora-backend {args.sglang_lora_backend} --sglang-reasoning-parser glm45 --sglang-tool-call-parser glm47 "
 
     if args.fp8_rollout:
-        # Point the sglang engines at the pre-converted fp8 checkpoint via the --sglang-config
-        # YAML (model_path override). update_weights must stay ON so the per-step LoRA adapter
-        # sync still reaches the engine (a model_path != hf_checkpoint otherwise auto-disables
-        # it); the base-weight sync is already skipped under --lora-base-cpu-backup +
-        # --colocate (skip_base_sync in update_weight_from_tensor.py), so the quantized base
-        # on the engine is never overwritten by bf16 training weights.
+        # Serve the fp8 ckpt via --sglang-config; update_weights stays on so the per-step LoRA
+        # sync reaches the engine (the bf16 base sync is already skipped under colocate + backup).
         sglang_config_path = f"{load_save_path}/sglang_fp8_rollout.yaml"
         os.makedirs(load_save_path, exist_ok=True)
         with open(sglang_config_path, "w") as f:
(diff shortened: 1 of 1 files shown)


yushengsu-thu:
Pushed the working fp8-rollout implementation (0efa72c7), e2e-validated on GLM-5.2_5layer (gsm8k, 8xH200 single node):

| rollout config | train_rollout_logprob_abs_diff | train_rollout_kl |
|---|---|---|
| bf16 (default) | 0.0105 | 1.2e-4 |
| **fp8 (--fp8-rollout, default profile)** | **0.271** | **0.060** |

The fp8 gap is flat across steps — the constant weight-quantization offset, not LoRA-sync drift. Given kl ~0.06, real fp8 RL runs should consider TIS on top of routing replay.

Notes:
- The engines are pointed at `<hf_checkpoint>_fp8` via a generated `--sglang-config` YAML with `update_weights: true` (LoRA adapters still sync per step; base sync is already skipped under `--lora-base-cpu-backup` + `--colocate`, so the quantized base is never overwritten).
- `SGLANG_SHARED_EXPERT_TP1=1` is required: the quantized load path builds `mlp.shared_experts` TP-sharded while the LoRA B buffer is sized from the global gate_up dim → engine-init crash in `set_lora_info` without it.
- The big-engine fp8 profile (engine=8 / kv fp8_e4m3 / flashmla_kv) is now full-model-only: on the toy, ep8/dp8 serving bisects to abs_diff 2.19 / kl 1.64 (ruled out: kv dtype, decode backend, LoRA layout — attention-only LoRA reproduces it; ep2/dp2 restores 0.27). Worth checking whether the full model shows the same topology sensitivity before trusting fp8 numbers at engine=8.
- `scripts/slice_fp8_checkpoint.py` builds the toy fp8 ckpt from the official GLM-5.2 fp8 release (GLM-5.2_5layer is a verbatim slice of layers 0-4, so tensors + block scales copy 1:1).

yushengsu-thu:
Full-model follow-up (4 nodes x 8 H200, TP8/DP4/EP32 train, single 32-GPU sglang engine ep32/dp32): the engine-topology fp8 anomaly **reproduces on the full GLM-5.2** — it is not toy-specific.

| full GLM-5.2, same 32-GPU engine topology | rollout logprob | abs_diff | kl | raw_reward |
|---|---|---|---|---|
| bf16 rollout | -0.22 | 0.0055-0.026 | 2.2e-4 - 3.5e-3 | 0.08-0.40 |
| fp8 rollout (official GLM-5.2_fp8) | -9.0 (garbage) | 2.61-2.77 | ~2.0 | 0.0 |

wandb: fp8 https://wandb.ai/glm-zero/miles-glm5_att_unfused_moe/runs/y05bmwaw, bf16 baseline https://wandb.ai/glm-zero/miles-glm5_att_unfused_moe/runs/yb14lo5f

Same topology in bf16 is clean, and fp8 at ep2/dp2 (toy) shows only the expected quantization offset (0.27) — so the breakage is fp8 serving x large ep/dp, independent of LoRA (attention-only reproduces), kv dtype and decode backend. Until the sglang side is fixed, fp8 rollout at engine>=8-GPU topologies should be considered unusable for RL; the fp8 default profile in this PR being gated to full-model-only should probably be gated OFF entirely or behind an explicit override.

yushengsu-thu:
**Correction of the earlier topology attribution** — root cause isolated with standalone serving bisect on the full GLM-5.2 fp8 checkpoint (tp8/ep8/dp8, dp-attention, no RL in the loop):

| standalone fp8 serving, gsm8k greedy (16 prompts) | reward | output |
|---|---|---|
| no LoRA, no env | **0.375** (matches bf16 RL raw_reward 0.398) | coherent, correct math |
| no LoRA, **SGLANG_SHARED_EXPERT_TP1=1** | **0.000** | pure garbage ("0000202000ITA...") |

So `SGLANG_SHARED_EXPERT_TP1=1` — the workaround this PR sets to get past the fp8 LoRA-init shape crash — is itself what corrupts fp8 serving at multi-GPU engine topologies (TP1-replicated shared experts appear to get double-counted in the dp-attention/EP reduce path). It happened to be benign at the toy's ep2/dp2, which misled the earlier bisect. Raw fp8 serving and the LoRA machinery are both fine.

Consequence for this PR: the TP1 env should be dropped once the sglang side fixes the real bug (LoRA B buffer sized from the global gate_up dim vs TP-sharded shared-expert partitions under the quantized load path — needs local-partition-sized buffers or per-rank slicing in set_lora_info). Until then fp8_rollout + MoE-layer LoRA targets can't produce usable rollouts on multi-GPU engines.

yushengsu-thu:
Pushed d236dab1: drop the SGLANG_SHARED_EXPERT_TP1 workaround (root-caused upthread — it double-counts TP1-replicated shared experts under deferred-reduce topologies). The underlying fp8 LoRA-init crash is now fixed properly in sglang (sgl-project/sglang#31251, commit 2a56d3cd9); `--fp8-rollout` requires an sglang build with that fix.

End-to-end validation with the fix, no env:
| run | abs_diff | kl | gsm8k raw_reward |
|---|---|---|---|
| toy fp8+LoRA (tp2/ep2) | 0.0347 flat | 9.8e-4 | — |
| toy bf16 regression | 0.0104 (unchanged) | 1.2e-4 | — |
| **full 744B fp8+LoRA RL (2×16-GPU engines)** | **0.0196** | 2.7e-3 | **0.3125** (bf16 baseline 0.398) |

wandb: fp8-fixed https://wandb.ai/glm-zero/miles-glm5_att_unfused_moe/runs/1pm52qfm · bf16 baseline https://wandb.ai/glm-zero/miles-glm5_att_unfused_moe/runs/yb14lo5f

Note: at 32-GPU engines the blockwise-fp8 divisibility check rejects the shared-expert shard (2048/32=64 < 128); 16-GPU engines divide cleanly (2048/16=128) and fit colocate memory (~44GB weights/GPU at mem-frac 0.5) — worth encoding that constraint in the runner's fp8 profile as a follow-up.

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -260,6 +279,27 @@ def _train(args: ScriptArgs):
     else:
         sglang_args = f"--rollout-num-gpus-per-engine {args.rollout_num_gpus_per_engine} --sglang-mem-fraction-static {args.sglang_mem_fraction_static} --sglang-cuda-graph-max-bs 64 --sglang-moe-runner-backend triton --sglang-disable-shared-experts-fusion --sglang-lora-backend {args.sglang_lora_backend} --sglang-reasoning-parser glm45 --sglang-tool-call-parser glm47 "
 
+    if args.fp8_rollout:
+        # Point the sglang engines at the pre-converted fp8 checkpoint via the --sglang-config
```
Zhichenzzz: could you squash those comments here and below?

Zhichenzzz (approved):

scripts/run_glm5_2_744b_a40b_lora.py
yushengsu-thu: Done — squashed both comment blocks (the sglang-config one here and the SGLANG_SHARED_EXPERT_TP1 one in extra_env_vars) down to two lines each in 922bcc28.

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -239,11 +252,17 @@ def _train(args: ScriptArgs):
...
-        _eng = min(8, args.num_gpus_per_node) if args.fp8_rollout else args.rollout_num_gpus_per_engine
-        _decode = "flashmla_kv" if args.fp8_rollout else "flashmla_sparse"
-        _cg = 256 if args.fp8_rollout else 64
-        _kv = "--sglang-kv-cache-dtype fp8_e4m3 " if args.fp8_rollout else ""
+        # fits engine=min(8, ngpu) on one node.
+        # The big-engine fp8 profile is FULL-MODEL ONLY: on the 5-layer toy the ep8/dp8
+        # engine gives a large rollout<->train deviation (abs_diff 2.19 / kl 1.64 vs
+        # 0.27 / 0.06 at ep2/dp2; bisected: not the kv dtype, not the decode backend,
+        # not the LoRA layout), so the toy keeps the small-engine bf16-style serving
+        # profile and only swaps the weights to fp8.
```
yushengsu-thu: clean this comment

yushengsu-thu: Done — condensed to the essential constraint in b5f1e9b1.

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -283,6 +315,8 @@ def _train(args: ScriptArgs):
             "INDEXER_ROPE_NEOX_STYLE": "0",
             "SGLANG_NSA_FORCE_MLA": "1",
             # PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True breaks torch_memory_saver
+            # No SGLANG_SHARED_EXPERT_TP1: it double-counts shared experts under dp-attention/EP;
+            # the fp8 LoRA-init crash it papered over is fixed in sglang lora/mem_pool.py.
```
yushengsu-thu: remove this

scripts/run_glm5_2_744b_a40b_lora.py
```diff
@@ -239,11 +252,13 @@ def _train(args: ScriptArgs):
 
     if _is_full:
         # mirrors run_glm5_744b_a40b.py; bf16 ~1488GB needs >=~22 GPUs/engine while fp8
-        # fits engine=min(8, ngpu) on one node
-        _eng = min(8, args.num_gpus_per_node) if args.fp8_rollout else args.rollout_num_gpus_per_engine
-        _decode = "flashmla_kv" if args.fp8_rollout else "flashmla_sparse"
-        _cg = 256 if args.fp8_rollout else 64
-        _kv = "--sglang-kv-cache-dtype fp8_e4m3 " if args.fp8_rollout else ""
+        # fits engine=min(8, ngpu) on one node. Big-engine fp8 profile is full-model only:
+        # the toy at ep8/dp8 shows a large rollout<->train deviation (abs_diff 2.19 vs 0.27).
```
yushengsu-thu: remove this

scripts/run_glm5_2_744b_a40b_lora.py
yushengsu-thu: Removed in ee41ad28.

scripts/run_glm5_2_744b_a40b_lora.py
yushengsu-thu: Removed in ee41ad28.

PR #1680, opened 2026-07-15
[AMD] fix GPU memory release for TMS pause on ROCm 7.2

XinyuJiangCMU:

## What
Adds a build-arg-gated step to `docker/Dockerfile.rocm` that installs a rebuilt `libhsa-runtime64` fixing the ROCm 7.2 ROCr VMM-pause regression. `docker/build.py` turns it on for the `rocm720-mi35x` variant; the ROCm 7.0 variants leave it off.

## Why
On ROCm 7.2, ROCr's `~MappedHandleAllowedAgent()` CPU-agent branch early-returns without unmapping the imported dma-buf:

```cpp
Runtime::MappedHandleAllowedAgent::~MappedHandleAllowedAgent() {
  if (targetAgent->device_type() == kAmdCpuDevice) return;  // leaks the imported dma-buf FD
  hsa_status_t status = targetAgent->driver().ReleaseShareableHandle(shareable_handle);
  assert(status == HSA_STATUS_SUCCESS);
}
```

The leaked FD pins the physical GPU memory, so `hipMemUnmap` + `hipMemRelease` free nothing while the address range stays reserved (ROCm/ROCm#6021). `torch_memory_saver.pause()` then frees nothing and colocate RL OOMs handing memory back to the trainer, which is the ROCm 7.2 regression reported in fzyzcjy/torch_memory_saver#83. ROCm 7.0 is unaffected. The upstream one-function fix (ROCm/rocm-systems#4363) is only on `develop`, not in any released 7.2.x (checked through 7.2.4), so bumping the base ROCm does not help.

## How
- The rebuilt `.so` ships as a release asset in `miles-wheels-rocm` (`rocm720-gfx950-v0.5.14`), built by its `build_rocr_vmmfix.py`: clone ROCR-Runtime @ `rocm-7.2.0`, apply the same ROCm/rocm-systems#4363 fix rebased onto that tag, then cmake.
- `Dockerfile.rocm` downloads it and overwrites `libhsa-runtime64` in place.

## Verification (gfx950 / MI355X)
A minimal test that reserves memory, maps it, then unmaps and releases it (what `pause()` does), same GPU: stock ROCm 7.2.0 frees **0 MB**, while this fix frees the full **8192 MB**, matching ROCm 7.0.0. `torch_memory_saver` `pause()`/`resume()` then works end to end: frees on pause, restores on resume, tensor contents intact with cpu_backup. This is also the first confirmation the fix works on MI355X (gfx950); upstream had only tested it on consumer GPUs.

## Removal
Remove this step once a released ROCm version includes the fix.

### DIFF
--- a/docker/Dockerfile.rocm
+++ b/docker/Dockerfile.rocm
@@ -26,6 +26,10 @@ ARG MEGATRON_BRANCH=miles-main
 ARG MILES_COMMIT=main
 
 ARG GPU_ARCH=gfx950
+# 1 on ROCm 7.2 to bake the ROCr VMM-pause fix; ROCm 7.0 has no such regression, leave 0.
+ARG APPLY_ROCR_VMMFIX=0
+ARG WHEELS_REPO=XinyuJiangCMU/miles-wheels-rocm
+ARG WHEELS_TAG_ROCM=rocm720-gfx950-v0.5.14
 ARG MAX_JOBS=128
 
 ARG RCCL_TESTS_REPO=https://github.com/ROCm/rocm-systems.git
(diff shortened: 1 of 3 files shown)


guapisolo (approved):
LGTM

PR #1683, opened 2026-07-15
[tml] Inkling model support

Zhichenzzz:
Inkling model + **full-parameter** RL support. The native LoRA layer is stacked on top in #2122.

https://www.lmsys.org/blog/2026-07-15-inkling-day0-support

https://github.com/sgl-project/sglang/pull/31681

https://github.com/radixark/Megatron-LM/pull/68

PR #1684, opened 2026-07-15
[tml] Inkling model doc

Zhichenzzz:
Adds the Inkling (975B) model recipe page under docs/models/thinkingmachines/, registers it in the models index and navigation. Companion to #1683.


guapisolo (approved):
LGTM.

PR #1686, opened 2026-07-15
[AMD] use prebuilt wheels for flash-attn and sgl-model-gateway

XinyuJiangCMU:

Install flash-attn and sgl-model-gateway from the prebuilt `miles-wheels-rocm` wheels instead of building them from source, mirroring the NV `docker/Dockerfile` (`SGL_ROUTER_USE_WHEELS`, wheels by default).

Stacked on #1680.


guapisolo (approved):
LGTM.

PR #1688, opened 2026-07-15
[AMD] Restore Python 3.10 StrEnum compatibility

Xinyu-Kang:
## Summary

The ROCm Docker image uses Python 3.10, where `enum.StrEnum` is unavailable. Unconditional imports in three fault-tolerance modules therefore block Miles from starting.

Use `backports.strenum` on Python 3.10 while retaining the standard-library implementation on Python 3.11+.

## Validation

Verified that all affected modules import successfully on Python 3.10.12.

### DIFF
--- a/miles/backends/megatron_utils/ft/types.py
+++ b/miles/backends/megatron_utils/ft/types.py
@@ -1,4 +1,9 @@
-from enum import StrEnum, auto
+from enum import auto
+
+try:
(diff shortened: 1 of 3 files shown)


XinyuJiangCMU:
LGTM

guapisolo (approved):
LGTM.

PR #1690, opened 2026-07-15
[router] set manual policy (sticky + min_load) as default agentic routing policy

yueming-yuan:
before:
before
after:
after

### DIFF
--- a/miles/backends/sglang_utils/arguments.py
+++ b/miles/backends/sglang_utils/arguments.py
@@ -147,5 +147,10 @@ def validate_args(args):
     if args.sglang_dp_size > 1:
         assert args.sglang_enable_dp_attention
 
+    if args.sglang_router_policy is None and args.use_session_server:
+        args.sglang_router_policy = "manual"
+        if args.router_assignment_mode == "random":
+            args.router_assignment_mode = "min_load"
+
     if getattr(args, "sglang_router_ip", None):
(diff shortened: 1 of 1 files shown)


yueming-yuan:
Heads up on two interactions with #1703 / radixark/sgl-router-for-miles#8:

1. The `and not enable_experimental_rollout_refactor()` guard here exists to dodge the #891 assert (`--sglang-router-policy` blocked under `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1`). #1703 removes that assert and gives the inference_rollout stack full routing-key support, so after it lands this guard can be dropped — otherwise flag=0 and flag=1 session-server runs get silently different default policies.
2. #1703 unifies every key site behind `policy_uses_routing_key(args)` (`consistent_hashing` or `manual`) in `generate_endpoint_utils` — covering both stacks' group/eval assignment and prefill recompute, which this PR's two legacy-stack sites don't reach. After rebasing on #1703, the `in ("consistent_hashing", "manual")` checks here collapse into that predicate (also note: `session_id` is renamed to `routing_key`). Router-side, sgl-router-for-miles#8 will reject keyless requests under `manual` by default, which this unification makes safe.

guapisolo (approved):
LGTM.

Zhichenzzz (approved):
LGTM

PR #1701, opened 2026-07-16
Point SGLANG_BRANCH back to sglang-miles

yueming-yuan:
sglang-miles has been repointed to the v0.5.15 stack (same commit as sglang-miles-v0.5.15, `5bca6f8`). Restore the default branch reference before the transitional `sglang-miles-v0.5.15` branch is deleted.


guapisolo (approved):
Approved

PR #1710, opened 2026-07-17
openenv/tbench2: per-task sandbox image recipe + Daytona materialization

nblintao:
First of a three-PR stack (this → TTL/ownership hardening → #1675 rebased on top). Pure additive: nothing imports the modules yet; the openenv TB2 adapter's per-task backend (#1675) adopts them in the follow-up.

## What

The build recipe for **per-task TB2 sandboxes**, next to its only consumer (the miles openenv TB2 adapter). TB2 is a per-task-image benchmark — every task pins its official runtime image in `task.toml` — so a cloud sandbox serving the env must be built per task: **the official task image ⊕ a tbench2_env server layer**, one layer, no DinD.

Two files, one per concern (per review):

- **`tb2_task_recipe.py` — provider-agnostic recipe**: `server_layer_commands()` emits plain shell commands — uv-managed venv at `/opt/envserver` running the installed `tbench2_env` package's source (embedded into the build, so each sandbox runs exactly the version installed locally, no released package needed; the tbench2_env features this relies on are upstream since huggingface/OpenEnv#965 + #972), task directory staged from the tasks checkout's pinned-SHA GitHub tarball with `solution/` excluded. Nothing Daytona-specific; the same layers can back a Dockerfile or another provider.
- **`tb2_task_sandbox.py` — Daytona materialization**: `create_task_sandbox()` creates per-episode declaratively from the `Image` definition — no named snapshots, so no org snapshot quota; repeat creates hit Daytona's build cache (~1 min warm, ~10 min first build per task). Daytona doesn't run the image CMD, so it execs `server_cmd()` and waits for `/health`. A bake CLI (`python tb2_task_sandbox.py`) can pre-register named snapshots as a warm cache.
- **Verifier-asset hygiene** (mirrors the official harness's stage-at-verify model): `solution/` never enters the image; `server_cmd()` sets `TB2_WITHHOLD_TESTS=1` so the server pulls `tests/` into memory at `reset()` and keeps it off the agent-visible filesystem outside the verify window.

The embedded env source is located from the installed `tbench2_env` package and requires an editable/checkout install; `_env_src_dir()` fails fast with instructions instead of erroring mid-image-build.

## Testing

`pytest examples/experimental/openenv/tests/test_tb2_task_recipe.py -q` → 3 passed (solution-exclusion anchored to the one task; withhold gate and staged-task default present in `server_cmd()`). Embedding smoke against a real editable install: source resolves and tars to ~16KB, well under Daytona's 64KB Dockerfile-line ceiling.

The recipe itself is the one already validated end-to-end on the #1675 branch: golden sweep 82/89 with 0 infra errors, full 89-task live-policy sweep with 0 systematic infra failures.

**Post-reorg sanity (2026-07-17)**: fresh end-to-end validation of the full stack — golden regression, live TTL/label checks, API-policy episodes — is recorded in #1675's Validation section.

### DIFF
--- a/examples/experimental/openenv/tb2_sandbox_daytona.py
+++ b/examples/experimental/openenv/tb2_sandbox_daytona.py
@@ -1,7 +1,7 @@
 """Daytona materialization of the per-task Terminal-Bench-2 sandbox recipe.
 
 The recipe itself — the shell layers that turn a task's official image into a
-combined task+env-server image — lives in ``tb2_task_recipe`` (sibling module)
+combined task+env-server image — lives in ``tb2_sandbox_recipe`` (sibling module)
 and is provider-agnostic. This module is everything Daytona-specific about
 turning that recipe into a running cloud sandbox:
 
@@ -11,7 +11,7 @@ turning that recipe into a running cloud sandbox:
       the declarative path avoids the quota entirely, and repeat creates hit
       Daytona's build cache (~1min after the first build). Daytona does not
       run the image CMD, so this execs ``server_cmd()`` and waits for /health.
-  bake CLI (``python tb2_task_sandbox.py ...``)  optionally pre-register
+  bake CLI (``python tb2_sandbox_daytona.py ...``)  optionally pre-register
       named snapshots ``<prefix><task-id>`` as a warm cache.
 """
 
@@ -22,7 +22,7 @@ import shlex
 import sys
 from pathlib import Path
 
-from tb2_task_recipe import (
+from tb2_sandbox_recipe import (
     read_task_config,
     resolve_docker_image,
     server_cmd,
--- a/examples/experimental/openenv/tb2_task_recipe.py
+++ b/examples/experimental/openenv/tb2_task_recipe.py
@@ -1,0 +1,255 @@
+"""Provider-agnostic image recipe for per-task Terminal-Bench-2 sandboxes.
+
+TB2 is a per-task-image benchmark: every task pins its official runtime image
+in ``task.toml`` (``[environment].docker_image`` — the exact image the TB2
+harness itself runs). A sandbox serving this env must therefore be built per
+task: **the official task image ⊕ this env's server layer**, one layer, no
+DinD. This module owns that recipe as plain shell commands — nothing
+provider-specific — so the same layers can back a Dockerfile, a Daytona
+declarative build (``tb2_task_sandbox``, the sibling module), or another
+provider.
+
+  ``server_layer_commands(task_dir)``  shell commands that turn the official
+      task image into a combined task+env-server image: a uv-managed Python
+      venv at ``/opt/envserver`` running the INSTALLED tbench2_env package
+      (source embedded into the build, so a patched checkout ships without a
+      released package), plus the task directory staged at ``/opt/tb2-tasks/<id>``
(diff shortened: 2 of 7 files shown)


Shi-Dong (commented):
I'd actually suggest splitting the Daytona support and the Daytona-agnostic generic sandbox support into two different files (still in the same PR).

nblintao:
> I'd actually suggest splitting the Daytona support and the Daytona-agnostic generic sandbox support into two different files (still in the same PR).

Good point! Done.

examples/experimental/openenv/tb2_sandbox_daytona.py
Shi-Dong, line 1: Perhaps rename to `tb2_task_daytona.py` to show that this script is for Daytona only? We might need other adapters if we switch to a different sandbox provider.

examples/experimental/openenv/tb2_task_recipe.py
```diff
@@ -0,0 +1,257 @@
...
+    return base64.b64encode(raw).decode()
+
+
+def _task_layer_command(task_dir: Path) -> str:
+    """Build command that stages the task dir at /opt/tb2-tasks/<id>.
+
+    One uniform path for every task: download the checkout's pinned-commit
+    GitHub tarball and extract just this task. Deterministic (the SHA pins
+    the content — note: the committed tree, not uncommitted local edits) and
+    payload-free, so build commands stay far from Daytona's 64KB
```
Shi-Dong: Nit: you might want to remove the reference to Daytona in the docstring since this script is supposed to be sandbox-agnostic.

examples/experimental/openenv/tb2_task_recipe.py
```diff
@@ -0,0 +1,257 @@
...
+def _tar_filter(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo | None:
+    name = Path(tarinfo.name).name
+    if name in {"__pycache__", ".initial_env"} or name.endswith((".pyc", ".egg-info")):
+        return None
+    return tarinfo
+
+
+def _dir_tar_b64(paths: list[Path], arcnames: list[str], max_bytes: int) -> str:
+    buf = io.BytesIO()
+    with tarfile.open(fileobj=buf, mode="w:gz") as tar:
```
Shi-Dong: Seems that `tarfile` automatically encodes current timestamp into the tarball, which means that every tarball produces a different base64 hash.  It feels to me that this might make it impossible for Daytona to reuse images.

Shi-Dong (approved):
Approved with minor comments.

examples/experimental/openenv/tb2_sandbox_daytona.py
nblintao: Fixed. Renamed the 2 files to `tb2_sandbox_recipe` and `tb2_sandbox_daytona` to make it more clean.

examples/experimental/openenv/tb2_task_recipe.py
nblintao: Done.

examples/experimental/openenv/tb2_task_recipe.py
nblintao: Good catch! Fixed.

PR #1711, opened 2026-07-17
openenv/tbench2: orphan TTL + ownership labels for per-task sandboxes

nblintao:
Second of the three-PR stack (#1710 → this → #1675 rebased on top). Base is `tao/tb2-sandbox-recipe`; only the last commit is new here.

## Problem

A trainer that dies without reaching its `daytona.delete()` (SIGKILL, OOM, node loss) leaks its sandbox **forever**: `create_task_sandbox` armed `auto_stop_interval=0`, so Daytona never reclaims it, and a running orphan bills CPU+memory+disk indefinitely. We hit exactly this: a batch launch died ~1 minute in and left 30 orphans running 25+ hours until swept by hand. In a shared Daytona org there was also no way to tell whose sandboxes they were — the API records no creator.

## Fix

**Orphan TTL as a dead-man's switch.** Creates now arm `auto_stop_interval=30` + `auto_delete_interval=120` (both overridable kwargs). The subtlety: Daytona's auto-stop clock counts **only SDK interactions — preview-proxy traffic does not reset it** ([docs](https://www.daytona.io/docs/en/sandboxes/)), and an episode's I/O runs entirely over the signed preview URL, so a bare TTL would kill healthy episodes longer than the interval. A keepalive **daemon thread** therefore beats `sandbox.refresh_activity()` every 5 minutes for as long as the creating process lives:

- live episode of any length → timer keeps resetting, never stopped mid-run
- trainer hard-killed → beats stop with it (daemon thread) → Daytona stops the sandbox within 30 min of the last beat, deletes the stopped remains 120 min later
- episode ends normally → caller deletes the sandbox → refreshes fail persistently → thread exits (3 consecutive failures tolerated, so a 15-minute API blip doesn't strand a live episode)

**Ownership labels.** Each sandbox now carries, alongside the existing `openenv-tbench2-task`:
- `openenv-launcher` — `$OPENENV_LAUNCHER` if set (recommended on shared hosts where the unix user is a generic account), else the local unix user
- `openenv-run-id` — `$OPENENV_RUN_ID`, when set

so shared-org cleanup sweeps can target one launcher or one run (`daytona.list(labels={"openenv-launcher": ...})`) instead of guessing.

## Testing

`pytest examples/experimental/openenv/tests/test_tb2_task_sandbox.py -q` → 7 passed. New coverage: labels default to the unix user / honor explicit env vars / omit run-id when unset; creates arm a positive TTL by default; keepalive beats while the sandbox is alive and exits after persistent failures once it's deleted. `create_task_sandbox`'s new params are keyword-only with defaults — the adapter callsite needs no change.

**Post-reorg sanity (2026-07-17)**: labels + TTL verified live on a real sandbox (openenv-launcher/run-id present, auto_stop=30 / auto_delete=120 armed, keepalive thread beating, clean delete); full-stack results in #1675's Validation section.


examples/experimental/openenv/tb2_sandbox_daytona.py
```diff
@@ -61,29 +68,96 @@ def task_resources(task_dir: Path):
...
+    The Daytona API records no creator, so in a shared org labels are the only
+    attribution. ``openenv-tbench2-task`` keys sweep/cleanup tooling to exactly
+    the sandboxes this recipe created (shared orgs run other workloads).
+    ``openenv-launcher`` is OPENENV_LAUNCHER when set — do set it on shared
+    hosts, where the unix user is a generic account — else the local unix
+    user. ``openenv-run-id`` (OPENENV_RUN_ID, optional) additionally groups
+    one run's sandboxes for targeted sweeps.
+    """
+    try:
+        user = getpass.getuser()
```
Shi-Dong, line 83: This makes sense for public users, but it seems that within our devbox, `getpass.getuser()` always returns `root`.  I'm worried that we'd still be unable to distinguish the sandbox owners after this PR.

Shi-Dong (approved):
LGTM.

examples/experimental/openenv/tb2_sandbox_daytona.py
nblintao: Yes. We'll use `OPENENV_LAUNCHER` below to distinguish. But I know it's not very convenient, since there is no default for it. Need to figure out a better way (Claude Code suggested some ways. But I don't think it's good idea to implement them on Miles. It should be changed on devbox's setup).

For now, it's not a blocker since we already have TTL, ownership label is no longer critical.

Shi-Dong: Kangyan just told me that we can use `RADIXARK_USER_EMAIL` to identify username within a devbox.  Would you mind opening a new PR to update this?

Shi-Dong: Nvm I think this script is public facing and it'd be weird to have an internal env var hardcoded here.

PR #1714, opened 2026-07-18
fix(rollout): recover session rollback on router-timeout desync

layahaasini:
## Summary

Fixes #955.

A router timeout can leave the engine holding an assistant reply the client never received. The client then retries with a shorter message list (a strict prefix of stored history), and session rollback raised a non-recoverable **400**, abandoning the whole rollout trial and wasting GPU compute.

This makes rollback recover: on a strict-prefix retry it resets the trajectory to zero assistant checkpoints and re-renders, so the rollout continues. Recovery is GPU-cost-neutral (the re-sent prompt is a radix-cache hit — no re-prefill).

## Change

In `miles/rollout/session/linear_trajectory.py`:
- Reset to zero assistant checkpoints when the retry is a non-empty strict prefix with no assistant in the matched prefix. Still bounded by `MAX_ASSISTANT_ROLLBACK_STEPS`, so it only fires for the single orphaned turn — a genuine divergence, or a 2+-assistant rollback, still raises.
- Reorder `prepare_pretokenized` so a reset re-renders from scratch instead of an empty-prefix `merge_tokens`.

No error-hierarchy or HTTP-contract change. A distinct `WARNING` keeps these desyncs countable in logs.

## Why not the proposed 409

A 409 is a reasonable ergonomics layer, but insufficient alone: a retried 409 hits the same path and only succeeds *because of* this rollback. The server-side fix is the necessary core; a 409 can layer on later.

## Limitations

- Covers the single-orphaned-assistant case; deeper multi-turn desyncs stay rejected.
- Root cause is the timeout itself (#920) — this is a backstop, not the root fix.

## Tests

3 new tests (repro, reset-then-continue, 2-assistant guard) + existing suite green (43/43). Fast tests use a stub tokenizer; the real-tokenizer path after reset runs only in CI.

Related: #920

### DIFF
--- a/miles/rollout/session/linear_trajectory.py
+++ b/miles/rollout/session/linear_trajectory.py
@@ -61,6 +61,10 @@ class LinearTrajectory:
 
         Must be called under ``self.lock``.
         """
+        # 1. Detect agent retries and roll back (at most one assistant step).
+        self._try_detect_and_rollback_to_assistant_checkpoint(request_messages)
+
+        # No stored prefix (first turn, or a rollback reset the trajectory): render fresh.
         if not self.token_ids:
             return tito_tokenizer.render_messages(
                 request_messages,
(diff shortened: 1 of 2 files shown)


layahaasini:
Closing: #1826 landed the same fix with equivalent test coverage.

PR #1716, opened 2026-07-18
Move fully-async rollout from examples into miles/rollout

yueming-yuan:
## Motivation

Fully-async rollout is used by production runs (the swe-agent-v2 GLM-4.7 async scripts) but its implementation lives in `examples/fully_async/`, wired up via a PYTHONPATH hack in every launch script. Core arguments it consumes (`--max-weight-staleness`) are already defined in core `arguments.py` while their only consumer sits in examples. First step of moving it into the core library.

## What this PR does

**Pure mechanical move** — no logic changes:

- `git mv examples/fully_async/fully_async_rollout.py → miles/rollout/fully_async_rollout.py` (100% rename similarity; blame/history preserved).
- All references switch to `--rollout-function-path miles.rollout.fully_async_rollout.generate_rollout_fully_async` and the example-dir PYTHONPATH entries are removed: the two qwen launch scripts, the two swe-agent-v2 async scripts (`FULLY_ASYNC_DIR` hack deleted), and the docs pages that referenced the old path.
- `examples/fully_async/` keeps the launch scripts + README, now pointing at the core module.

A follow-up PR (stacked) rewrites the module on the class-based rollout API.

## Testing

Pure rename + path updates; `pre-commit run --all-files` passes. The moved module is import-compatible (same content, same imports).


docs/user-guide/fully-async.md
```diff
@@ -32,7 +32,7 @@ function that owns the background worker:
 ```diff
 - python3 train.py ...
 + python3 train_async.py ...
-+   --rollout-function-path fully_async_rollout.generate_rollout_fully_async
++   --rollout-function-path miles.rollout.fully_async_rollout.generate_rollout_fully_async
```
guapisolo, line 35: I suggest we directly add a args `--fully-async` to make sure all user know there is a fully async example.

guapisolo: Also we need to add a fully async CI on dapo math, will submit a PR later.

guapisolo:
> I suggest we directly add a args --fully-async to make sure all user know there is a fully async example.

I am let codex doing this.

guapisolo:
Submitted to #1924 stack on #1782 to avoid cascade rebase

guapisolo (approved):
Handle fully async flag name in later PR. So approve this as-is.

PR #1630, opened 2026-07-11
(4/4) feat(ci): metric-history gate — GATE_DEFAULTS one-liners + standard-metric sweep (M4)

guapisolo:
> **Depends on:** #1517
> Stacked on the parent PR's branch. Review / merge the parent first.

## Summary
One-line `register_ci_gate` declarations for standard metrics, swept across the CUDA RL e2e tests.

## Motivation
Requiring every test to write `steps` + `constraint` literals makes blanket coverage expensive: the standard metrics (`train/grad_norm`, `train/ppo_kl`, logp-diff, `rollout/raw_reward`) want the same defaults everywhere. A per-`metric_key` defaults table beside the parser supplies them, so the one-liner is a complete minimal declaration. The sweep lands in the same PR so the table ships together with its consumers.

## Usage
```python
register_cuda_ci(est_time=600, suite="stage-c-8-gpu-h100")

register_ci_gate(metric_key="rollout/raw_reward")  # steps + constraint from GATE_DEFAULTS
```

## Design Notes
- **Key choices:** a defaulted field keys on the TABLE literal (same canonical-JSON derivation as a written literal), so editing a table entry re-keys every declaration relying on it — one global reset lever per standard metric.
- A written literal always wins over its table default.
- A metric with no `GATE_DEFAULTS` entry must still write both fields.
- Table keys stay within the capture whitelist (test-enforced).
- The sweep declares all five metrics on each of the 22 targets: the sglang rollout path requests `return_logprob` unconditionally, every target runs `policy_loss`, so all five keys exist in every target's record.
- Deliberately skipped: disabled tests, serving / precision files (no training metrics), `test_mimo_7B_mtp_only_grad` (zero grads by design), `test_qwen3_4B_ckpt` (multi-stage record merge needs its own look).

## Verification
- **Tests added:** `test_one_liner_fills_from_gate_defaults`, `test_partial_default_written_literal_wins`, `test_one_liner_without_table_entry_rejected`, `test_gate_defaults_within_capture_whitelist_and_valid`, `test_one_liner_runtime_is_noop`.
- All 22 swept files machine-verified: each parses into 5 valid specs with the `register_cuda_ci` registry intact.
- Full `tests/ci/test/` run passes (247 passed, 1 skipped).

## Review Focus
- Scrutinize the `GATE_DEFAULTS` band values in `register.py` — shadow-calibration starting points; test owners tune or veto their line in the sweep commit.
[...]

PR #1719, opened 2026-07-20
(5/5) feat(ci): metric-history gate — two-sided ref-based band constraint (M5)

guapisolo:
Stacked on #1630 (M4). Shadow-phase design fix for the gate constraint, before any real history accumulates.

## Motivation

Two defects in the M2 constraint family, found while reviewing gate semantics:

1. **The improving side was unbounded.** Under `higher_is_worse`, a metric that collapses far below baseline (usually a broken metric — data not loaded, degenerate rollouts — not a genuine improvement) passed unconditionally. On a nightly run that run is marked trusted and its value enters the baseline pool, dragging the mean; the broken run passes, then subsequent healthy runs fail against the polluted baseline.
2. **The band scaled from `max(|cur|, |ref|)`**, so a regressing value widened its own tolerance: with `rel=0.5` (the GATE_DEFAULTS starting point) the effective ceiling was `2×ref`, not the declared `1.5×ref`.

## Modifications

- `constraints.py` — every constraint is two-sided: `cur` must land in `[ref − band_down, ref + band_up]`; each side declared independently as `rel_up` / `abs_floor_up` / `rel_down` / `abs_floor_down` (≥1 param per side, parser-enforced); each side's band scales from the reference only: `band = max(rel·|ref|, abs_floor)`. The `direction` abstraction is removed — a lenient side gets a wide band, not no band. `ConstraintOutcome` now carries the corridor `(lo, hi)`.
- `register.py` — `GATE_DEFAULTS` rewritten two-sided: worse-side values carried over unchanged; the formerly unbounded side gets a loose guard (`rel_down=0.8` for grad_norm / KL-family, `rel_up=0.5` for `raw_reward`).
- `gate.py` — failure reason reports the corridor: `outside [lo, hi]`.
- `docs/ci/03-metric-history-gate.md` — constraint section rewritten (doc-dev bound).
- Tests migrated + new coverage: ref-only band scaling (the `2×ref` case now FAILs), asymmetric corridor at gate level (collapse/jump both flagged), per-side missing-param rejection, `direction` unknown-key regression.

**Baseline impact:** all e2e declarations are GATE_DEFAULTS one-liners, so the standard-metric coordinates re-key and cold-start. Deliberate: cheapest during shadow calibration, while bands are still being tuned and no enforcement exists.

## Verification

- `pytest tests/ci/test/` — 310 passed, 1 skipped.
- All 85 `register_ci_gate` specs across the 17 e2e files parse against the new schema.
- Next nightly shadow round: watch for guard-side FAILs on real runs (expected rare; bands deliberately loose).


guapisolo:
Fused into the stack instead of landing as a separate M5 PR (per review discussion): the two-sided ref-based constraint now lives in each owning layer — core redesign in #1516 (ci-history-gate-m2 @ 91d7ffe4d4), harness-layer test literals in #1517 (ci-history-gate-m3 @ 67470db158), two-sided GATE_DEFAULTS in #1630 (ci-history-gate-m4 @ 5501321274). The superseded symmetric-band commit b644da0206 was dropped. Final m4 tree = this PR's tree, plus one drift-test comment fix this branch had missed.

PR #1722, opened 2026-07-20
[fix] gate LoRA checkpoint writers on dp and cp, not dp alone

yushengsu-thu:
save_lora_checkpoint gates its writers on effective_dp.rank == 0, but adapter weights are replicated across CP ranks — with context-parallel-size > 1 every CP rank on dp rank 0 writes the same adapter_megatron_tp{t}_pp{p}.pt and adapter_model.bin concurrently. Add cp.rank == 0 to the gate.

### DIFF
--- a/miles/backends/megatron_utils/lora_utils.py
+++ b/miles/backends/megatron_utils/lora_utils.py
@@ -383,9 +383,12 @@ def save_lora_checkpoint(
     from miles.utils import megatron_bridge_utils
 
     save_path = Path(save_dir)
-    is_dp_rank_0 = get_parallel_state().effective_dp.rank == 0
-    tp_rank = get_parallel_state().tp.rank
-    pp_rank = get_parallel_state().pp.rank
+    # Adapter weights are replicated across CP ranks, so the writer gate must
+    # include CP: on DP alone, every CP rank writes the same file concurrently.
+    parallel_state = get_parallel_state()
+    is_dp_rank_0 = parallel_state.effective_dp.rank == 0 and parallel_state.cp.rank == 0
+    tp_rank = parallel_state.tp.rank
+    pp_rank = parallel_state.pp.rank
 
     # Create directory on dp_rank=0, then synchronize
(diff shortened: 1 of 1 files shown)


miles/backends/megatron_utils/lora_utils.py
```diff
@@ -383,9 +383,12 @@ def save_lora_checkpoint(
     from miles.utils import megatron_bridge_utils
 
     save_path = Path(save_dir)
-    is_dp_rank_0 = get_parallel_state().effective_dp.rank == 0
-    tp_rank = get_parallel_state().tp.rank
-    pp_rank = get_parallel_state().pp.rank
+    # Adapter weights are replicated across CP ranks, so the writer gate must
+    # include CP: on DP alone, every CP rank writes the same file concurrently.
```
yushengsu-thu: clean the comments

miles/backends/megatron_utils/lora_utils.py
```diff
@@ -383,9 +383,12 @@ def save_lora_checkpoint(
...
+    parallel_state = get_parallel_state()
+    is_dp_rank_0 = parallel_state.effective_dp.rank == 0 and parallel_state.cp.rank == 0
+    tp_rank = parallel_state.tp.rank
+    pp_rank = parallel_state.pp.rank
```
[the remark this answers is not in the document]
yushengsu-thu: verified in the bridge: the adapter export broadcasts across PP (every rank yields the full tensor set), so with PP > 1 each stage's tp0 rank was writing the same adapter_model.bin. Added pp_rank == 0 to the HF gate and renamed the flag to is_dp_cp_rank_0. The native .pt writes stay per-PP-rank on purpose — the filename carries pp_rank.

miles/backends/megatron_utils/lora_utils.py
yushengsu-thu: Done.

maocheng23 (approved):

PR #1728, opened 2026-07-20
fix(megatron): serialize Hugging Face metadata loads per node

kevinamzn:
## Summary

- limit serialized Hugging Face config/tokenizer loading to one wave per GPU position within a node
- allow matching rank positions on different nodes to load concurrently

## Why

The existing loop performs one metadata-loading wave per global rank. Trainer ranks are ordered contiguously by node and GPU, so `global_rank % num_gpus_per_node` remains unique within each node while allowing the same position on different nodes to run concurrently. This reduces startup serialization from world size to GPUs per node.

In a 64-rank, 8-node benchmark, metadata-loading critical-path time decreased from 118.39s to 28.12s (4.21x faster; 90.27s / 76.3% reduction). All 64 ranks completed one config and tokenizer load.

The existing global barrier after each wave is preserved. The optimization assumes the Hugging Face cache state being protected is node-local.

## Validation

- `pre-commit run --files miles/backends/megatron_utils/actor.py`
- `python -m py_compile miles/backends/megatron_utils/actor.py`
- 64-rank, 8-node baseline and optimized startup runs completed successfully

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -17,7 +17,7 @@ from miles.utils.argparse_utils import inplace_modify_args
 from miles.utils.audit_utils.event_logger.logger import event_logger_context
 from miles.utils.audit_utils.witness.allocator import WitnessInfo
 from miles.utils.context_utils import with_defer
-from miles.utils.distributed_utils import get_gloo_group, init_process_group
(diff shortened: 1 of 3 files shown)


kevinamzn:
Closing this draft; we are not pursuing this upstream change.

kevinamzn:
Superseded by #1753, which carries the original minimal per-node implementation as a clean two-line diff.

PR #1733, opened 2026-07-21
[AMD] Drop inert DSv4 rollout knobs and add an MTP recipe

XinyuJiangCMU:

Most of the rollout knobs the DeepSeek-V4 AMD script sets are inert. For `DeepseekV4ForCausalLM` on HIP, `ServerArgs` force-sets the same variables to the same values right after, so the script is writing environment variables that get overwritten a moment later. This drops those 13, plus three CLI workarounds that are no longer needed: `--sglang-dsa-topk-backend torch`, `--sglang-disable-custom-all-reduce`, and `--use-miles-router`.

It also adds `--enable-mtp`, which serves the checkpoint's own MTP block as an EAGLE draft with 3 speculative steps, top-k 1, and 4 draft tokens. The draft is not trained. On gfx950 the aiter all-gather deadlocks the rollout engines once speculative decoding is on, so the flag also sets `SGLANG_USE_AITER_AG=false` to fall back to RCCL for the all-gather.

I ran the MTP recipe on 4 nodes, 32 MI355X, DeepSeek-V4-Flash-FP8 colocate GRPO on dapo_aime. It completed 3 training steps with no engine restarts, at an accepted length of 2.75 to 2.79.


guapisolo (approved):
LGTM

PR #1735, opened 2026-07-21
[PPO] Share Actor/Critic GPUs

xiuhu17:
# [PPO] Share Actor/Critic GPUs

## Summary

- Let PPO actor and critic time-share the same GPU placement through sequential offload/onload, eliminating dedicated critic GPU allocation.
- Replace the actor–critic NCCL synchronization group with a Ray-based handoff: the critic returns CPU-backed value tensors, which are restored on the actor GPUs for advantage computation.
- Align rollout placement, checkpoint save, and TMS lifecycle handling with the shared-GPU design.
- Add focused lifecycle and placement tests covering actor/critic sharing and rollout GPU accounting.

## References

- [THUDM/slime#1856 — refactor/ppo](https://github.com/THUDM/slime/pull/1856): core shared actor/critic placement and critic-value handoff design.
- [THUDM/slime#1882 — fix ppo value offload bugs](https://github.com/THUDM/slime/pull/1882): PPO offload lifecycle and rollout GPU placement fixes.
- [THUDM/slime#1888 — fix checkpoint save with offloaded training](https://github.com/THUDM/slime/pull/1888): wake/sleep handling around checkpoint saves.
- [THUDM/slime#1950 — fix colocated rollout GPU accounting](https://github.com/THUDM/slime/pull/1950): correct rollout GPU resource calculation when actor and critic share placement.

## Validation

- PPO fast tests: `21 passed`
- 4×H200 PPO E2E: `3` rollouts completed with full actor and critic training
- Weight versions advanced without mixed-version samples
- Qwen3-30B-A3B PPO run with CPU memory profiling on miles dashboard: https://github.com/radixark/miles/pull/1654, https://github.com/radixark/miles/pull/1736
- Qwen3-4B-Thinking 30-step PPO run completed successfully with a reward curve consistent with the previous implementation

Screenshot 2026-07-18 at 5 19 04 PM
Screenshot 2026-07-18 at 5 19 18 PM
Screenshot 2026-07-18 at 5 19 26 PM
Screenshot 2026-07-18 at 5 19 34 PM
Screenshot 2026-07-18 at 5 19 42 PM
Screenshot 2026-07-18 at 5 19 48 PM
Screenshot 2026-07-18 at 5 19 55 PM
Screenshot 2026-07-18 at 5 20 13 PM
Screenshot 2026-07-20 at 7 24 56 PM


tests/e2e/megatron/test_qwen3_4B_ppo.py
guapisolo: as we colocated actor and critic, we can set the pp higher to 2 here.

xiuhu17: it makes sense,  just changed.

guapisolo:
Some user suggest that we should allow critic model use different parameter in Megatron. We can do this in later PRs.

guapisolo:
Megatron bridge cannot serve critic model. Will push a commit to fail loud.

xiuhu17:
> Some user suggest that we should allow critic model use different parameter in Megatron. We can do this in later PRs.

Good suggestions, do in subsequent pr.

guapisolo:
Pushed 4 commits on top of 7aceff030 (`7aceff030..0ac8303a2`):

- **032784e63 `[PPO] fix train_async: serialize critic->actor and pass critic values`** — `train_async.py` still had the old design: critic and actor training launched concurrently via `eager_create_task`, and the actor never received the critic's values. With the shared actor/critic GPU placement in this PR, concurrent training would have both models `wake_up()` on the same GPUs simultaneously (OOM), and `train_actor` now requires the critic's per-rank values as `external_data` to compute advantages. The async entry now mirrors `train.py`: `values = await critic_model.train(...)` then `actor_model.train(..., external_data=values)`. The generation/training overlap of the async loop is unaffected.
- **5af98b5d2 `[test] adapt shared-PPO lifecycle stubs to get_rollout_data tuple return`** — since the mooncake transfer change (#591), `get_rollout_data` returns `(rollout_data, store_get_result)`; the stubs in `test_shared_ppo_lifecycle.py` still returned a bare dict, so both tests failed on unpack. They pass now.
- **0ac8303a2 `[PPO] inline the critic-only warmup gate in the train loop`** — `actor_trains` was only consumed inside the `if args.use_critic:` branch where its `(not args.use_critic)` disjunct is always false; inlined the surviving `rollout_id >= args.num_critic_only_steps` condition at its single use site.
- **035670206 `[doc] clarify --num-critic-only-steps help text`** — documents that it is the critic value-function warmup period (actor frozen) and only takes effect with `--advantage-estimator ppo`.

All four verified locally: clean cherry-picks onto the current head, pre-commit hooks pass, `py_compile` on the touched entry points, and the two lifecycle tests go red → green.

guapisolo:
`miles_validate_args` force-setting `args.offload_train = True` under `use_critic` silently overrides an explicit `--no-offload-train`. Running shared-PPO without offload is a legitimate configuration for offload debugging — e.g. to rule out offload overhead / offload bugs — as long as both models fit on the shared GPUs, so the flag's contract shouldn't be swallowed.

Changed the behavior to:
- `offload_train` unset → still defaults to `True` (normal shared-PPO path unchanged);
- explicit `--no-offload-train` → kept, with a warning that this configuration is reserved for offload debugging (both models stay resident on the shared train GPUs).

Pushed 2 commits on top of `0ac8303a2` (`0ac8303a2..a34579210`):
- **8e101855b `[test] cover bridge-mode critic rejection`** — the fail-loud test promised above
- **a34579210 `[PPO] respect explicit --no-offload-train under critic`** — the behavior change described here

miles/backends/megatron_utils/actor.py
```diff
@@ -594,7 +615,10 @@ def save_model(self, rollout_id: int, force_sync: bool = False) -> None:
             return
 
         # torch dist may trigger nccl communication during saving.
-        if self.args.offload_train:
+        auto_offload = self.args.use_critic and self.args.offload_train
```
guapisolo: Also here, we'd better add "wake_up: bool = False" instead of both `use_critic` and `offload_train`

miles/backends/megatron_utils/actor.py
```diff
@@ -358,12 +359,25 @@ def train(
...
-                return self.train_actor(rollout_id, rollout_data, witness_info=witness_info, attempt=attempt)
+                result = self.train_actor(
+                    rollout_id,
+                    rollout_data,
+                    external_data=external_data,
+                    witness_info=witness_info,
+                    attempt=attempt,
+                )
+
+            if self.args.use_critic and self.args.offload_train:
```
guapisolo: I think we should add a param `sleep: bool = False` here, and control this flag in train.py / train_async.py when critic is used. This can make sure code logic is self-included.

miles/utils/arguments.py
```diff
@@ -2709,10 +2715,18 @@ def miles_validate_args(args):
...
+            args.megatron_to_hf_mode != "bridge"
+        ), "Critic models are not supported with --megatron-to-hf-mode bridge"
         args.critic_num_gpus_per_node = args.actor_num_gpus_per_node
-    if args.critic_num_nodes is None:
         args.critic_num_nodes = args.actor_num_nodes
+    else:
+        if args.critic_num_gpus_per_node is None:
+            args.critic_num_gpus_per_node = args.actor_num_gpus_per_node
+        if args.critic_num_nodes is None:
+            args.critic_num_nodes = args.actor_num_nodes
```
guapisolo: Why is there an else here? If there is no critic, these args should not be set at all — nothing reads `critic_num_gpus_per_node` / `critic_num_nodes` outside `use_critic`: `create_training_models` only touches them under `use_critic`, and the `critic_train_only` branches in `rollout_server.py` are never enabled anywhere. This was left over from main, where the critic could run on separate GPUs. Dropped the else in 4d72f2e3d so they stay `None` when there is no critic.

miles/ray/actor_group.py
```diff
@@ -74,14 +74,38 @@ async def init(self):
             indep_dp_info=indep_dp_info,
         )
 
-    async def train(self, rollout_id, rollout_data_pack):
+    async def train(self, rollout_id, rollout_data_pack, external_data=None):
```
guapisolo, line 77: This might make ft v2 crash. I will push a fix.

guapisolo (approved):
LGTM now. Better let code owner's review.

miles/utils/arguments.py
```diff
@@ -2573,10 +2573,15 @@ def miles_validate_args(args):
         )
 
     args.use_critic = args.advantage_estimator == "ppo"
-    if args.critic_num_gpus_per_node is None:
+    if args.use_critic:
+        assert args.train_backend == "megatron", "Shared Actor/Critic PPO requires the Megatron backend"
```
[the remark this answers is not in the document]
Shi-Dong: Might be worth fixing.

Zhichenzzz:
It still has some conflicts and ci fails.

miles/ray/placement_group.py
```diff
@@ -158,30 +151,32 @@ async def create_training_models(args, pgs, rollout_manager):
         rollout_manager=rollout_manager,
         with_opd_teacher=args.use_opd and args.opd_type == "megatron",
     )
+    actor_start_rollout_ids = await actor_model.init()
+
     if args.use_critic:
+        critic_args = copy.deepcopy(args)
+        critic_args.kl_coef = 0
```
yueming-yuan, line 158: Should we add a kl_coef == 0 assertion, given we force the critic model not to use kl in the colocate mode, while the old mode allows using kl?

miles/backends/megatron_utils/actor.py
```diff
@@ -482,11 +514,13 @@ def train_actor(
...
 
                 if self.args.use_critic:
-                    sync_actor_critic_data(
-                        self.args,
-                        rollout_data,
-                        self._actor_critic_groups,
-                    )
+                    if external_data is not None and get_parallel_state().is_pp_last_stage:
+                        values_ref = external_data.get("values")
+                        if values_ref is not None:
```
yueming-yuan: will there be any case that it satisfies `external_data is not None and get_parallel_state().is_pp_last_stage` but we get `values_ref == None`? If not, should we assert values_ref is not None and remove this if branch?

yueming-yuan (approved):
approved with small comments

miles/ray/placement_group.py
guapisolo: Very good finding. I just notice `kl_coef` and `kl_loss_coef ` are two different things. The preivous one will be applied to ppo gae, and the latter one is used for kl penalty.. So there should be an assert here.

guapisolo:
@xiuhu17 will continue push some code simplify

xiuhu17:
@guapisolo It seems there are some issues in the ci envitonment.

The CI image is based on SGLang v0.5.15 and includes:
sglang-kernel==0.4.4
flashinfer==0.6.12
The workflow then checks out the rolling sglang-miles source branch, which has since moved to SGLang v0.5.16 and requires:
sglang-kernel==0.4.5
flashinfer==0.6.14

guapisolo:
> @guapisolo It seems there are some issues in the ci envitonment.
> 
> The CI image is based on SGLang v0.5.15 and includes: sglang-kernel==0.4.4 flashinfer==0.6.12 The workflow then checks out the rolling sglang-miles source branch, which has since moved to SGLang v0.5.16 and requires: sglang-kernel==0.4.5 flashinfer==0.6.14

It's mainly b/c the CI run when we merged bump sgl 0.5.16 PR but the new docker is not created with `dev` flag yet. Naively rerun can fix this.

guapisolo:
I have another PR #1959 to speed up docker build. The blocker is download speed of wheels. After that we can rerun the CI

guapisolo:
I give up, let simply rerun the CI.

xiuhu17:
> I give up, let simply rerun the CI.

I have reran the 4x test with pp size = 2. 
Screenshot 2026-07-29 at 11 37 13 PM

xiuhu17:
> I give up, let simply rerun the CI. 

For the CI test failure, I reproduce locally and it seems that the Docker image contains multiple cuDNN versions: system cuDNN 9.13 and wheel cuDNN 9.22. Transformer Engine loads the older system library because its .so has no RPATH pointing to the wheel directory. GPT-OSS exercises TE fused attention and fails with CUDNN_STATUS_BAD_PARAM, while Qwen uses FlashAttention and is unaffected. Prefixing the cuDNN 9.22 wheel path in LD_LIBRARY_PATH fixes the test. CI should also avoid downgrading cuDNN during dependency reconciliation.

guapisolo:
> > I give up, let simply rerun the CI.
> 
> For the CI test failure, I reproduce locally and it seems that the Docker image contains multiple cuDNN versions: system cuDNN 9.13 and wheel cuDNN 9.22. Transformer Engine loads the older system library because its .so has no RPATH pointing to the wheel directory. GPT-OSS exercises TE fused attention and fails with CUDNN_STATUS_BAD_PARAM, while Qwen uses FlashAttention and is unaffected. Prefixing the cuDNN 9.22 wheel path in LD_LIBRARY_PATH fixes the test. CI should also avoid downgrading cuDNN during dependency reconciliation.

OK then I think it's just environment issue?

guapisolo:
let me rerun again

xiuhu17:
> * PR Test / stage-c-8-gpu-h200 (1) / run (pull_request)

now it seems the weight update has some issues for dpsk v32 fp8. I am going to look at it

xiuhu17:
Fixed in: https://github.com/radixark/miles/pull/2014

PR #1741, opened 2026-07-21
Add examples/swe-agent: GLM-4.7-Flash agentic training with Harbor

Shi-Dong:
Adds `examples/swe-agent`: a GLM-4.7-Flash agentic training example that runs synchronous GRPO in Miles while a [Harbor](https://github.com/harbor-framework/harbor) agent server creates the task sandboxes, runs the agents, and returns verifier rewards.

This PR is purely additive — 7 new files, no deletions and no edits to existing examples.

## Why

**Miles currently has no Harbor agentic *training* example outside `examples/experimental/`.** Every Harbor-based training example lives under `examples/experimental/`: `swe-agent-v2`, `swe-agent-v2-amd`, and `openenv`. The only Harbor reference outside `experimental/` is `examples/eval/terminal_bench_via_agent_server`, which is evaluation-only and does no training.

That matters because this is the reference multi-turn agentic setup on the session-server TITO path — `docs/user-guide/agentic-chat-template.md` and `docs/user-guide/rollout-endpoints.md` both send readers here as *the* worked example of a custom generate function, a custom agent function, and the abort hook. Users following those docs land in `experimental/`, in a directory that also contains an unrelated GSM8K reasoning launcher and a launcher that depends on a private repository they cannot clone.

## Naming

The directory is `swe-agent`, not `swe-agent-v2`. The `v2` suffix only made sense inside `experimental/`, where it separated this Harbor-based generation from the older Nemo-Gym-based `experimental/swe-agent`. Neither of those is being promoted, so a version suffix at the top level would point readers at a v1 that does not exist there.

## What is in the example

| File | Purpose |
| --- | --- |
| `run.py` | Synchronous GLM-4.7-Flash launcher (the validated recipe). |
| `run-glm47-flash-agentic-async.py` | Disaggregated fully asynchronous launcher. |
| `swe_agent_function.py` | Sends each rollout to the Harbor agent server; implements the abort hook. |
| `generate.py` | Builds rewards, metrics, and training samples. |
| `download_and_process_data.py` | Converts supported datasets to Miles JSONL. |
| `README.md` | End-to-end recipe: agent server, data prep, launch, verification. |

It points at the public `harbor-framework/harbor` branch `harbor-miles-v0.20.0`, so the recipe is runnable from a clean checkout. Agent-server hostnames, checkpoint paths, and the W&B project name are placeholders rather than the values the validation run happened to use.

Both launchers route rollouts through sgl-router. The deprecated `--use-miles-router` flag that the experimental copy still passes has been dropped from both, matching the configuration the validation run below actually used (`use_miles_router=False`, `router_backend=sglang`).

## Validation

A GLM-4.7-Flash Terminal-Bench 2 run on one node of 8xH200 has been training continuously on these files for several days and is past 124 synchronous GRPO steps, with checkpoints saved on schedule and TITO session tracking healthy throughout. The README documents that run's actual shape: 4 prompts x 8 samples = 32 trajectories per step, `--global-batch-size 32`, `--max-seq-len 65536`, `--num-rollout 200`, `--save-interval 20`. `run.py` defaults match.

That run surfaced two operational lessons the README passes on in general terms. Synchronous rollout waits for the slowest trajectory in the batch, so a step that draws an unusually slow task can take several times the typical 10 minutes. And W&B can silently stop accepting some metric rows while others keep arriving, which looks exactly like a frozen reward curve — so "Verify progress" says to confirm a suspected stall against the on-disk `train_data/<step>` and `rollout_data/<step>.pt` dumps rather than a dashboard.

The README also explains how to size the agent server's `--max-concurrent` (one sandbox per trajectory in a step) and why `--agent-timeout` must be generous.

## Follow-up

Deleting the now-duplicated `examples/experimental/swe-agent-v2` and repointing the two docs pages at `examples/swe-agent` will come as a separate PR, so this one stays reviewable as an addition.

## Test plan

- [x] Long-running GLM-4.7-Flash Terminal-Bench 2 training run on these files (124+ GRPO steps, one node, 8xH200)
- [x] `python -m py_compile` on all launchers
- [x] `ruff check examples/swe-agent/`
- [x] `git diff --check`
- [x] Net diff against `main` confirmed additive-only (no deletions, no modifications to existing files other than the `examples/README.md` index entry)


examples/swe-agent-v2/run.py
```diff
@@ -0,0 +1,264 @@
...
+        "--adam-beta2 0.98 "
+    )
+
+    sglang_args = (
+        "--rollout-num-gpus-per-engine 1 "
+        "--sglang-mem-fraction-static 0.7 "
+        "--sglang-tool-call-parser glm47 "
+        "--sglang-reasoning-parser glm45 "
+        "--sglang-router-port 31000 "
+        # TODO: speculative decoding has issue, need to fix later
```
Shi-Dong: Remove this line.

guapisolo (approved):
Approve to unblock

PR #1759, opened 2026-07-21
(2/2) refactor(session): assemble training samples on the session server; records never leave it

guapisolo:
> **Depends on:** #1916

## Summary

Assemble agentic training samples on the owning session server over HTTP.

## Motivation

The rollout driver previously downloaded every session record before computing, truncating, and merging the training sample. Because each record contains the full trajectory prefix, that transfer grows quadratically with the number of turns and makes the driver serialize, transfer, and parse data whose only consumer is sample assembly.

This PR establishes HTTP as the complete baseline transport. It intentionally adds no Mooncake dependency or runtime assumption; a later PR can add Mooncake as an optional optimization without changing the assembly contract.

## Before / After

- **Before / After:** Replace driver-side records download with server-side compute → truncate → unconditional merge.
- The endpoint returns one safetensors payload described by `SAMPLES_VALUE_SPEC`.
- The driver overlays the computed fields onto a deepcopy of its input `Sample`, so input-only fields remain local.
- Server-produced metadata crosses as a delta before downstream agent/session overlays.
- `OpenAIEndpointTracer.collect_samples` performs one non-retrying binary HTTP POST.

## Behavior Preservation

- **How we know:** Golden tests compare exact post-merge `Sample` values, including truncation.
- Metadata authority is `input < server < agent < session`.
- Lifecycle/debug messages survive the wire.
- `no_records` and `all_truncated` remain singleton `ABORTED` results.
- POST timeouts, transport failures, or non-2xx replies raise instead of silently losing a sample.
- Session DELETE is attempted on every path, while a DELETE failure cannot hide the POST result.

## Verification

- **Existing test suite:** 143 passed; 17 skipped across session assembly/codec, router samples-op, OpenAI endpoint client, generate-hub, plus rollout integration tests.
- Pre-commit passed.

## Review Focus

- Scrutinize `SessionCore.collect_samples`: assembly stays synchronous on the event loop so its lock-free read sees one coherent records list.
- Scrutinize `SAMPLES_VALUE_SPEC`: it is the sole field allowlist plus dtype/codec contract for the wire.
- Confirm the transport boundary is HTTP-only in this PR; Mooncake remains a separate optional follow-up.

### DIFF
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -54,6 +54,7 @@ class OpenAIEndpointTracer:
     async def collect_samples(self, input_sample: Sample, *, max_seq_len: int | None) -> SamplesReply:
         """Fetch server-assembled training samples for this session."""
         try:
+            # `asyncio.TimeoutError` propagates after cleanup is attempted for `agentic_tool_call.generate` to handle.
             payload = await post_bytes_no_retry(
                 f"{self.base_url}/samples",
                 {"max_seq_len": max_seq_len},
--- a/miles/rollout/generate_utils/openai_endpoint_utils.py
+++ b/miles/rollout/generate_utils/openai_endpoint_utils.py
@@ -7,7 +7,7 @@ import logging
 import random
 from argparse import Namespace
 
-from miles.rollout.session.samples.codec import SamplesReply, decode_samples_reply
+from miles.rollout.session.samples.codec import SamplesReply, decode_samples_and_merge_input_sample
 from miles.utils.http_utils import post, post_bytes_no_retry
 from miles.utils.types import Sample
 
@@ -78,4 +78,4 @@ class OpenAIEndpointTracer:
             except Exception as e:
                 logger.warning(f"Failed to delete session {self.session_id} after collecting samples: {e}")
 
-        return decode_samples_reply(payload, input_sample)
+        return decode_samples_and_merge_input_sample(payload, input_sample)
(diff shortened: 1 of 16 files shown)


miles/rollout/session/samples/codec.py
```diff
@@ -0,0 +1,222 @@
+"""Black-box sample packing for the `POST /sessions/{id}/samples` reply; only the
```
guapisolo: This is a vibed file, but verified by e2e CIs. No tito mismatch rate regression.

miles/rollout/generate_utils/openai_endpoint_utils.py
```diff
@@ -50,39 +51,21 @@ async def create(args: Namespace):
...
+                f"{self.base_url}/samples",
+                {"max_seq_len": max_seq_len},
                 timeout=_SESSION_REQUEST_TIMEOUT,
             )
-        except asyncio.TimeoutError:
-            logger.error(
-                f"Timed out waiting for session {self.session_id} records after {_SESSION_REQUEST_TIMEOUT}s "
-                f"(likely stale HTTP keepalive connection). Returning empty records."
-            )
-            # Still attempt to clean up the session.
```
Shi-Dong, line 64: Something I don't understand: what if `TimeoutError` is raised here? Wouldn't the error be propagated out and crash the rollout group?

guapisolo: let me carefully check this.

Shi-Dong (approved):
LGTM!

PR #1761, opened 2026-07-22
[feat] Muon support for multi-LoRA per-slot optimizers

yushengsu-thu:
Multi-LoRA currently rejects `--optimizer muon` at launch. Since the per-slot construction already goes through the stock `get_megatron_optimizer`, this just opens the same factory to muon: `dist_muon` collapses to plain `muon` (multi-LoRA applies LayerWise once over all slots itself), each per-slot child now uses its own family-matched `init_state_fn`, and slot-retirement cleanup additionally zeroes muon's `momentum_buffer` so a reused slot doesn't inherit the previous adapter's momentum. The adam/AdamW path is unchanged, and the stepping machinery and per-adapter LR scheduler were already optimizer-agnostic. Selecting muon requires the `emerging-optimizers` package, same as the non-LoRA muon path.

### DIFF
--- a/miles/backends/megatron_utils/model.py
+++ b/miles/backends/megatron_utils/model.py
@@ -164,17 +164,19 @@ def setup_model_and_optimizer(
     config = OptimizerConfig(**kwargs)
     config.timers = None
 
-    if _is_muon_optimizer(config.optimizer):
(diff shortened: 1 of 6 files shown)


Zhichenzzz:
I think current radixark/megatron version did not support muon well (we will upstream the latest one soon). so maybe after that we could check this muon for multi-lora then

PR #1762, opened 2026-07-22
feat(session): serve stream:true chat completions as fake streaming

guapisolo:
## Summary
Serve `stream: true` chat completions from the session server as fake streaming (single SSE chunk).

## Motivation
Streaming clients hitting POST /sessions/{id}/v1/chat/completions got a
deterministic 400 today: the session server always injects
return_meta_info=True, and sglang rejects that combination with
stream=true. TITO also needs the complete non-streaming response
(choices[0].message + meta_info.output_token_logprobs), so real
streaming passthrough cannot work either.

## Usage
```bash
curl -N http://<router>/sessions/<id>/v1/chat/completions \
  -d '{"messages": [{"role": "user", "content": "hi"}], "stream": true}'
```
Returns `text/event-stream` with one `chat.completion.chunk` carrying the whole message as one delta, then `data: [DONE]`. TITO tracking, `SessionRecord`, `GET /sessions/{id}` all stay identical to a non-streaming request.

## Design Notes
- **Key choices:** `chat_completions` pops `stream`/`stream_options` in Phase 1, keeping the backend call plus the `SessionRecord` non-streaming.
- `_response_to_stream_chunk` (adapted from NVIDIA-NeMo/ProRL-Agent-Server & THUDM/slime) puts all tool_calls in the single chunk with `index` set because fragmented arguments across chunks break real clients.
- Streaming clients rebuilding the assistant message drop the wire-only tool_call `index` that sglang always serializes, so `message_matches` now ignores `index` plus null-valued keys inside tool_calls.
- `MockSGLangServer` now emits `index` like real sglang, making the fast suite exercise the true wire shape.
- Errors keep their real status codes as JSON because every failure occurs before the SSE body is built.
- **Alternatives considered:** real streaming passthrough — sglang rejects `return_meta_info` with `stream=true`; per-chunk responses break TITO validation.
- Heartbeat / token slicing commits HTTP 200 before the result exists, degrading upstream errors to in-stream events.

## Verification

| test | asserts |
|---|---|
| `TestChatFakeStreaming` (tests/fast/router/test_sessions.py) | single-chunk parity with non-streaming; stream flags never reach backend or record; tool_calls in one chunk with parseable arguments; usage and `finish_reason="length"` passthrough; backend 400 passes through as JSON without a record; `stream: false` regression; openai SDK accumulation; streamed tool-call turn then tool_result continuation extends the session without rollback |
| `TestMessageMatches` (tests/fast/utils/chat_template_utils/test_template.py) | tool_call `index` int/null/absent all match; real `id`/`arguments`/`content` diffs still mismatch |

- Affected fast suites (router, mock server, generate_hub, chat template): 457 passed.

## Review Focus
- Scrutinize `_normalize_tool_calls` in miles/utils/chat_template_utils/template.py: it must ignore only wire-only artifacts, never real tool_call differences.
- Scrutinize the `_response_to_stream_chunk` delta shape (role/content/reasoning_content/tool_calls with `index`) against real client accumulators.
- Scrutinize the three `_chat_client_response` call sites: the `closing` / `num_assistant` race paths must also honor `client_stream`.

### DIFF
--- a/miles/utils/chat_template_utils/template.py
+++ b/miles/utils/chat_template_utils/template.py
@@ -134,13 +134,8 @@ def apply_chat_template_from_str(
 
 _TEMPLATE_RELEVANT_KEYS = ("role", "content", "reasoning_content", "tool_calls")
 
-# Wire-only tool_call keys ignored when matching a stored message against a
-# client replay.  The OpenAI non-streaming message shape has no ``index`` on
-# tool_calls — it is an sglang extension (ToolCall.index, the tool's position
-# in the tools list, int or null) — so protocol-faithful clients (SSE
-# accumulators rebuilding the message from deltas) legitimately replay
-# tool_calls without it, or with a renumbered value.  No chat template reads
-# it either way.
+# SGLang serializes `index` on non-streaming tool calls, while accumulated
+# streaming messages may omit or renumber it; no chat template reads it.
 _WIRE_ONLY_TOOL_CALL_KEYS = ("index",)
 
 
@@ -163,9 +158,7 @@ def _normalize_value(value: Any) -> Any:
 def _normalize_tool_calls(value: Any) -> Any:
     """Project tool_calls down to template-relevant content for comparison.
 
-    Drops wire-only keys plus null-valued keys from each tool_call dict:
-    Jinja2 renders an absent key and a None-valued key identically, so
-    neither may distinguish a stored message from its replay.
+    Only keys in `_WIRE_ONLY_TOOL_CALL_KEYS` are removed; all other values remain part of history matching.
 
     Deliberately a comparison-time projection, NOT a repair of the incoming
     message from stored state.  Matching only decides whether a replay is
@@ -180,11 +173,7 @@ def _normalize_tool_calls(value: Any) -> Any:
     if not isinstance(value, list):
         return value
     return [
-        (
-            {k: v for k, v in call.items() if k not in _WIRE_ONLY_TOOL_CALL_KEYS and v is not None}
-            if isinstance(call, dict)
-            else call
-        )
+        ({k: v for k, v in call.items() if k not in _WIRE_ONLY_TOOL_CALL_KEYS} if isinstance(call, dict) else call)
         for call in value
     ]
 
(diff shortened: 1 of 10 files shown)


miles/utils/test_utils/openai_stream_client.py
```diff
@@ -0,0 +1,87 @@
...
+        "created": first.get("created"),
+        "model": first.get("model"),
+        "choices": [{"index": 0, "message": message, "finish_reason": finish_reason}],
+    }
+    if usage is not None:
+        response["usage"] = usage
+    return response
+
+
+async def stream_chat_completions(client, url: str, payload: dict, *, label: str = "") -> dict:
```
Shi-Dong: Nit: please annotate `client` as well.

miles/utils/chat_template_utils/template.py
```diff
@@ -151,16 +160,52 @@ def _normalize_value(value: Any) -> Any:
     return value
 
 
+def _normalize_tool_calls(value: Any) -> Any:
+    """Project tool_calls down to template-relevant content for comparison.
+
+    Drops wire-only keys plus null-valued keys from each tool_call dict:
+    Jinja2 renders an absent key and a None-valued key identically, so
```
Shi-Dong: Are you sure? I think `None` is rendered as `"None"` in Jinja, and an absent key is rendered as an empty string.

Shi-Dong (approved):
Conditional LGTM. Please take a look at `_normalize_tool_calls`.

miles/utils/chat_template_utils/template.py
guapisolo: Sry this is vibed. (

miles/utils/test_utils/openai_stream_client.py
guapisolo: Added the concrete `httpx.AsyncClient` annotation and import in 42a6a1d609d05de1747fc4061a28ae4a934750ed.

miles/utils/chat_template_utils/template.py
guapisolo: You're right—Jinja distinguishes an explicit `None` from an absent key. In 42a6a1d609d05de1747fc4061a28ae4a934750ed, `_normalize_tool_calls` now removes only the wire-only `index`; every other key/value remains part of history matching. I also added a regression test for `id=None` versus an absent `id`, while preserving the intended `index` int/null/absent equivalence.

PR #1586, opened 2026-07-06
feat(megatron): add post-save checkpoint hook

waitong94:
## Summary

- Add a `--custom-megatron-post-save-hook-path` callback that runs on rank 0 after checkpoint saves.
- Document the CLI flag, customization hook signature, and argument parsing/validation.

## Why

This gives users a generic extension point for post-checkpoint work such as exporting, copying, or indexing saved checkpoints without patching the Megatron save path.

## Hook Behavior

When `--custom-megatron-post-save-hook-path` is set, the Megatron train actor runs the hook after the normal checkpoint save path completes. The hook only runs on global rank 0 so distributed jobs do not duplicate uploads, exports, or file writes.

If `--async-save` is enabled, the actor first blocks on `maybe_finalize_async_save(blocking=True)` before invoking the hook. This ensures the checkpoint files are fully written before user code observes or copies them.

The hook is loaded with `load_function` from the configured module path and called as:

```python
hook(args, rollout_id, checkpoint_dir, hf_checkpoint_dir)
```

Hook inputs:

- `args`: the runtime argument namespace for the training actor.
- `rollout_id`: the checkpoint iteration that just completed.
- `checkpoint_dir`: the Megatron checkpoint directory for `rollout_id`.
- `hf_checkpoint_dir`: the Hugging Face checkpoint directory when `--save-hf` is configured for the actor, otherwise `None`.

## Validation

- `python -m pytest tests/fast/utils/test_arguments.py -q`
- `python -m py_compile miles/backends/megatron_utils/actor.py miles/utils/arguments.py tests/fast/utils/test_arguments.py`
- `black tests/fast/utils/test_arguments.py`
- `SKIP=black pre-commit run --files miles/backends/megatron_utils/actor.py miles/utils/arguments.py tests/fast/utils/test_arguments.py docs/user-guide/cli-reference.md docs/user-guide/customization.md`
- `git diff --check`

PR #1763, opened 2026-07-22
fix(megatron): finalize async saves on all ranks before hooks

HJSang:
## Summary

- finalize Megatron asynchronous checkpoint saves on every training rank before invoking a post-save hook
- keep post-save hook execution restricted to rank 0
- add a regression test covering rank 0 and nonzero-rank ordering

## Why

Megatron distributed-checkpoint finalization performs distributed coordination. The post-save hook path previously called blocking finalization only on rank 0, while the other ranks returned from `save_model`. Multi-rank jobs could therefore hang indefinitely before the hook ran.

This preserves the existing contract introduced in https://github.com/radixark/miles/pull/1586: asynchronous saves are fully written before the hook observes them, and the hook itself runs once on rank 0. The only change is that checkpoint finalization now executes collectively.

## Validation

- `pytest -q --confcutdir=tests/fast/backends/megatron_utils tests/fast/backends/megatron_utils/test_actor_checkpoint.py` (2 passed in `radixark/miles:dev`)
- `pre-commit run --files miles/backends/megatron_utils/actor.py tests/fast/backends/megatron_utils/test_actor_checkpoint.py`
- multi-rank reproduction: before the fix, rank 0 remained in Megatron async-call finalization after peer ranks returned; collective finalization completed normally with the fix

## GPU CI

A maintainer must add `run-ci-ckpt` and `run-ci-megatron` before this draft is marked ready for review.


HJSang:
Can someone take a look at this simple PR? it fixed async bug. Thanks!

PR #1767, opened 2026-07-22
fix: filter_long_prompt invalid return types and multimodal handling

mathewjhan:
Currently, filtering returns a boolean in some cases when it should return `List[Sample]`. In addition, it doesn't handle processors correctly for some multimodal cases. This PR ports over the slime fixes to filter_long_prompt [see](https://github.com/THUDM/slime/blob/main/slime/utils/data.py#L81).


yushengsu-thu (approved):

yushengsu-thu:
heads up — tested this after merge and the multimodal case still crashes when a length limit is set.

by the time we're in the processor branch, `sample.prompt` is already the rendered template *string* (the `isinstance` guard above returned otherwise), and `qwen_vl_utils.extract_vision_info` can't take a string:

```
TypeError: string indices must be integers, not 'str'
```

repro: any dataset with actual image samples + `--rollout-max-prompt-len` (or `--rollout-max-context-len`, which auto-fills it). The examples don't set either flag so the filter is skipped entirely — that's why CI doesn't catch it.

To be fair this crashed before this PR too (even for text-only samples, which this PR did fix), so main is strictly better now — but the multimodal half of the fix never actually kicks in.

I think the loop should just reuse `sample.multimodal_inputs` (already computed from the original message list in `Dataset.__init__`) with `call_processor`, same as what rollout does in `sglang_rollout.py:151`:

```python
for sample in multimodal:
    processor_output = call_processor(processor, sample.prompt, sample.multimodal_inputs)
    ...
```

that would also fix the token count — right now images never get passed, so the image token expansion isn't counted at all.

can send a follow-up PR if that sounds right

PR #1777, opened 2026-07-23
feat(tito): remove --tito-allow-append-roles arg

guapisolo:
> **Depends on:** #1628

## Summary
Default TITO allowed append roles flip from tool-only to tool plus user.

## Motivation
Every TITO family (V3.2 included, since its vendored encoder honors render-level `drop_thinking=False`) now registers an append-only `{tool, user}` surface, so the conservative tool-only default just forces every multi-user-turn config to hand-pin the flag.

## Usage
```bash
# user appends accepted by default:
python train.py --use-session-server --tito-model deepseekv32 ...
# explicit roles without a session server now fail loud:
python train.py --tito-allowed-append-roles tool ...
```

The second command raises at argument validation instead of silently passing.

## Design Notes
- **Key choices:** the CLI default becomes a `None` sentinel; the misconfig gate keys on "explicitly passed".
- **Key choices:** an explicit `--tito-allowed-append-roles tool` without `--use-session-server` now fails loud where it previously slipped through.
- **Key choices:** the user-append render warning narrows to `--tito-model=default` under a session server; fixed families hard-fail unsupported role combos at template resolution instead.
- **Key choices:** `ModelConfig.allowed_append_roles` gains the default `("tool", "user")`; per-model explicit overrides stay untouched.

## Verification
- **Tests added:** `test_default_policy_allows_user_append` pins the new default policy.
- **Tests added:** the tool-only policy tests move to an explicit `registry_tool_only` fixture.
- **Existing test suite:** 2978 fast tests pass across `tests/fast/utils`, `tests/fast/router`, `tests/fast/rollout`.

## Review Focus
- Scrutinize the `None`-sentinel misconfig gate rework in `miles/utils/arguments.py`.
- Scrutinize the narrowed user-append warning conditions.


miles/rollout/session/core.py
```diff
@@ -223,20 +223,48 @@ async def chat_completions(
...
             request_body["no_stop_trim"] = False
-            # Chat template kwargs should also be forwarded to sglang to make sure
-            # parsers work correctly.
-            server_ctk = self.registry.tito_tokenizer.chat_template_kwargs
-            if server_ctk:
-                request_body["chat_template_kwargs"] = {
-                    **server_ctk,
-                    **(request_body.get("chat_template_kwargs") or {}),
-                }
+            # Client chat_template_kwargs override launch kwargs into one
```
guapisolo: Keep request's chat_template_kwargs and override some key fields by tito tokenizer chat template kwargs

PR #1778, opened 2026-07-23
refactor(tito): allowed append roles become a pure harness gate

guapisolo:
> **Depends on:** #1777

## Summary
Template resolution becomes role-independent; append roles only gate harness behavior.

## Motivation
`resolve_fixed_chat_template` keyed template rows on `allowed_append_roles`, forcing every new append shape (mid-session system injection, assistant input for compaction or few-shot) through template-registration ceremony. With preserve-think kwargs pinned as family constants, one template per family serves every role set. Default-config renders stay unchanged: after #1777 they already rode the multi-role rows this PR folds down.

## Before / After
- **Before / After:** before, each family registered `SUPPORTED_TEMPLATES` rows per role set, resolved by smallest superset; after, each family registers one `FixedTemplate`, resolved by `resolve_fixed_chat_template(tito_model)` with no roles argument.
- **What moved where:** preserve-think kwargs fold from the multi-role rows onto the single `FIXED_TEMPLATE`; `tokenize_additional_non_assistant` becomes `tokenize_additional_messages`; `VALID_APPEND_ROLES` (now including `assistant`) becomes the single source for the CLI choices.
- **What moved where:** `qwen3.5_fixed.jinja` renders mid-session system messages instead of raising; the vendored dsv32 reasoning-required raise narrows to the upstream `drop_thinking=True` path; MiniMax tool-only configs swap from the HF-native template to the fixed jinja.
- **What moved where:** injected assistant input joins the prompt region of the next sample; the loss mask only covers generated response tokens, so it trains with loss 0 without new masking code.

## Behavior Preservation
- **How we know:** per-family byte tests prove intermediate-system, injected-assistant, assistant-then-user appends extend renders byte-for-byte on all nine jinja/HF families plus both DeepSeek encoder bridges.
- **How we know:** MiniMax native-vs-fixed renders verified byte-identical on tool-loop histories regardless of the generation prompt; Kimi K2.5/K2.6 are untouched because their single `{tool, user}` row already served every subset request.
- **How we know:** explicit tool-only configs change render bytes only when the initial prompt carries historical user/assistant thinking turns — the folded kwargs previously applied to multi-role surfaces only.

## Verification
- **Existing test suite:** 3015 fast tests pass across `tests/fast/utils`, `tests/fast/router`, `tests/fast/rollout`; `test_fixed_templates.py` is rewritten as a full family matrix over the role-free resolver.

## Review Focus
- Scrutinize the collapsed per-family `FIXED_TEMPLATE` kwargs against the previous multi-role rows.
- Scrutinize the `qwen3.5_fixed.jinja` mid-session system block.
- Scrutinize the narrowed dsv32 raise gate.


guapisolo:
Superseded by #1777: the flag-as-harness-gate design evolved into binding append roles to fixed templates (f230f074ae) and removing --tito-allowed-append-roles outright with FIXED_TEMPLATE-derived verify surfaces (19c31708b6). The still-valid test coverage from this branch (restricted-family schedule derivation, assistant-input rollback/regenerate, injected-assistants-are-not-generated-checkpoints) has been ported into those commits on #1777.

PR #1784, opened 2026-07-24
feat(docker): overlay mooncake structured object store module pending release

guapisolo:
## Summary
Ship the mooncake structured object store API in the CUDA image for the mooncake rollout backend.

## Motivation
miles' mooncake rollout-data object store backend needs the structured
object store API (FieldSchema, export_ref/import_ref, unified put/get,
release_result) that is merged upstream in kvcache-ai/Mooncake #2907,
#3013 and #3023 but missing from every released wheel: the v0.3.12
release branch was cut before those merges landed.

## Usage
```bash
docker build -f docker/Dockerfile .
# inside the built image, the API the mooncake object-store backend imports resolves:
python3 -c "from mooncake.structured_object_store import FieldSchema, MooncakeBundleTransfer, export_ref, import_ref"
```
This unblocks `--object-store-backend mooncake` (PR #591) on images built from this Dockerfile.

## Design Notes
- **Key choices:** overlay the single pure-Python module `structured_object_store.py` at pinned upstream commit `4dbe5a4c` over the installed package.
- The upstream API commits touch only that module.
- The module runs unchanged on the C++ core the sglang base image already ships.
- The new RUN layer import-checks every required symbol, so an incomplete overlay fails the image build.
- **Alternatives considered:** bumping to the v0.3.12 wheel was rejected because its release branch predates the API merges.
- Rebuilding mooncake from source was rejected as too heavy for a temporary gap that one release will close.
- `docker/Dockerfile.rocm` is unchanged: whether its base image ships mooncake is unverified.

## Verification
- **Tests added:** none; the RUN layer's import check gates every image build.
- Local round-trip with the overlaid module passes against the exact wheel the current image ships (`mooncake-transfer-engine-cuda13` 0.3.11.post1).
- The round-trip covers put with field schemas, ref export/import, get, `release_result`, `cleanup_dataproto` over tcp against a local `mooncake_master`.

## Review Focus
- Scrutinize the `MOONCAKE_SOS_COMMIT` pin: it must cover every symbol `miles/utils/object_store.py` imports when PR #591 lands.
- Scrutinize the import-check line in the new RUN layer; it is the only gate keeping an incomplete overlay out of runtime.
- Scrutinize the drop condition: the layer comment binds removal to the first mooncake release containing the API.


yueming-yuan (approved):

Zhichenzzz:
wait for a second to merge after this docker ci https://github.com/radixark/miles/pull/1791 merged

PR #1790, opened 2026-07-24
openenv/tbench2: score the shared-server leg natively; retire the adapter compensation

nblintao:
Follow-up to #1675, resolving [the review thread](https://github.com/radixark/miles/pull/1675#discussion_r3645321595) about retiring the `native_evaluate=False` branch. **Draft until the shared-server deployment upgrades** (gate 1 of 2 — the upstream contract — cleared when huggingface/OpenEnv#1012 merged; see Landing order below).

## What

Both agent-function legs now speak ONE scoring protocol: raw exec commands (the server resolves the task image's real WORKDIR) and the standard `evaluate` action (canonical `tests/test.sh`, run server-side). The adapter-side compensation machinery the shared leg carried for older `tbench2_env` deployments is deleted:

- `_apply_workdir` + the `OPENENV_TASK_WORKDIR` / `OPENENV_TB2_TESTS_SRC` knobs
- `_CANONICAL_EVAL_CMD` + the reward / test.sh-rc markers and their parsers
- the `testsh_rc` metrics plumbing
- the `native_evaluate` episode-wiring parameter itself — legs now differ only in `run_body` (how an env comes into being) and `post_episode` (the shared server keeps its trial-dir purge; a per-episode sandbox needs none)

Plus two consistency fixes in the same spirit: `scan_golden.py` adopts the loop's no-verdict semantics (a server-side scoring failure or non-canonical harness prints as `ERR`, not as a fake `0.0` in a golden baseline), and the docs/README describe the server requirement as a version contract (tbench2_env ≥ the OpenEnv#1012 merge, `04d259ea6`) instead of narrating a particular old deployment's behavior.

## Why deletion needed an upstream PR first

The premise "the fixes are already in OpenEnv" held for **local mode** only (huggingface/OpenEnv#965 + #972 — what the Daytona leg runs). **Docker mode** — what the shared server runs — still scored via bare `pytest tests/` from `/task` and executed agent commands there, so flipping the shared leg to native scoring would have traded canonical-in-\/app for bare-pytest-in-\/task: a fidelity regression, not a cleanup. huggingface/OpenEnv#1012 brings docker mode up to the same contract; this PR is its miles-side consumer.

## Runtime contract guard (replaces an impossible preflight)

The Daytona leg preflights the *installed server source* before launch; against a **remote** shared server that's impossible — and an old deployment doesn't fail, it returns a plausible-looking (mis-scored) reward for every episode. The guard: `evaluate` must carry the canonical harness marker (`info.harness == "tests/test.sh"`, emitted by current upstream). Anything else — an older server's bare-pytest info, a server-side scoring error, a missing reward — drops the sample with a warning. An out-of-date deployment therefore surfaces as **every sample dropping loudly**, never as a believable reward curve. Composes with #1675's existing error/None-reward drop.

Known behavior change: task dirs shipping no `tests/test.sh` (scored by the server's pytest *fallback*, which carries no harness marker) are also dropped — deliberate, since that path is indistinguishable from an old server; all 89 official TB2 tasks and the synth pools ship `test.sh`.

## Validation

**Post-merge re-validation** (2026-07-29, after OpenEnv#1012 landed — everything below runs against a `TB2_MODE=docker` server freshly installed from **merged upstream main**, i.e. exactly what the upgraded deployment will run; also re-proves the README install instructions as written):

- **Offline**: `pytest examples/experimental/openenv/tests/ -q` → 21 passed, incl. a test asserting an old-server-shaped `evaluate` reply (valid-looking reward, `{tests_passed, exit_code}` info) yields `reward=None`.
- **Server contract (no LLM)**: golden replay of official solutions — fix-git **1.0** (agent cwd probed as `/app/personal-site`, the image's real WORKDIR), chess-best-move **1.0**, no-solution control **0.0**, imageless task dir rejected loudly at reset.
- **miles integration (this branch's `run_episode`, real episode path)**: stub policy control → reward **0.0** (guard passes a canonical server; genuine failure scored, not dropped); DeepSeek-V4-Flash as policy → **fix-git solved in 13 turns, 1/1, 0 errors**.

Earlier rounds (against the pre-merge #1012 branch) additionally demonstrated the **negative case live**: the same stub against the currently-deployed older shared server → `evaluate produced no canonical verdict (error='', harness=''); dropping episode`, reward **None** — the exact silent mis-scoring this guard exists to block.

Not covered without GPU (deferred to a 4-GPU smoke at cutover): the session-server TITO wiring + GRPO loop of `run()` — shared code across both legs, already exercised by #1675's 4×H200 validation.

## Landing order (why draft)

1. ~~huggingface/OpenEnv#1012 merges (docker-mode canonical contract).~~ **Done** — merged as `04d259ea6`.
2. The shared-server deployment reinstalls `tbench2_env` from upstream ≥ `04d259ea6` and restarts — this also closes the old deployment's reward-hacking hole (tests/ and solution/ visible to the agent in `/task`).
3. This PR lands. Note the cutover couples client and server: the OLD client's compensation breaks against the NEW server (tests withheld from `/task`), and this NEW client drops everything against the OLD server — so step 2 and 3 should happen together, coordinated with whoever operates the deployment.

### DIFF
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -21,14 +21,14 @@ Env vars:
   AGENT_MODEL_NAME   model name sent to the policy (default: "model")
   MILES_ROUTER_EXTERNAL_HOST  optional host rewrite for off-cluster agents
 
-Server contract: the env server must run CURRENT upstream tbench2_env (the
-same pinned install the README describes for the Daytona leg) — canonical
-tests/test.sh scoring inside the standard ``evaluate`` action, task WORKDIR
-resolved server-side, verifier assets withheld. Since huggingface/OpenEnv#965
-+ #972 that is what local mode provides, and huggingface/OpenEnv#1012 brings
-docker mode to the same contract. Episodes against an OLDER deployment are
-not silently mis-scored: its ``evaluate`` carries no canonical-harness marker,
-so every episode is dropped with a warning (see the guard in _multi_turn).
+Server contract: the env server must run tbench2_env at or after the
+huggingface/OpenEnv#1012 merge (04d259ea6; install per the README) —
+canonical tests/test.sh scoring inside the standard ``evaluate`` action, task
+WORKDIR resolved server-side, verifier assets withheld. The adapter verifies
+the contract on every episode rather than trusting the deployment: an
+``evaluate`` reply without the canonical-harness marker is treated as no
+verdict and the episode is dropped with a warning (see the guard in
+_multi_turn).
 
 Daytona-sandbox variant: ``openenv_daytona_agent_function`` (sibling module)
 is a drop-in ``--custom-agent-function-path`` alternative that runs every
@@ -322,12 +322,12 @@ async def _multi_turn(
         #   - reward=None / `error` set: the scoring step itself errored
         #     server-side (toolkit timeout, staging I/O) -- not tests failing.
         #   - harness marker absent: the server scored, but not through the
-        #     canonical tests/test.sh (an older tbench2_env deployment whose
-        #     `evaluate` is bare pytest in the wrong dir, or a task dir
-        #     without test.sh scored by the server's pytest fallback). The
-        #     reward LOOKS valid, which is exactly why it must not be trusted:
-        #     source preflight is impossible against a remote server, so this
-        #     marker is the contract check.
+        #     canonical tests/test.sh (a tbench2_env install predating the
+        #     contract in the module docstring, or a task dir without test.sh
+        #     scored by the server's pytest fallback). The reward LOOKS
+        #     valid, which is exactly why it must not be trusted: source
+        #     preflight is impossible against a remote server, so this marker
+        #     is the contract check.
         raw_reward = getattr(eval_result, "reward", None)
         eval_error = _obs_field(eval_result, "error")
         harness = str(_obs_info(eval_result).get("harness", ""))
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -173,11 +173,6 @@ _DEFAULT_ENV_URL = "http://localhost:8003"
 #                       shared server accumulates trial dirs; a Daytona
 #                       sandbox lives only for its episode and needs nothing).
 #
-# Scoring is identical on both legs: the standard `evaluate` action against a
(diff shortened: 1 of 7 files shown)


examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -82,86 +86,6 @@
...
-        if _REWARD_MARKER in line:
-            raw = line.split(_REWARD_MARKER, 1)[1].strip()
-            if not raw:
-                return None
-            try:
-                return float(raw)
-            except ValueError:
-                return None
-    return None
-
```
Shi-Dong, line 151: Seems that the behavior is a little bit different [in OpenEnv](https://github.com/huggingface/OpenEnv/blob/04d259ea692240160f6034a3f97dcf2180bf66b1/envs/tbench2_env/server/tbench2_env_environment.py#L204-L212)?  If `reward.txt` is missing, OpenEnv gives 0.0 reward, whereas this deleted function gives `None`?

examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -82,86 +86,6 @@
...
-            try:
-                return int(raw)
-            except ValueError:
-                return None
-    return None
-
-
 # Per-message WS recv timeout. Docker-mode tbench2 reset (container create),
 # exec, and evaluate (pytest) each routinely exceed the EnvClient default of 60s.
 _MESSAGE_TIMEOUT_S = float(os.getenv("OPENENV_MESSAGE_TIMEOUT_S", "600"))
```
Shi-Dong: Seems that OpenEnv [is giving](https://github.com/huggingface/OpenEnv/blob/04d259ea692240160f6034a3f97dcf2180bf66b1/envs/tbench2_env/server/tbench2_env_environment.py#L1011) the verifier 900 seconds by default?

Shi-Dong (approved):

examples/experimental/openenv/openenv_agent_function.py
nblintao: Oh, good catch. Let me fix that in openenv. Could you review it? https://github.com/huggingface/OpenEnv/pull/1025

examples/experimental/openenv/openenv_agent_function.py
nblintao: Updated to 1200s. It should always be greater than the timeout on openenv side. It would be better if this can be managed automically. Created an issue at https://github.com/huggingface/OpenEnv/issues/1026

PR #1791, opened 2026-07-24
ci(docker): PR-side image build and run the GPU matrix inside the freshly built image

Zhichenzzz:
## Problem

`build-and-push` only runs after merge, so Dockerfile breakage lands on `main` first — and the tests that do run on a docker PR execute in the *old* `dev` image, proving nothing about the image the PR produces.

## Change

**PR-side docker CI, integrated into `pr-test.yml`.** When a PR touches `docker/Dockerfile`, `docker/build.py`, `docker/patch/**`, or `requirements.txt`:

- `docker-paths` detects the change; `docker-build` builds `cu13-x86` and pushes a PR-scoped `radixark/miles:pr-<num>` tag (same-repo PRs; forks skip and stay on `dev`).
- `resolve-ci-image` waits for the build and redirects **every GPU suite into the freshly built image**; a failed build stops the matrix instead of green-lighting the stale image. An explicit `ci-image-tag:` PR-body directive still wins.
- `docker-pr-tag-cleanup.yml` deletes the tag when the PR closes; while the PR is open the tag stays available for re-runs.

Carried fixes, all surfaced by this pipeline validating itself:

- TE patch step resolves the install dir via `importlib.util.find_spec` — TE 2.17 dlopens `libcuda.so.1` at import, which doesn't exist in the GPU-less build container.
- NVFP4 reference-quantizer import moved to `quantization_ref_nvfp4` (TE 2.17 module rename); caught by the GPU matrix running inside the new image, where the old path no longer exists.
- `docs/ci/02-docker-build.md` documents the new flow.

## Verification

Self-validating on this PR (head `6bc79c4`): `docker-build` 23:25→23:35, `pr-1791` tag pushed 23:35:06, `resolve-ci-image` 23:35:34, GPU suites from 23:40 — strictly ordered after the fresh image. Full matrix (stage-b 2-GPU, stage-c 2/4/8-GPU on H100+H200) green inside `radixark/miles:pr-1791`.


guapisolo:
Use https://github.com/radixark/miles/blob/main/.claude/skills/doc-dev/SKILL.md to change CI docs.

guapisolo:
Looks like the docker re-built is no exec before gpu job run.

guapisolo:
Could we modify `.github/workflows/pr-test.yml` and redirect `resolve-ci-image` to `docker-build-pr` job ?

guapisolo (approved):
LGTM.

PR #1792, opened 2026-07-25
[lora] Native LoRA for standard Megatron-core raw mode models

Zhichenzzz:
## Motivation

LoRA in miles requires `--megatron-to-hf-mode bridge`, so models trained through miles' own model provider (raw mode) cannot use it at all: `setup_model_and_optimizer` only reaches the LoRA setup through `_setup_lora_model_via_bridge`, and the raw branch silently builds a full-parameter model even when `--lora-rank` is set.

This PR adds a native LoRA path for raw mode, shipped as a plugin (`miles_plugins/lora/`). It attaches adapters directly to the mcore model the provider builds, before the Float16Module / DDP wrap, so DDP sees an already-frozen base and only allocates grad buffers for adapter params. **Bridge-mode LoRA is untouched** — it keeps its own model setup, exporter, and PP behavior.

The branch absorbed three follow-ups as it matured: #2017 (extraction into the plugin), #2072 (function-level redesign + fixes from validating four models end-to-end), and — after merging main's Inkling LoRA (#2122) — a rewrite of Inkling onto the plugin architecture, replacing the model-specific monolith.

## Layout

```python
miles_plugins/lora/
    __init__.py
    lora.py              # Single public entry point and lifecycle orchestration
    config.py            # LoRAConfig: rank, alpha, dropout, and target modules
    registry.py          # model_type -> LoRAArchSpec; launcher helpers
    hf_adapter.py        # HF/PEFT naming (incl. megatron->HF conversion), import, export
    sglang_adapter.py    # SGLang serving export (fused-buffer zero-fill, target expansion)
    checkpointing.py     # Adapter checkpoint state dicts and per-rank shard naming
    distributed.py       # TP/EP/SP communication and gradient handling (ParallelGather)

    spec/
      base.py            # Contracts: ProjectionSpec, AttachContext, protocols, enums
      layout.py          # Declarative layouts: dim resolvers, bindings, attach walk, spec bases
      attention.py       # GQA / MLA / GDN / hybrid attention spec classes
      mlp.py             # Fused gate/up + down spec class
      moe.py             # Routed-expert per-layer policy; Inkling grouped experts
      lm_head.py         # Output-head specs (Inkling's muP lm_head)

    modules/
      linear.py          # LoRALinear + LoRASplitAdapter (SplitQKV / SplitFC1)
      moe.py             # Grouped-GEMM expert, shared-experts, and lm_head adapters

```

## Design

**Classes carry the taxonomy; every fact is declared once and derived everywhere else.**

- **Declarative layouts** (`spec/layout.py`): an architecture spec is a class holding a `ModuleLayout` table — which projections exist, on which physical linear, with which shard geometry. One shared `attach_layout()` implements the existence-check → target-filter → guard → build → hook walk; the dim-resolver section is the only spec-side reader of MCore attribute names. A new layout is ~15 lines of facts.
- **Spec hierarchy** (`LayoutSpec` → `AttentionSpecBase` → `GQA/MLA/GDN/Inkling`, `HybridGQAGDN(GQAAttentionSpec)`, `SharedOuterExpertMoESpec(GeneralExpertMoESpec)`): `supported_targets`, the canonical `--target-modules` order, and SGLang's fused-family expansion all derive from the layout declaration; a new architecture's fused groups reach serving-side target expansion automatically. `ShardLayout`/`AttentionFamily` str-enums replace bare strings; no parallel name tables in the exporters.
- **Guards instead of silent wrongness**: layouts the specs would get wrong (e.g. `num_query_groups < TP`, where mcore re-gathers and re-slices qkv) fail at startup with a pointer to bridge mode instead of mis-sharding silently.
- **Public export descriptors**: adapters expose `exports() -> ProjectionExport` (with `SGLangFusedGroup` metadata); `sglang_adapter.py` consumes only that API.
- **Adapter IO hooks**: modules whose export shape isn't a per-projection column/row gather override `export_plan(gather)` / `load_plan_custom(take)`; `hf_adapter.py` stays generic. `ParallelGather` batches async all-gathers over the TP *or* EP group behind one `request(tensor, dim, group=...)` call.
- **Escape hatches for non-table structure**: `ModuleLayout.hf_block_prefix` and `ProjectionBinding.adapter_class` override HF naming and the module class per binding; `LoRAArchSpec` carries `moe.attach(...)` and an optional `lm_head` spec so a fully custom family still routes through the one orchestrator in `lora.py`.
- **Launcher helpers**: `default_target_modules(hf_checkpoint)` and `preflight_native_lora(...)` (registry ∧ mbridge bridge ∧ model-args audit, no GPU) — launch scripts stop re-declaring per-family target strings and fail in seconds when a conversion bridge is missing.

Two numerics details worth calling out:

- **Fused-qkv layout.** mcore emits qkv per query group (`q1 q2 k1 v1 | q3 q4 k2 v2 | ...`). The split adapter keeps `B` in plain per-projection order and permutes the *delta* at forward time, so export/load stay simple row gathers.
- **Replicated-param grads.** A column-parallel `A` is replicated across TP while each rank holds only its slice of `B`, so every rank computes a partial `dL/dA` and the true gradient is their sum; the same applies to replicated `B` (row-parallel) and to MLA's replicated down-projections once sequence parallelism shards the sequence. Tagged at creation and summed in `finalize_model_grads` before the DP reduce-scatter; a no-op when nothing is tagged.

## Architecture coverage

| spec | model_type |
|---|---|
| GQA (fused qkv, shared-outer MoE) | `llama`, `qwen2`, `qwen2_moe`, `qwen3`, `qwen3_moe`, `mimo`, `glm4`, `glm4_moe` |
| hybrid GQA(+output gate)/GDN | `qwen3_5`, `qwen3_5_moe`, `qwen3_6`, `qwen3_6_moe`, `qwen3_next` |
| MLA (+DSA) | `deepseek_v3`, `deepseek_v32`, `glm4_moe_lite`, `glm_moe_dsa`, `kimi_k2`, `kimi_k25`, `joyai_llm_flash` |
| Inkling TML (4-way fused attn, grouped experts, lm_head) | `inkling_mm_model` |

Which of these have an end-to-end run on record is the Validation section below.

## Inkling on the plugin (replaces the #2122 monolith)

`miles_plugins/models/inkling/lora.py` (759 lines, its own attach/export/IO stack) is deleted; Inkling's specs live in the role files next to the base classes they inherit from (`InklingAttentionSpec` in `spec/attention.py`, `InklingDenseMLPSpec` in `spec/mlp.py`, `InklingMoESpec` in `spec/moe.py`, `InklingLMHeadSpec` in `spec/lm_head.py`):

- **Attention**: 4-way fused `[q;k;v;r]` declared as one `FusedAttach` — Inkling's fusion is a plain concat (no GQA group-permute), so the split-adapter row table `{q,k,v,r}` is the entire model-specific code. TML export names (`attn.wq_du/wk_dv/wv_dv/wr_du/wo_ud`) come straight from the `ProjectionSpec`s.
- **Dense MLP**: fused gate/up exports as the single TML `gate_up_proj` projection via an `adapter_class` override whose `export_plan` gathers the gate/up halves separately (a plain dim-0 TP gather would interleave ranks).
- **Routed experts**: `LoRAGroupedFC1/FC2` — shared-A, per-expert-B grouped-GEMM deltas (`F.grouped_mm` over the router's token offsets), EP-aware export/load. **Shared experts**: one adapter with per-sub-expert B slices, hooked into each sub-expert's forward.
- **lm_head**: `LoRAOutputHead` with muP logit scaling and unpadded-vocab trim.
- **Serving**: TML names are auto-detected by SGLang, so the spec declares `sglang_lora_target_modules=("all",)` and `sglang_adapter.sglang_target_modules()` answers launch/sync target lists for every native run — core carries no model-specific serving logic.

## Fixes found by running the models natively

- **`miles_plugins/mbridge/kimi_k25.py`** (new): raw-mode K2.5 conversion needs an mbridge bridge (bridge mode uses megatron.bridge's own `KimiK25VL` bridge and never converts). Thin `DeepseekV3Bridge` subclass: `text_config` descent + `language_model.` prefix remap.
- **`megatron_to_hf` kimi dispatch**: `_convert_to_hf_core` matched only `kimi_k25`, but the direct weight iterator passes `kimik25config` (dead branch); the converter also lacked the raw text-only provider's unprefixed top-level weight names.
- **`convert_kimi_int4_to_bf16.py` strips `quantization_config`**: a surviving one sends SGLang through its CompressedTensors path, which serves the BF16 checkpoint with a degenerate context-free forward → garbage rollouts, train/rollout `logprob_abs_diff` ~2.2. Verified by a megatron-vs-HF-vs-SGLang three-way logprob comparison.
- **`torch_dsa_topk` clamps k to the key length**: short sequences (< `index_topk` keys) crashed `torch.topk` in raw-mode log-prob compute.
- **`run_glm5_2_744b_a40b_lora_native.py`** (new): raw-mode GLM-5.2 launcher; its toy conversion must be single-rank (any PP split of the 5-layer toy starts a stage on a DSA skip layer).

## Also in this change

- **`--debug-lora-train-only`**: train adapters in Megatron while rollout stays on the frozen base policy. The rollout-facing gate becomes `lora_rollout_enabled()`, so SGLang's `enable_lora`, the per-request `lora_path`, and the adapter sync all follow it. Useful for isolating training-side numerics.
- **`--check-lora-weight-equal`**: sha256 comparison of every adapter tensor between Megatron's export and the engine's received copy, verified to fail (not pass vacuously) under single-tensor fault injection.

## CI

`tests/e2e/lora_native/` (label `lora-native`, trigger `run-ci-lora-native`): qwen3, qwen3.5, glm4.7, glm5.2-5layer, kimi-2.5-2layer, and inkling-small-4layer — native runs checking logprob parity and weight-update equality (vision towers and engine-transformed params skip-listed).

## Validation

**Numerics vs dense reference** (`tests/manual/lora/verify_lora_native.py`): fresh-adapter no-op exact 0.0; column/row/MLA deltas at 1e-7 relative error; export identical on every TP rank; export→load round-trip; replicated-param grad sums exercised on *distinct* per-rank partials — across TP1, TP2, TP2+SP, ± MLA. The plugin extraction was additionally checked side-by-side against the pre-plugin implementation: bit-identical adapter params, forward outputs, backward grads, and HF exports.

**End-to-end**: 20-rollout native GRPO runs (wandb `ch271828n-team/miles_lora_native`), acceptance ≈ 0.01 train/rollout `logprob_abs_diff`:

| model | logprob_abs_diff |
|---|---|
| Qwen3-8B | 0.0099 |
| Qwen3.5-35B-A3B | 0.0105 |
| Qwen3.5-9B (dense hybrid GDN) | 0.0045–0.0090 (mean 0.0063), grad_norm ≤ 0.032 throughout — the recorded raw-backward divergence does not reproduce |
| GLM-5.2 5-layer | 0.0096 |
| Kimi-K2.5 2-layer | 0.0124 (after the quantization_config fix) |
| Inkling-Small (full 42-layer, 4 nodes / 32×H200, TP4·EP4·PP8, 16-GPU engines) | 0.0091–0.0123 (mean 0.0107) over 10 rollouts; reward stable ~0.5 on dapo-math |

The Inkling run exercises the whole custom stack at multi-node scale: 4-way fused [q;k;v;r] export, grouped-GEMM expert adapters under EP4, shared-experts hooks, the muP lm_head, and SGLang serving via the `["all"]` sentinel. (SGLang note: LoRA-on-MoE serving currently supports only the triton MoE runner — `flashinfer_trtllm_routed` fails engine init with NotImplementedError.)

`tests/fast/miles_plugins/lora`: 64 passed (layout locks, target normalization, guard behavior, Inkling spec). Adapter attribute names are pinned, so checkpoint keys are unchanged.


Zhichenzzz:
> [!WARNING]
> Parts of this comment are **corrected** in a later comment: https://github.com/radixark/miles/pull/1792#issuecomment-5137653815

Ran this on an 8×H200 devbox. Summary: the adapter math, the sharding, and the sync-to-SGLang loop all check out; the RL run proves plumbing and gradient flow but not learning quality (see the caveat at the end).

## Numerical verification vs dense reference

`tests/manual/lora/verify_lora_native.py` (added in this PR) builds a real megatron-core `GPTModel`, attaches adapters, and compares the adapter branch against independently computed dense math. The base GEMM is subtracted out, so only this PR's math is under test.

| check | TP1 | TP2 | TP2+SP |
|---|---|---|---|
| fresh adapter (B zero-init) is a no-op | **0.0** exact | **0.0** exact | **0.0** exact |
| column-parallel (fc1) delta vs dense ref | 3.5e-7 rel | 3.3e-7 rel | 3.3e-7 rel |
| row-parallel (fc2) delta vs dense ref | 1.2e-7 rel | 2.2e-7 rel | 1.8e-7 rel |
| exported tensors identical on every TP rank | ✓ | ✓ | ✓ |
| export → load round-trip (params / outputs) | ✓ | ✓ | ✓ |
| `dL/dA == 0` while `dL/dB != 0` (fresh adapter) | **0.0** exact | ✓ | ✓ |
| grads nonzero once B randomized | ✓ | ✓ | ✓ |
| replicated-param grads agree across TP | ✓ | ✓ | ✓ |
| TP sum combined *distinct* per-rank partials | n/a | ✓ | ✓ |

That last row matters: without it, "grads agree across ranks" could pass trivially on identical partials, so it would not actually test `reduce_marked_lora_grads`.

The tagged-param count goes 10 → 14 when sequence parallelism is on, which is exactly the row-parallel `B` params becoming TP-summed as the docstring describes.

Unit tests: **24 passed** (`tests/fast/backends/megatron_utils/test_lora_native.py`).

## End-to-end RL

`scripts/run_qwen3_lora_native.py` with Qwen3-0.6B, TP2 + sequence parallel, gsm8k, colocated SGLang, `--check-lora-weight-equal`:

```
[lora-native] rank=16 alpha=32 scale=2.000 targets=[down_proj, gate_proj, ...]   # both TP ranks
[lora-native] exported 392 adapter tensors in 7.1 ms                            # 28 layers x 14
POST /load_lora_adapter_from_tensors HTTP/1.1" 200 OK                           # engine accepted it
Job 'raysubmit_yL8227QpP3XUN5nK' succeeded
```

392 is the expected count: per layer 6 (q/k/v × A,B) + 2 (o) + 4 (gate/up × A,B) + 2 (down) = 14, over 28 layers.

Across 5 optimizer steps: `train_rollout_logprob_abs_diff` 0.012–0.017, `train_rollout_kl` ~5e-4, `grad_norm` 0.27–0.55.

The logprob agreement is the informative number — megatron's LoRA forward under TP2+SP matching SGLang's rollout using the adapter that was shipped over the wire. A wrong export or shard mapping would put that at 0.5+, not 1e-2.

## Caveat, stated plainly

Reward mean stayed 0.0 (Qwen3-0.6B does not solve gsm8k at these lengths), so **this run does not demonstrate reward-driven learning** — only that the loop runs, the adapter reaches the engine intact, and gradients reach the optimizer. Correctness of the adapter math and the TP grad summation rests on the numerical harness above, which does not depend on reward variance.

One loose end I did not chase because it is framework accounting rather than this PR: at one step `pg_loss` was ~0 while `grad_norm` was 0.27. Worth a look by someone who knows that path, but it reproduces independently of native LoRA.

Not covered by any of the above: PP > 1 (the `_pp_assemble_full_adapter` path), and loading a real pretrained PEFT adapter from disk via `--lora-adapter-path`.

Zhichenzzz:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Merged `main` into this branch (it was `CONFLICTING`, so no `pull_request` workflow had ever been scheduled for it), then triaged the one GPU failure the first real run produced. Please push back if any conclusion is off.

## Status

`main` is merged in and the branch is `MERGEABLE`, so CI runs for the first time: [checks](https://github.com/radixark/miles/pull/1792/checks). All CPU stages pass. One GPU job failed and I am classifying it as **infra (RDMA), not this PR**.

## The failure

[`stage-c-8-gpu-h100 (1) / run`](https://github.com/radixark/miles/actions/runs/30433884179/job/90518695053) (runner `novita-host2`), in `tests/e2e/megatron/test_glm47_flash/test_r3_mtp_deepep.py`.

The RDMA endpoint could not be reached, the P2P weight transfer that depends on it failed, and training then asserted on all-zero metrics:

```
worker_pool.cpp:448] Worker: Cannot make connection for endpoint: 172.18.0.2:16104@mlx5_bond_0, no alternate peer RNIC is ava...
p2p_transfer_utils.py:180 - [P2P] Transfer future failed: [P2P-Shared] Transfer failed for session 172.18.0.2:1610...
AssertionError: log_dict={'train/loss': 0.0, 'train/pg_loss': 0.0, 'train/entropy_loss': 0.0, 'train/pg_clipfrac': 0.0, ...}
```

The all-zero `log_dict` is the symptom, not the cause: the P2P transfer failure precedes it by ~1 minute.

## Why this is not this PR

- This PR touches no P2P, DeepEP, or GLM-4.7-flash code — `gh pr diff 1792 --name-only` matches nothing for `p2p|deepep|glm47|transfer`.
- The same test on the same lane passes on a different branch at roughly the same time: [`stage-c-8-gpu-h100 (1) / run` on #1683](https://github.com/radixark/miles/actions/runs/30425817767/job/90493028127) is SUCCESS.

## Next

Rerunning the failed job once the rest of this run finishes, so live sibling jobs are not disturbed. If the rerun reproduces the same RDMA fingerprint I will re-triage rather than keep retrying.

Zhichenzzz:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Second GPU failure on this branch; it reproduces on an unrelated PR too, which is the evidence for calling it infra. Please push back if any conclusion is off.

A second job failed, and the same test fails the same way on an unrelated branch — which is the strongest evidence available that it is not either PR:

| PR | job | runner | signature |
| --- | --- | --- | --- |
| #1792 (this one) | [`stage-c-4-gpu-h200 (1)`](https://github.com/radixark/miles/actions/runs/30433884179/job/90518695125) | `scitix-72-1` | `ActorUnavailableError ... RpcError` ×3 |
| [#1683](https://github.com/radixark/miles/pull/1683) | [`stage-c-4-gpu-h200 (1)`](https://github.com/radixark/miles/actions/runs/30425817767/job/90493028137) | `scitix-72-0` | `ActorUnavailableError ... RpcError` + `BrokenPipeError` |

Same test both times, `tests/e2e/megatron/test_qwen3_30B_A3B/test_baseline.py`:

```
ray.exceptions.ActorUnavailableError: The actor b2f9bdce2e3a48c2c2fec56e02000000 is unavailable:
The actor is temporarily unavailable: RpcError: RPC ...
```

The two branches have almost nothing in common — #1792 is the general native-LoRA extraction, #1683 is Inkling support — so a shared code cause is implausible; the shared factor is the CI environment. In both logs training was healthy right up to the actor going away (`outcome=NORMAL`, non-zero `train/loss`), and there is no OOM (`OutOfMemory`, `oom-killer`, `CUDA out of memory` all match 0). This PR touches neither `train.py` nor anything under `miles/ray/`.

Also still red from before, and unrelated to this: [`stage-c-8-gpu-h100 (1)`](https://github.com/radixark/miles/actions/runs/30433884179/job/90518695053) with the RDMA/P2P fingerprint in [my earlier comment](https://github.com/radixark/miles/pull/1792#issuecomment-5116048602).

Rerunning both once the last shard of this run finishes. Everything else is green: [checks](https://github.com/radixark/miles/pull/1792/checks).

Zhichenzzz:
> [!WARNING]
> Parts of this comment are **corrected** in a later comment: https://github.com/radixark/miles/pull/1792#issuecomment-5137653815

> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Recording that CI is fully green on this commit, with links, so the state is verifiable later. Please push back if any conclusion is off.

## Green on `eb462dc3ed686c2a3e5bfd16b260efb8b4028d6e`

All 24 non-skipped checks pass: [checks](https://github.com/radixark/miles/pull/1792/checks) · [per-commit view](https://github.com/radixark/miles/pull/1792/commits/eb462dc3ed686c2a3e5bfd16b260efb8b4028d6e). Every GPU stage is included:

| stage | result |
| --- | --- |
| `stage-c-2-gpu-h200` (0, 1) | SUCCESS |
| `stage-c-4-gpu-h200` (0, 1, 2) | SUCCESS |
| `stage-c-8-gpu-h100` (0, 1) | SUCCESS |
| `stage-c-8-gpu-h200` (0, 1) | SUCCESS |

Plus lint, docker, CodeQL and all CPU shards.

The two failures I triaged earlier both passed on rerun, which is what their infra classification predicted:

- [`stage-c-8-gpu-h100 (1)`](https://github.com/radixark/miles/actions/runs/30433884179/job/90518695053) — RDMA/P2P, [comment](https://github.com/radixark/miles/pull/1792#issuecomment-5116048602) → now SUCCESS
- [`stage-c-4-gpu-h200 (1)`](https://github.com/radixark/miles/actions/runs/30433884179/job/90518695125) — `ActorUnavailableError`, [comment](https://github.com/radixark/miles/pull/1792#issuecomment-5118937798) → now SUCCESS

## What changed to get here

CI had never run on this branch: it was `CONFLICTING`, and GitHub does not schedule `pull_request` workflows for a branch that cannot be merged. Merging `main` in was the unblock. The one conflict was in `generate_endpoint_utils.py`'s import block — `main` added `extract_multimodal_train_inputs` while this branch swapped `is_lora_enabled` for `lora_rollout_enabled`; both are kept, the same resolution [#1683](https://github.com/radixark/miles/pull/1683) took for the identical conflict.

Also aligned the shared LoRA code with #1683 so the two do not drift: `miles/utils/lora.py` is now byte-identical between the branches, and the explanatory comments this branch added are stripped the same way (directive comments and docstrings kept). `tests/fast/backends/megatron_utils` passes 267 tests locally after that.

Worth noting for review: two bugs I found and fixed in #1683 do **not** apply here — this branch keeps `main`'s `FlattenedTensorBucket` adapter transport (so no cuMem/CUDA-IPC problem under `torch_memory_saver`) and has no missing-checkpoint gate in `model.py` (so no silent skip of `load_checkpoint` for bridge-mode runs). Extracting the general path kept it clear of both.

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+    it back to the full sequence the adapter's replicated ``A`` expects.
+    """
+    gamma = getattr(module, "layer_norm_weight", None)
+    if gamma is not None:
+        x = _rmsnorm(x, gamma, spec.eps, spec.zero_centered_gamma)
+    if spec.sequence_parallel:
+        from megatron.core.tensor_parallel.mappings import gather_from_sequence_parallel_region
+
+        x = gather_from_sequence_parallel_region(x)
+    return _dropout(x, spec, module.training)
```
yushengsu-thu: SP off + TP>1: the branch's input grad is never summed across TP. Wrap x in copy_to_tensor_model_parallel_region (true-on-policy forces SP off, so this path is real).

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+    """Adapters on the fused qkv (column-parallel) and the output proj (row-parallel)."""
+    from megatron.core import parallel_state as ps
+
+    n = 0
+    tp_rank = ps.get_tensor_model_parallel_rank()
+    num_q = attn.num_attention_heads_per_partition
+    num_kv = attn.num_query_groups_per_partition
+    head_dim = attn.hidden_size_per_attention_head
+
+    if spec.wants("q_proj", "k_proj", "v_proj"):
```
yushengsu-thu: q,v-only targeting silently trains and exports k too, while adapter_config says q,v. Reject fused subsets or record the actual targets.

miles/backends/megatron_utils/lora_utils.py
```diff
@@ -387,42 +457,44 @@ def save_lora_checkpoint(
...
+                show_progress=False,
+            ):
+                lora_state_dict[hf_name] = weight
+    else:
+        from .lora_native import resolve_lora_provider
+
+        for hf_name, weight in resolve_lora_provider(args).export_lora_hf_named(model):
+            lora_state_dict[hf_name] = weight.cpu()
+
     if is_dp_cp_rank_0 and tp_rank == 0 and pp_rank == 0:
```
yushengsu-thu, line 594: Under PP>1 this writes stage-0 layers only — needs the same PP assembly the sync path got.

miles/backends/megatron_utils/update_weight/hf_weight_iterator_bridge.py
```diff
@@ -70,7 +70,16 @@ def get_hf_weight_chunks(self, megatron_local_weights, weight_type: str = "base"
...
+        """Normalise names, postprocess, and quantize one exported weight at a time.
+
+        megatron-bridge PEFT reports a LoRA-wrapped module's base weight as
+        ``<module>.to_wrap.weight``. The quantizer regexes here and the atomic-group
+        suffixes used by ``_stream_atomic_units`` are both written against unwrapped
+        names, so the segment is stripped first; leaving it in silently disables both
+        (the distributed iterator does the same in ``_get_weight_transfer_update_units``).
+        """
+        for hf_param_name, weight, wrapped_megatron_name in named_weights:
+            megatron_param_name = wrapped_megatron_name.replace(".to_wrap.", ".")
```
yushengsu-thu, line 82: Quantized params emit multiple tensors per name; after the unwrap they collide in _stream_atomic_units and trip the pending assert. Groups need to accept multi-tensor params.

miles/backends/megatron_utils/update_weight/hf_weight_iterator_direct.py
```diff
@@ -32,6 +32,12 @@ def __init__(self, *args, **kwargs):
     def get_hf_weight_chunks(self, megatron_local_weights, weight_type="base"):
         rank = dist.get_rank()
 
+        if weight_type == "lora":
+            from ..lora_native import resolve_lora_provider
+
+            yield resolve_lora_provider(self.args).export_lora_hf_named(self.model)
```
yushengsu-thu: Colocate + non-adam (or --disable-weights-backuper): adapter pages are memory-saver-paused during export — crash or garbage sync. Worth a startup guard.

yushengsu-thu (commented):
here

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+    return torch.tensor(index, dtype=torch.long, device=device)
+
+
+def _wrap_forward(module: nn.Module, delta_fn, scale: float) -> None:
+    """Add ``scale * delta_fn(x)`` to ``module``'s output, keeping its (out, bias) contract."""
+    original = module.forward
+
+    def forward(x, *args, **kwargs):
+        out, bias = original(x, *args, **kwargs)
+        return torch.add(out, delta_fn(x), alpha=scale), bias
```
yushengsu-thu: The adapter module is never __ call __  — its params are only read inside this closure — so under --overlap-param-gather no forward pre-hook ever dispatches its bucket's all-gather; with DP>1 every step-1+ forward runs on stale shards. Guard the combo for now.

miles/backends/megatron_utils/model.py
```diff
@@ -969,6 +977,16 @@ def initialize_model_and_optimizer(
...
         )
+    if (
+        is_lora_enabled(args)
+        and role == "actor"
+        and args.megatron_to_hf_mode != "bridge"
+        and getattr(args, "lora_adapter_path", None)
+    ):
+        from .lora_native import resolve_lora_provider
+
+        resolve_lora_provider(args).load_lora_adapter_hf(model, args.lora_adapter_path)
```
yushengsu-thu, line 955: Needs optimizer.reload_model_params() after this load — the mains were snapshotted from init values inside load_checkpoint, so step 1 wipes the loaded adapter (same pattern as checkpoint.py:190 and multi_lora_utils).

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+            else:
+                mixer_only_layers.append(layer.layer_number - 1)
+
+        mlp = layer.mlp
+        if hasattr(mlp, "linear_fc1"):
+            assert getattr(mlp.config, "gated_linear_unit", True), "native LoRA assumes a gated (SwiGLU) MLP"
+            wrapped += _attach_mlp(mlp, hf_layer + "mlp.", spec)
+        shared = getattr(mlp, "shared_experts", None)
+        if shared is not None and hasattr(shared, "linear_fc1"):
+            wrapped += _attach_mlp(shared, hf_layer + spec.shared_expert, spec)
```
yushengsu-thu: With --moe-shared-expert-overlap the dispatcher owns the shared-expert comms but spec.sequence_parallel still reads the global flag: SP on → _branch_input gathers an already-gathered input, shape crash at torch.add; SP off → the down_proj delta gets all-reduced twice (post_forward_comm re-reduces it, ×TP). Gate the combo, or derive gather/reduce from the module's effective parallel mode.

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+            rank=rank,
+            scale=float(args.lora_alpha) / rank,
+            dropout=float(getattr(args, "lora_dropout", 0.0) or 0.0),
+            a_init=getattr(args, "lora_A_init_method", "xavier") or "xavier",
+            eps=config.layernorm_epsilon,
+            hidden=config.hidden_size,
+            sequence_parallel=bool(config.sequence_parallel),
+            zero_centered_gamma=bool(getattr(config, "layernorm_zero_centered_gamma", False)),
+            tp_size=ps.get_tensor_model_parallel_world_size(),
+            targets=frozenset(args.target_modules or ()),
```
yushengsu-thu: Megatron-style names (linear_qkv, linear_fc1, …) aren't normalized here — they match nothing, silently, while sglang still gets the full list. Normalize or hard-error on unmatched targets.

miles/backends/megatron_utils/lora_native.py
```diff
@@ -0,0 +1,848 @@
...
+
+        _wrap_forward(module, delta, spec.scale)
+        return 1
+
+    if hasattr(attn, "linear_q_down_proj"):
+        if spec.wants("q_a_proj"):
+            n += add_replicated(attn.linear_q_down_proj, _MLA_Q_A, "q_a_proj", config.q_lora_rank)
+        if spec.wants("q_b_proj"):
+            n += add_column_parallel(attn.linear_q_up_proj, _MLA_Q_B, config.q_lora_rank, heads_local * q_head_dim)
+    elif spec.wants("q_proj") and hasattr(attn, "linear_q_proj"):
```
yushengsu-thu: q_lora_rank=None MLA (DeepSeek-V2-Lite / Moonlight) exports unfused q_proj + kv_a_proj_with_mqa; sglang-miles' loader assumes the fused qkv_a layout and can't ingest it — first sync fails. Reject these models for now.

yushengsu-thu (commented):
here

examples/lora/run_lora_native.py
yueming-yuan, line 1: suggest to move this file to /example/lora

miles/utils/arguments.py
```diff
@@ -1456,6 +1456,29 @@ def add_lora_arguments(parser):
...
+                help=(
+                    "Dotted module path implementing the native-LoRA provider protocol "
+                    "(wrap_model_provider_with_lora / load_lora_adapter_hf / export_lora_hf_named). "
+                    "Defaults to miles.backends.megatron_utils.lora_native, which covers standard "
+                    "Megatron-core GPT models; point this at a model plugin whose module structure "
+                    "diverges from plain mcore. Only used with --megatron-to-hf-mode raw."
+                ),
+            )
+            parser.add_argument(
+                "--lora-train-only",
```
yueming-yuan: Suggest adding a "debug" prefix, to match the convention of this kind of args e.g. "debug-train-only", "--debug-skip-weight-update", "--debug-disable-optimizer",

Zhichenzzz:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Corrects two errors in my own earlier comments and records where the review findings stand. Please push back if any conclusion is off.

## Corrections to my earlier comments

**1. The "all 24 checks green" comment overstated the GPU coverage.** I reported `stage-c-8-gpu-h200 (0, 1)` as SUCCESS. Only shard 0 ran GPU tests; shard 1 was `/ run-cpu → skipped`, and no `stage-c-8-gpu-h200 (1) / run` job existed in that run at all. I read the `run-cpu` variant as the `run` job. So that lane was half as covered as I claimed.

**2. The verification comment's round-trip row was not actually verified.** It lists `export → load round-trip (params / outputs)` as passing. That check covered the load and the first weight sync, both of which happen *before* any optimizer step. It did not cover whether the adapter survives step 1 — and it does not: `load_lora_adapter_hf` runs after `load_checkpoint` has already snapshotted the fp32 main params from the adapter's init values, so step 1 wrote them back and discarded the loaded adapter. Fixed in 203c61187 with `optimizer.reload_model_params()`, matching what `checkpoint.py` and `multi_lora_utils` already do. The unit-test count in that comment (24) is also stale; it is 275 now.

## Review findings

All ten of @yushengsu-thu's comments are addressed across 203c61187, 6ff9ad40f and b477320ad. The three that failed silently rather than loudly:

| Finding | Status |
|---|---|
| `--lora-adapter-path` discarded at step 1 | fixed |
| Megatron-style `--target-modules` attached nothing | fixed: normalised through `convert_target_modules_to_hf`, unimplemented targets rejected at startup |
| Quantized params tripped the atomic-group assert | fixed; this one was a regression from my own `.to_wrap.` strip |
| SP off with TP > 1 sent upstream layers 1/TP of the adapter gradient | fixed |
| PP > 1 saved stage 0's layers only | fixed |
| `--overlap-param-gather`, `--moe-shared-expert-overlap`, colocate without the backuper, MLA without `q_lora_rank` | fail-fast guards, as suggested |

On the quantized-param finding: the old grouping dropped 4 of 12 tensors for a grouped MLA pair (later ones overwrote earlier in the same slot) and then failed `assert not pending`. Before the `.to_wrap.` strip the suffixes never matched, so the collision was unreachable — the strip is what exposed it.

Verification, on 2xB200: `verify_lora_native.py` passes across `--tp 1`, `--tp 1 --mla`, `--tp 2`, `--tp 2 --sp`, `--tp 2 --sp --gate`, `--tp 2 --sp --mla`. `--tp 2` without SP had never been run, which is why the gradient gap survived the original matrix. A new first-activation-gradient check fails with spread 4.232e-03 when the fix is reverted and reports 0.000e+00 with it; the pre-existing checks could not catch it because they cover the adapter's own parameter gradients, not what flows upstream. A new `verify_pp_adapter_assembly.py` covers the PP path on a real 2-rank pipeline group.

@yueming-yuan's `--debug-` prefix suggestion is applied in 6ff9ad40f. Worth noting the name still reads as a tautology: LoRA already trains only the adapters, so "train only" does not describe what the flag changes — its sole effect is that `lora_rollout_enabled` turns false, disabling SGLang LoRA serving and the adapter sync while the rollout keeps serving the frozen base. Happy to rename it to something like `--debug-lora-no-rollout-serving` if you prefer. The suggestion to move `scripts/run_lora_native.py` under `examples/lora/` is not done yet.

## One red check I cannot attribute

`stage-c-8-gpu-h200 (1) / run` fails on `tests/e2e/megatron/model_scripts/test_deepseek_v32_5layer_fp8.py`, reproducibly (failed again on rerun), with FP8 weight-equality drift across several tensors (`max_abs_err` 0.0068 to 0.101, including `self_attn.indexer.wk.weight` under `--freeze-indexer`).

What I can establish: none of my changed code is on that test's path. The iterator selector resolves `raw` to `HfWeightIteratorDirect` and `bridge` to `HfWeightIteratorBridge`; the test uses the default `raw`, so the `_stream_atomic_units` change is never loaded. The test and `scripts/run_deepseek_v32.py` contain no LoRA flags, so the native-LoRA changes never run. The one shared-file edit (moving `pp_assemble_full_adapter` into `lora_utils`) is import-level, and its call site sits inside the LoRA adapter-sync block.

What I cannot establish: that the test passes without my commits. It has **never run on this branch before** — the shard composition changed when my added tests shifted the time estimates, so it landed in shard 1 and executed for the first time. `main` does not run this suite on push, and the test needs a single node with 8 GPUs, which I do not have. So this is "newly exercised and failing", not "was green and I broke it", and it needs someone who knows the DSA indexer path. A frozen weight drifting at all looks worth a look independently of this PR.

miles/backends/megatron_utils/update_weight/hf_weight_iterator_bridge.py
```diff
@@ -116,14 +145,15 @@ def _stream_atomic_units(items, atomic_update_groups):
...
             continue
         group, idx, suffix = match
         prefix = megatron_name[: -len(suffix)]
         slots = pending.setdefault((prefix, group.key), [None] * len(group.suffixes))
-        slots[idx] = (hf_name, weight)
+        assert slots[idx] is None, f"duplicate {suffix} for {prefix} in atomic group {group.key}"
+        slots[idx] = tensors
         if None not in slots:
-            yield list(slots)
+            yield [pair for slot in slots for pair in slot]
```
yushengsu-thu, line 156: if you modified this, could it still pass lora e2e case in megatron-bridge path?

yushengsu-thu (approved):

HJSang (approved):
I like this design a lot: clean and highly structured!
wondering how do you approach this design?

PR #1795, opened 2026-07-25
Bump sglang to v0.5.16

yueming-yuan:
Bumps the sglang base image and the sglang-miles branch to v0.5.16.

- `SGLANG_IMAGE_TAG`: `v0.5.15` → `v0.5.16` (Dockerfile + `build.py` cu12-x86)
- `SGLANG_BRANCH`: `sglang-miles` → `sglang-miles-v0.5.16`

Follows the v0.5.15 cycle (#1655 then #1701): this PR pins the versioned branch, and a
follow-up points `SGLANG_BRANCH` back to `sglang-miles` once it is promoted.

## sglang side

`sglang-miles-v0.5.16` rebases the 31 sglang-miles commits from v0.5.15 onto v0.5.16,
landing as 22 commits: https://github.com/sgl-project/sglang/tree/sglang-miles-v0.5.16

Three PRs were dropped because v0.5.16 already carries them:

| PR | superseded by |
|---|---|
| #32108 sampling mask | upstream #27408 — all three sampler helpers, the env var and the test are present |
| #31831 expected_checksums | arrived via #31681 — identical sha256 block in `tp_worker.py` |
| #28524 dumper/comparator | `get_grad` callable, `threshold_dsl.py` and its test are present (its two `/pull_weights` fixups were kept) |

Notable reimplementation, because v0.5.16 refactored the same areas independently:

- Weight updating moved into `ModelRunner.weight_updater`, so the spec-draft work
  (#27749/#28575/#27750) is reimplemented there: the receive/load split lands on
  `WeightUpdater`, and `SchedulerWeightUpdaterManager` owns the target-receives-once /
  load-into-each-runner fan-out. The worker-level `update_weights_from_*` entry points
  are dropped so there is only one path.
- The remote-instance transfer engine moved into `RemoteInstanceWeightTransporter`, so
  #21278's per-rank `RankParallelismConfig` publishing moved with it.
- `#29874`'s `free_lora` is dropped: v0.5.16's `LoRAMemoryPool.remove_lora()` already
  releases the slot on unload and additionally zeroes the buffers for cuda-graph replay.
- `#30421` was extended to v0.5.16's new `deepseek_v4_dspark.py`, which had the same
  stale `_attn_sink_local` bug.
- All new `get_global_server_args()` call sites moved to
  `runtime_context.get_server_args()`; v0.5.16 ratchet-pins the legacy accessor to a
  single call site and the test fails in both directions.

ci-sglang-pr: sglang-miles-v0.5.16

[...]

PR #1800, opened 2026-07-26
docker: install transformer_engine_torch runtime deps

yueming-yuan:
## Problem

#1796 switched TransformerEngine to the rolling wheels, installed with
`--no-deps`. Pinning the three TE dists to exactly those wheels is the point of
that flag, but it also drops `transformer_engine_torch`'s own runtime
requirements:

```
torch>=2.1, einops, onnxscript, onnx, packaging, pydantic,
nvdlfw-inspect, transformer_engine_cu13==2.17.0
```

`transformer_engine.pytorch` imports `onnxscript` unconditionally on its core
import path:

```
transformer_engine/pytorch/__init__.py:21   -> module/__init__.py
  module/layernorm_linear.py:21 -> module/base.py:23 -> module/_common.py:15
    export.py:61 -> onnx_extensions.py:26 -> import onnxscript
```

so every image built since #1796 dies as soon as anything touches TE — which is
every GPU test:

```
File ".../transformer_engine/pytorch/onnx_extensions.py", line 26, in <module>
    import onnxscript
ModuleNotFoundError: No module named 'onnxscript'
```

Seen on #1795's `stage-c-*-gpu-h200` jobs (run 30190570673), where it takes out
the run before any test body executes. It is not specific to that PR — it
reproduces on any image built from current main.

## Why the build stayed green

`verify_transformer_engine.py` compared versions and metadata and then called

```python
importlib.util.find_spec("transformer_engine")
```

`find_spec` resolves the package directory without importing it, so it succeeds
against an installation that cannot actually be imported. The verifier passed,
the image was pushed, and the failure surfaced hours later in GPU tests.

## Change

- Install the non-TE runtime deps after the pinned `--no-deps` triplet.
- Have the verifier assert that every non-TE requirement of
  `transformer_engine_torch` resolves, so a future `--no-deps` gap fails at the
  layer that introduced it rather than shipping a green image.

Deliberately not dropping `--no-deps`: letting pip resolve the triplet is what
that flag exists to prevent.

## Verification

Ran the new verifier inside a miles image that has the deps, then hid
`onnxscript`'s dist-info to reproduce exactly the state #1796 produces:

```
--- deps present (expect PASS) ---
PASS
--- simulate the #1796 state: hide onnxscript ---
AssertionError: transformer_engine_torch runtime deps not installed: ['onnxscript']
--- restored ---
PASS again
```

### DIFF
--- a/docker/verify_transformer_engine.py
+++ b/docker/verify_transformer_engine.py
@@ -2,10 +2,16 @@
 
 import importlib.metadata as metadata
 import importlib.util
+import re
 import sys
 
 
 VERSION = "2.17.0"
+
+
+def _dist_name(requirement: str) -> str:
+    """'onnxscript', 'torch>=2.1', 'foo; extra == \"bar\"' -> normalised dist name."""
+    return re.split(r"[<>=!~;\[ ]", requirement.strip(), maxsplit=1)[0].lower().replace("_", "-")
 
 
 def verify(core_dist: str) -> None:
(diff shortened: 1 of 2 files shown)


docker/verify_transformer_engine.py
guapisolo, line 1: Is it possible to add this file to CI?

guapisolo (approved):
LGTM. wait for CI green and e2e pass then we can merge.

Zhichenzzz:
Looks great! should we merge it now?

PR #1809, opened 2026-07-27
docs: fix factual errors in fully-async example page

Shi-Dong:
## Summary

Fixes factual errors in `docs/examples/fully-async.md` found during a fact-check of every page under `docs/examples/` against the actual code in `examples/fully_async/`.

- **Quick-start log output**: the page showed a fabricated `[trainer] iter 1/3000 ...` log block. Replaced with the real stdout prints from `fully_async_rollout.py` (`Creating new global async worker...`, `Starting async rollout generation for N groups`, `Rollout completed in ...s! Global worker queue size: N`).
- **Worker constructor**: the doc passed a nonexistent `concurrency=args.sglang_server_concurrency` kwarg; `AsyncRolloutWorker.__init__` only takes `(args, data_buffer)`.
- **Worker loop**: the doc showed an invented `_producer` coroutine. Replaced with a condensed version of the real `continuous_worker_loop` (reap finished tasks, top up in-flight groups via `generate_and_rm_group`, done-callback onto the output queue), including the real in-flight cap: `--rollout-batch-size` groups, or `--async-max-concurrent-samples ÷ n_samples_per_prompt` when set.
- **Entry function**: the doc showed a made-up async signature. The real `generate_rollout_fully_async(args, rollout_id, data_buffer, evaluation=False)` is synchronous, rejects evaluation mode, drains `--rollout-batch-size` groups, recycles aborted groups, and sorts by prompt index.
- **What to watch**: the doc listed fabricated `async/*` wandb metrics that the example never logs. Replaced with the real stdout signals plus the standard `perf/rollout_time` vs `perf/actor_train_time` comparison.
- **MoE variation**: the doc claimed `tp=4 ep=8`, a nonexistent `--sglang-enable-ep-moe` flag, and a batch-size recommendation not in the script. `run_qwen3_30b_a3b_fully_async.py` actually uses TP 8 / EP 8, one 8-GPU SGLang engine, `--use-tis`, and demonstrates the `--pause-generation-mode` × `--update-weight-transfer-mode` combinations — every pairing except `in_place` + `p2p`, which the script rejects.
- **Partial rollout**: the doc claimed unfinished trajectories are "saved to disk and resumed". In reality this example recycles aborted groups to the in-memory data buffer but calls `reset_for_retry()` first, so they regenerate from the prompt; true resume-from-partial is the stock `--partial-rollout` behavior of the built-in rollout function, which this custom drain loop bypasses. The section now states exactly that.

## Test plan

- [ ] Verified all replacement snippets against `examples/fully_async/fully_async_rollout.py` and `run_qwen3_30b_a3b_fully_async.py` on current main
- [ ] Docs render check

### DIFF
--- a/docs/examples/fully-async.md
+++ b/docs/examples/fully-async.md
@@ -49,7 +49,9 @@ You should see:
 ```text
 Creating new global async worker...
 Continuous async rollout worker started
-[trainer] iter 1/3000 | drained 32 samples (queued: 18)
+Starting async rollout generation for 32 groups
+...
+Rollout completed in 41.23s! Global worker queue size: 3
 ```
 
 ## What changes vs. the default recipe
@@ -77,8 +79,7 @@ def get_global_worker(args, data_buffer):
     with _worker_lock:
         if _global_worker is None or not _global_worker.worker_thread.is_alive():
             print("Creating new global async worker...")
-            _global_worker = AsyncRolloutWorker(args, data_buffer,
-                                                concurrency=args.sglang_server_concurrency)
+            _global_worker = AsyncRolloutWorker(args, data_buffer)
             _global_worker.start()
         return _global_worker
 ```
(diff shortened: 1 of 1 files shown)


Zhichenzzz:
LGTM! Lets wait till https://github.com/radixark/miles/pull/1717 and https://github.com/radixark/miles/pull/1740 are merged

PR #1814, opened 2026-07-27
Move in-model CP layout helpers into miles_plugins/models/cp_utils.py

yueming-yuan:
## What

Pure refactor, no behavior change.

`miles/backends/training_utils/cp_utils.py` owns the CP helpers the training backend applies **around** the model — `slice_with_cp`, `natural_to_zigzag_slice`, `slice_loss_masks_for_local_cp`, `all_gather_with_cp`, `allgather_cp_redistribute`, `get_logits_and_tokens_offset_with_cp`, … All of them operate on tokens / masks / logprobs / logits outside `forward()`, and are called from `loss.py`, `data.py`, `losses.py`.

`build_gdn_cp_context` was the one exception. It runs *inside* a layer's forward, its only callers are `miles_plugins/models/qwen3_next.py` and `qwen3_5.py`, and it forced an optional flash-linear-attention dependency into a core training module:

```python
try:
    from fla.ops.cp import build_cp_context as _fla_build_cp_context
except ImportError:
    _fla_build_cp_context = None
```

The zigzag ↔ contiguous-shard relayout has the same shape but the opposite problem: it lived as private helpers inside `miles_plugins/models/hf_attention.py`, so only models built on `HuggingfaceAttention` could reach it. It is not HF-attention-specific — it bridges Megatron's zigzag CP layout to the contiguous rank-local shard every fla CP operator expects, and any model with an fla linear-attention layer needs it.

## Change

- `miles_plugins/models/cp_utils.py` becomes the module for CP helpers that run inside a forward: `build_gdn_cp_context` (moved from core) plus the relayout helpers `zigzag_to_packed_shard` / `packed_shard_to_zigzag` and their internals (moved from `hf_attention.py`, underscore prefix dropped).
- `detect_and_setup_hybrid_cp` moves the other way, out of `miles_plugins/models/cp_utils.py` into `hf_attention.py`. It is defined in terms of `isinstance(module, HuggingfaceAttention)`, so it belongs next to that class — and moving it is what breaks the import cycle `hf_attention → cp_utils → hf_attention`.
- Core `cp_utils.py` loses `build_gdn_cp_context`, the `fla.ops.cp` try/except, and its now-unused `torch.nn` import.
- Import sites updated: `qwen3_next.py`, `qwen3_5.py`, `miles/backends/megatron_utils/actor.py`, `tests/e2e/precision/test_hf_attention_cp_relayout.py`.

## Verification

- Every relocated definition is byte-identical to its original apart from the underscore renames — checked by extracting both versions with `ast` and comparing source text (10/10 identical).
- Static module-level import-graph check: no cycles involving any touched module. `miles_plugins.models.cp_utils` now has zero intra-repo imports; core `cp_utils` imports only `.parallel`.
- Behavioral coverage is unchanged and comes from the existing `tests/e2e/precision/test_hf_attention_cp_relayout.py` round-trip test, which now imports the public names. It needs GPUs, so it runs in CI rather than locally.

## Motivation

New model support adds a third model that needs the relayout, and it is not a `HuggingfaceAttention` subclass. Without this, it would either duplicate the permutation logic or import a private symbol across plugins. This PR stands on its own, though — it removes an optional-dependency leak from a core training module either way.


Zhichenzzz (approved):
LGTHuman

PR #1815, opened 2026-07-27
refactor(tito): rename tokenize_additional_non_assistant to tokenize_additional_messages

guapisolo:
## Summary
Rename `tokenize_additional_non_assistant` to `tokenize_additional_messages`.

## Motivation
Appended input may now include assistant turns, so the method name and its docstrings must stop claiming "non-assistant". Mechanical rename of the def, its DSv4 override, every call site, and the wording lines that made the same claim; no behavior change.

## Before / After
- **Before / After:** the renamed def in `tito_tokenizer.py`, its DSv4 override, every call site now use `tokenize_additional_messages`.
- Docstring lines that claimed "non-assistant" input now describe appended messages.
- **What moved where:** nothing moved; 5 files change identifier/wording lines only (+30/−30).

## Behavior Preservation
- **How we know:** the diff is symmetric rename/wording lines; the covering suites pass with only call-site renames.

## Verification
- **Existing test suite:** `pytest tests/fast/utils/chat_template_utils/test_tito_tokenizer.py tests/fast/utils/chat_template_utils/test_pretokenized_via_tito.py tests/fast/router/test_linear_trajectory.py` — 139 passed at this commit.

## Review Focus
- Scrutinize `tito_tokenizer.py` for any hunk that changes more than the identifier or docstring wording.

PR #1816, opened 2026-07-27
feat(chat-template): DSv3.2 encoder honors drop_thinking=False when preserving thinking

guapisolo:
> **Depends on:** #1815

## Summary
DSv3.2 encoder honors the `drop_thinking=False` preserve-thinking path.

## Motivation
drop_thinking=False is the history-preserving mode of the vendored DSv3.2 encoder: think blocks in assistant messages before the last user message stay in the rendered prompt, so incremental tokenization reuses generated tokens verbatim. The upstream reasoning-required invariant (an assistant message after the last user message must carry reasoning_content or tool_calls) belongs to the upstream drop path; enforcing it under the preserving mode rejected renderable histories. The gate is now scoped to drop_thinking=True; on the preserving path a reasoning-free assistant message renders with its thinking slot closed as </think>. V4 gains tests only.

## Usage
Run DSv3.2 with `drop_thinking=False`; a history whose trailing assistant message lacks `reasoning_content` renders instead of raising `DS32EncodingError`, while think blocks already present in earlier assistant messages are preserved.

## Design Notes
- **Key choices:** only the gate's scope changes in `render_message`; the upstream `drop_thinking=True` path keeps its invariant untouched.

## Verification
- **Tests added:** `test_drop_thinking_false_injected_appends_are_append_only` in `test_deepseek_v32.py` + `test_deepseek_v4.py` — reasoning-free assistant messages render append-only on the preserving path (129 passed at this commit).

## Review Focus
- Scrutinize the gate condition in `encoding_dsv32.py` `render_message` for `drop_thinking=True` histories that must still raise.

### DIFF
--- a/miles/utils/chat_template_utils/templates/encoding_dsv32.py
+++ b/miles/utils/chat_template_utils/templates/encoding_dsv32.py
@@ -282,7 +282,10 @@ def render_message(
 
         summary_content = content or ""
 
-        if thinking_mode == "thinking" and index > last_user_idx:
+        # Miles modification: only the upstream drop path enforces the
+        # reasoning-required invariant; drop_thinking=False accepts injected
+        # assistant input without reasoning_content (renders as `</think>`).
+        if drop_thinking and thinking_mode == "thinking" and index > last_user_idx:
             if not (reasoning_content or tool_calls):
                 raise DS32EncodingError(
                     f"ThinkingMode: {thinking_mode}, invalid message without reasoning_content/tool_calls `{msg}` after last user message"
--- a/miles/utils/chat_template_utils/templates/encoding_dsv32.py
+++ b/miles/utils/chat_template_utils/templates/encoding_dsv32.py
@@ -282,9 +282,7 @@ def render_message(
 
         summary_content = content or ""
 
-        # Miles modification: only the upstream drop path enforces the
-        # reasoning-required invariant; drop_thinking=False accepts injected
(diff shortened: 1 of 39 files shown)


miles/utils/chat_template_utils/templates/encoding_dsv32.py
```diff
@@ -282,7 +282,10 @@ def render_message(
 
         summary_content = content or ""
 
-        if thinking_mode == "thinking" and index > last_user_idx:
+        # Miles modification: only the upstream drop path enforces the
+        # reasoning-required invariant; drop_thinking=False accepts injected
+        # assistant input without reasoning_content (renders as `</think>`).
```
Shi-Dong: Very nit: this three-line comment is quite abstruse. Probably it'd be simpler to say "we disallow drop thinking."

Shi-Dong (approved):
LGTM with nit.

miles/utils/chat_template_utils/templates/encoding_dsv32.py
guapisolo: agree.

guapisolo: Simplified the comment to state the policy directly: TiTO disallows `drop_thinking=True`. Landed in 389f1bdb45832afb7410d92815e062cd6584df60, with you credited as a co-author.

PR #1817, opened 2026-07-27
refactor(tito): bind append roles to fixed templates

guapisolo:
## Summary
Each TITO model binds to one fixed chat template; multi-template selection is removed.

## Motivation
`--tito-allowed-append-roles` was an overdesign: it treated chat-template behavior as caller policy and let one model pair with multiple `SUPPORTED_TEMPLATES` rows. Each family now binds its model to exactly one `FixedTemplate` owning the template path, fixed kwargs, and allowed_append_roles (restricted families record the template's real capability); the one-model-multiple-templates implementation is removed, resolution keys on `--tito-model` alone, and conflicting caller values fail loud. Shrinking the customization space makes the TITO tokenizer easier to reason about. The CLI flag still parses here and is removed in the follow-up commit; until then `scripts/tools/verify_chat_template.py` and `tests/manual/session/bench_session_server_overhead.py` still call the two-argument resolver.

## Before / After
- **Before / After:** one model could select among multiple `SUPPORTED_TEMPLATES` rows, keyed by model plus caller-supplied roles.
- Now each family binds one `FIXED_TEMPLATE`; resolution keys on `--tito-model` alone.
- **What moved where:** append-role policy moved from CLI/session configuration into the family classes in `tito_tokenizer.py`; the `FixedTemplate` type lives in `template.py`; 16 files, +449/−569.

## Behavior Changes and Verification
- **Intentional behavior change:** `--tito-model=default` keeps the model's native template, while each named family now unconditionally uses its registered template path, required kwargs, and append-role surface; custom template paths and conflicting required kwargs fail loud.
- **How we know:** migrated assertions in `test_fixed_templates.py` plus `test_tito_tokenizer.py` pin the expected template paths, kwargs, and role sets for every family.

## Verification
- **Existing test suite:** `pytest tests/fast/utils/chat_template_utils tests/fast/router` — 1625 passed at this commit.

## Review Focus
- Scrutinize each family's `FIXED_TEMPLATE` registration in `tito_tokenizer.py` against the `SUPPORTED_TEMPLATES` rows it replaces.
- Scrutinize the fail-loud path for caller role values that previously selected a different row.

### DIFF
--- a/miles/utils/test_utils/chat_template_verify.py
+++ b/miles/utils/test_utils/chat_template_verify.py
@@ -137,9 +137,12 @@ def verify_append_only(
 # Trajectories expose two class attributes used for verify-layer filtering:
 #
 #   * ``APPEND_ROLES: frozenset[str]`` — non-assistant roles that appear after
-#     the first assistant message.  Drives ``--tito-allowed-append-roles``.
+#     the first assistant message.  Compared with FixedTemplate capability.
 #   * ``IS_THINKING: bool`` — any assistant carries ``reasoning_content``.
 #     Drives ``--thinking`` and whether ``enable_thinking`` kwarg is passed.
+#
+# This shared verifier intentionally covers only non-assistant append trajectories;
+# dedicated family CPU tests and the session verifier cover injected assistant input.
 #
 # Both are declared on the trajectory class (mock_trajectories.py), alongside
 # ``TOOLS`` / ``PRETOKENIZE_POSITIONS`` / ``MESSAGES``.  This file only lists
@@ -316,9 +319,8 @@ def check_coverage(
 ) -> CoverageReport:
     """Enumerate ``thinking × append-role-subset`` combinations and report gaps.
 
-    Used as a sanity check that every meaningful combination of
-    ``--tito-allowed-append-roles`` and ``--thinking`` is backed by at least
-    one trajectory — otherwise certain CLI settings would be no-ops.
+    Used as a sanity check that every meaningful append-role and thinking
+    combination is backed by at least one trajectory.
     """
     if cases is None:
         cases = ALL_CASES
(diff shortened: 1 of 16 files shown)


miles/utils/test_utils/chat_template_verify.py
```diff
@@ -372,7 +374,9 @@ def run_all_checks(
     through the CLI.
     """
     if allowed_append_roles is None:
-        allowed_append_roles = {"tool"}
+        from miles.utils.chat_template_utils.tito_tokenizer import ALL_APPEND_ROLES
```
Shi-Dong: Please hoist this import to the beginning

miles/utils/test_utils/chat_template_verify.py
```diff
@@ -560,26 +560,29 @@ def run_all_checks_via_tito(
...
-    calling this — this function does not consult ``SUPPORTED_TEMPLATES``.
+    calling this — this function does not consult ``FIXED_TEMPLATE``.
     """
-    if allowed_append_roles is None:
-        allowed_append_roles = {"tool"}
     if thinking not in THINKING_MODES:
         raise ValueError(f"thinking must be one of {THINKING_MODES}; got {thinking!r}")
     extra = extra_template_kwargs or {}
 
+    from miles.utils.chat_template_utils import TITOTokenizerType
```
Shi-Dong: Please hoist this import to the beginning

Shi-Dong (approved):
LGTM.
Note that the PR description claims that "default configurations resolve as before", but I think many defaults have been modified. You might want to update the description.

Shi-Dong (approved):
Also please hoist the imports.

guapisolo:
@Shi-Dong Thanks — I hoisted both imports to module scope in `18dc019e42344081a33a6a9fffa107696db68af6`. I also updated the PR description to remove the merged #1816 dependency note and replace the “default configurations resolve as before” claim with the actual default-versus-named-family behavior.

miles/utils/test_utils/chat_template_verify.py
guapisolo: Hoisted `TITOTokenizerType` to module scope in `18dc019e42344081a33a6a9fffa107696db68af6`.

miles/utils/test_utils/chat_template_verify.py
guapisolo: Hoisted `ALL_APPEND_ROLES` to module scope in `18dc019e42344081a33a6a9fffa107696db68af6`.

PR #1818, opened 2026-07-27
refactor(tito): remove --tito-allowed-append-roles and derive append surfaces from fixed templates

guapisolo:
## Summary
Remove `--tito-allowed-append-roles`; verify surfaces derive schedules from `FIXED_TEMPLATE`.

## Motivation
With append roles bound to the family template, the flag is dead caller policy: remove it from argparse, validation, examples, fixtures, the verify runner, bench, and docs. The verify surface (GPU harness, e2e configs, verify scripts) now derives its role schedule from FIXED_TEMPLATE, and injected-assistant driving becomes a terminal verify action. Flag removal and surface derivation are one contract move; separating them leaves the GPU session-verify chain reading a removed argument in between.

## Before / After
- **Before / After:** callers passed `--tito-allowed-append-roles`; the session-verify harness read the roles from args.
- Now the harness derives append surfaces from the family's `FIXED_TEMPLATE`; injected-assistant input becomes a terminal schedule action.
- **What moved where:** schedule derivation lives in `session_verify_agent.py` (`fixed_template_append_roles`, `select_schedule`); flag plumbing is deleted from `arguments.py`, examples, fixtures, runner, bench, docs; the ownership wording in `agentic-chat-template.md` follows; 27 files, +392/−169.

## Behavior Preservation
- **How we know:** with the flag unset (the only configuration consistent with the parent PR's binding), derived surfaces equal the previous defaults.
- `TestTitoFixedTemplateConfiguration` pins the per-family derived configuration; the harness tests pin the schedules.

## Verification
- **Existing test suite:** `pytest tests/fast/utils/test_arguments.py tests/fast/utils/test_utils` — 148 passed at this commit.
- **New tests:** `test_session_verify_agent.py` covers restricted-family schedules (`test_minimax_schedule_excludes_system_and_keeps_assistant_rollback`, Qwen3.5 variant) plus injected-assistant rollback (`test_assistant_input_generated_response_is_rolled_back_and_regenerated`).

## Review Focus
- Scrutinize the schedule derivation in `session_verify_agent.py` for families whose templates restrict append roles.
- Scrutinize the misconfiguration gate rework in `arguments.py` for configurations the removed flag used to reject.

PR #1819, opened 2026-07-27
refactor(session): track generated checkpoints separately from client assistant history

guapisolo:
> **Depends on:** #1818

## Summary
Rollback reads recorded `generated_checkpoint_message_ends` instead of scanning assistant roles.

## Motivation
Fixed templates may append client-injected assistant messages, but only backend-generated responses establish token checkpoints. Counting assistant roles during rollback would retain stale token/record state; generated_checkpoint_message_ends becomes the checkpoint source of truth. The rollback API and the behavior of sessions without injected assistants are unchanged.

## Before / After
- **Before / After:** rollback scanned the matched prefix for the last `role == "assistant"` message; every assistant message counted as a checkpoint.
- `LinearTrajectory` now records `generated_checkpoint_message_ends` when it appends a generated response; rollback takes the last recorded end inside the matched prefix.
- **What moved where:** checkpoint bookkeeping is a `LinearTrajectory` field updated on append/rollback in `linear_trajectory.py`; +137/−22 across the class plus its tests.

## Behavior Preservation
- **How we know:** the pre-existing `TestRollback` cases pass unchanged; sessions without injected assistants keep identical checkpoint positions.

## Verification
- **Existing test suite:** `pytest tests/fast/router/test_linear_trajectory.py` — 44 passed at this commit.
- **New test:** `test_injected_assistants_are_not_generated_checkpoints` — injected assistants create no checkpoints; rollback/regeneration walk the recorded `generated_checkpoint_message_ends` transitions.

## Review Focus
- Scrutinize the rollback path in `linear_trajectory.py` for history shapes where injected assistants sit between generated responses.


miles/rollout/session/linear_trajectory.py
```diff
@@ -194,23 +201,25 @@ def _try_detect_and_rollback_to_assistant_checkpoint(
...
+        # Only responses generated by this session create checkpoints. Assistant
+        # messages supplied by the client are ordinary prompt history.
+        checkpoint_index = next(
+            (
+                i
+                for i in range(len(self.generated_checkpoint_message_ends) - 1, -1, -1)
+                if self.generated_checkpoint_message_ends[i] <= match_len
+            ),
+            -1,
+        )
```
Shi-Dong: Nit: This line is too long and takes a human a while to parse. I'd suggest splitting it up into multiple lines to make the logics clearer.

Shi-Dong (approved):
LGTM!

miles/rollout/session/linear_trajectory.py
guapisolo: Good point — unfolded it into an explicit reverse loop (and tightened the comment) in edfe502ed9.

PR #1827, opened 2026-07-27
fix: run PPO GAE over trainable tokens only

Shi-Dong:
## Summary

PPO's GAE currently treats every response token as an MDP transition and injects the terminal reward at the last response token, regardless of the loss mask. In multi-turn / agentic rollouts, tool and environment observation tokens are masked (`loss_mask == 0`), which breaks GAE in three ways:

1. **Terminal reward can be attenuated or lost.** If the response ends with masked tokens (e.g. a final env observation), the scalar reward lands on a masked position. It only reaches trainable tokens after `(gamma * lambd)^gap` decay — and with `gamma = 0` it is lost entirely.
2. **Masked tokens leak signal into the recursion.** KL-shaped rewards and critic value deltas at masked positions (tokens the policy never produced) flow into the advantages of neighboring trainable tokens, and each masked token adds one step of `gamma * lambd` decay, diluting credit across long observation blocks.
3. **Truncation semantics were undefined.** Nothing documented or tested what happens when a rollout hits the length limit.

This PR defines and enforces the semantics in `get_advantages_and_returns_batch`:

- GAE runs on the compressed subsequence of trainable tokens; masked tokens carry no reward (including KL shaping), contribute no value delta, and the GAE carry crosses them with exactly one `gamma * lambd` decay per trainable transition (undecayed pass-through).
- The terminal reward is injected at the **last trainable token**.
- Fully masked samples get zero advantages/returns (their terminal reward is dropped) instead of training on garbage.
- Truncated sequences keep the existing zero-bootstrap treatment (value after the last trainable token taken as 0, observed reward still applied) — now documented and pinned by a test, no behavior change.
- The function outputs zero advantages/returns at masked positions. Downstream transforms can still shift these entries to nonzero values — advantage whitening applies its affine transform to every position, and the on-policy distillation KL penalty is added per token — but the whitening statistics are mask-weighted, so the injected zeros do not bias them. Correctness relies on the losses: the policy loss re-zeros advantages at inactive tokens and every loss reducer weights by `loss_mask`, so masked positions never receive gradient.

For fully unmasked responses the results are bit-identical to the previous implementation (covered by a regression test). The `vanilla_gae` / `chunked_gae` kernels are untouched — masking happens by compressing each sample before the batched scan, so the FLA-style chunked kernel keeps working as-is.

## Test plan

- [x] New `tests/fast/backends/training_utils/test_ppo_gae_masks.py` (all parametrized over `chunked` True/False):
  - internal mask gap matches a serial reference on the compressed subsequence; masked positions get 0
  - carry crosses a mask gap with exactly one `gamma * lambd` decay, independent of gap length
  - terminal reward lands on the last trainable token when the tail is masked (`gamma = 0` would previously lose it)
  - fully masked sample yields zeros without disturbing its batch neighbors
  - all-ones mask reproduces the previous unmasked behavior exactly
  - `gamma = lambd = 1` telescopes to `sum(rewards) + terminal - V_t`, pinning zero-bootstrap truncation semantics
- [x] New 2-rank Gloo CP test in `test_ppo_cp_advantages.py`: masked GAE under CP == single-rank baseline, masked positions zero, exact values for `gamma = lambd = 0`
- [x] Full CPU baseline suite green (64 passed): `tests/test_chunked_gae.py`, `test_ppo_gae_masks.py`, `test_ppo_cp_advantages.py`, `test_train_data_conversion.py`, `test_ppo_ratio_numerics.py`, `test_true_on_policy_loss_metrics.py`
- [x] Re-ran `test_ppo_gae_masks.py` + `test_ppo_cp_advantages.py` after merging `origin/main` (17 passed)
- [x] `ruff check` and `black --check` clean on all touched files

### DIFF
--- a/miles/backends/training_utils/loss_hub/math_utils.py
+++ b/miles/backends/training_utils/loss_hub/math_utils.py
@@ -611,7 +611,7 @@ def get_advantages_and_returns_batch(
     """
     Batched GAE with CP support, computed over trainable tokens only.
 
-    Semantics (PPO-002):
+    Semantics:
       - Masked tokens (`loss_mask == 0`, e.g. tool/env observations in
         multi-turn rollouts) are not MDP transitions. GAE runs on the
         subsequence of trainable tokens, so masked tokens carry no reward
--- a/miles/backends/training_utils/loss_hub/math_utils.py
+++ b/miles/backends/training_utils/loss_hub/math_utils.py
@@ -624,8 +624,15 @@ def get_advantages_and_returns_batch(
       - Truncated sequences use the same zero bootstrap as terminated ones:
         the value after the last trainable token is taken as 0 and the
         observed terminal reward is still applied.
-      - Advantages and returns at masked positions are 0; the policy and value
-        losses mask them out, so they are never trained.
+      - This function outputs zero advantages and returns at masked positions.
+        Downstream transforms may still shift these entries to nonzero values
+        (advantage whitening applies its affine transform to every position,
+        and the on-policy distillation KL penalty is added per token), but the
+        whitening statistics themselves are mask-weighted, so the injected
+        zeros do not bias them. Correctness relies on the policy and value
+        losses masking these positions out (the policy loss re-zeros
+        advantages at inactive tokens and all loss reducers weight by
+        `loss_mask`), so masked positions never receive gradient.
 
     C_i is the length of values_list[i] and rewards_list[i] on the current CP rank.
     Input:
--- a/miles/backends/training_utils/loss_hub/math_utils.py
+++ b/miles/backends/training_utils/loss_hub/math_utils.py
@@ -603,12 +603,30 @@ def get_advantages_and_returns_batch(
     terminal_rewards,
     qkv_format,
     max_seq_lens,
+    loss_masks,
     gamma,
     lambd,
     chunked: bool = True,
 ):
     """
-    Batched GAE with CP support.
+    Batched GAE with CP support, computed over trainable tokens only.
+
+    Semantics (PPO-002):
+      - Masked tokens (`loss_mask == 0`, e.g. tool/env observations in
+        multi-turn rollouts) are not MDP transitions. GAE runs on the
+        subsequence of trainable tokens, so masked tokens carry no reward
+        (including KL shaping), contribute no value delta, and the GAE carry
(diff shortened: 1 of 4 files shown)


miles/backends/training_utils/loss_hub/math_utils.py
```diff
@@ -605,19 +605,38 @@ def get_advantages_and_returns_batch(
...
+        crosses them without extra `gamma * lambd` decay.
+      - The terminal reward is added at the last trainable token, not the last
+        response token.
+      - Fully masked samples get zero advantages and returns; their terminal
+        reward is dropped.
+      - Truncated sequences use the same zero bootstrap as terminated ones:
+        the value after the last trainable token is taken as 0 and the
+        observed terminal reward is still applied.
+      - Advantages and returns at masked positions are 0; the policy and value
+        losses mask them out, so they are never trained.
```
Shi-Dong: Seems that downstream OPD and advantage whitening may still shift them to nonzero values.

miles/backends/training_utils/loss_hub/math_utils.py
```diff
@@ -657,77 +684,83 @@ def get_advantages_and_returns_batch(
...
+        # Compress each sample to its trainable positions so that masked
+        # tokens do not act as MDP transitions in the GAE recursion.
+        trainable_indices = [
+            loss_masks[i][: response_lengths[i]].to(device).nonzero(as_tuple=True)[0] for i in range(B)
+        ]
+        trainable_lengths = [idx.numel() for idx in trainable_indices]
+
         # pad to max_len for batched GAE
-        max_len = max(response_lengths)
+        max_len = max(trainable_lengths)
```
guapisolo, line 700: I have a small question here:
- For rollout samples ending with `loss_mask=1`. e.g. Tool response
 - If it's mistakenly added, I think we should trim it in session server, or `convert_samples_to_train_data` function (rollout sample -> train samples)
 - If it's intentionally added, I think we shouldn't set `max_len = max(trainable_lengths)`.. let me discuss with cc.

guapisolo: Ignore this cmt. My idea is not correct.

guapisolo (approved):
Goodfix. LGTM.

Zhichenzzz (approved):

PR #1843, opened 2026-07-28
Wire up InferenceController and RolloutExecutor and make the controller API async

fzyzcjy:
Part of #1837

### DIFF
--- a/docs/advanced/fault-tolerance.md
+++ b/docs/advanced/fault-tolerance.md
@@ -49,10 +49,10 @@ Each loop iteration does:
 ## Engine recovery
 
 When `--use-fault-tolerance` is on, `MegatronActor.update_weights` calls
-`rollout_manager.recover_updatable_engines` on rank 0 before each weight
+`inference_controller.recover_updatable_engines` before each weight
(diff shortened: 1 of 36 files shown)


guapisolo:
Could we pull two small, dependency-light fixes forward and place them immediately after #1843, so this prefix can merge without waiting for the full refactor stack?

1. Extract the minimal process-seam fix from #2614: keep the controller-side and executor-side EvalFleet state separate, and pass only data such as EvalFleetInfo / EvalFleetPin across the process boundary. This closes the serialized-copy/state-divergence bug introduced here. This should be described as a process-boundary correctness fix only: #1847 and #2497 intentionally remove the old eval-fleet FT path, so it must not claim that dedicated eval recovery is restored.
2. Preserve replay compatibility for persisted audit JSONL with component="rollout_manager", either through a legacy alias or an explicit migration to rollout_executor. #2503 still accepts only rollout_executor, so older records remain unreadable or get dropped.

Ideally these are minimal follow-up commits/PRs directly on top of #1843, and we merge that smaller safe prefix before the remaining #1837 stack.

PR #1845, opened 2026-07-28
Support saving and loading for rollout functions and add a BaseRolloutFn base class

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/rollout_executor.py
+++ b/miles/ray/rollout/rollout_executor.py
@@ -236,13 +236,24 @@ class RolloutExecutor:
 
     # -------------------------- checkpointing -----------------------------
 
+    # TODO the train and eval rollout functions will become one object, so one save/load is enough here
     def save(self, rollout_id):
         if self.args.rollout_global_dataset:
             self.data_source.save(rollout_id)
+        if not self.use_legacy_rollout_v1:
+            if self.generate_rollout is not None:
+                self.generate_rollout.save(rollout_id)
+            if (eval_fn := self.eval_generate_rollout) is not None and eval_fn is not self.generate_rollout:
+                eval_fn.save(rollout_id)
         event_logger_checkpoint.snapshot(self.args, rollout_id)
 
     def load(self, rollout_id=None):
         self.data_source.load(rollout_id)
+        if not self.use_legacy_rollout_v1:
+            if self.generate_rollout is not None:
+                self.generate_rollout.load(rollout_id)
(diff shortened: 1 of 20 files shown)


miles/ray/rollout/rollout_executor.py
```diff
@@ -236,13 +236,18 @@ async def _get_rollout_data(self, rollout_id):
...
         if self.args.rollout_global_dataset:
             self.data_source.save(rollout_id)
+        if not self.use_legacy_rollout_v1:
+            self.generate_rollout.save(rollout_id)
         event_logger_checkpoint.snapshot(self.args, rollout_id)
 
     def load(self, rollout_id=None):
         self.data_source.load(rollout_id)
+        if not self.use_legacy_rollout_v1:
+            self.generate_rollout.load(rollout_id)
```
fzyzcjy: [P1] Guard checkpoint hooks when no rollout function exists

When `--load-debug-rollout-data` is set on the default non-legacy path, `__init__` intentionally leaves `self.generate_rollout` as `None`, and the unconditional startup call from `create_training_models()` reaches this line. This raises `AttributeError` before replay training starts, while `save()` would fail identically at the first checkpoint. Deliver-1 `deec6fce05` and deliver-2 retain this dereference, so skip rollout hooks when no rollout object was constructed.

miles/ray/rollout/rollout_executor.py
```diff
@@ -236,13 +236,18 @@ async def _get_rollout_data(self, rollout_id):
 
     # -------------------------- checkpointing -----------------------------
 
+    # TODO the train and eval rollout functions will become one object, so one save/load is enough here
     def save(self, rollout_id):
         if self.args.rollout_global_dataset:
             self.data_source.save(rollout_id)
+        if not self.use_legacy_rollout_v1:
+            self.generate_rollout.save(rollout_id)
```
fzyzcjy: [P2] Checkpoint a distinct eval rollout instance

When `--eval-function-path` differs from `--rollout-function-path`, `__init__` creates a distinct `eval_generate_rollout`, but checkpointing invokes hooks only on `generate_rollout`. A stateful eval implementation therefore never writes or restores its state despite implementing the new contract, so resumed evaluations silently restart from constructor state. This remains in deliver-1 `deec6fce05` and deliver-2; also checkpoint `eval_generate_rollout` when it is not the shared train instance.

fzyzcjy (commented):
The patch introduces a startup crash in supported debug-replay mode and omits checkpointing for separately configured eval rollout objects. Both defects persist at deliver-1, so the op is not safe to deliver.

PR #1848, opened 2026-07-28
Drop the node_rank guards and access only node-0 engines

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No blocker attributable to this op remains at deliver-1. The intervening cell/client refactors make node-0 addressing explicit, and no deliver-2 fix needs to be moved forward for this op.

PR #1852, opened 2026-07-28
Add a general httpx client provider to http_utils

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/http_utils.py
+++ b/miles/utils/http_utils.py
@@ -183,6 +183,29 @@ def terminate_process(process: multiprocessing.Process, timeout: float = 1.0) ->
         process.join()
 
 
+class GeneralHttpClientProvider:
+    _CONNECT_TIMEOUT = 10.0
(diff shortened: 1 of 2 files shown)


guapisolo (dismissed):

miles/utils/http_utils.py
```diff
@@ -183,6 +183,29 @@ def terminate_process(process: multiprocessing.Process, timeout: float = 1.0) ->
         process.join()
 
 
+class GeneralHttpClientProvider:
+    _CONNECT_TIMEOUT = 10.0
+    _TIMEOUT = httpx.Timeout(None, connect=_CONNECT_TIMEOUT)
```
guapisolo: **[P1] Keep finite write and pool deadlines**

`httpx.Timeout(None, connect=_CONNECT_TIMEOUT)` disables the `read`, `write`, and `pool` timeouts together. A stalled request-body write or pool wait can therefore block the shared SGLang control path indefinitely. Could we keep `read=None` for long-running responses while assigning finite deadlines to `write` and `pool`, and add a focused test for the resulting timeout policy?

PR #1853, opened 2026-07-28
Make SGLangApiClient an async worker handle over httpx

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: #1861 removes the Ray actor boundary that made `HTTPStatusError` unserializable, and the fix remains present through the checked chain tip.

PR #1854, opened 2026-07-28
Make the rollout-side offload/onload fan-out async to avoid blocking sync code

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: the async fan-out preserves the existing filtering, concurrency, and result semantics, with no blocking finding through the checked chain tip.

PR #1855, opened 2026-07-28
Add the sync-to-async bridge for fire-then-rendezvous call sites

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: the sync-to-async bridge preserves the intended fire-then-rendezvous semantics, with no blocking finding through the checked chain tip.

PR #1856, opened 2026-07-28
Extract the engine url formatting for reusing

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: the shared URL builder preserves IPv4, hostname, and IPv6 behavior, with no blocking finding through the checked chain tip.

PR #1858, opened 2026-07-28
Avoid using SGLangEngine actor to call SGLang APIs and directly call in rollout process

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/server_engine.py
+++ b/miles/ray/rollout/server_engine.py
@@ -5,6 +5,8 @@ import logging
 import ray
 from pydantic import BaseModel, ConfigDict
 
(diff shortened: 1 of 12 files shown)


guapisolo (approved):
Approving this PR in the context of the #2176 safe-merge checkpoint. Moving rollout control calls from Ray actor forwarding to `SGLangApiClient` removes an unnecessary scheduling and serialization hop while keeping process lifecycle ownership in the actor. The original concern—that this migration rejected the documented `host`/`port` override contract—is fixed by #2082 before #2176 and remains fixed through the checked chain tip.

PR #1860, opened 2026-07-28
Pause health monitoring before snapshotting the updatable engines to fix racing condition

fzyzcjy:
Part of #1837


guapisolo (approved):
good fix.. 
health monitor = engine set writer, weight updater = engine set reader
Should pause write first, then read engine set snapshot, and then update weight.

PR #1862, opened 2026-07-28
Register the engines with the router from the rollout process instead of ray actors

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving this PR in the context of the #2176 safe-merge checkpoint. Moving router-membership ownership into the rollout process is the right direction, but the original ordering could mark an engine alive before router registration succeeded. #2090 commits serving state only after successful router publication; that fix lands before #2176 and remains present through the checked chain tip.

PR #1864, opened 2026-07-28
Introduce ServerCell and store the group's engines as cells

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving this PR in the context of the #2176 safe-merge checkpoint. Introducing `ServerCell` is the right ownership model, but the original floor-division grouping could silently drop a trailing engine wrapper for an uneven multi-node topology. #1879 adds the required divisibility fail-fast and focused coverage before #2176, and that fix remains present through the checked chain tip.

PR #1865, opened 2026-07-28
Make ServerGroup callers cell-aware and drop the all_engines property

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context. The stale `all_engines` consumer introduced here is fixed by #1883, before safe merge checkpoint #2176. The remaining cell-aware migration looks good to me.

PR #1866, opened 2026-07-28
Move vanilla methods onto ServerCell

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in the context of the #2176 safe-merge checkpoint. This change moves cell-scoped offload/onload/check_weights dispatch into ServerCell while preserving the primary-engine target, allocation filters, fan-out, and result shape. Fresh native and direct review found no actionable issue in this increment.

PR #1867, opened 2026-07-28
Move engine teardown into ServerCell.stop

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: #1870 moves teardown selection to cell indices, eliminating the partial-engine unregister-before-validation state, and the fix remains present through the checked chain tip. This finding therefore no longer blocks the PR in stack context.

PR #1868, opened 2026-07-28
Extract the Ray actor construction into launch_sglang_ray_actor

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/server_cell.py
+++ b/miles/ray/rollout/server_cell.py
@@ -1,10 +1,15 @@
 import logging
+import os
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 import ray
+from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
+from miles.backends.sglang_utils.sglang_engine import SGLangEngine
 from miles.ray.rollout.server_engine import ServerEngine
+from miles.ray.utils import NOSET_VISIBLE_DEVICES_ENV_VARS_LIST
+from miles.utils import dumper_utils
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


guapisolo (dismissed):
Reviewed in the #1837 stack context: this is a behavior-preserving extraction of the SGLang Ray actor construction, with placement, environment, and lifecycle flow preserved. No active finding needs to be carried into safe merge checkpoint 1 (#2176).

miles/ray/rollout/server_cell.py
```diff
@@ -57,6 +62,67 @@ def flatten_cells(cells: list[ServerCell]) -> list[ServerEngine]:
...
+    pg: Any,
+    global_rank: int,
+    gpu_index: int,
+    worker_type: str,
+    sglang_overrides: dict,
+    num_gpus_per_engine: int,
+) -> ray.actor.ActorHandle:
+    pg, reordered_bundle_indices, reordered_gpu_ids = pg
+
+    num_gpus = 0.2
```
guapisolo, line 77: [P1] Ban the RDT path before landing this refactor

RDT is not part of this refactor plan and conflicts directly with the actor-ownership changes in this stack. Do not port the #1313 launch logic into `launch_sglang_ray_actor`.

Minimal revert instruction:

1. Remove `"rdt"` from the `--update-weight-transfer-mode` choices so startup rejects this path.
2. Resolve this hunk entirely in favor of this PR: drop `rdt_reuse_pg`, the RDT `num_gpus=0` branch, `rdt_pg_kwargs`, and `pg_bundles` forwarding.
3. Keep `launch_sglang_ray_actor` as the single non-RDT actor-launch path; do not add an RDT compatibility branch.

The remaining unreachable RDT implementation can be removed separately. The acceptance condition here is that `--update-weight-transfer-mode rdt` is rejected and no RDT-specific rollout-launch behavior survives this rebase.

PR #1871, opened 2026-07-28
Rename PortCursors to PortAllocator and key it by the node's ip

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in the #1837 stack context. This cleanly migrates port-allocation identity from group-local indices to node IPs; #1872 completes the shared allocator mechanics before the #2176 safe-merge checkpoint, and the behavior remains present through the checked chain tip.

PR #1872, opened 2026-07-28
Add PortAllocator.alloc and thread the shared allocator through allocation

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context. The shared per-node allocator change is sound, with no active findings requiring follow-up before safe merge checkpoint #2176. LGTM.

PR #1873, opened 2026-07-28
Serve HTTP from the mock engine on its allocated port

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/test_utils/mock_sglang_engine.py
+++ b/miles/utils/test_utils/mock_sglang_engine.py
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 import logging
-import threading
 import ray
+
+from miles.utils.misc import get_free_port
+from miles.utils.test_utils.mock_sglang_http_server import MockSGLangHttpServer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


miles/utils/test_utils/mock_sglang_engine.py
```diff
@@ -70,10 +80,7 @@ def simulate_crash(self):
 
     def _get_current_node_ip_and_free_port(self, start_port: int = 15000, consecutive: int = 1):
         self._record("_get_current_node_ip_and_free_port", (), {"start_port": start_port, "consecutive": consecutive})
-        with self._lock:
-            port = max(self._port_seq, start_port)
-            self._port_seq = port + consecutive
-            return ("127.0.0.1", port)
+        return ("127.0.0.1", get_free_port(start_port=start_port, consecutive=consecutive))
```
guapisolo, line 83: [P1] Codex review (reproduced): this actor still publishes `127.0.0.1`, while the new HTTP server also binds only to loopback. `tests/conftest.py` supports connecting to an existing cluster through `RAY_ADDRESS`; when Ray places the actor on another node or network namespace, the driver resolves this URL to its own loopback. I reproduced this with separate server and driver network namespaces: the server returned HTTP 200 locally, while two fresh driver namespaces both got `ConnectionRefusedError` on the advertised address. Please return `get_current_node_ip()` here, bind the mock server to that address (or `0.0.0.0`), and add a remote-placement case.

guapisolo (commented):
Codex review: one reproduced P1 finding.

miles/utils/test_utils/mock_sglang_engine.py
guapisolo: not fixed in the later PRs

PR #1874, opened 2026-07-28
Raise for external rollout mode before launching actors

fzyzcjy:
Part of #1837


guapisolo (approved):
External rollout will be re-supported in #2514. approve as-is.

PR #1876, opened 2026-07-28
Make start_engines async and await engine init inside it

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context. This PR safely moves init waiting into ServerGroup; the later stack completes the lifecycle-ownership cleanup before the safe-merge checkpoint at #2176. No active finding remains. LGTM.

PR #1877, opened 2026-07-28
Restructure start_engines into per-cell blocks and assume engines in one cell have same lifecycle

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: #1886/#2115 make cell startup atomic with rollback, and #1889 restores cross-cell startup concurrency; both fixes remain present through the checked chain tip.

PR #1878, opened 2026-07-28
Move per-cell engine startup into ServerCell.start_engines

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/server_group.py
+++ b/miles/ray/rollout/server_group.py
@@ -1,17 +1,14 @@
 import asyncio
 import dataclasses
-import functools
 import logging
 from typing import Any
 
-import ray
 from sglang.srt.constants import GPU_MEMORY_TYPE_WEIGHTS
 
-from miles.backends.sglang_utils.sglang_engine import build_server_url
 from miles.backends.sglang_utils.sglang_router_api_client import SGLangRouterApiClient, use_legacy_router_api
 from miles.ray.rollout.addr_allocator import PortAllocator
-from miles.ray.rollout.server_cell import SHUTDOWN_TIMEOUT, ServerCell, flatten_cells, launch_sglang_ray_actor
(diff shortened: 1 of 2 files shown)


miles/ray/rollout/server_group.py
```diff
@@ -82,67 +79,37 @@ async def start_engines(
...
             if cell.is_allocated:
                 continue
 
-            new_entries: list[tuple[int, ServerEngine, Any]] = []
-            for local_index, engine_slot in enumerate(cell.engines):
-                flat_index = cell_index * self.nodes_per_engine + local_index
-                global_rank = self.rank_offset + flat_index
-                rollout_engine = launch_sglang_ray_actor(
+            started_cell_indices.append(cell_index)
+            cell_starts.append(
```
guapisolo, line 92: [P1][Checkpoint status: unresolved at #2176; fixed later in #2622]

This moves the lifecycle claim into the coroutine passed to `asyncio.gather()`. Two concurrent start/recover calls can both select the same stopped cell before either child claims it; a concurrent stop can hit the same window, so the second start may assert or a completed stop may be undone. Please serialize start/stop, or claim synchronously before yielding with an idempotent loser recheck. The #2176 path still calls `RayWorkerManager.start_cells()` / `stop_cells()` directly without lifecycle serialization; the shared membership lock only lands later in #2622.

guapisolo (changes requested):
Blocking for safe-merge checkpoint #2176: this refactor defers the lifecycle claim into the coroutine passed to asyncio.gather(), so concurrent start/recover calls can select the same stopped cell and a concurrent stop can observe stale state. The direct RayWorkerManager.start_cells()/stop_cells() path still bypasses serialization at #2176; the shared membership lock only arrives in #2622. Please bring that fix before #2176, or move the checkpoint past the complete fix.

PR #1879, opened 2026-07-28
Sink the startup config from ServerGroup into ServerCell

fzyzcjy:
Part of #1837


guapisolo (dismissed):
Approved for this incremental change in the #1837 stack. Moving startup config into ServerCell preserves rank/GPU placement and engine metadata. The inherited #1878 lifecycle race remains separate and is only fixed by #2622, after safe-merge checkpoint #2176.

PR #1880, opened 2026-07-28
Address workers by cell index and let the cell describe itself to the router

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: the cell-index contract is carried consistently through lifecycle and router membership, and the checked chain keeps that ownership model intact.

PR #1881, opened 2026-07-28
Move per-cell recovery into ServerCell.recover

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in stack context: #2117 completes the provider-reconcile replacement of the aggregate-recovery path, and #2094 fixes missed engine reconnects; both land before safe merge checkpoint #2176 and remain present through the checked tip.

PR #1882, opened 2026-07-28
Move has_new_engines to RolloutServer

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in stack context. The migration residue is removed in #1883, and the partial-recovery dirty-state gap is replaced by the snapshot/connection-status source of truth in #2094. Both fixes land before safe merge checkpoint #1 (#2176).

PR #1883, opened 2026-07-28
Delete ServerGroup and let RolloutServer own a flat list of cells

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in stack context: this flattens RolloutServer → ServerGroup → ServerCell → ServerEngine into RolloutServer → ServerCell → ServerEngine. The migration gaps are repaired by #2063, #2064, #2100, and #2108 before safe merge checkpoint #2176; the remaining cancellation-safety P2 is non-blocking and still present at the checked #1837 tip.

PR #1884, opened 2026-07-28
Extract the rollout cell state machine into cell_state.py

fzyzcjy:
Part of #1837


guapisolo (approved):
LGTM in the #1837 stack context. This cleanly extracts the rollout state schema without changing lifecycle semantics, and the later cell-level ownership refactors retain it through safe-merge checkpoint #2176.

PR #1885, opened 2026-07-28
Drive the cell state machine through whole-cell transitions

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context: #1885 can leak staged actors on launch failure and its restart test is invalid, both fixed by #2112 with rollback, retry, and whole-cell restart coverage; #2115/#2175 further harden and cover the failure path. These fixes land before safe merge ckpt 1 (#2176) and remain through the checked chain tip #2649, so the findings no longer block this PR within the safe-merge range.

PR #1886, opened 2026-07-28
Lift the state machine onto ServerCell and delete ServerEngine

fzyzcjy:
Part of #1837


miles/ray/rollout/rollout_server.py
```diff
@@ -219,9 +217,9 @@ class RolloutServer:
     _port_allocator: PortAllocator = dataclasses.field(default_factory=PortAllocator)
 
     @property
-    def engines(self) -> list[ServerEngine]:
-        """All node-0 engines across all cells."""
-        return [cell.primary_engine for cell in self.server_cells]
+    def api_clients(self) -> list[SGLangApiClient]:
```
guapisolo, line 220: `RolloutServer.engines` is removed here, but `EvalFleet._pin_fleet()` still reads `self._srv.engines` before entering its retry block. With `--eval-num-gpus > 0`, the first snapshot eval raises `AttributeError` instead of degrading to `EvalSkip`. Please migrate this path to `api_clients` before the #2176 safe-merge checkpoint; the equivalent repair currently appears only in #2614.

miles/ray/rollout/server_cell.py
```diff
@@ -142,62 +166,93 @@ async def stop(self, router_api_client: SGLangRouterApiClient) -> None:
...
+        logger.info(f"{debug_name} start old={self._state}")
+        assert isinstance(self._state, old_state_cls), f"{self._state=}"
+        self._state = new_state
+        logger.info(f"{debug_name} end new={self._state}")
+
+    async def probe_and_mark_dead(self) -> None:
+        if not self.is_allocated:
+            return
+        try:
+            await asyncio.wait_for(self.primary_actor_handle.get_weight_version.remote(), timeout=60)
```
guapisolo, line 224: `ServerCell.probe_and_mark_dead()` now probes only `primary_actor_handle`. In a multi-node cell, a secondary actor can die while the cell remains allocated/alive, so `recover()` never restarts it. Please preserve all-worker liveness detection before the #2176 safe-merge checkpoint; the equivalent `RayWorkerManager` repair currently appears only in #2622.

PR #1887, opened 2026-07-28
Key a server's cells by a globally unique string id instead of numerical indices

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context: the dict migration here left the topology-test helper iterating keys and probe_and_mark_dead() treating keys as cells; #2063 and #2067 respectively fix those issues before safe-merge checkpoint #2176. Both repairs remain present through the checked chain tip.

PR #1888, opened 2026-07-28
Start the rollout engines without blocking the caller

fzyzcjy:
Part of #1837


guapisolo:
Non-blocking: #1889 made the node probes concurrent, but dynamic port allocation still calls blocking `ray.get()` from async `_BaseActorManager.alloc_ports()` through `PortAllocator.alloc()`; this remains present at #2176 and the checked #2649 tip. Could we make port allocation awaitable, or explicitly offload the blocking call, so the `RayWorkerManager` event loop cannot be stalled?

PR #1889, opened 2026-07-28
Probe every node of a cell concurrently to avoid serialized initialization

fzyzcjy:
Part of #1837


guapisolo:
This incremental change itself reviewed cleanly, but I cannot approve it for the #2176 safe-merge checkpoint: the lifecycle race introduced in #1878 is still reachable through the direct RayWorkerManager.start_cells()/stop_cells() path at #2176. #2108 only protects the older controller path; the complete shared membership lock does not land until #2622. Please move that fix before #2176, or move the checkpoint past the complete fix.

PR #1891, opened 2026-07-28
Store per-token weight version spans per generation call on Sample

fzyzcjy:
Part of #1837

### DIFF
--- a/examples/infra_features/random_async/random_async_rollout.py
+++ b/examples/infra_features/random_async/random_async_rollout.py
@@ -40,7 +40,7 @@ from random_async_sglang_metrics import SGLangMetricsReporter, record_agent_requ
 from miles.rollout.data_source import DataSource
 from miles.utils.async_utils import run
 from miles.utils.http_utils import post as http_post
-from miles.utils.types import Sample
+from miles.utils.types import Sample, WeightVersionsPerCall
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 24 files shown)


guapisolo:
Holding approval for now. I checked the single #1891 delta (`3f255213` -> `65ce548b`) against current `main@2c78e2a`, rather than merging the stale stack branch.

- **Semantics:** there is no fundamental conflict with fully async. `DefaultDataBuffer` only consumes `Sample.oldest_weight_version`; legacy scalar metadata becomes one span, while a call containing multiple versions makes group staleness use the oldest span. That is intentionally more conservative and preserves the existing group-level contract.
- **Current-main port:** 23 of 24 changed files apply cleanly. `miles/dashboard/dump_reader.py` has four content conflicts: imports, `TrainRow.weight_versions`, `SUMMARY_VERSION`, and `_summary_row`. The resolution needs to preserve current main CP/alignment and rollout `staleness`, while flattening spans for `weight_version`, `weight_version_min`, `mixed_version`, and `turns`, then bumping the summary cache version.
- **Test migration:** current-main fully async, session, and router fixtures still construct or assert flat `list[str]`. With the new representation those paths fail at `call.spans` or `call.to_dicts()`. They need typed fixtures/expectations, plus coverage for one generation call containing multiple spans and using the oldest span for staleness.
- **Stack boundary:** flat-string dump compatibility is repaired by #1892, so I am not asking #1891 to duplicate that fix. The multimodal span-remapping limitation is already documented by the TODO and is not an additional blocker here.

GitHub correctly reports this PR mergeable into its stacked base; the conflict above is with the eventual current-main port. The data-model change looks compatible, but I would like the rebase/port and migration coverage represented before approval.

PR #1892, opened 2026-07-28
Load rollout dumps written before per-call weight version spans

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in stack context. This restores pre-span dump compatibility without fabricating token ranges; the remaining legacy-metrics gap is non-blocking for safe merge checkpoint #1 (#2176).

PR #1893, opened 2026-07-28
Temporarily publish the weight version when no weight update can carry it

fzyzcjy:
Part of #1837


guapisolo (approved):
Weight-version numbering is not yet consistent across update modes (`disk-delta` starts at `v0`, while distributed modes start at `v1`). Mixed-mode weight updates are not currently supported, so this is non-blocking for the stack through safe-merge checkpoint #2176; please align the version semantics before mixed-mode support is introduced.

PR #1894, opened 2026-07-28
Add assertions to reject training data generated under the sglang default weight version

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in the #1837 stack through #2649, including the #2176 safe-merge boundary. The guard rejects uninitialized SGLang weight versions without rejecting fully-async numeric staleness; no unresolved P1 remains, with missing-provenance and CI-injection coverage tracked as P2 follow-ups.

PR #1899, opened 2026-07-28
Snapshot the external commands of every shell launch script

fzyzcjy:
Part of #1837


yueming-yuan:
I'm kind of concerned that keeping this snapshot test after we merge this series of PR will make the future CI test more complex. Should we remove this after merging the PR?

yueming-yuan:
(we might also remove all .sh scripts and convert them to .py ones)

PR #1901, opened 2026-07-28
Snapshot the commands and generated configs of every python launch script

fzyzcjy:
Part of #1837


yueming-yuan:
similar concern to #1899, after this script refactor, I think people will frequently change the .py scripts, so if they don't regenerate the snapshot each time, the CI will fail and make it a bit more complex.

I think for most script changes, the semantic changes are intentional (and just a few lines), so should we move this test to manual and only run it when people work on large changes/refactors?

fzyzcjy:
(discussed offline)

PR #1915, opened 2026-07-28
fix(lora): honor --attention-backend in the bridge LoRA path

yushengsu-thu:
LoRA runs silently ignore `--attention-backend`: `_setup_lora_model_via_bridge` never copies it onto the provider, unlike `model_provider.py:78`. Megatron maps that field to the `NVTE_*_ATTN` vars, so the dropped value let TE pick a backend the caller had opted out of — which is why `test_lora_qwen2.5_0.5B.py` dies in the fused-attention backward on `main` nightly.

Honoring it makes one previously inert request take effect. gpt-oss runs with `--softmax-type learnable`, where TE 2.17 disables FlashAttention outright and reports no fused kernel, leaving `UnfusedDotProductAttention` as the only option; its CI test asked for `fused`, so it now asks for `auto`.

### DIFF
--- a/miles/backends/megatron_utils/bridge_lora_helpers.py
+++ b/miles/backends/megatron_utils/bridge_lora_helpers.py
@@ -101,6 +101,7 @@ def _setup_lora_model_via_bridge(args: Namespace) -> list:
     provider.recompute_num_layers = args.recompute_num_layers
     provider.recompute_modules = args.recompute_modules
     provider.distribute_saved_activations = args.distribute_saved_activations
+    provider.attention_backend = args.attention_backend
     provider.variable_seq_lengths = True
     provider.moe_token_dispatcher_type = "alltoall"
     provider.moe_router_load_balancing_type = "none"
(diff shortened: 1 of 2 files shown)


miles/backends/megatron_utils/bridge_lora_helpers.py
```diff
@@ -101,6 +101,10 @@ def _setup_lora_model_via_bridge(args: Namespace) -> list:
     provider.recompute_num_layers = args.recompute_num_layers
     provider.recompute_modules = args.recompute_modules
     provider.distribute_saved_activations = args.distribute_saved_activations
+    # attention kernel selection (parity with model_provider.py): Megatron turns this
+    # into NVTE_FLASH_ATTN/NVTE_FUSED_ATTN, so dropping it leaves TE's heuristic free
+    # to pick a backend the user explicitly opted out of.
```
yushengsu-thu: remove the comments

yushengsu-thu: Removed in 80df93674 — the PR is now the single assignment line.

guapisolo (approved):
LGTM.

guapisolo:
let's wait for CI pass and then merge

yushengsu-thu:
@guapisolo  passed

Zhichenzzz (approved):

PR #1918, opened 2026-07-28
Remove the experimental swe-agent examples

Shi-Dong:
## Summary

Removes all three swe-agent examples under `examples/experimental/` and repoints every doc reference at `examples/swe-agent`.

- `examples/experimental/swe-agent` — the original Nemo-Gym-based generation. Last substantive change was 2026-03-25, and it was the only thing in the repo using git submodules.
- `examples/experimental/swe-agent-v2` — the Harbor recipe that was promoted to `examples/swe-agent` in #1741.
- `examples/experimental/swe-agent-v2-amd` — a near-copy of the above for AMD.

`.gitmodules` is deleted as well: its only two entries were the submodules belonging to the first directory, both pointing at personal forks, so the file is empty once that tree is gone. **Clones no longer need `--recurse-submodules`.**

## Why

`examples/experimental/swe-agent-v2` was described in the docs as "the maintained recipe" for Harbor training, but it lived under a directory whose README says the contents are "not fully verified, only for experimental use". #1741 promoted it to `examples/swe-agent`, so keeping the experimental copies means two divergent copies of the same recipe and docs that send readers to the one we no longer want them to use.

The other two directories are not superseded so much as finished: the Nemo-Gym example targets an environment the repo no longer integrates with, and the AMD variant is a fork of the promoted tree.

## What is not carried over

`examples/swe-agent` covers the GLM-4.7-Flash Harbor path (sync + fully-async launchers). These files existed only in the removed trees and have no equivalent there, so this PR does drop them:

- two GLM-4.7 **reasoning** launchers (sync and async) — a different model configuration from the Flash agentic recipe
- a training recipe targeting an internally-hosted Harbor server
- a standalone agent-server implementation from the AMD variant

All of it stays in git history and can be restored or promoted separately if any of it is still wanted. Flagging it explicitly because it is easy to miss in a deletion-only diff.

## Docs updated

`docs/user-guide/harbor.md`, `docs/user-guide/agentic-chat-template.md` and `docs/user-guide/rollout-endpoints.md` all pointed at the experimental path; they now point at `examples/swe-agent`. One docstring cross-reference in the OpenEnv example was updated for the same reason.

## Test plan

- [x] `git grep` finds zero remaining references to the removed paths anywhere in the tree
- [x] `.gitmodules` is empty after the submodule paths are removed, and no gitlink entries remain in the index
- [x] `examples/swe-agent` is untouched by this PR
- [x] `uvx ruff check` passes on the one modified Python file
- [ ] Docs build / link check in CI


guapisolo (approved):
LGTM. time to say goodbye to old files

nblintao (approved):

PR #1919, opened 2026-07-28
examples: add swe-agent-harbor-daytona (Harbor sandboxes on Daytona)

Shi-Dong:
## Summary

Adds `examples/swe-agent-harbor-daytona`, a variant of `examples/swe-agent` whose
task sandboxes run on [Daytona](https://www.daytona.io/) cloud sandboxes instead
of local Docker.

Two things differ from `examples/swe-agent`:

- **Daytona sandboxes.** The agent-server host needs outbound HTTPS but no
  Docker daemon, no local image builds, and no local disk for task images. This
  is the practical option when the trainer runs on a GPU node where you cannot
  run Docker-in-Docker.
- **terminus-2 agent.** terminus-2 runs as a host process and calls the model
  endpoint itself rather than from inside the sandbox. It appends both `user`
  and `tool` turns, so the session server needs
  `--tito-allowed-append-roles user tool`. Without it every append is rejected,
  trajectories collapse to a single turn, and the session server rolls them
  back. That flag is the reason this needs its own launcher instead of a README
  pointing at `examples/swe-agent/run.py`.

Everything else — TITO, the session server, GRPO, the reward path — is shared
with `examples/swe-agent`, so `run.py` puts that example on `PYTHONPATH` and
imports its `generate.py` and `swe_agent_function.py` rather than keeping a
second copy in sync. Data prep also reuses
`examples/swe-agent/download_and_process_data.py`, with `--agent-name terminus-2`.

## Contents

| File | Purpose |
| --- | --- |
| `README.md` | Provision Daytona, start the agent server, prepare data, launch training, verify progress, troubleshoot. |
| `run.py` | Training launcher for Daytona-backed terminus-2 runs. |
| `launch_agent_server.sh` | Starts the Harbor agent server in Daytona mode. |

The README also writes down the failure modes that cost the most time on a
multi-day run of this shape:

- `--rollout-max-response-len` and `AGENT_MAX_OUTPUT_TOKENS` cap a **single
  turn**, not the whole trajectory. Under `HARBOR_RESPONSE_LENGTH_POLICY=abort`
  an over-cap turn ends the trial with none of that turn's tool calls performed,
  so the trial scores 0 and dilutes its GRPO group.
- Read `rollout/raw_reward` for the solve rate; `rollout/rewards` is the
  GRPO-centered advantage and sits near zero by construction.
- When censusing trial directories by outcome, key on the trial's **start** time
  (the mtime of its `config.json`), because writing `exception.txt` bumps the
  directory mtime and makes any mtime-sorted listing look like everything is
  failing.

Two launcher settings that are hardcoded in `examples/swe-agent/run.py` are
exposed as flags here, since they are the ones you actually end up tuning:
`--rollout-max-response-len` and `--max-tokens-per-gpu` (plus `--lr` and
`--sglang-mem-fraction-static`).

`--router-external-host` defaults to the resolved local IP. sgl-router parses
this into a Rust `SocketAddr`, so it must be a numeric IP and a hostname fails
to bind. The default is computed in a helper guarded against resolution
failure, so an unresolvable hostname cannot break module import.

## Test plan

- [x] `python -m py_compile examples/swe-agent-harbor-daytona/run.py`
- [x] `bash -n examples/swe-agent-harbor-daytona/launch_agent_server.sh`
- [x] `ruff check examples/swe-agent-harbor-daytona/run.py`
- [x] Pipeline validated end-to-end by a multi-day GLM-4.7-Flash run on one node
      of 8 H200 GPUs against Terminal-Bench 2 tasks, reaching non-zero
      `rollout/raw_reward` with the agent server colocated on the trainer host.
- [ ] Docs-and-example change; no library code is touched, so no existing test
      covers it.


examples/swe-agent-harbor-daytona/launch_agent_server.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+# Start the Harbor agent server in Daytona mode.
+#
+# Run this from a harbor checkout, before launching run.py. Trials are graded
```
Shi-Dong: Should specify the harbor branch, e.g. `harbor-miles-v0.20.0`.

examples/swe-agent-harbor-daytona/README.md
```diff
@@ -0,0 +1,186 @@
...
+- A `stopped` sandbox still consumes quota. Only deletion, or expiry of the
+  sandbox's auto-delete interval, frees it.
+- The quota is shared by everything using the same account. If other jobs use
+  the key, your sandbox creations will start failing for reasons that have
+  nothing to do with your run — so check total usage by owner before assuming
+  you are leaking sandboxes.
+
+An over-quota creation is rejected quickly rather than hanging, and surfaces as
+`DaytonaValidationError`, or as `EnvironmentStartTimeoutError` when creations
+are merely slow because the account is near its limit.
```
Shi-Dong: We don't need this part.

examples/swe-agent-harbor-daytona/README.md
```diff
@@ -0,0 +1,186 @@
...
+Run the agent server under a process supervisor or a detached terminal
+multiplexer on its own host, not in a foreground shell over SSH: if that shell
+dies it takes the agent server and every live sandbox with it, and the trainer
+then starves without an obvious error.
+
+Verify `http://<agent-server>:11000/health` before launching Miles.
+
+## 3. Prepare data
+
+`../swe-agent/download_and_process_data.py` converts a local JSONL into Miles
```
Shi-Dong: Don't use relative path. Specify full path (from repo root).

examples/swe-agent-harbor-daytona/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Troubleshooting
+
+| Symptom | Cause |
+| --- | --- |
+| `DaytonaValidationError` on sandbox create | Account-wide disk quota exhausted; see step 1. |
+| `EnvironmentStartTimeoutError` in bursts | Sandbox creation is slow because the account is near its quota. |
+| `SingleTurnMaxSeqLenExceededError` | Per-turn output cap too low; see the sizing section. |
+| `ContextLengthExceededError` | `AGENT_MAX_INPUT_TOKENS` below the observed context length. |
+| Trajectories all one turn, many session rollbacks | `--tito-allowed-append-roles user tool` missing, so terminus-2's appends are rejected. |
```
Shi-Dong: Remove this row.

examples/swe-agent-harbor-daytona/README.md
```diff
@@ -0,0 +1,173 @@
+# SWE-Agent training with Harbor on Daytona sandboxes
+
+This example trains GLM-4.7-Flash on agentic terminal and coding tasks, with
+task sandboxes hosted on [Daytona](https://www.daytona.io/) instead of local
+Docker. Miles runs synchronous GRPO and serves the policy through its session
+server; a Harbor agent server drives the agent and returns verifier rewards.
+
+It is the [`examples/swe-agent`](../swe-agent) pipeline with two changes:
```
Shi-Dong: Don't use relative path anywhere in this PR.

examples/experimental/swe-agent-harbor-daytona/README.md
```diff
@@ -0,0 +1,173 @@
+# SWE-Agent training with Harbor on Daytona sandboxes
+
+This example trains GLM-4.7-Flash on agentic terminal and coding tasks, with
+task sandboxes hosted on [Daytona](https://www.daytona.io/) instead of local
+Docker. Miles runs synchronous GRPO and serves the policy through its session
+server; a Harbor agent server drives the agent and returns verifier rewards.
```
Shi-Dong, line 6: Mention here that this example is supposed to be run on a single H200 node (8 GPUs).

examples/experimental/swe-agent-harbor-daytona/run.py
```diff
@@ -0,0 +1,284 @@
+"""SWE-Agent launcher (GLM-4.7-Flash) for Harbor tasks on Daytona sandboxes.
+
+Same Miles <-> Harbor pipeline as examples/swe-agent, with two differences:
```
nblintao: Do you think we should rename `swe-agent` to something like `swe-agent-harbor-docker`? or should we should better keep the url slug unchanged?

examples/experimental/swe-agent-harbor-daytona/run.py
```diff
@@ -0,0 +1,284 @@
...
+        "--custom-agent-function-path swe_agent_function.run "
+        "--custom-rm-path generate.reward_func "
+        "--rollout-function-path generate.RolloutFn "
+        "--dynamic-sampling-filter-path miles.rollout.filter_hub.dynamic_sampling_filters.check_no_aborted "
+        "--tito-model glm47 "
+        "--use-session-server "
+        "--session-server-port 30000 "
+        # terminus-2 appends both user and tool turns; without this the session
+        # server rejects the append and rolls the trajectory back.
+        "--tito-allowed-append-roles user tool "
```
nblintao: https://github.com/radixark/miles/pull/1818 already removed this option. Do we still need this `run.py`?

examples/experimental/swe-agent-harbor-daytona/launch_agent_server.sh
```diff
@@ -0,0 +1,43 @@
...
+export AGENT_MAX_OUTPUT_TOKENS="${AGENT_MAX_OUTPUT_TOKENS:-8192}"
+export HARBOR_RESPONSE_LENGTH_POLICY=abort
+
+mkdir -p "$TRIALS_DIR"
+
+exec python miles_agent_server.py \
+    --host 0.0.0.0 \
+    --port "$PORT" \
+    --max-concurrent "$MAX_CONCURRENT" \
+    --agent-timeout 5400 \
```
nblintao, line 40: `examples/swe-agent/swe_agent_function.py` harcoded to 
```
        response = await asyncio.wait_for(
            _post_agent_server(f"{agent_server_url}/run", request),
            timeout=3600,  # 1 hour max per trial
```
so the server will still be running after the agent function gives up.

examples/experimental/swe-agent-harbor-daytona/launch_agent_server.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
nblintao, line 1: Are we supposed to converge sh scripts to py?

nblintao (changes requested):

examples/experimental/swe-agent-harbor-daytona/launch_agent_server.sh
Shi-Dong: This script is for launching the agentic server instead of the training job. Converting to `.py` is mainly for the purpose of CI.

examples/experimental/swe-agent-harbor-daytona/run.py
Shi-Dong: Done. Removed `run.py`.

nblintao (approved):

examples/experimental/swe-agent-harbor-daytona/run.py
Shi-Dong: Good idea. Opened PR 2233 to handle the renaming.

examples/experimental/swe-agent-harbor-daytona/launch_agent_server.sh
Shi-Dong: Opened a follow-up PR 2228 to address this comment.

PR #1825, opened 2026-07-27
Kimi-K3 lora RL day-0 support

yueming-yuan:
ci-sglang-pr: #37704
ci-megatron-pr: #94

### Blog
check details and experiment results in:
https://www.lmsys.org/blog/2026-07-27-kimi-k3-day0-support

### Depends on

- radixark/Megatron-LM#94 (the K3 backend primitives: `gated_activation_func`, `moe_latent_use_norm`, `sum_gradients_across_tp_domain`)
- radixark/miles#3194 (native LoRA fixes: the engine skips startup `lora_paths` when a trainer pushes the adapter, which the K3 launchers rely on with `--lora-adapter-path`)
- sgl-project/sglang#37704 (colocated K3 LoRA RL fixes on `sglang-miles`)

Runs on the stock `radixark/miles:dev` image with those two branches; the old `sglang-miles-k3` branch and the `radixark/miles:kimi-k3` image are retired.

### About

Megatron training backend for Kimi K3 plus the colocated SGLang rollout path
it needs for RL. KDA and MLA attention per layer, the attention-residual
snapshot bank, and native LoRA adapters under TP/EP/PP/CP — with shared-A /
per-expert-B factors for the 896-expert MoE, exported to the rollout engines
as HF-named chunks over CUDA IPC.

### Reproduce

One launcher, `scripts/run_kimi_k3.py`: `--model-name Kimi-K3|Kimi-K3-4layer`, `--train-mode lora|full`.
The native MXFP4 checkpoint, its BF16 dequantization and the `torch_dist` conversion derive from
`--model-dir`/`--model-name` and each has a `prepare-*` command; the four-layer prune is
`Pinaster/Kimi-K3-4layer` (1 dense + 3 MoE layers):

```bash
python scripts/run_kimi_k3.py prepare-download --model-name Kimi-K3-4layer --task gsm8k
python scripts/run_kimi_k3.py prepare-bf16 --model-name Kimi-K3-4layer
python scripts/run_kimi_k3.py prepare-torch-dist --model-name Kimi-K3-4layer
python scripts/run_kimi_k3.py train --model-name Kimi-K3-4layer --train-mode lora --mode normal --task gsm8k
```

The full-model recipe (validated on 16 nodes x 4 GPUs) is in `docs/models/kimi/kimi-k3.md` section 4.

### Validation

- **Four-layer, rollout TP16, GB300 4 nodes** (the layout on which sgl-project/sglang#37704's Marlin
[...]

PR #1922, opened 2026-07-29
Kimi K3: --colocate-memory-peak-device — overlap the trainer/rollout handoff on the GPU

yueming-yuan:
Stacked on the `kimi-k3` branch (#1825). One squashed commit. The persistent CUDA-IPC LoRA transfer buffer was split out to #1923.

### What

`--colocate-memory-peak-device {cpu,gpu}` (default `cpu` = existing behavior).

In colocated RL, each trainer<->rollout handoff briefly holds two host copies:
the engine's weight mirror (built when the engine releases its weights under
`--enable-weights-cpu-backup`) and the trainer's own TMS backup. On hosts where
the two together exceed the memory budget (GB300: 4x140 GB mirror + trainer
backup vs the job cgroup), `gpu` mode reorders both handoffs so the overlap
lives on the GPU instead and the host copies never coexist:

- rollout->train: drop the engine KV cache + CUDA graphs (pure GPU memory),
  wake the trainer while the engine weights are still resident, then release
  the weights — their mirror replaces the trainer's just-freed backup.
- train->rollout: the trainer returns its allocator caches and the no-backup
  LoRA grad-buffer region to the device, the engine resumes its weights, and
  only then does the trainer sleep into the space the mirror vacated.
- init: the engine weights stay resident through the trainer checkpoint load
  (their mirror cannot coexist with the load's footprint).

`RolloutManager` gains split `offload_kv`/`offload_weights`/`onload_weights`;
the trainer actor gains an idempotent `sleep`/`wake_up` pair plus
`offload_grad_buffer`; the K3 launcher hardcodes `gpu`.

### Validation

16 nodes x 4 GPU (GB300), full Kimi-K3 LoRA RL (together with #1923): both
handoffs stable cycle over cycle, host anon+shmem peaking ~770 GB of the
898 GB cgroup with a single large host copy at any time; `eval/aime`
0.367 -> 0.467 by the second eval. This is the configuration inside the
released `radixark/miles:kimi-k3` image, which the Reproduce section of
#1825 describes.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -253,10 +253,31 @@ class MegatronTrainRayActor(TrainRayActor):
 
         return start_rollout_id
 
+    # Make sleep()/wake_up() idempotent: gpu peak mode wakes the trainer early.
+    _train_offloaded = False
+    _grad_buffer_offloaded = False
+
+    @with_logs
+    @timer
+    def offload_grad_buffer(self) -> None:
+        """Release the no-backup grad buffers ahead of the engine weight
+        resume: free host-side, and the heavy PP stages need the bytes."""
+        assert self.args.offload_train
+        if self._train_offloaded or self._grad_buffer_offloaded:
+            return
+        if not is_lora_enabled(self.args):
+            return
+        print_memory("before offload grad_buffer")
+        torch_memory_saver.pause(tag="grad_buffer")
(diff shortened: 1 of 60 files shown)


yueming-yuan:
Superseded by #3100 (--colocate-memory-peak-device gpu under LoRA, merged to main); the K3 branch now carries it via main.

PR #1923, opened 2026-07-29
Kimi K3: ship LoRA chunks through a persistent CUDA-IPC transfer buffer

yueming-yuan:
Stacked on the `kimi-k3` branch (#1825). One commit, one file. Independent of #1922.

### What

The chunked LoRA push previously allocated a fresh flat bucket per chunk for
CUDA-IPC. Every bucket rotates an IPC handle and a `/torch_*` ref-counter
file, and under consumer-side release lag this races the producer's
reclamation: premature reap -> freed-then-reused addresses -> stale-handle
`cudaIpcOpenMemHandle` failures on the receiving schedulers, ENOENT
`shm_unlink` aborts inside noexcept paths (`__cxa_call_terminate` ->
SIGABRT), and up to a whole update's buckets left pinned on the GPU past the
sync.

All LoRA chunks now ship through one 512 MB/GPU storage, allocated once and
never freed: one handle and one ref-counter file for the whole run, nothing
enters the CUDA-IPC limbo, so there is nothing to race or reap. The caller's
per-chunk ACK barrier makes overwriting safe; an explicit stream sync after
each copy covers the fact that the IPC sync event is recorded only at the
first share of a storage.

### Validation

16 nodes x 4 GPU (GB300), full Kimi-K3 LoRA RL: the weight-sync failure
family above stopped reproducing with this change; producer allocated returns
to baseline + 0.5 GB after every update. Shipped inside the released
`radixark/miles:kimi-k3` image.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -55,7 +55,6 @@ from .parallel import verify_megatron_parallel_state
 from .replay_utils import register_replay_list_moe
 from .update_weight.common import named_params_and_buffers
 from .update_weight.update_weight_from_distributed.broadcast import UpdateWeightFromDistributed
-from .update_weight.update_weight_from_distributed.p2p import UpdateWeightP2P
 from .update_weight.update_weight_from_tensor import UpdateWeightFromTensor
 
 if TYPE_CHECKING:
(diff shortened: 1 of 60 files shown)


yueming-yuan:
Superseded by the unified weight-update stack on main plus #3147 (IPC producer release, per-bucket adapter release); the persistent transfer buffer is no longer needed.

PR #1927, opened 2026-07-29
[feat] Support training with variable global batch size

Zhichenzzz:
## Motivation

RL rollout production is inherently variable: dynamic-sampling filters drop samples, agentic / multi-turn rollouts emit a variable number of training samples per execution, and collected batches rarely land on an exact multiple of the training batch size. The legacy pipeline assumed the opposite — a fixed `global_batch_size` counted in samples, every DP rank holding the same sample count, per-sample loss weighting. The consequences: trailing samples were trimmed away (wasted rollouts), a rollout execution that produced N training samples was counted N times in the loss (over-weighting long executions), and its sibling samples could be split across training steps or trim boundaries.

This PR first moves DP/microbatch scheduling to the rollout side (commit 1), then makes the **rollout execution the unit of accounting** end to end, and adds the packing/throughput knobs that fall out of having a global scheduler. The core design follows slime's variable-global-batch-size series ([THUDM/slime#1930](https://github.com/THUDM/slime/pull/1930), [THUDM/slime#1933](https://github.com/THUDM/slime/pull/1933)), adapted to miles's backends and constraints; the FLOPs-balanced packing ports slime [#2017](https://github.com/THUDM/slime/pull/2017)/[#2029](https://github.com/THUDM/slime/pull/2029). FFD packing and the partial final step are miles extensions on top.

## Worked examples

**1. Trailing rollouts: dropped vs trained.** 32 rollouts collected, `global_batch_size=24`:

```
before:  train [24] + drop 8            -> 25% of generation wasted
after:   train [24] + train [8]         -> --allow-partial-train-step
```

The final step is mathematically honest, not padded: its loss normalizer is 8 and the LR scheduler advances by 8, so the gradient is "the mean over these 8 rollouts" — the same estimator as a full step, just with fewer rollouts.

**2. One execution split into N samples: loss double-counting.** An agentic execution (rollout 7) is too long for one training sample, so the generate fn splits it into siblings A (500 loss tokens) and B (300); C (400 tokens) is an independent execution:

```
before:  loss = (mean(A) + mean(B) + mean(C)) / 3     -> rollout 7 counted twice
after:   loss = (sum(A)/800 + sum(B)/800 + sum(C)/400) / 2
                 \_____ = rollout 7's token mean ____/
```

`rollout_mask_sums` broadcasts the whole-rollout total (800) to every sibling as its denominator, so the execution is counted exactly once — regardless of which micro-batch, DP rank, or CP shard each sibling lands on: the partial `sum/800` contributions re-assemble under the existing all-reduce.

**3. Siblings split by a batch boundary.** 5 samples where the last two are siblings, legacy `global_batch_size=4` counted in samples:

```
before:  train [s0 s1 s2 s3a] , drop s3b     -> half an execution trained
after:   group {s0} {s1} {s2} {s3a+s3b}, gbs counts rollouts
         -> siblings enter one step together, or drop together
```

**4. Pack first, distribute second.** 6 samples of lengths `[9,7,5,3,2,1]`, token cap 10, dp2:

```
before (split by position, pack per rank):
  rank0 [9,7,5] -> bins [9][7][5]   (3 micro-batches, 21 tokens)
  rank1 [3,2,1] -> bins [3+2+1]     (1 micro-batch,    6 tokens)
  -> PP requires equal counts: rank1 idles; 21 vs 6 load skew

after (global FFD pack, then distribute):
  bins [9+1][7+3][5+2] -> split one bin to reach dp multiple ->
  rank0: [9+1][5] = 15 tokens   rank1: [7+3][2] = 12 tokens
  -> equal micro-batch counts by construction, balanced load
```

**5. The common case is a strict no-op.** With 1 rollout = 1 sample and counts divisible by `global_batch_size`, rollout groups are singletons, each denominator is the sample's own mask sum, and the loss reduces to the legacy per-sample mean term by term — verified numerically below (30B: identical `logprob_abs_diff` to main).

### Foundation: rollout-side scheduling (commit 1)

Scheduling becomes a single pure function on the rollout side, `miles/utils/dp_schedule.py::build_dp_schedule`:

- For each training step (a `global_batch_size` chunk of samples), samples are assigned to DP ranks (Karmarkar-Karp under `--balance-data`, strided otherwise), then each rank's samples are grouped into micro-batches — fixed `micro_batch_size` chunks (static) or first-fit under `max_tokens_per_gpu * cp` with a DP-wide MAX and largest-bin splitting for PP/VPP alignment (dynamic).
- The rollout manager packages the result into the per-rank data shards: each shard carries its `micro_batch_indices` plus the shared per-step `num_microbatches`. `get_data_iterator` on the training side just consumes them — the `all_reduce(MAX)` sync point is gone.
- The training backend advertises eligibility by exposing its full parallel config (`dp/cp/vpp/mb_group`). Paths that can't precompute keep the legacy train-side scheduling untouched: multi-LoRA (slot-contiguity constraints), fsdp/torchtitan, indep_dp/fault-tolerance (dp_size can change at runtime), multimodal batches (train-side media-token expansion changes sequence lengths), and non-divisible sample counts.

Being pure Python with no ray/sglang imports, the scheduler is unit-testable on CPU: invariants (full coverage, disjoint partitions, per-rank micro-batch-count equality for PP sync, token-cap respect with the oversized-sample exception, VPP rounding) are asserted across static/dynamic/balance scenarios, plus a randomized equivalence sweep showing `num_microbatches` matches the legacy train-side computation exactly.

### Variable global batch size (commit 2)

Three layers, all computed on the rollout side and shipped to trainers inside the data shards:

**1. Scheduling — pack first, distribute second.**
Samples are grouped by `Sample.rollout_id` (defaults to `sample.index`, so the classic one-sample-per-rollout path is unchanged; compact rollout functions stamp the same id on every sibling, enforced by shape validation before flattening). Training steps are formed from `global_batch_size` ROLLOUTS — a rollout's samples never split across steps. Each step's samples are packed into `K` micro-batches with one global first-fit-decreasing pass under the token cap (`max_tokens_per_gpu * cp`); `K` is aligned up to `dp_size * mb_group` by splitting the largest bins (cap-preserving); micro-batches are then distributed across DP ranks — strided, Karmarkar-Karp on micro-batch token sums (`--balance-data`), or on per-sample FLOPs estimates (`--balance-by-flops`, capturing the quadratic attention term when lengths vary widely). Trailing rollouts that don't fill a step drop whole — or, with `--allow-partial-train-step`, train as one smaller final step: genuinely variable per-step batch sizes at zero sample waste.

**2. Accounting — per-rollout loss weighting.**
`rollout_mask_sums` — the whole-rollout loss-mask total, broadcast to every sibling sample — is precomputed at the step level, where all of a rollout's samples are visible. The loss reducer divides each sample's masked contribution by its rollout's total, so partial contributions summed across micro-batches (and across DP/CP ranks under the existing all-reduce) reconstruct exactly one token-weighted mean per rollout, no matter how the packer scattered the siblings. With 1 rollout = 1 sample this equals the legacy per-sample mean exactly.

**3. Normalization and metrics — true per-step counts.**
The loss normalizer and the LR scheduler increment use the actual per-step rollout count (`num_rollouts`, threaded from the schedule through the data shards via `get_num_rollouts`), not `args.global_batch_size`, so uneven steps stay correctly scaled. Metric reduction switches from mean-of-rank-means to count-weighted `(sum, count)` aggregation, unbiased when ranks or steps hold different counts.

## Compatibility

- **Strict no-op for existing workloads** — verified numerically below.
- Legacy train-side path preserved for multi-LoRA, fsdp/torchtitan, indep_dp/fault-tolerance, and multimodal; `--use-dynamic-global-batch-size` / `--disable-rollout-trim-samples` keep working for those paths.
- One deliberate semantic change: in the TIS decoupled path, modified response masks correct the numerator only; denominators stay the precomputed rollout totals, keeping a single normalizer across the loss.

## Validation

Three dedicated CI e2e tests ship with this PR:
- `tests/e2e/short/test_qwen2.5_0.5B_compact_rollout.py` — a compact generate fn (`tests/manual/compact_split_generate.py`) splits every rollout execution into two sibling samples; validates by-rollout step accounting and per-rollout loss weighting end to end.
- `tests/e2e/megatron/test_qwen3_4B_variable_gbs.py` — `global_batch_size=24` over 32 collected rollouts with `--allow-partial-train-step` and trim disabled, under cp2: trains uneven `[24, 8]` steps.
- `tests/e2e/megatron/test_qwen3_30B_A3B/test_dp_attention.py` — sglang dp-attention (`--sglang-enable-dp-attention`, dp4) with the rollout-side schedule active: dp-aware request dispatch reshuffles sample completion order; the schedule must stay order-insensitive (grouping keys on rollout_id).

CPU: 366 tests green at the tip (schedule invariants incl. compact grouping and whole-rollout trailing trim, randomized sweeps with random rollout groupings, weighted-reduction units, loss snapshots, session merge, FFD property tests, argument validation).

GPU e2e, 8xH200, `--ci-test`:

| run | topology | result |
|---|---|---|
| Qwen2.5-0.5B gsm8k (GRPO) | dp8, dynamic batch, mooncake | ✅ green at every layer of the series |
| Qwen3-30B-A3B (GRPO) | dp2 + cp2 + pp2 + ep2, metric gates | ✅ see no-op table below |
| Qwen3.5-35B-A3B GDN (GRPO) | cp2 | ✅ succeeded |
| Qwen3-4B PPO | actor cp2 + separate critic, `num_critic_only_steps 1` | ✅ succeeded on this rebased tree |
| compact gain demo | 0.5B + `tests/manual/compact_split_generate.py` | ✅ |
| compact stress | 288 samples / 64 rollouts, cp2, tight cap | ✅ |
| partial-step demo | gbs=24, trim off | ✅ |

### No-op check — 30B dp2+cp2+pp2+ep2, identical topology across the series layers

| tree | `train_rollout_logprob_abs_diff` | `grad_norm` |
|---|---|---|
| scheduling refactor only (#1927) | 0.0178 | 0.13 |
| + per-step batch-size threading | 0.0171 | 0.099 |
| + per-rollout denominators | 0.0177 | 0.099 |
| + pack-first / by-rollout steps | 0.0176 | 0.13 |

All within run-to-run sampling noise: 1-rollout-=-1-sample training is untouched.

### Compact rollout accounting (the new capability)

The compact fixture (`tests/manual/compact_split_generate.py`) splits each execution into 2 training samples sharing `rollout_id`:

```
Rollout-side DP schedule: num_samples=64, num_rollouts=[32], num_microbatches=[1]
```

32 executions → 64 samples → ONE step of 32 rollouts, per-rollout weighting; rewards stable (0.44-0.56) over 3 rollout batches. Previously this workload trained as 64 independent samples: two steps per batch, siblings split across step/trim boundaries, multi-sample rollouts over-weighted proportionally to their sample count.

Stress: `1 + (index % 8)` siblings per rollout (variable lengths), cp2, `max_tokens_per_gpu=2048` so siblings scatter across micro-batches AND DP ranks, NaN checking on:

```
Rollout-side DP schedule: num_samples=288, num_rollouts=[32, 32], num_microbatches=[3, 3]
```

No NaN, rewards 0.52-0.62, job succeeded — cross-mb/cross-rank denominators reconstruct exactly.

### Partial final step — variable gbs in action

gsm8k 0.5B, `--global-batch-size 24 --allow-partial-train-step --disable-rollout-trim-samples`:

```
Rollout-side DP schedule: num_samples=32, num_rollouts=[24, 8], num_microbatches=[1, 1]
```

Three rollout batches trained with uneven `[24, 8]` steps; the 25% of rollouts previously discarded now train; LR/loss follow the true per-step counts; CI logprob asserts pass.

### FFD packing gain (CPU sweep, 32 samples/step, 2000 steps/regime, cap 9216)

| length regime | bins saved vs first-fit |
|---|---|
| short responses (mean ~350) | 0% (both optimal) |
| lognormal mean ~2500 | 2.9% |
| lognormal mean ~4500 | 4.9% |
| uniform 2k-6k (items ~cap/2) | 6.8% |
| items ~cap | 0.6% |

Fewer bins = fewer gradient-accumulation passes per step; never worse (property-tested).

### PPO (Qwen3-4B, actor cp2 + separate critic)

- Schedule active on both actor and critic groups: `num_rollouts=[32, 32]`, micro-batches `[7,8] / [6,8] / [6,6]` across 3 rollout batches
- Critic trains through the same machinery: `critic-value_loss` 0.56 → 0.32-0.64 band, `value_clipfrac` 0.0, `critic-lr` 1e-5 as configured
- Actor rewards 0.5 → 0.72; weight sync clean

### DIFF
--- a/miles/utils/dp_schedule.py
+++ b/miles/utils/dp_schedule.py
@@ -1,0 +1,81 @@
+"""Per-rollout DP/microbatch scheduling, computed on the rollout side."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from miles.utils.seqlen_balancing import expand_bins_by_splitting, first_fit_pack, get_seqlen_balanced_partitions
+
+SCHEDULE_CONFIG_KEYS = ("dp_size", "cp_size", "vpp_size", "microbatch_group_size_per_vp_stage")
+
+
+def has_full_schedule_config(train_parallel_config: dict | None) -> bool:
+    """True when the backend advertised every field build_dp_schedule needs."""
+    if not train_parallel_config:
+        return False
+    return all(key in train_parallel_config for key in SCHEDULE_CONFIG_KEYS)
+
+
+def build_dp_schedule(
+    args: Any,
+    train_parallel_config: dict,
+    total_lengths: list[int],
+    *,
+    global_batch_size: int,
+) -> tuple[list[list[int]], list[list[list[int]]], list[int]]:
+    """Compute per-rank ``(partitions, micro_batch_indices, num_microbatches)``."""
+    dp_size = train_parallel_config["dp_size"]
+    cp_size = train_parallel_config["cp_size"]
+    vpp_size = train_parallel_config["vpp_size"] or 1
+    mb_group = train_parallel_config["microbatch_group_size_per_vp_stage"]
+
+    assert len(total_lengths) % global_batch_size == 0, (
+        f"num_samples ({len(total_lengths)}) is not a multiple of global_batch_size "
+        f"({global_batch_size}); trim upstream before scheduling."
+    )
+    num_steps = len(total_lengths) // global_batch_size
+
+    if args.use_dynamic_batch_size:
+        assert args.max_tokens_per_gpu is not None
+        max_per_bin = args.max_tokens_per_gpu * cp_size
+
+    partitions: list[list[int]] = [[] for _ in range(dp_size)]
+    micro_batch_indices: list[list[list[int]]] = [[] for _ in range(dp_size)]
+    num_microbatches: list[int] = []
+
+    for step_i in range(num_steps):
+        step_start = step_i * global_batch_size
+        step_lengths = total_lengths[step_start : step_start + global_batch_size]
+
+        if args.balance_data:
+            rank_parts = get_seqlen_balanced_partitions(step_lengths, dp_size, equal_size=True)
+        else:
+            rank_parts = [list(range(r, global_batch_size, dp_size)) for r in range(dp_size)]
+
+        if not args.use_dynamic_batch_size:
+            mbs = args.micro_batch_size
+            n = len(rank_parts[0])  # gbs / dp, same for every rank
+            assert n % mbs == 0, (
+                f"per-rank batch ({n} = global_batch_size {global_batch_size} / dp_size {dp_size}) "
+                f"is not a multiple of micro_batch_size ({mbs})"
+            )
+            rank_mbs = [[list(range(i, i + mbs)) for i in range(0, n, mbs)] for _ in range(dp_size)]
+            num_mbs_per_rank = n // mbs
+        else:
+            rank_lens = [[step_lengths[i] for i in rank_parts[r]] for r in range(dp_size)]
+            rank_mbs = [first_fit_pack(rank_lens[r], max_per_bin) for r in range(dp_size)]
+            num_mbs_per_rank = max(len(b) for b in rank_mbs)
+            if vpp_size > 1:
+                num_mbs_per_rank = max(num_mbs_per_rank // mb_group * mb_group, 1)
+            for r in range(dp_size):
+                expand_bins_by_splitting(rank_mbs[r], num_mbs_per_rank, rank_lens[r])
+
+        num_microbatches.append(num_mbs_per_rank)
+
+        for r in range(dp_size):
+            for mbs_local in rank_mbs[r]:
+                local_start = len(partitions[r])
+                partitions[r].extend(step_start + rank_parts[r][i] for i in mbs_local)
+                micro_batch_indices[r].append(list(range(local_start, local_start + len(mbs_local))))
+
+    return partitions, micro_batch_indices, num_microbatches
--- a/miles/utils/dp_schedule.py
+++ b/miles/utils/dp_schedule.py
@@ -39,15 +39,15 @@ def build_dp_schedule(
     vpp_size = train_parallel_config["vpp_size"] or 1
     mb_group = train_parallel_config["microbatch_group_size_per_vp_stage"]
 
-    # mbs count per step must divide evenly across ranks and, under VPP, mb groups.
+    # micro-batch count per step must divide evenly across ranks and, under VPP, mb groups.
     align_to = dp_size * (mb_group if vpp_size > 1 else 1)
 
     # Group sample positions by rollout id (first-occurrence order) so a rollout's
     # samples always land in the same training step.
-    rollout_to_samples: dict[int, list[int]] = {}
+    rollout_id_to_sample_index: dict[int, list[int]] = {}
     for sample_pos, rid in enumerate(rollout_indices):
-        rollout_to_samples.setdefault(rid, []).append(sample_pos)
-    rollout_ids = list(rollout_to_samples.keys())
+        rollout_id_to_sample_index.setdefault(rid, []).append(sample_pos)
+    rollout_ids = list(rollout_id_to_sample_index.keys())
 
     # Plan the rollout count of each training step: full steps of global_batch_size,
     # plus (under --allow-partial-train-step) one smaller final step for the trailing
@@ -60,7 +60,7 @@ def build_dp_schedule(
     num_rollouts = [global_batch_size] * num_full_steps
     leftover = len(rollout_ids) - num_full_steps * global_batch_size
     if leftover and getattr(args, "allow_partial_train_step", False) and args.use_dynamic_batch_size:
-        leftover_samples = sum(len(rollout_to_samples[rid]) for rid in rollout_ids[-leftover:])
+        leftover_samples = sum(len(rollout_id_to_sample_index[rid]) for rid in rollout_ids[-leftover:])
         if leftover_samples >= dp_size:
             num_rollouts.append(leftover)
         else:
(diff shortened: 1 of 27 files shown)


guapisolo (dismissed):
recover some key cmts to improve readability. LGTM now ,but better see #2135

miles/utils/dp_schedule.py
```diff
@@ -0,0 +1,167 @@
...
+        step_rollouts = rollout_ids[rollout_cursor : rollout_cursor + step_rollout_count]
+        rollout_cursor += step_rollout_count
+        sample_indices = [pos for rid in step_rollouts for pos in rollout_to_samples[rid]]
+        step_lengths = [total_lengths[i] for i in sample_indices]
+        assert len(sample_indices) >= dp_size, (
+            f"step of {step_rollout_count} rollouts has {len(sample_indices)} samples < dp_size {dp_size}; "
+            f"each step needs at least one sample per rank."
+        )
+
+        step_mbs = _pack_step(args, step_lengths, cp_size)
```
guapisolo: mbs is confusing "micro batch size" or "micro batches". Will commit fix.

yushengsu-thu:
I need this pr for enabling variable global batch size in multi-lora https://github.com/radixark/miles/pull/2137. waiting for this pr merging

miles/utils/dp_schedule.py
```diff
@@ -0,0 +1,141 @@
...
+        # Grow the micro-batch count to a multiple of align_to by splitting multi-sample micro-batches.
+        target = max((len(step_micro_batches) + align_to - 1) // align_to * align_to, align_to)
+        if target != len(step_micro_batches):
+            if not args.use_dynamic_batch_size:
+                raise AssertionError(
+                    f"static path: micro-batch count ({len(step_micro_batches)}) is not a multiple of {align_to}; "
+                    f"adjust the config so step_size % (dp_size * micro_batch_size * mb_group) == 0."
+                )
+            expand_bins_by_splitting(step_micro_batches, target, step_lengths)
+            assert (
```
guapisolo: The hard assert is required, but it's very fragile. We should directly ban non multi lora path of use dynamic global batch size. There is no way to ensure `len(step_micro_batches) % align_to == 0` or fit in target. I will submit a PR.

guapisolo (approved):

guapisolo:
Approved. only need to check #2197

yushengsu-thu (approved):

yueming-yuan (approved):

PR #1930, opened 2026-07-29
[2/N] Sample.rollout_id plumbing for compact / subagent rollouts

Zhichenzzz:
## Summary

Stacked on #1929. Part A of slime #1933: identify which rollout execution each training sample came from, so later PRs can aggregate loss/metrics per rollout instead of per sample.

- `Sample.rollout_id` field, default `None`; the pipeline falls back to `sample.index`, so the default one-sample-per-rollout path is unchanged.
- `convert_samples_to_train_data` emits `train_data["rollout_ids"]`; the DP split carries it per shard.
- `validate_rollout_id_annotated` walks the nested rollout output before flattening: the default shapes (flat `list[Sample]`, `list[list[Sample]]`) skip validation; a compact leaf (`list[Sample]` at depth >= 2 with more than one sibling) must have all siblings share a non-None `rollout_id`.

No numerics change — plumbing only.

## Validation

- CPU: 68 tests green (`test_rollout_data_conversion.py` — 6 new validation cases; `test_train_data_conversion.py` — index fallback + explicit rollout_id).
- GPU e2e: Qwen2.5-0.5B gsm8k short (dp8, mooncake) succeeded — `rollout_ids` round-trips through the object store.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -140,10 +140,8 @@ class MegatronTrainRayActor(TrainRayActor):
                 )
             dist.barrier(group=get_gloo_group())
 
-        # A full config (all miles.utils.dp_schedule.SCHEDULE_CONFIG_KEYS) advertises
-        # that the rollout side may precompute the DP/mbs schedule. indep_dp keeps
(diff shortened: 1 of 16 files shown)


Zhichenzzz:

PR #1931, opened 2026-07-29
[3/N] Per-rollout loss denominators and constant-divisor metric reduction

Zhichenzzz:
## Summary

Stacked on #1930. Parts B+D of slime #1933: aggregate loss and metrics per rollout instead of per sample. Strict no-op while 1 rollout = 1 sample (the default path); the semantics only change for compact / subagent rollouts that split one execution into multiple training samples.

- `convert_samples_to_train_data` precomputes `rollout_mask_sums` — whole-rollout loss-mask totals, broadcast per sample — at the step level, so a rollout whose samples land in different micro-batches still gets a full denominator. The DP split carries it; `get_rollout_data` promotes it to a GPU tensor.
- `get_sum_of_sample_mean` gains `sample_denoms`: `None` keeps the legacy per-sample mean; `rollout_mask_sums` yields one token-weighted mean per rollout.
- `loss_function` and the TIS decoupled path use the precomputed denominators. TIS note (deliberate alignment with slime #1933): modified response masks now correct the numerator only; the denominator stays the whole-rollout total from the original masks.
- `aggregate_train_losses` supports a constant `step_global_batch_size` divisor (producer leaves a `0` placeholder in `values[0]`; no CP factor — CP-chunked partial numerators reconstruct exactly once under the DP*CP all-reduce). Megatron threads it through both the normal and the indep_dp aggregation paths; fsdp / torchtitan stay on the legacy reduction.
- `log_rollout_data` token metrics use the same denominators, so reported means live in the same space as the gradient signal.

## Validation

- CPU: 84 tests green, including new `test_rollout_mask_sums.py` (precompute grouping, masked tokens, per-sample == per-rollout equivalence in the 1:1 case, constant-divisor reduction incl. legacy-producer guard), loss snapshot tests, true-on-policy dtype test.
- GPU e2e on 8xH200 (`--ci-test`, checker + metric gates on):

| run | topology | result |
|---|---|---|
| Qwen2.5-0.5B gsm8k short | dp8, dynamic batch, mooncake | ✅ succeeded |
| Qwen3-30B-A3B baseline | dp2 + cp2 + pp2 + ep2, metric gates | ✅ succeeded, `logprob_abs_diff` 0.0177 (base branch: 0.0171/0.0178 same topology) |

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -38,7 +38,13 @@ from miles.utils.types import RolloutBatch
 
 from ...utils.profile_utils import TrainProfiler
 from ...utils.tensor_backper import TensorBackuper
-from ..training_utils.data import DataIterator, get_data_iterator, get_rollout_data, sync_actor_critic_data
+from ..training_utils.data import (
+    DataIterator,
+    get_data_iterator,
+    get_global_batch_sizes,
+    get_rollout_data,
+    sync_actor_critic_data,
+)
 from ..training_utils.log_utils import log_cpu_memory, log_perf_data, log_rollout_data
 from ..training_utils.loss import (
     compute_advantages_and_returns,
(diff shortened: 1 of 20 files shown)


Zhichenzzz:

PR #1938, opened 2026-07-29
[7/N] Port --balance-by-flops for FLOPs-balanced micro-batching

Zhichenzzz:
## Summary

Stacked on #1937. Port of slime #2017 + #2029 (staying in sync with upstream): `--balance-by-flops` switches the dynamic-batch micro-batch partitioner from token packing to FLOPs-balanced Karmarkar-Karp, and distributes mbs across DP ranks by FLOPs weight.

- Per-sample workload = `calculate_fwd_flops` (miles's version — covers MoE, MLA/q_lora, GQA), capturing the quadratic attention term that token counts miss when sequence lengths vary widely.
- `num_mbs = ceil(total_tokens / cap)`; with at least as many mbs as samples, every sample gets its own mbs.
- Requires `--use-dynamic-batch-size` (validated at parse time). NOTE (upstream behavior, documented in the help text): FLOPs balancing does not enforce the per-mbs token cap — a partition can exceed `max_tokens_per_gpu` when the cap is tight.
- Mutually composes with the rest of the series: rollout grouping, partial steps, and FFD (FFD applies only on the non-FLOPs path).

## Validation

- CPU: 103 tests green, incl. FLOPs packing invariants (coverage/PP-sync; cap intentionally unchecked in this mode) and the singleton fallback.
- GPU e2e: Qwen2.5-0.5B gsm8k short + `--balance-by-flops` (dp8, mooncake, `--ci-test`) ✅ succeeded.

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -946,6 +946,18 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
             )
 
             parser.add_argument(
+                "--balance-by-flops",
+                action="store_true",
+                default=False,
+                help=(
+                    "Use FLOPs-based workload estimation for micro-batch partitioning via "
+                    "Karmarkar-Karp instead of first-fit token packing, and distribute mbs "
+                    "across DP ranks by FLOPs. Captures the quadratic attention cost when "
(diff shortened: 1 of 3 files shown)


Zhichenzzz:

PR #1940, opened 2026-07-29
Make the structured log prefix a caller-supplied tag

fzyzcjy:


guapisolo (approved):
Approved in stack context: the caller-owned structured-log tag contract introduced here remains unchanged through the checked #2176 tip, and the later RPC work beginning in #1942 consumes the same contract with `tag="rpc"`.

PR #1941, opened 2026-07-29
Move load_function into miles.utils.function_registry

fzyzcjy:


guapisolo (commented):
[P2][Unresolved through #2649] This refactor still leaves six `load_function` imports in `verifiers_rollout.py`, `checkpoint_eval.py`, `fully_async_data_buffer.py`, `session/v2/core.py`, and `message_matcher_hub/funcs.py` pointing at `miles.utils.misc`, so those consumers still pull in the heavy `misc`/Ray dependency. Please migrate them to `miles.utils.function_registry` and extend the isolation coverage to catch stale `misc` imports.

PR #1942, opened 2026-07-29
Implement a worker RPC server that submits a call and polls for its outcome

fzyzcjy:


guapisolo (approved):
Approved in stack context. The sync, dedup, cancellation, retention, and non-finite wire gaps are fixed by #1944/#1947/#1948/#1952/#1950 before #2176; return-type validation lands later in #2609, with all repairs retained through #2649.

PR #1943, opened 2026-07-29
Implement a typed RPC client handle

fzyzcjy:


miles/utils/workers/rpc/client/handle.py
```diff
@@ -0,0 +1,80 @@
...
+        self._transport = RpcTransport(server_url=server_url, http_client=http_client)
+
+    def __getattr__(self, name: str) -> Callable[..., Awaitable[Any]]:
+        if name.startswith("_"):
+            raise AttributeError(name)
+        spec = self._specs.get(name)
+        if spec is None:
+            raise AttributeError(f"{self._worker_cls_name} has no rpc method {name!r}")
+
+        async def call(**kwargs: Any) -> Any:
```
guapisolo, line 50: Could we preserve the worker method positional-call contract here? This wrapper only accepts `**kwargs`, so `await handle.foo(1)` fails locally before submission even when the worker method accepts positional arguments. The gap is still present at safe-merge checkpoint #2176; #2313 is the first complete fix, so please move or backport that argument canonicalization before the checkpoint.

PR #1944, opened 2026-07-29
Support sync worker methods on per-group executor threads

fzyzcjy:


guapisolo (approved):
Reviewed in the #1837 stack context. The wrapper-metadata issue is fixed in [#1949](https://github.com/radixark/miles/pull/1949/files) and remains resolved through the #2176 safe-merge checkpoint and current stack tip #2649. No blocking issues remain.

PR #1945, opened 2026-07-29
Retry a submit only when it provably never reached the server

fzyzcjy:


guapisolo (approved):
Reviewed in the #1837 stack context through safe-merge checkpoint #2176. Submit retries are limited to failures that prove the request never reached the worker, while ambiguous or post-send failures remain single-shot; no blocking issues found.

PR #1946, opened 2026-07-29
Long-poll from the client and tolerate a blip mid-poll

fzyzcjy:


miles/utils/workers/rpc/client/call.py
```diff
@@ -92,29 +100,46 @@ async def _submit_attempt(self, *, request: SubmitRequest, remaining: float) ->
         log_structured(logger.debug, op="submit", phase="accepted", **self._log_fields)
 
     async def _poll_until_done(self) -> CallStatusResponse:
-        expires_at = time.monotonic() + self._call_timeout_seconds
+        try:
+            return await retry_until_deadline(
+                self._poll_once,
+                total_seconds=self._call_timeout_seconds,
```
guapisolo, line 110: Codex review: Please reject non-finite `call_timeout_seconds` before entering this retry loop. With `NaN`, `retry_until_deadline` can never decide that the deadline has expired, so a pending call keeps polling indefinitely; I reproduced hundreds of polls until an outer cancellation.

miles/utils/workers/rpc/client/call.py
```diff
@@ -92,29 +100,46 @@ async def _submit_attempt(self, *, request: SubmitRequest, remaining: float) ->
...
-            **self._log_fields,
-            timeout_s=self._call_timeout_seconds,
-        )
-        raise TimeoutError(
-            f"{self._method_label} (call id {self._call_id}) still pending after {self._call_timeout_seconds}s"
-        )
+        except (TimeoutError, asyncio.TimeoutError) as e:
+            raise _CallStillPendingError(f"long poll timed out after {poll_seconds:.1f}s") from e
+        except (httpx.TransportError, RetryableResponseError) as e:
+            await asyncio.sleep(RETRY_INITIAL_DELAY_SECONDS)
```
guapisolo: Codex review: Please bound this sleep by `remaining`, or move the backoff into the deadline-aware retry helper. A transport error in the last <1s of the call budget still sleeps the full 1s; with a 50ms budget I measured about 1.0s before timeout.

PR #1947, opened 2026-07-29
Refuse a duplicate call id with 409 instead of running it twice

fzyzcjy:


guapisolo (approved):
Reviewed in stack context through safe-merge checkpoint #2176. PR #1947 rejects duplicate call IDs before execution while keeping the original call pollable; no actionable issues found.

PR #1948, opened 2026-07-29
Pin a client to one server process with a boot uuid

fzyzcjy:


guapisolo (approved):
Reviewed in the #1837 stack context through safe merge checkpoint #2176. With a frozen code version and no old-checkpoint or mixed-version rollback requirement, the boot-UUID fencing holds and I found no blocking issue.

PR #1949, opened 2026-07-29
Carry rich argument and result types and refuse ambiguous signatures

fzyzcjy:


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at deliver-1 within the requested scope. Deliver-2 changes address later integration requirements rather than a blocker for this op.

PR #1950, opened 2026-07-29
Support non-finite floats across the wire

fzyzcjy:


guapisolo (approved):
Reviewed in stack context through the safe-merge checkpoint #2176. The serialization design holds, and no end-to-end-backed blocker remains. LGTM.

PR #1951, opened 2026-07-29
Support worker env vars before importing the server

fzyzcjy:


guapisolo (commented):
[P1][Unresolved through #2649] `serve.py` writes the complete own/worker argv, computed environment values, and exec argv to stdout; #2446 also wires this entrypoint into the Helm production launch path. Please log only non-sensitive phase metadata, identities, or counts, and add a sentinel-secret test proving neither argv nor environment values reach the logs.

[P2][Fixed only after checkpoint by #2345] This remains broken at the #2176 checkpoint. #2345 removes the hidden inner CLI contract by moving both entrypoints onto the shared `--specs`/`--pool-id` parser, so #2345 or an equivalent correction must land with the checkpoint.

[P2][Unresolved through #2649] The exec chain still rebuilds `[sys.executable, "-m", ...]`, so Python flags such as `-O`, `-u`, `-I`, `-E`, and `-X` are dropped. Please preserve the interpreter-option prefix from `sys.orig_argv` while replacing only the module target, and cover one observable flag in a subprocess test.

PR #1952, opened 2026-07-29
Expire finished calls that were never retrieved

fzyzcjy:


guapisolo (approved):
Approving in stack context: #1952 fixes the unbounded retention of finished, unpolled RPC calls from #1942 by adding a finished-record TTL. The repair lands before safe-merge checkpoint #2176 and remains present through the checked #2649 tip.

PR #1953, opened 2026-07-29
refactor(examples): group examples into infra_features/ and experimental/

Shi-Dong:
## Summary

The top level of `examples/` had grown to 24 entries with no signal about which are
verified recipes, which are infrastructure plumbing, and which are unverified
experiments. This sorts them so the top level mostly holds training recipes people are
expected to run.

**New `examples/infra_features/`** — runtime/infra plumbing rather than a model recipe:
`low_precision`, `p2p_weight_transfer`, `train_infer_mismatch_helper`,
`true_on_policy`, `random_async`.

**Moved to `examples/experimental/`** (joining the existing `openenv`):
`DrGRPO`, `eval`, `eval_multi_task`, `formal_math`, `multi_agent`, `reproducibility`,
`search-r1`, `strands_sglang`, `tau-bench`.

**Merged**: `geo3k_vlm_multi_turn/` → `geo3k_vlm/multi_turn/`, so the two geo3k
recipes live together.

**Deleted**: `retool` (superseded by `retool_v2`) and `true_on_policy_vlm`, plus the
now-orphaned doc page `docs/examples/retool.md`.

**Kept at top level**: `fully_async`, `geo3k_vlm`, `lora`, `multi_lora`,
`on_policy_distillation`, `retool_v2`, `swe-agent`.

`examples/README.md` is rewritten to match, and now lists `lora`, `multi_lora`,
`p2p_weight_transfer`, `random_async` and `retool_v2` — all of which were missing from
the old index. It also fixes a stale `strands-agents` link (the directory is
`strands_sglang`).

## `fully_async` is deliberately left alone

A separate in-flight PR relocates `examples/fully_async/`, so this PR does not touch it.
`examples/fully_async/` and both `fully-async` doc pages are byte-identical to `main`,
and `examples/swe-agent/run-glm47-flash-agentic-async.py` keeps its original
`PYTHONPATH` entry — so the two changes should not conflict.

Worth knowing for whoever lands that move: `fully_async_rollout.py` is not purely
example code. It defines `generate_rollout_fully_async`, and there is no equivalent
under `miles/`. Two things outside the directory depend on it and will need updating
alongside any relocation:

* `examples/swe-agent/run-glm47-flash-agentic-async.py`, which puts the directory on
  `PYTHONPATH` and passes
  `--rollout-function-path fully_async_rollout.generate_rollout_fully_async`
* `docs/user-guide/fully-async.md`, a feature guide still in the nav, which documents
  that same flag

## Reference updates

[...]

PR #1954, opened 2026-07-29
refactor(examples)(temp): leave fully_async out of the reorg

Zhichenzzz:
Stacked on #1953 (`shi/260729-examples-reorg`). Restores every fully_async path to its state
on `main`, so the reorg no longer touches fully_async. Please squash this into #1953, or merge
it into that branch, whichever you prefer.

## Why hold fully_async back

**1. It collides with #1716.** That PR promotes the rollout module into the core library:

```
examples/fully_async/fully_async_rollout.py
  ├─ #1716 →  miles/rollout/fully_async_rollout.py           (into core)
  └─ #1953 →  examples/infra_features/fully_async/...        (stays in examples)
```

#1716 also rewrites `--rollout-function-path` from the bare module name
`fully_async_rollout.generate_rollout_fully_async` to the fully qualified
`miles.rollout.fully_async_rollout.generate_rollout_fully_async`, which removes the existing
dependency on the example directory being on `sys.path`. Six files are edited by both PRs, so
whichever lands second would conflict, or silently undo the other.

**2. Content was deleted with no replacement.** Nothing in the tree recreates these:

| Path | Lines | In #1953 |
|---|---|---|
| `docs/examples/fully-async.md` | 205 | deleted |
| `examples/fully_async/run_qwen3_30b_a3b_fully_async.py` | 176 | deleted |
| `examples/fully_async/run-qwen3-4b-fully_async.sh` | 142 | deleted |
| `examples/fully_async/README.md` | 45 | deleted |
| `examples/infra_features/fully_async/README.md` | 43 | added |

`docs/user-guide/fully-async.md` gained 2 net lines, all link rewrites, so none of the deleted
material was folded into it. That leaves the feature with no runnable recipe and no examples
docs page.

## What this PR does

Restores to their `main` state: `docs/examples/fully-async.md`,
`docs/user-guide/fully-async.md`, and all four files under `examples/fully_async/`. Drops
`examples/infra_features/fully_async/`. Restores the `examples/fully-async` navigation entry in
`docs/docs.json` and the Card in `docs/examples/index.md`.

It also repairs the references that would otherwise dangle once that directory is gone:

* `train_async.py`
* `examples/infra_features/random_async/README.md` and its rollout module
* `examples/README.md`, where the entry moves to the top level list with its path corrected

Verified that the fully_async paths are now byte identical to `main`, that no reference to
`infra_features/fully_async` remains anywhere in the repository, that `docs.json` still parses,
and that pre-commit passes on every touched file.

## Left alone deliberately

The `docs/examples/index.md` link to `/user-guide/fully-async` stays as you set it. That page
exists either way, and the reworded question reads better than the original.

## Suggested sequencing

Land #1716 first, since it is 11 files and unambiguous. Once it is in, moving what remains of
`examples/fully_async/` (a README and two launch scripts) under `infra_features/` is a small
independent change, with the rollout module already in core and out of the way.


Zhichenzzz:
Superseded by 84bf09c1e on the base branch, which does everything this PR did. Closing.

Verified against the current tip of `shi/260729-examples-reorg`:

* the fully_async paths are byte identical to `main`
* `examples/infra_features/fully_async/` is gone
* no reference to `infra_features/fully_async` remains in `train_async.py`,
  `examples/README.md`, or the `random_async` README and rollout module
* the `examples/fully-async` nav entry and the index Card are back

Rebasing this branch onto that tip leaves an empty diff, with the only conflict being a
line wrap in `train_async.py` where both sides make the same change.

Two things your revert caught that this PR missed, worth recording:

1. **`examples/swe-agent/run-glm47-flash-agentic-async.py`** builds `FULLY_ASYNC_DIR` from
   `"infra_features" / "fully_async"` as separate string literals, so my
   `grep infra_features/fully_async` never matched it. That one is executable code rather
   than prose, so it is the most consequential of the set.
2. **`docs/user-guide/training-script-walkthrough.md`** had a `/user-guide/fully-async` link
   I did not check.

You also fully reverted the `docs/examples/index.md` intro sentence and the "where to start"
link, where I had deliberately kept the reworded version. A clean full revert is the better
call, since it leaves no partial state to reason about.

The sequencing suggestion stands: land #1716 first, then moving what remains of
`examples/fully_async/` under `infra_features/` is a small independent change with the rollout
module already in core.

PR #1964, opened 2026-07-30
[TOP 2/N] Add batch-invariant scorer integration coverage

kaixih:
## Summary

Add one GPU integration test for the existing Miles true-on-policy scorer.

With Megatron batch-invariant mode enabled, the scorer keeps its existing path:

```text
TP full-vocab gather
-> padded-vocab truncation
-> torch.log_softmax
-> selected-token gather
```

The test uses BF16 logits with the DSV4 129280-token vocabulary. It verifies
that Miles reaches Megatron's registered BI `log_softmax` implementation and
that the same row scored alone or in a 17-row batch produces a bitwise-identical
selected-token logprob.

## Scope

- No runtime code or CLI changes.
- No duplicate BI kernel or custom op.
- DSV4 prefill-vs-trainer E2E parity remains a follow-up.

This is split 2/N from #1788 and is orthogonal to #1912.

## Validation

- New GPU integration test: passed on GB200 and in `stage-b-2-gpu-h200` CI.
- Focused local scorer/config tests: 59 passed.
- GitHub pre-commit and CPU matrix: passed.


kaixih:
Closing this test-only split because no additional Miles runtime integration is required. Megatron-LM's [`enable_batch_invariant_mode()`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/custom_layers/batch_invariant_kernels.py#L1072-L1085) globally registers the accelerator implementation of `aten::_log_softmax`; therefore Miles' existing `torch.log_softmax` scorer automatically uses the batch-invariant kernel under `--batch-invariant-mode`. This conclusion is also recorded in #1788.

PR #1966, opened 2026-07-30
[Bug Fix][OPD] Stop gradients through old-policy and teacher scores

ChenyangWangjohn:
## Motivation

OPD and PPO-style policy-gradient training treat teacher, reference-policy, pre-update policy, rollout-policy, and advantage tensors as fixed training inputs. Although standard scoring paths create these values without gradients, the training boundary should enforce that contract for custom and fused paths as well.

Without an explicit boundary, a live old-policy tensor can enter the PPO/GSPO importance ratio and retain an unnecessary graph alongside the differentiable current-policy branch.

## What's in this PR

- Detach `log_probs`, `rollout_log_probs`, `ref_log_probs`, and `teacher_log_probs` at the persistent rollout-data boundary.
- Defensively detach advantages and fixed score inputs at the policy-loss boundary while preserving gradients through current-policy logits.
- Keep detached teacher scores on their original device in persistent rollout data; device alignment is local to OPD computation.
- Reuse the detached values in PPO, OPSM, GSPO, TIS, KL loss, debug dumps, and mismatch metrics.
- Preserve custom TIS functions that intentionally do not consume trainer-scored log-probs; strict input checks remain on the built-in TIS implementation.
- Add non-leaf graph-retention and policy-loss backward coverage, including the reference-policy KL path and custom TIS compatibility.

The OPD objective, reverse-KL sign, clipping, coefficients, and current-policy forward values are unchanged.

## Validation

- Reran the official Qwen3-8B student / Qwen3-32B teacher OPD recipe on an NVIDIA H200 GPU cluster with the fixed gradient boundary; training completed successfully.
- Regression coverage exercises persistent non-leaf tensors and the real policy-loss backward path for trainer-scored, rollout-scored, reference, and teacher inputs.


Shi-Dong (approved):
LGTM

maocheng23 (approved):

PR #1967, opened 2026-07-30
[Bug Fix] Always release rollout-engine broadcast lock on failure

ChenyangWangjohn:
## Motivation

`_update_weight_implementation` acquires a shared rollout-engine lock before starting a distributed weight broadcast. Previously, an exception during broadcast setup or while waiting for rollout-engine completion skipped the release.

Because `Lock.acquire()` is polled until it succeeds, a leaked lock makes the next weight sync wait forever and hides the original failure.

## What's in this PR

- Wrap broadcast submission and completion in `try/finally`.
- Release the rollout-engine lock after every successful acquire.
- Preserve success-path ordering:
  - Converted tensors are cleared only after broadcast completion.
  - The progress bar advances only after a completed sync.
- Preserve the original broadcast exception when lock release succeeds.
- Cover successful sync, lock contention, setup failure, engine-side failure, retry after failure, and argument forwarding.

## Validation

- Reran the official fully-async training recipe on an NVIDIA H200 GPU cluster through repeated weight synchronizations; training completed successfully.
- Failure-path coverage verifies that both submission and engine-side exceptions release the lock and remain visible to the caller.

### DIFF
--- a/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
+++ b/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
@@ -105,21 +105,24 @@ class UpdateWeightFromDistributed(DistBucketedWeightUpdateMixin):
     def _update_weight_implementation(
         self, converted_named_tensors: list[tuple[str, torch.Tensor]], pbar: tqdm | None = None
     ) -> None:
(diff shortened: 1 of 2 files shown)


Shi-Dong (approved):
Nice fix. LGTM.

maocheng23 (approved):

PR #1968, opened 2026-07-30
[Bug Fix] Reduce distributed min/max metrics as extrema

ChenyangWangjohn:
## Motivation

`gather_log_data` averaged every gathered scalar across effective DP/CP ranks. That is correct for ordinary mean metrics, but wrong for rank-local extrema: averaging local maxima under-reports the global maximum, while averaging local minima over-reports the global minimum.

## What's in this PR

- Add per-key `mean`, `min`, and `max` reduction to `gather_log_data`, with `mean` as the backward-compatible default.
- Derive the multi-turn reduction policy from `_min` and `_max` metric suffixes so future extrema cannot silently fall back to averaging.
- Raise a clear error when gathered metric schemas differ across ranks.
- Cover reducer semantics and the actual `log_multi_turn_data` wiring.

Metric names and existing mean call sites are unchanged. The implementation adds no collective; it only changes root-rank reduction of values that have already been gathered.

## Validation

- Reran the repository's distributed training recipe on an NVIDIA H200 GPU cluster and confirmed that multi-turn minima and maxima reflect global rank extrema while mean metrics retain their existing behavior.
- Regression coverage exercises both the reducer and its production call site.


miles/backends/training_utils/log_utils.py
```diff
@@ -328,7 +369,10 @@ def log_multi_turn_data(rollout_id: int, args: Namespace, rollout_data: RolloutB
                 log_dict["multi_turn_metric/round_number_mean"] = np.mean(round_number_array)
                 log_dict["multi_turn_metric/round_number_max"] = np.max(round_number_array)
                 log_dict["multi_turn_metric/round_number_min"] = np.min(round_number_array)
-        gather_log_data("multi_turn", args, rollout_id, log_dict)
+        reduction_by_key = {
+            key: "max" if key.endswith("_max") else "min" for key in log_dict if key.endswith(("_max", "_min"))
+        }
```
Shi-Dong: Wouldn't it be better to set a default for reduction key inside `_reduce_gathered_log_dicts`? E.g. if a metric's name ends with `_min`, then the default reduction becomes `min`.

Shi-Dong (approved):
LGTM with minor comment.

miles/backends/training_utils/log_utils.py
ChenyangWangjohn: Good Suggestion. I moved the default reduction policy into _reduce_gathered_log_dicts: keys ending in _min or _max now default to the corresponding extrema reduction, while other keys continue to use mean. Explicit reduction_by_key entries still take precedence. This also removes the per-caller wiring from log_multi_turn_data, and I added tests for both suffix inference and explicit overrides.

miles/backends/training_utils/log_utils.py
```diff
@@ -22,18 +22,70 @@
...
+    for rank, rank_metrics in enumerate(gathered_log_dict[1:], start=1):
+        if rank_metrics.keys() != expected_keys:
+            raise ValueError(
+                f"Metric keys differ across ranks: rank 0={list(expected_keys)}, "
+                f"rank {rank}={list(rank_metrics.keys())}."
+            )
+
+    reduced = {}
+    for key in expected_keys:
+        values = [rank_metrics[key] for rank_metrics in gathered_log_dict]
```
maocheng23: This process would be tooo broad. We may not only use the key name to determine how to reduce values.

maocheng23 (changes requested):

miles/backends/training_utils/log_utils.py
ChenyangWangjohn: Agreed. I restored mean as the default for the generic helper and made log_multi_turn_data explicitly pass min or max only for its known extrema metrics. Existing callers that do not provide a reduction policy retain their previous behavior.

maocheng23 (approved):

PR #1969, opened 2026-07-30
[OPD] Accelerate external Top-K scoring with position-blocked candidate unions

ChenyangWangjohn:
## Motivation

The official SGLang Top-K OPD recipe becomes impractical for sufficiently long responses. In the validated long-thinking settings, the legacy scorer could not finish the first training step within a 600-second scoring deadline.

The bottleneck is the representation used for cross-model candidate scoring. OPD needs at most `K` candidate log-probabilities at each of `T` response positions, so the useful result has shape `[T, K]`. The legacy implementation instead unions the candidate IDs from the entire response into a set `U`, asks SGLang to evaluate every ID in `U` at every position, and receives a `[T, U]` candidate table:

```text
useful result [T, K]
        |
        v
response-wide candidate union U
        |
        v
materialize and transport [T, U]
        |
        v
discard all but [T, K]
```

The request sends the `U` candidate IDs once, but the response contains `T x U` `(logprob, token_id)` entries. Since `U` can approach `T x K`, candidate extraction and transport can grow toward `O(T^2 K)`, even though only `T x K` entries survive. This inflates temporary results, device-to-host transfer, JSON serialization, network traffic, and client parsing; it does not describe the Transformer forward complexity.

Position blocking trades one oversized scoring request for multiple smaller requests. Those requests also need explicit concurrency and end-to-end time bounds, while incomplete or misaligned scores must be rejected before they can enter training.

This PR replaces the unbounded response-wide representation with a bounded position-blocked representation and adds those operational safeguards around the external scorer. It preserves the exact `[T, K]` scores and existing OPD objective, requires no SGLang API or trainer-backend change, and reduces worst-case candidate-response growth to `O(T B K)`. In the validated non-thinking runs, the same change reduced step time by 2.7x for student-selected candidates and 3.1x for teacher-selected candidates; previously timing-out thinking settings also completed.

### 1. Why this PR uses position blocks

SGLang accepts one candidate-ID set per scoring request and evaluates it at every scored position. It cannot directly represent the different `K` candidates needed at each of `T` positions without a new position-aware scoring contract.

Adding that contract would require coordinated SGLang request and runtime changes. This PR keeps the SGLang and trainer interfaces unchanged and bounds the existing scoring representation inside Miles by grouping positions into blocks.

### 2. Replace the global union with position blocks

The response is divided into blocks of `B` consecutive positions. Each block keeps the original position-local candidate rows, but builds a union using only that block:

```text
[T, K] position-local candidates
        |
        +--> positions [0, B)     -> U0 -> score [B, U0] -> compact [B, K]
        +--> positions [B, 2B)    -> U1 -> score [B, U1] -> compact [B, K]
        +--> ...
                                                        |
                                                        v
                                                restored [T, K]
```

For a block covering positions `[start, end)`, Miles:

1. Computes `U_block = union(C_start, ..., C_end-1)`.
2. Sends the input prefix ending at `end`, sets the scored response window to the `B` block positions, and sends `token_ids_logprob=U_block`.
3. Receives `[B, U_block]` candidate log-probabilities from SGLang.
4. Selects only the original `C_t` entries from each position's returned row.
5. Compacts the block back to `[B, K]` and concatenates all blocks into `[T, K]`.

The final candidate IDs and log-probabilities are unchanged. The existing weighting modes and OPD reward calculation receive the same position-local structure as before; only the external scoring and transport strategy changes.

### 3. Make external scoring bounded and fail loudly

Position blocking issues multiple smaller requests, so Miles reuses its shared HTTP client, caps per-process concurrency, and applies one end-to-end deadline across queueing, retries, backoff, and transport. Each response is limited to the requested token window and is rejected before training if any score is missing, non-finite, or token-misaligned.

These controls are exposed through `--opd-scoring-timeout`, `--opd-scoring-max-inflight`, `--opd-scoring-retries`, and `--opd-top-k-scoring-block-size`.

## SGLang compatibility

The `only-teacher` path uses SGLang's existing `token_ids_logprob` API to rescore teacher-selected IDs on the student. This was validated together with [sgl-project/sglang#32859](https://github.com/sgl-project/sglang/pull/32859), which prevents valid host-list result rows from crashing SGLang response normalization. That fix does not change the scoring API or returned values.

## Scope

This PR does not change the `only-student` or `only-teacher` candidate semantics, Top-K reward weighting, OPD objective, gradient application, trainer backend, model-parallel topology, or SGLang API.

## Validation

The official Qwen3-8B student / Qwen3-32B teacher OPD recipe was rerun on an NVIDIA H200 GPU cluster with Top-K 16, block size 32, and up to eight scoring requests in flight.

| Strategy and template mode | Legacy response-wide union | Position blocks |
| --- | ---: | ---: |
| `only-student`, non-thinking | 6/6 steps, 289 s/step | 6/6 steps, 108 s/step (2.7x faster) |
| `only-teacher`, non-thinking | 6/6 steps, 285 s/step | 6/6 steps, 93 s/step (3.1x faster) |
| `only-student`, thinking | Failed in step 0 at the 600 s scoring deadline | Completed two test steps |
| `only-teacher`, thinking | Failed in step 0 at the 600 s scoring deadline | Completed 2/2 test steps, 492 s/step |

Thinking mode produced substantially longer responses in these runs. These results establish behavior for the tested layout and deadline; they are not a universal latency bound.

Regression coverage includes both Top-K strategies, all three reward-weight modes, position-local restoration, stable student-version assembly, response-window and token-alignment validation, and bounded deadline and retry behavior.

### DIFF
--- a/docs/advanced/on-policy-distillation.md
+++ b/docs/advanced/on-policy-distillation.md
@@ -12,6 +12,10 @@ On-policy distillation (OPD) trains a student model on its own rollouts while us
 | `--opd-log-prob-top-k` | Number of top-k tokens retained for the Rethinking OPD token reward. `0` uses sampled-token OPD; `16` matches the paper recipe default. |
 | `--opd-top-k-strategy` | Top-k token set strategy: `only-student`, `only-teacher`, `intersection`, `union`, or `xor`. |
 | `--opd-reward-weight-mode` | Weighting scheme for top-k rewards: `student_p`, `teacher_p`, or `none`. |
+| `--opd-top-k-scoring-block-size` | Response positions grouped into one arbitrary-ID scoring request for `only-student` and `only-teacher` (default: `32`). `0` restores the legacy response-wide candidate union. |
+| `--opd-scoring-timeout` | Total deadline in seconds for one external scoring request, including in-flight queueing, retries, and transport (default: `600`). |
+| `--opd-scoring-max-inflight` | Maximum concurrent external scoring requests per process (default: `8`). `0` disables the bound. |
+| `--opd-scoring-retries` | Number of retries for a failed external request or a mixed-version blocked student-scoring attempt (default: `1`). Each external request retains its own total deadline; `0` fails after the first attempt. |
 | `--opd-teacher-load` | Path to teacher Megatron checkpoint. **Required** when `--opd-type=megatron`, **must not be set** when `--opd-type=sglang`. |
 | `--opd-teacher-ckpt-step` | Optional checkpoint step for teacher model. |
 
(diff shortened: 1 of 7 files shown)


maocheng23:
Overlap with PR 1298?

PR #1970, opened 2026-07-30
Give every launch a machine-readable run uuid

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in the #1837 stack context through safe-merge checkpoint #2176. The launch UUID contract is consumed by #1998 for deterministic session-server identities and remains intact at the checked tip; no blocking issues from me.

PR #1971, opened 2026-07-30
Rename control_server to api_server for general purpose

fzyzcjy:
Part of #1837


guapisolo:
Blocking in stack context: #1971 removes the v0.1.0 `--control-server-port` / `control_server_port` boundary without a deprecated alias or config migration. It is still unfixed at #2176 (`97bba455`) and through #2649; #2307 only changes the new port default, so please preserve and normalize the legacy spelling, with explicit conflict handling, before the safe-merge boundary.

PR #1973, opened 2026-07-30
Avoid session server input using full args

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/rollout/session/config.py
+++ b/miles/rollout/session/config.py
@@ -1,0 +1,57 @@
+from typing import Any
+
+from miles.utils.pydantic_utils import FrozenStrictBaseModel
+
+
+class SessionServerConfig(FrozenStrictBaseModel):
+    host: str
+    port: int
+    instance_id: str | None
+    backend_url: str
+    timeout: float | None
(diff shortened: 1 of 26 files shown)


yueming-yuan (approved):

miles/rollout/session/config.py
```diff
@@ -0,0 +1,57 @@
...
+    apply_chat_template_kwargs: dict[str, Any] | None
+    use_rollout_routing_replay: bool
+    use_rollout_indexer_replay: bool
+    sglang_speculative_algorithm: str | None
+    num_layers: int | None
+    moe_router_topk: int | None
+    save_debug_trajectory_data: str | None
+    lora_rank: int
+    lora_adapter_path: str | None
+    use_session_server: str | None
```
guapisolo: `--use-session-server` uses `const=True`, so the documented bare form reaches this field as `True` and Pydantic rejects `str | None` before launch. This is fixed only in #2425, after safe merge checkpoint #2176; please either include #2425 or accept/normalize the boolean compatibility value here.

PR #1975, opened 2026-07-30
Add launch_bound_subprocess with binding a child's lifetime to its parent

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/process_utils.py
+++ b/miles/utils/workers/process_utils.py
@@ -1,0 +1,46 @@
+import atexit
+import ctypes
+import os
+import signal
+import subprocess
+import sys
+from functools import partial
+
+_PR_SET_PDEATHSIG = 1
+
+_LIBC = ctypes.CDLL(None, use_errno=True) if sys.platform == "linux" else None
+
+
+def launch_bound_subprocess(argv: list[str], *, envs: dict[str, str]) -> subprocess.Popen:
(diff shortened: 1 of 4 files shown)


yueming-yuan (approved):

miles/utils/workers/process_utils.py
```diff
@@ -0,0 +1,46 @@
...
+_LIBC = ctypes.CDLL(None, use_errno=True) if sys.platform == "linux" else None
+
+
+def launch_bound_subprocess(argv: list[str], *, envs: dict[str, str]) -> subprocess.Popen:
+    parent_pid = os.getpid()
+    process = subprocess.Popen(
+        argv,
+        env={**os.environ, **envs},
+        start_new_session=True,
+        preexec_fn=partial(_set_parent_death_signal, expected_parent_pid=parent_pid) if _LIBC is not None else None,
```
guapisolo: `launch_bound_subprocess()` is later called from `CommandActor`, whose Ray worker process already has runtime threads. CPython documents `preexec_fn` as unsafe in the presence of threads because the child can deadlock before `exec`; in that case `Popen()` never returns and the server readiness timeout cannot start. This remains unchanged at #2176. Could we move the `prctl` setup into a small execed launcher/supervisor and add a threaded-launch regression?

PR #1976, opened 2026-07-30
Launch the sgl and miles routers as command-line subprocesses

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -1,17 +1,22 @@
 import logging
 import multiprocessing
 import random
+import sys
 import uuid
 
-from sglang_router.launch_router import RouterArgs
-
+from miles.backends.sglang_utils.router_args_utils import compute_sglang_router_args, router_args_to_argv
 from miles.rollout.session.config import compute_session_server_config
 from miles.rollout.session.server import run_session_server
 from miles.router.config import compute_miles_router_config
-from miles.router.router import run_router as run_miles_router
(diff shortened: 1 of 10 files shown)


yueming-yuan (approved):

miles/ray/rollout/router_manager.py
```diff
@@ -63,14 +62,7 @@ def start_router(args, *, has_pd_disaggregation: bool = False, force_new: bool =
...
 
-    # spawn (not fork): the child must not inherit threads/finalizers from this
-    # Ray actor (e.g. wandb's service thread), which deadlock a forked child.
-    process = multiprocessing.get_context("spawn").Process(
-        target=run_router,
-        args=(router_args,),
-    )
-    process.daemon = True
-    process.start()
+    process = process_utils.launch_bound_subprocess(launch_argv, envs={})
```
guapisolo, line 70: [P1] On Linux, `launch_bound_subprocess()` passes a Python `preexec_fn` to `Popen`. This call runs after Ray/tracking have started threads, and CPython warns that `preexec_fn` can deadlock before `exec`; in that case `Popen` never returns, so the 30s readiness timeout never starts. Could we preserve `PR_SET_PDEATHSIG` via an exec-first trampoline (or equivalent) without running Python `preexec_fn` in the multithreaded parent? I traced this through #2176 and #2649; #1988 only moves the same helper call into a Ray `CommandActor`, so it remains unresolved.

PR #1977, opened 2026-07-30
Refactor session servers to be launched as command-line subprocesses

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -1,12 +1,10 @@
 import logging
-import multiprocessing
 import random
 import sys
 import uuid
 
 from miles.backends.sglang_utils.router_args_utils import compute_sglang_router_args, router_args_to_argv
 from miles.rollout.session.config import compute_session_server_config
-from miles.rollout.session.server import run_session_server
 from miles.router.config import compute_miles_router_config
 from miles.utils.http_utils import (
     _wrap_ipv6,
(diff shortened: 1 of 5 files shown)


yueming-yuan (approved):

miles/ray/rollout/router_manager.py
```diff
@@ -129,10 +125,8 @@ def start_session_server(args):
         config = compute_session_server_config(
             args, host=ip, port=port, instance_id=instance_id, backend_url=router_url
         )
-        process = multiprocessing.get_context("spawn").Process(target=run_session_server, args=(config,))
-        process.daemon = True
-        process.start()
-        processes.append((port, process))
+        launch_argv = [sys.executable, "-m", "miles.rollout.session.server", *config_to_argv(config)]
+        processes.append((port, process_utils.launch_bound_subprocess(launch_argv, envs={})))
```
guapisolo, line 133: [P1][unresolved at #2176 and through #2649] This drops the explicit multiprocessing `spawn` boundary and routes startup through `launch_bound_subprocess()`, whose Linux path runs a Python `preexec_fn` after fork. In a threaded Ray process this can deadlock before `exec`; #1988 only moved the call into `CommandActor`, and the same unsafe boundary remains. Please keep this path spawn-safe, or remove Python `preexec_fn` execution after fork before migrating it.

miles/ray/rollout/router_manager.py
```diff
@@ -129,10 +125,8 @@ def start_session_server(args):
         config = compute_session_server_config(
             args, host=ip, port=port, instance_id=instance_id, backend_url=router_url
         )
-        process = multiprocessing.get_context("spawn").Process(target=run_session_server, args=(config,))
-        process.daemon = True
-        process.start()
-        processes.append((port, process))
+        launch_argv = [sys.executable, "-m", "miles.rollout.session.server", *config_to_argv(config)]
```
guapisolo, line 132: [P2][unresolved at #2176 and through #2649] Reconstructing the child command as `[sys.executable, "-m", ...]` drops interpreter flags that multiprocessing spawn preserved, including `-O`, `-I`, `-E`, `-W`, and supported `-X` options. Please preserve the multiprocessing-compatible interpreter option prefix and cover one observable flag in a real child process.

PR #1978, opened 2026-07-30
Add worker specs describing how each worker kind is launched

fzyzcjy:
Part of #1837


guapisolo (approved):
Approving in stack context: #1978 moves worker launch knowledge into declarative specs, giving the later worker-manager PRs a common boundary before safe-merge checkpoint #2176. That boundary remains in use through the checked #2649 tip.

PR #1979, opened 2026-07-30
Add CommandActor that binds its lifetime to a launched subprocess

fzyzcjy:
Part of #1837


guapisolo (approved):
Checked through the #2176 safe-merge checkpoint. The raw command/API-key logging remains, but is accepted as non-blocking under our in-house threat model.

PR #1980, opened 2026-07-30
Build the dashboard engine topology on the driver instead of the actor

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in the #1837 stack context: moving topology ownership to the driver preserves the existing GPU, multi-node, and recovery behavior. No active findings before safe merge checkpoint #2176.

PR #1981, opened 2026-07-30
Delete the external-engine path from SGLangEngine

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context: PR #1981’s orphan external-check computation and stale tuple-unpacking test helper are both fixed by PR #2096, before #2176, and the corrected files remain unchanged through checked tip #2649.

PR #1982, opened 2026-07-30
Add the sglang server argv boundary with a mandatory production roundtrip

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/sglang_utils/server_args_utils.py
+++ b/miles/backends/sglang_utils/server_args_utils.py
@@ -1,0 +1,40 @@
+import argparse
+
+from sglang.srt.server_args import ServerArgs
+
+from miles.utils.workers.argv_utils import render_cli_argv
+
+
+def server_args_to_argv(server_args: ServerArgs) -> list[str]:
+    return render_cli_argv(
+        server_args,
+        make_parser=_make_cli_parser,
+        from_parsed=ServerArgs.from_cli_args,
+        required_argv=_required_argv(server_args),
+    )
+
+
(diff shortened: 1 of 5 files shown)


miles/backends/sglang_utils/server_args_utils.py
```diff
@@ -0,0 +1,40 @@
+import argparse
+
+from sglang.srt.server_args import ServerArgs
+
+from miles.utils.workers.argv_utils import render_cli_argv
+
+
+def server_args_to_argv(server_args: ServerArgs) -> list[str]:
+    return render_cli_argv(
```
fzyzcjy, line 9: [P2] Catch parser exits at the server argv boundary

When a per-group SGLang override survives direct `ServerArgs` construction but violates an argparse type or `choices` constraint, this mandatory roundtrip raises `SystemExit`. Deliver-1 (`deec6fce05`) still calls it inside `RayWorkerManager`, whose startup rollback catches only `Exception`; the manager actor can therefore die without Miles reporting the failure or cleaning up partially started cells. Deliver-2 fixes this in `1f5c71148a`; move only its `SystemExit`-to-`AssertionError` wrapper into this op's nested `parse()` (`shlex` is needed only for the rendered-argv diagnostic), rather than depending on the later renderer refactor.

fzyzcjy (commented):
The server argv roundtrip leaves argparse's process-terminating `SystemExit` uncontained. This remains in deliver-1 and can kill the worker-manager actor while bypassing startup rollback.

PR #1983, opened 2026-07-30
Dissolve RayActor and probe workers through NodeProbeMixin

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context: #1985 (f9a8ed45) fixes #1983's missing CommandActor import, and the fix remains intact through #2649, including safe merge checkpoint #2176.

PR #1984, opened 2026-07-30
Bracket ipv6 addressing where it is built instead of in engine init

fzyzcjy:
Part of #1837


guapisolo (approved):
#1984 correctly moves IPv6 bracketing to the address-construction boundary, while the launch path still strips brackets before socket binding. The contract remains intact through the checked #2176 checkpoint; LGTM.

PR #1985, opened 2026-07-30
Launch sglang engines as command subprocesses under CommandActor

fzyzcjy:
Part of #1837


fzyzcjy (commented):
GPU localization now depends on the wrong process environment in supported multi-node layouts, and optional SGLang credentials are copied into logged process command lines. Both regressions remain in safe deliver-1.

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -87,87 +62,45 @@ def build_server_url(host: str, port: int) -> str:
...
+    args,
+    *,
+    rank: int,
+    worker_type: str,
+    base_gpu_id: int,
+    sglang_overrides: dict,
+    num_gpus_per_engine: int,
+    addr_and_ports: dict,
+) -> EngineLaunchPlan:
+    server_args_dict = _compute_server_args(
```
fzyzcjy, line 77: [P1] Resolve GPU IDs in the target actor environment

When the driver and rollout host use different visibility masks—the existing `examples/lora/run-qwen2.5-3B-megatron-lora-disaggregated-multi-node.sh` permits, for example, head `CUDA_VISIBLE_DEVICES=0` and rollout `4,5`—this pre-dispatch call runs `_to_local_gpu_id` against the driver's mask, so the valid rollout placement base `4` is rejected instead of becoming target-local `0`. The previous actor-side call saw the rollout host's mask. Resolve the ID in the target actor; safe deliver-1 only moves this calculation to the unscheduled `RayWorkerManager`, so the blocker remains.

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -87,87 +62,45 @@ def build_server_url(host: str, port: int) -> str:
...
+        base_gpu_id=base_gpu_id,
+        engine_info_bootstrap_port=addr_and_ports.get("engine_info_bootstrap_port"),
+        sglang_overrides=sglang_overrides,
+        num_gpus_per_engine=num_gpus_per_engine,
+    )
 
-        logger.info(f"Simulating crash on engine {self.server_host}:{self.server_port}...")
-        self.shutdown()
+    launch_args = {**server_args_dict, "host": server_args_dict["host"].strip("[]")}
+    cmd = shlex.join([sys.executable, "-m", "sglang.launch_server", *server_args_to_argv(launch_args)])
```
fzyzcjy, line 93: [P1] Keep credentials out of the engine command line

When an engine is configured with an `api_key` or `ssl_keyfile_password`, `server_args_to_argv` places the plaintext value in this command. `CommandActor.run` logs the entire command at INFO and the child exposes it through `/proc/*/cmdline` on every engine host; unlike the former in-process launch, this leaks credentials supplied through the SGLang config even when SGLang logging is disabled. Pass sensitive fields through a non-argv channel and redact the actor log; safe deliver-1 retains both paths, so this is still a blocker.

PR #1986, opened 2026-07-30
Forbid positional arguments when constructing ray actors

fzyzcjy:
Part of #1837


guapisolo:
**P1 — unresolved at safe checkpoint #2176 and current tip #2649.** This PR establishes a kwargs-only Ray actor-constructor contract, but two core actors still keep a positional channel: `_PrometheusCollector` uses `.remote(args)` with `__init__(self, args)`, and `_HttpPosterActor` uses `.remote(per_actor_conc)` with `__init__(self, concurrency)`. No later incremental PR repairs either pair. Please make both constructors keyword-only and pass `args=args` / `concurrency=per_actor_conc` before treating #2176 as the safe merge checkpoint.

PR #1987, opened 2026-07-30
Move PortAllocator into miles.utils.workers

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: this is an ownership-only move with no behavior change, and all in-repo imports are migrated. It sits before the safe merge checkpoint at #2176. LGTM.

PR #1990, opened 2026-07-30
Move sglang config resolution into sglang_config.py

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: the stale `_resolve_sglang_config` test calls introduced here are fixed by #2108 before safe-merge checkpoint #2176, and the fix remains intact through checked tip #2649. LGTM.

PR #1991, opened 2026-07-30
Cover sglang config parsing with unit tests through resolve_sglang_config

fzyzcjy:
Part of #1837


guapisolo (approved):
Codex stack review: the resolution-coverage gap in #1991 is closed by #1992, #2063, and #2100. I verified the restored contracts at safe-merge checkpoint #2176 and through the checked tip #2649.

PR #1992, opened 2026-07-30
Resolve the sglang config at construction time instead of mid-run

fzyzcjy:
Part of #1837


guapisolo (approved):
Looks good in stack context. Moving model resolution into `resolve_sglang_config` keeps startup behavior intact, validates every model before any router starts, and is present at safe merge checkpoint #2176.

PR #1993, opened 2026-07-30
Parse the raw sglang config into a dedicated pydantic model

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context: #1994 (a0e8a39a) replaces #1993's frozen eval-config mutation with model_copy transformations, and the fix remains intact through #2649, including safe merge checkpoint #2176.

PR #1994, opened 2026-07-30
Convert the resolved sglang config to pydantic models

fzyzcjy:
Part of #1837


guapisolo (approved):
The stale ModelConfig.resolve(args) caller is fixed in #2108, and that repair remains present through the checked #2176 checkpoint. The remaining zero-value validation gap is non-blocking; LGTM in stack context.

PR #1995, opened 2026-07-30
Move needs_offload and overrides into sglang config resolution

fzyzcjy:
Part of #1837


guapisolo (approved):
This cleanly moves stable per-group deployment decisions into config resolution while preserving GPU ordering, offload classification, and override precedence. Reviewed at `d42a9e3` with #2176 as the safe-merge boundary; no actionable findings.

PR #1996, opened 2026-07-30
Drop external router and fixed router ip mode

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -23,22 +23,10 @@ logger = logging.getLogger(__name__)
 _SERVER_READY_TIMEOUT_SECS = 120
 
 
-def start_router(args, *, has_pd_disaggregation: bool = False, force_new: bool = False) -> tuple[str, int]:
-    """Start sgl router or miles router and return (router_ip, router_port).
-
-    If ``args.sglang_router_ip`` is already set and ``force_new`` is False,
-    skip launching and return the existing values.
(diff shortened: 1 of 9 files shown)


miles/ray/rollout/router_manager.py
```diff
@@ -18,22 +18,10 @@
...
+def start_router(args, *, has_pd_disaggregation: bool = False) -> tuple[str, int]:
+    """Start sgl router or miles router and return (router_ip, router_port)."""
     router_ip = _wrap_ipv6(get_host_info()[1])
-    if force_new:
-        router_port = find_available_port(random.randint(3000, 4000))
-    else:
-        router_port = args.sglang_router_port
-        if router_port is None:
-            router_port = find_available_port(random.randint(3000, 4000))
+    router_port = find_available_port(random.randint(3000, 4000))
```
guapisolo, line 29: [P1][unresolved through #2649; not fixed at #2176] Later stack PRs partially repair this: #2051 makes router launch consume the pinned `PortInfo`, #2053 removes `--sglang-router-port` from the external-router ban, and #2096 covers `base + model_idx`. However, static router ports still bypass the node-local availability/reservation check, while readiness only TCP-connects, so a stale listener can make startup accept the wrong process. Please fail before launch if any pinned `base + i` port is occupied (or reserve and verify ownership), and update `training-backend.md` so only `--sglang-router-ip` is described as removed.

PR #1997, opened 2026-07-30
Add worker specs for the sglang and miles routers

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: #2006 completes the worker-spec context contract, and #2176 completes the trainer/controller migration. Both fixes are present at safe merge checkpoint #2176 and remain through #2649.

PR #1998, opened 2026-07-30
Derive session server instance ids from the run uuid and index

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed [PR #1998 changes](https://github.com/radixark/miles/pull/1998/files) in stack context: the run UUID and stable instance index are used consistently by the session-server launch config and tracer metadata, with no repair dependency before #2176. The same derivation remains present through checked tip #2649.

PR #1999, opened 2026-07-30
Implement a worker spec for the session server

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/worker_spec.py
+++ b/miles/utils/workers/worker_spec.py
@@ -50,9 +50,23 @@ class BaseWorkerSpec(FrozenStrictBaseModel):
     scheduling: SchedulingSpec
 
 
-class LaunchCommandContext(FrozenStrictBaseModel):
+class HostAndPort(FrozenStrictBaseModel):
     host: str
-    ports: dict[str, int]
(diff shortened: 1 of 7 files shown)


miles/utils/workers/worker_spec.py
```diff
@@ -50,9 +50,23 @@ class BaseWorkerSpec(FrozenStrictBaseModel):
...
+
+
+# dict key: name
+NamedHostAndPorts = dict[str, HostAndPort]
+
+
+class LaunchCommandContext(FrozenStrictBaseModel):
+    cell_index: int
+    self_addrs: NamedHostAndPorts
+    spec_addrs: dict[str, list[NamedHostAndPorts]]
```
guapisolo, line 69: Could we rename this to `pool_addrs` or `pool_id_to_addrs` and update its callers? This map is keyed by pool IDs (for example, `compute_router_pool_id(0)`), so `spec_addrs` treats a spec as an identity and conflicts with `.claude/rules/pool-cell-worker-names.md`. I checked the stack through safe-merge checkpoint #2176 and the checked #2649 tip; no later PR repairs this naming.

PR #2000, opened 2026-07-30
Use SGLANG_DG_CACHE_DIR_PER_PROCESS instead of hand-built cache dirs

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context through safe-merge checkpoint #2176 and checked tip #2649. The SGLang-owned per-process DeepGEMM cache isolation is correctly wired; no actionable findings.

PR #2001, opened 2026-07-30
Delete the unused get_base_gpu_id function

fzyzcjy:
Part of #1837


guapisolo (approved):
Codex review: `base_gpu_id` now comes exclusively from the actual Ray placement, and all current callers already pass it explicitly. Removing the unreachable rank-based fallback looks good in the #1837 stack context before safe-merge checkpoint #2176.

PR #2002, opened 2026-07-30
Stop passing a random seed and let each sglang engine pick its own

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/sglang_utils/sglang_engine.py
+++ b/miles/backends/sglang_utils/sglang_engine.py
@@ -106,7 +106,8 @@ def _compute_server_args(
     kwargs = {
         "model_path": args.hf_checkpoint,
         "trust_remote_code": True,
-        "random_seed": args.seed + rank,
+        # NOTE: do not pass random seed and let SGLang pick random ones
+        # "random_seed": args.seed + rank,
         # memory
         "enable_memory_saver": args.offload_rollout,
         # distributed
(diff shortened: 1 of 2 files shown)


guapisolo (commented):
One blocking reproducibility issue.

miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -110,7 +110,8 @@ def _compute_server_args(
     kwargs = {
         "model_path": args.hf_checkpoint,
         "trust_remote_code": True,
-        "random_seed": args.seed + rank,
+        # NOTE: do not pass random seed and let SGLang pick random ones
```
guapisolo, line 109: **[P1] Preserve a reproducible engine RNG until every path is request-seeded**

`sampling_seed` bypasses the engine RNG only in SGLang's regular deterministic sampler. The live `sglang-miles` branch still generates EAGLE verification coins with engine-global `torch.rand` ([source](https://github.com/sgl-project/sglang/blob/a6100a0228e281e34b8320be61f95ae00859db03/python/sglang/srt/speculative/eagle_utils.py#L720-L745)); upstream [#30822](https://github.com/sgl-project/sglang/pull/30822) is not in that branch and explicitly leaves draft-side rejection-sampling RNG unseeded. Letting SGLang choose a fresh engine seed therefore makes non-greedy speculative rollouts differ across otherwise identical restarts despite a fixed request seed. Please retain the deterministic per-engine seed here until all production RNG consumers are request-seeded, or gate this behavior to configurations where that invariant holds and cover it end to end.

PR #2003, opened 2026-07-30
Pass node_rank instead of rank when computing the engine launch command

fzyzcjy:
Part of #1837


guapisolo (approved):
Stack-verified: #2063 removes the obsolete global-rank alignment guard, and #2096 repairs the stale test caller and adds cross-cell coverage. Both fixes land before #2176 and remain intact through #2649.

PR #2004, opened 2026-07-30
Extract compute_inference_engine_env_vars and forbid compute_server_args defaults

fzyzcjy:
Part of #1837


guapisolo (approved):
The missing miles.rollout.session.ports import exposed by #2004 is removed in #2006 before #2176, and the repair remains present through the checked #2649 tip. LGTM in stack context.

PR #2005, opened 2026-07-30
Derive the engine api key from the model config

fzyzcjy:
Part of #1837


guapisolo (approved):
The shared rollout fixture gap is fixed by #2063 before the #2176 safe-merge checkpoint and remains fixed through checked tip #2649. API-key precedence is preserved across global and per-group config, so this looks good in stack context.

PR #2006, opened 2026-07-30
Launch inference engines through the inference engine worker spec

fzyzcjy:
Part of #1837


guapisolo (approved):
issue fixed before #2176

PR #2013, opened 2026-07-30
[tito] Add the Inkling TITO family (Inkling / Inkling-Small)

Zhichenzzz:
Registers `TITOTokenizerType.INKLING` so the session server can drive multi-turn / agentic rollouts on [Inkling](https://huggingface.co/thinkingmachines/Inkling) and [Inkling-Small](https://huggingface.co/thinkingmachines/Inkling-Small).

The family needs no fixed-template override and no boundary glue: the HF repos ship a `chat_template.jinja` that renders equivalently to sglang's token-level Inkling renderer, and the appended-suffix diff is naturally append-only (thinking content is preserved in history). The subclass only pins `assistant_start_str` / the message-boundary special tokens and binds sglang's `inkling` reasoning and tool-call parsers.

Two files, +38 lines: the subclass + enum + registry entry, and the `_EXPECTED_FIXED_TEMPLATES` row the family-coverage test requires.

## Verification

**Stage 2 — CPU chat-template roundtrip** (`scripts/tools/verify_chat_template.py --model thinkingmachines/Inkling-Small --tito-model inkling --thinking both`): **74/74 PASS**, covering multi-turn, long-chain, multi-user-turn, intermediate-system, no-tool, and mixed-role sequences in both thinking modes. `tests/fast/utils/chat_template_utils/` is green.

**Stage 3 — session-server e2e**, driven manually on a 6-layer surgery checkpoint of Inkling-Small (8×H200, two sglang engines at TP4/EP4, megatron actor at TP4+SP): **128 samples, 0 token-sequence mismatches** across every driver action the family's role surface allows — `append_user`, `append_system`, `append_assistant`, `force_final`, `rollback`, and the append_tool sentinel path.

No CI e2e registration is included: the full 276 B weights do not fit the runner fleet, and a slice-based registration would encode a configuration nobody has run. Worth adding once the full weights can be verified somewhere that fits them.

Two properties of the slice-based run, recorded so the next person doesn't chase them — neither is a TITO defect:
- The harness's coverage gate ("no sample produced an append_tool action") cannot be satisfied by a 6-layer slice; it emits no well-formed tool calls. Genuine tool-call coverage needs the full weights.
- Under sampling (temperature 0.7) the slice's near-random output occasionally emits Inkling's own message-boundary special tokens, which re-tokenize as real separators and surface as a `special_token_count` segment-count mismatch (37 vs 39 segments). Greedy decoding removes it entirely.

Serving-side parsers come from sgl-project/sglang#32958.

### DIFF
--- a/miles/utils/chat_template_utils/templates/inkling_fixed.jinja
+++ b/miles/utils/chat_template_utils/templates/inkling_fixed.jinja
@@ -1,0 +1,133 @@
+{%- set effort_map = {"none": 0.0, "minimal": 0.1, "low": 0.2, "medium": 0.7, "high": 0.9, "max": 0.99} -%}
+{%- set role_token = {"user": "<|message_user|>", "assistant": "<|message_model|>", "system": "<|message_system|>", "tool": "<|message_tool|>"} -%}
+
+{%- macro emit_thinking_effort() -%}
+    {%- set eff = reasoning_effort if reasoning_effort is defined and reasoning_effort is not none else 0.9 -%}
+    {%- if eff is string -%}
+        {%- set key = eff | trim -%}
+        {%- if key not in effort_map -%}
+            {{- raise_exception("Unknown reasoning_effort: " ~ eff) -}}
+        {%- endif -%}
+        {%- set num = effort_map[key] -%}
+    {%- else -%}
+        {%- set num = eff | float -%}
+    {%- endif -%}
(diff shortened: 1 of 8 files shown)


guapisolo (dismissed):
LGTM.

guapisolo (approved):
CI passed

PR #2022, opened 2026-07-31
dashboard: cache full-window engine series queries

yueming-yuan:
## Motivation

Opening the sglang metrics tab takes several seconds. The page fires one `/api/timeline/engine_series` request per metric key (~15), and each request independently re-concatenates every hourly partition block and re-filters the entire engine_series stream (9.2M rows / 1.4GB for a 12h 16-node run) before partitioning by engine. Derived metrics (ttft/e2e/rate families) do this twice (`_sum` + `_count`). At 0.3–4s per request, queued behind the browser's 6-connections-per-origin limit, the tab takes seconds to load — and re-pays the full cost on every poll tick.

## Modifications

Cache the merged full-window frame and the per-metric sorted partitions in `MetricStore`, versioned by the tuple of partition file sizes so appended scrapes (live/follow mode) invalidate the cache automatically. Only the `t0=t1=None` path (what the metrics page requests) uses the cache; bounded timeline-overlay queries keep the existing lazy per-partition path. `engine_metric_names()` (full-stream scan on every `/api/meta`) reuses the cached frame.

Measured on the 12h GLM-5.2 run dump (9.2M rows, 8 engines, 15 keys):

| | before | after |
|---|---|---|
| 15 concurrent requests (tab open) | ~6.5s | 0.39s |
| single request, warm | 0.3–4s | ~10ms |

`tests/fast/dashboard/test_store.py`: 14 passed.


miles/dashboard/store.py
```diff
@@ -1033,6 +1040,27 @@ def _engine_parts(self, metric: str, t0: float | None, t1: float | None) -> dict
             for key, part in sorted(frame.partition_by(["addr", "labels_json"], as_dict=True).items())
         }
 
+    def _engine_frame(self) -> pl.DataFrame:
```
Zhichenzzz, line 1043: **Race: a second thread can read `_engine_frame_cache` while it's still `None`.** generated by cc 🤖

The data endpoints are sync `def`, so FastAPI runs them in the anyio threadpool — genuinely concurrent, which is exactly the "15 requests when the tab opens" case this optimizes. Thread A sets `_engine_cache_version`, then spends 0.3–4s inside `window(None, None)`. Threads B..O see a matching version, skip the rebuild, and return `_engine_frame_cache` — still `None`. `_engine_parts_cached` then calls `.filter()` on it, and `AttributeError` isn't in `_translate_errors()`'s map, so it surfaces as a bare 500. First page load on a cold store hits this every time.

There's a second one hiding behind it: `_engine_parts_cached` can compute `parts` from frame V1 and then write them into the *new* dict another thread just installed for V2 — the attribute is resolved at assignment time, so stale partitions get served until the next version bump.

One lock fixes both, and serializing the rebuild is what we want anyway. The version assignment has to move last:

```python
with self._engine_cache_lock:
    if version != self._engine_cache_version:
        frame = reader.window(None, None)
        self._engine_frame_cache, self._engine_parts_cache = frame, {}
        self._engine_cache_version = version   # last, not first
    return self._engine_frame_cache
```

Zhichenzzz (approved):

PR #2023, opened 2026-07-31
dashboard: dp-aware engine metrics

yueming-yuan:
## Problem

With dp-attention (e.g. `--sglang-dp-size 4`), each dp rank's scheduler exports its own labeled series on `/metrics`, but the dashboard scraper kept only the `engine_type` label — every rank folded into one `(addr, metric)` series. Verified on a real 16-node GLM-5.2 run: every (tick, engine) held exactly 4 samples with the rank identity lost. Consequences:

- engine charts interleaved 4 ranks' values in one line (sawtooth artifacts, arbitrary rank per point after downsampling)
- derived rate metrics (`*_per_s`, latency means) diffed across different ranks' counters — mostly gaps and garbage
- the wandb snapshot `dashboard/engine_*_running_reqs` was last-writer-wins, i.e. one arbitrary rank
- advisory peak concurrency understated engine load by up to dp_size×
- the dp-rank starvation signal (one rank queueing while others idle — the actual congestion root cause on our run) was invisible

## Changes

- **scraper**: keep `dp_rank` in `KEPT_LABELS` so rank identity reaches the store
- **store**: `engine_series(..., per_dp_rank=False)` folds ranks into engine-level values — summed, or averaged for intensive metrics (`token_usage`, `cache_hit_rate`, `kv_transfer_latency_ms`); aggregation happens on raw samples before rate/mean derivation, so counter rates sum rank rates and histogram means weight by count. Folding groups on scrape ts (all ranks of an engine come from one fetch), which also repairs dumps written before this change (same-ts duplicates). `per_dp_rank=True` returns per-rank series with `dp_rank` in labels.
- **server**: `/api/timeline/engine_series` gains `per_dp_rank` (default off — response shape unchanged for existing consumers: metrics view, timeline overlay)
- **collector**: wandb snapshot sums same-tick samples per engine instead of last-writer-wins
- **advisory**: new dp-imbalance warning — busiest rank carrying real load while the idlest sits under 25% of it points at non-dp-aware routing (`--router-dp-aware`)
- **metrics view**: "split by dp rank" toggle in the engine legend; aggregated by default

## Validation

- 193 dashboard tests pass (10 new: fold/split/legacy-repair/rate/mean/advisory/scraper)
- replayed against the real dp4 run's dump: 4× duplicate samples collapse to one series per engine, gen-token rate recovers full coverage (27.6k points), advisory reports true engine concurrency


miles/dashboard/advisory.py
```diff
@@ -84,4 +100,23 @@ def compute_advisories(store: MetricStore, *, t0: float | None = None, t1: float
...
+        if busiest >= DP_IMBALANCE_MIN_RUNNING and idlest < DP_IMBALANCE_RATIO * busiest:
+            detail = ", ".join(
+                f"dp{rank}={mean:.1f}" for rank, mean in sorted(means.items(), key=lambda kv: int(kv[0]))
+            )
+            out.append(
+                Advisory(
+                    level="warning",
+                    message=(
+                        f"{addr}: dp ranks imbalanced (mean running reqs {detail}) — requests pile onto "
+                        "few ranks while others idle; check the router's dp-aware routing (--router-dp-aware)"
```
Zhichenzzz, line 118: also checking load balance strategy (maybe manual + min_load?)

miles/dashboard/collector.py
```diff
@@ -160,7 +160,10 @@ def __init__(
         # latest-value caches for the Prometheus forwarding snapshot; kept
         # separately because store buffers empty out on every flush
         self._latest_gpu: dict[tuple[str, int], GpuSample] = {}
-        self._latest_running_reqs: dict[str, float] = {}
+        # addr -> (scrape ts, sum over that tick's samples): dp-attention
```
Zhichenzzz, line 163: trim those comments?

Zhichenzzz (approved):

PR #2027, opened 2026-07-31
dashboard: advisory v2 — run-health alarms before config tuning

yueming-yuan:
## Motivation

The v1 advisories compare engine steady-state metrics against configured caps and read low utilization as reclaimable headroom. On the incidents that motivated this change, low utilization was a **symptom of starvation**, and the advice pointed backwards. Concretely, on a 744B agentic run whose rollout pipeline collapsed (dp-rank starvation → client-timeout storm → whole-group discards → backfill feedback loop), the training side sat in an open `train_wait` for **11.5 hours** while 35% of engine requests were client-aborted — and the advisory panel said only *"Peak concurrency (14) stayed under 30% of --sglang-max-running-requests (256); consider lowering it"*. Lowering the cap would have locked in the real bottleneck.

## Modifications

Order the rules by what they claim, and only emit config-tuning advice on a run with no active alarm — the same observation means opposite things on a healthy and a starved run:

**critical — pipeline liveness**
- Stall watchdog: an open phase running longer than max(6× its own median closed duration, 30min). Phases that are open by design (fully-async `rollout`) never close, so they have no baseline and produce no claim. `train_wait` gets a data-starvation hint.

**warning — compute being wasted**
- Engine abort ratio from the cumulative counters, with the group-discard amplification spelled out (one aborted sample discards its whole n-sample group; backfill resubmission amplifies engine load).
- KV-pool-bound concurrency: token usage pegged while peak running reqs sit far below the cap → *raise* `--sglang-mem-fraction-static`, explicitly **not** "lower the cap" (this signature previously resolved truncation / 400s / low concurrency / zero-grad in one change).
- Zero-std reward groups on the newest rollout (zero policy gradient), distinguishing all-zero rewards (systematic failure) from task difficulty.
- Truncation fraction, citing `--rollout-max-response-len` (the per-turn cap is usually the binding limit in multi-turn agentic runs).

**info — the v1 tuning suggestions, gated on no active alarm**

Supporting fixes:
- `DumpReader.groups()`: null-reward groups (train dumps missing/not yet synced) were flagged `zero_std=True` via `fill_null(0.0)` — missing data is not a degenerate group. The rollout view showed a false "8/8 zero-std groups" tile from exactly this.
- `MetricStore.phase_events()` public accessor for the raw (unexpanded) events.
- Persist `rollout_max_response_len` into the meta args snapshot so the truncation advisory can cite it.
- `/api/advisory` now passes the `DumpReader`; the panel renders levels distinctly (retitled "Run advisory").

## Validation

Replayed against the stalled run's real dump:

```
[critical] Phase 'rollout' has been open for 11.8h on 1 rank(s) — 42x its median duration (17min); the pipeline is stalled, not slow
[critical] Phase 'train_wait' has been open for 11.7h on 32 rank(s) — 51x its median duration (14min); the pipeline is stalled, not slow (training is starved of rollout data — check the data buffer and engine health)
[warning] 35% of engine requests were aborted client-side (3064/8826) — request timeouts are poisoning samples; one aborted sample discards its whole 8-sample group, and backfill resubmission amplifies the engine load
[warning] 25% of samples in rollout 8 are truncated
```

The two v1 info suggestions are correctly suppressed, and the null-reward guard keeps the zero-std rule quiet on a dump whose train shards are absent.

Note: a dp-rank queue-imbalance advisory (the other starvation signature we hit) needs per-dp-rank label granularity in the scraper — the collector currently reduces engine labels to `engine_type`. Left as a follow-up.

Tests: `tests/fast/dashboard/` advisory/store/dump_reader_views/server — 60 passed.


miles/dashboard/advisory.py
```diff
@@ -21,10 +38,165 @@
...
 
 
+def compute_advisories(
+    store: MetricStore, reader: DumpReader | None = None, *, t0: float | None = None, t1: float | None = None
+) -> list[Advisory]:
+    args = store.meta.args if store.meta else {}
+    alarms = _stall_advisories(store)
+    alarms += _engine_alarm_advisories(store, args, t0, t1)
+    if reader is not None:
+        alarms += _rollout_advisories(reader, args)
```
Zhichenzzz, line 52: **`DumpStillWriting` on the newest rollout takes the liveness alarms down with it.**

`_rollout_advisories` uses `train_ids[-1]` — the newest and so most likely still-being-written rollout — and `summary()` raises `DumpStillWriting` on it, which the server maps to 503.

The frontend retries 503 three times at 1.5s and then `catch {}` keeps the previous list, so the symptom is the advisory panel silently frozen on a stale snapshot, at a cost of 4 requests / 4.5s per follow tick. And the rules that go missing are the stall and abort alarms — the ones that only matter on a live run.

The dump-side rules should be allowed to be absent without taking the rest with them:

```python
if reader is not None:
    try:
        alarms += _rollout_advisories(reader, args)
    except (DumpStillWriting, FileNotFoundError, KeyError):
        pass
```

Minor, same function: `groups()` already calls `summary()` internally and `summary()`'s cache is on-disk parquet, so this reads the same parquet twice per request. Taking it once and passing it down would halve that.

Zhichenzzz (commented):
qq: will 2022-2027 be stacked or which order should we follow?

Zhichenzzz (approved):

PR #2030, opened 2026-07-31
[async] async data buffer: unified filters and better observability

yueming-yuan:
When rollout production outruns training consumption in fully-async mode, finished groups pile up in what is effectively an unbounded queue (`asyncio.Queue(maxsize=1000)` = 125 training steps of backlog), and FIFO consumption means training always eats the oldest, stalest data — with no bound and no visibility.

Stacked on #1717 (`yueming/fully-async-class-api`), where `fully_async_rollout.py` lives.

## Buffer capacity with staleness-ranked eviction

`--async-data-buffer-max-batches` caps the `DataBuffer` in multiples of `rollout_batch_size` (default `2`). On overflow the producer does not block; the buffer evicts stalest-first — recycling the evicted prompts into the data source for regeneration (same path as the existing abort/staleness recycling) — until nothing in the buffer is beyond `--max-weight-staleness` (when set and the engine version is known) and the buffer is back within capacity. Ranking:

1. the group whose stalest sample is oldest (min weight version),
2. ties by the larger summed staleness,
3. remaining ties oldest-arrival-first.

The ranking needs no engine-version query: ranking by (min weight version, summed weight versions) is identical to ranking by (largest sample staleness, largest summed staleness) — with equal group sizes the current version is a constant offset that cancels. The version is only consulted for the `--max-weight-staleness` threshold.

`0` disables the cap and restores the legacy behavior: a large bound that blocks the producer when full.

Evicting at insertion also relieves the drain-time `--max-weight-staleness` filter, which otherwise discovers stale groups only at consumption, one drain at a time, while training waits on the recycle-regenerate churn.

## LIFO option

`--async-data-buffer-order lifo` consumes the freshest group first, keeping updates closest to on-policy without throttling production. Old groups then sink and only leave via eviction or a late (stale) drain, so lifo is meant to pair with the cap and/or `--max-weight-staleness`; the help text says so. Default stays `fifo`.

## Metrics

- `rollout/fully_async/evicted_stale_groups`, `evicted_overflow_groups`, `evict_rate` (evictions / groups entering the buffer since the previous drain)
- `rollout/fully_async/buffer_avg_staleness`, `buffer_max_staleness` over the groups sitting in the buffer at drain time
- the consumed `avg_staleness` / `max_staleness` now report unconditionally, not only when `--max-weight-staleness` is set (previously the feature had to be on before you could see whether you needed it)

## Sizing note

Measured on the current GLM-5.2 16-node run: steady-state queue is 0–5 groups and the post-eval burst is bounded by the in-flight cap (16 groups at `--async-max-concurrent-samples 128`), so ~2 training batches of capacity absorbs every observed burst while bounding queue-induced staleness to ~2 steps — hence the default of 2.

## Tests

`DataBuffer` unit tests (overflow eviction, threshold-first eviction, summed-staleness tie-break, lifo order, staleness stats) plus a drain-level metrics-wiring test; existing tests updated for the two new args. 19 passed, 3 consecutive runs.


miles/rollout/fully_async_rollout.py
```diff
@@ -68,6 +71,112 @@ def group_oldest_weight_version(group: Group) -> int | None:
...
+    """Finished groups waiting between rollout production and training consumption.
+
+    Supported dataflow/staleness control options:
+
+    (1) max groups: use ``--async-data-buffer-max-batches`` to set the max size
+        of the buffer, in multiples of rollout_batch_size. On overflow the most
+        stale groups are evicted and their prompts recycled for regeneration;
+        0 disables eviction and blocks the producer when the buffer is full.
+    (2) order: use ``--async-data-buffer-order`` to set the consumption order,
+        fifo (default) or lifo. lifo trains on the freshest group first — pair
```
guapisolo: A dumb q, will anyone actually use LIFO? (

Shi-Dong: This is a good point... LIFO does seem unnecessary. I checked AReaL and they hardcoded FIFO.

miles/rollout/fully_async_rollout.py
```diff
@@ -230,20 +347,19 @@ async def _drain(self, rollout_id: int) -> RolloutFnTrainOutput:
...
-                            f"staleness={staleness} > max={args.max_weight_staleness})"
-                        )
-                        continue
+            oldest = group_oldest_weight_version(group)
+            current = await self._weight_version.get(args)
+            if oldest is not None and current is not None:
+                staleness = current - oldest
+                staleness_values.append(staleness)
+                if args.max_weight_staleness is not None and staleness > args.max_weight_staleness:
+                    self._recycle(prompt_group)
```
Shi-Dong: I'm a bit concerned about the recycle behavior.  Suppose that a task is genuinely hard and takes long to finish.  The current code paths suggest that the task is recycled when 1) a weight update happens when the task group is still in-flight, or 2) the task group finishes, but the data buffer finds out that it is too stale.  Does it mean that the task will be endlessly sent to the rollout machine?

Shi-Dong: Cross-checked with AReaL and PrimeRL and it seems that they don't have such recycling mechanism.  Probably fine to keep as-is for the v0.1 release, but might be worth thinking through in the future.

guapisolo (dismissed):
LGTM.

miles/rollout/fully_async_rollout.py
guapisolo: > 2) the task group finishes, but the data buffer finds out that it is too stale. Does it mean that the task will be endlessly sent to the rollout machine?

I have similar concern. 
- I think append to data source at the beginning, and retry asap is a possible way. It won't trigger "endlessly" retry. But I think it shouldn't be a default choice
- I think we should have more choices.
  - Add a `--async-stale-samples-handler` arg. It can include three modes "retry", "drop" and "mask". 
  - retry is the current behavior as-is
  - drop is drop the stale sample (default)
  - mask is mask the stale generation, and still train it. (like Kimi k25 report, can leave as TODO)
  - And it can also be a customizable function.

guapisolo: Also, I suggest we move dynamic filter function before the staleness handler. cuz if a prompt cannot produce useful signal. We should drop it first. No matter it's staleness or not.

miles/rollout/fully_async_rollout.py
guapisolo: > 1) a weight update happens when the task group is still in-flight,

I think it's a bug comment here.
- We've already ban `--partial-rollout` for fully async.
- Aborted sample generally caused by weight update mode == `abort`
- So, in `arguments.py` we should ban `--pause-generation-mode` to be not `abort`
```
if args.fully_async:
        assert ...
```
- Then the comment should be is "corner case handler", rather than handling weight update.

miles/rollout/fully_async_rollout.py
```diff
@@ -68,6 +71,112 @@ def group_oldest_weight_version(group: Group) -> int | None:
...
+                    await self._cond.wait()
+                self._entries.append(entry)
+            self.entered_groups += 1
+            self._cond.notify_all()
+
+    async def get(self) -> BufferEntry:
+        async with self._cond:
+            while not self._entries:
+                await self._cond.wait()
+            entry = self._entries.pop() if self._order == "lifo" else self._entries.pop(0)
```
guapisolo: I think the `--async-data-buffer-order` controls how trainer fetch data from data buffer, right? 
e.g. Some trainer want longer samples, which might have larger cp, other one might wanna shorter samples. 
So I think we can naively add a thin customizable func here instead of `--async-data-buffer-order`

miles/utils/arguments.py
```diff
@@ -680,6 +684,42 @@ def add_rollout_arguments(parser):
                     "(rollout_batch_size groups, i.e. rollout_batch_size * n_samples_per_prompt)."
                 ),
             )
+            parser.add_argument(
+                "--async-data-buffer-max-batches",
```
guapisolo: a dumb q, do you think it's better to set this param as `--async-data-buffer-max-groups` and default as `2 * rollout_batch_size`.

yueming-yuan: The cons of this are that we cannot access `rollout_batch_size` here,  so we have to set it to `None` as the default and set it to `2 * rollout_batch_size` later. I think it might be confusing because people will think `None` = no limit..

yueming-yuan: If the name is a bit misleading, how about changing it to `--async-data-buffer-capacity-factor` to emphasize that it's a factor?

guapisolo: Oh good finding lol. 
I think we can also make it a float datatype? and floor down here 
```
self._capacity = (
            args.async_data_buffer_max_batches * args.rollout_batch_size
            if args.async_data_buffer_max_batches
            else 1000  # legacy blocking bound
)
```

miles/rollout/fully_async_data_buffer.py
```diff
@@ -0,0 +1,233 @@
...
+    - aborted groups (the generate function gave up, e.g. an agentic collect timeout)
+    - groups ``--dynamic-sampling-filter-path`` does not keep
+
+    Rejected on get, because staleness depends on when the group is consumed:
+
+    - groups beyond ``--max-weight-staleness``
+
+    Dataflow control options:
+
+    (1) max groups: use ``--async-data-buffer-capacity-factor`` to set the max
```
guapisolo: Why the blocking logic removed here? 

In the current impl. When the buffer is full, we just evicted stale groups

This design also force data buffer hold `weight_version` info, which introduce some semantic confusion for `weight_version` in `DataBufferInput` class. (it's actually observed engine weight version rather than the group's min weight version.). 

What we achieve is just save some cpu memory when kick out staled groups.

miles/rollout/fully_async_data_buffer.py
```diff
@@ -0,0 +1,233 @@
...
+        size of the buffer, floor(factor * rollout_batch_size) groups. On
+        overflow the most stale groups are evicted.
```
guapisolo: I suggest the behavior change:

When the data buffer is full, block the `data_source -- prompt --> generation_pool` path to stop too many rollouts.

guapisolo:
also a tiny fix on this #2241

guapisolo (approved):
LGTM except for the weight version handling.

Shi-Dong (approved):
LGTM

PR #2034, opened 2026-07-31
[fp8] Gate the ue8m0 weight quantizer on the backend scale format

JessicaJiang-123:
Co-author-with: @XinyuJiangCMU 
## What

`_quantize_param` currently selects the ue8m0 weight quantizer whenever `weight_block_size == [128, 128]`. This gives all such checkpoints power-of-2 scales, even on backends whose GEMM consumes plain fp32 scales.

In SGLang, ue8m0 is enabled only on Blackwell:

```python
DEEPGEMM_SCALE_UE8M0 = ENABLE_JIT_DEEPGEMM and is_sm100_supported()
```

- Hopper DeepGEMM uses fp32 scales when `DEEPGEMM_SCALE_UE8M0` is false.
- ROCm AITER `gemm_a8w8_blockscale` also consumes fp32 block scales.
- Miles training uses `NVTE_FP8_BLOCK_SCALING_FP32_SCALES=1`, so training already produces continuous fp32 weight scales.

As a result, non-Blackwell rollout currently converts training scales to power-of-2 values unnecessarily.

## Changes

- Export `deep_gemm_wrapper` from `miles/backends/megatron_utils/sglang.py`.
- Use `per_block_cast_to_fp8` only when `deep_gemm_wrapper.DEEPGEMM_SCALE_UE8M0` is true.
- Otherwise use `blockwise_cast_to_fp8_triton`, which produces fp32 block scales with the same shape and dtype.

## Branch selection

| Target | `DEEPGEMM_SCALE_UE8M0` | Before | After |
|---|---:|---|---|
| Blackwell | `True` | ue8m0 | unchanged |
| Hopper | `False` | power-of-2 fp32 scales | continuous fp32 scales |
| sm89 / sm120 | `False` | power-of-2 fp32 scales | continuous fp32 scales |
| ROCm | `False` | power-of-2 fp32 scales | continuous fp32 scales |


wenchenvincent:
@JessicaJiang-123 For ROCm, we should also have different branches:
gfx942 (Hopper counterpart) should use continuous fp32 scale and gfx950 should use (Blackwell counterpart) should use power-of-2 fp32 scale.

yueming-yuan:
made an edit in this commit [d0dc7be](https://github.com/radixark/miles/pull/2034/commits/d0dc7be04326c384ec2cf30d9866a62ca6fcc657) to select from real TE setting.
AMD one needs check

yueming-yuan (approved):

PR #2046, opened 2026-08-02
Support DFlash rollouts without logprobs

zianglih:
## Summary

@humansand

- Disable rollout logprob requests for SGLang DFLASH and DSPARK, which do not
  support `return_logprob` during speculative generation.
- Consume SGLang's top-level `output_ids` and leave rollout logprobs optional so
  the training backend can compute actor logprobs.
- Reject features that genuinely require rollout-side logprobs, while retaining
  sampled-token OPD support.
- Cover shared and legacy rollout paths, lowercase algorithm names, empty
  completions, and multi-turn tool observations.

## Root cause

Miles unconditionally sent `return_logprob=true`. SGLang's DFlash family rejects
that request before generation, so deterministic HTTP 400 responses could also
trip the router circuit breakers and surface later as HTTP 503 errors.

## Validation

- `pytest -q tests/fast/rollout/generate_hub/test_single_turn.py tests/fast/rollout/generate_hub/test_multi_turn.py -k 'SpeculativeGenerationWithoutLogprobs or two_turns_without_rollout_logprobs'`: 18 passed, 1 skipped
- `python3 -m py_compile` on all six changed Python files
- `ruff check` on all six changed Python files
- `git diff --check`
- Eight-node/64-B200 colocated DeepSeek-V4 DSPARK integration
  (`raysubmit_2jzhLgZzcvEqGfdu`): exactly three 256/256 rollout generations,
  three actor/reference-logprob and backward/optimizer steps, and four 543/543
  target-weight syncs completed with all rank phases `ok=true`
- The integration run had no HTTP 400/503 responses or fatal errors, kept MTP
  training disabled, and left the raw DSPARK drafter metadata hashes unchanged

### DIFF
--- a/miles/rollout/generate_utils/generate_endpoint_utils.py
+++ b/miles/rollout/generate_utils/generate_endpoint_utils.py
@@ -13,6 +13,44 @@ from miles.utils.processing_utils import encode_image_for_rollout_engine, extrac
 from miles.utils.types import Sample
 
 
+_SPECULATIVE_ALGORITHMS_WITHOUT_LOGPROBS = {"DFLASH", "DSPARK"}
+
+
+def generation_return_logprob(args) -> bool:
+    algorithm = getattr(args, "sglang_speculative_algorithm", None)
+    if isinstance(algorithm, str):
+        algorithm = algorithm.upper()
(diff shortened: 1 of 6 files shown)


jvmncs:
fyi @zianglih https://github.com/sgl-project/sglang/pull/33459

PR #2047, opened 2026-08-02
fix glm47-flash: use paged MLA prefill on B200

guapisolo:
## Summary
Use paged MLA prefill for GLM-4.7-Flash on B200 without changing GPU topology.

## Symptom & Reproduction
- **Symptom:** B200 rollout prefill fails with `Unsupported head dimensions: head_dim_qk=256, head_dim_vo=256`.
- **Reproduction:** Run `python scripts/run_glm47_flash.py --hardware B200` on B200 with the documented setup.

## Root Cause
1. `flashinfer_mla_backend.py` selects the SM100 ragged CUTLASS prefill path.
2. `fmha_cutlass_sm100.cu` rejects QK and value head dimensions of 256.

## Fix
The launcher accepts B200 and adds `--sglang-flashinfer-mla-disable-ragged` only for that hardware, selecting SGLang's paged MLA prefill path while preserving the H200 default and existing 8-GPU parallelism.

## Verification
- **New / updated test:** No repository test was added because this launcher has no focused test harness.
- **Generated-argument probe:** B200 differs from H200 only by the new SGLang flag.
- **Preservation check:** node8/TP4/EP8/rollout4 remain for both hardware values.
- **Runtime smoke:** A two-B200 end-to-end run reached the first successful optimizer step.

## Review Focus
- Scrutinize the `args.hardware == "B200"` guard in `scripts/run_glm47_flash.py`.
- Scrutinize the preserved topology defaults in `scripts/run_glm47_flash.py`.


yueming-yuan (approved):
approve with nit: i think we might not need test for scripts args, given we will only edit it intentionally(

PR #2048, opened 2026-08-02
Rename PortAllocator internals and pass the actor positionally

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: #2053 removes the temporary `spec_session_server -> router_manager` back-edge introduced here. The fix lands before safe merge checkpoint #2176 and remains present through checked tip #2649.

PR #2049, opened 2026-08-02
Add a worker provider abstraction over RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in full-stack context: the initially premature provider abstraction is completed by #2343 and remains intact at the checked #2649 tip. That repair lands after the #2176 safe-merge checkpoint, so #2049 is not standalone-safe at that checkpoint.

PR #2050, opened 2026-08-02
Stop checking actor liveness while waiting for a tcp port

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context through #2649. #2050 cleanly narrows readiness to the TCP endpoint and enables the worker-provider transition in #2051. No blocking findings; the change lands before and is included in safe-merge checkpoint #2176.

PR #2051, opened 2026-08-02
Launch the sgl and miles routers through RayWorkerManager

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/specs/entrypoint.py
+++ b/miles/ray/specs/entrypoint.py
@@ -5,7 +5,9 @@ from miles.utils.workers.worker_spec import BaseWorkerSpec
 def compute_specs(args) -> list[BaseWorkerSpec]:
     return [
         *inference.specs_router(args),
-        inference.spec_session_server(args),
-        *inference.specs_inference_engine(args),
+        # TODO enable it
+        # inference.spec_session_server(args),
+        # TODO enable it
+        # *inference.specs_inference_engine(args),
         # TODO add more
     ]
--- a/miles/utils/workers/ray_worker_manager.py
+++ b/miles/utils/workers/ray_worker_manager.py
@@ -1,21 +1,169 @@
+import asyncio
+from collections.abc import Callable
+from dataclasses import dataclass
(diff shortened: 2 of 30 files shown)


miles/ray/specs/entrypoint.py
```diff
@@ -5,7 +5,9 @@
 def compute_specs(args) -> list[BaseWorkerSpec]:
     return [
         *inference.specs_router(args),
```
guapisolo, line 7: [P3] Keep `debug_train_only` free of rollout services

Codex review: `launch_worker_manager()` runs before the later `InferenceController.init()` guard, so `compute_specs()` still emits router workers in `--debug-train-only` mode. This does not affect training correctness, but it unnecessarily starts a router actor and binds ports despite the documented train-only contract; could we return no router specs when `args.debug_train_only` is set?

miles/utils/workers/ray_worker_manager.py
```diff
@@ -1,21 +1,169 @@
...
+                num_gpus=0,
+                runtime_env={"env_vars": self.spec.env_var()},
+            )
+            .remote()
+        )
+
+    async def alloc_ports(self) -> None:
+        self.self_addrs = {}
+
+        node_ip = await self.actor_handle._get_node_ip.remote()
```
guapisolo, line 149: [P2] Preserve the `MILES_HOST_IP` router override

Codex review: this now derives the router host only from `CommandActor._get_node_ip()`, whereas the previous path used `get_host_info()` and honored `MILES_HOST_IP`. This is non-blocking when Ray’s node IP is already usable, but deployments that set the override for a container/pod address will bind and publish the Ray-detected IP instead. The behavior remains present through #2176 and #2649; could the router spec carry an explicit bind/advertised-host override?

guapisolo (approved):
Codex stack review: #2054 completes the router/session placement transition, while #2053 and #2057 close the source-test gaps; all land before safe-merge checkpoint #2176 and remain intact through checked tip #2649. The two inline comments are non-blocking follow-ups.

PR #2052, opened 2026-08-02
Support pinning workers to the head node in worker manager

fzyzcjy:
Part of #1837


guapisolo (approved):
Verified that #2052 preserves the router's head-node placement through the RayWorkerManager migration, and that the contract remains intact at the checked #2176 tip (97bba455). LGTM in the #1837 stack context.

PR #2053, opened 2026-08-02
Drop support of configurable session server ip and port

fzyzcjy:
Part of #1837


guapisolo (approved):
[P1] In isolation, the configured-port regression could make session servers unreachable under fixed network policies; #2054 fixes it before the #2176 safe-merge checkpoint, and the fix remains present through #2649. The remaining CLI validation and documentation follow-ups are non-blocking.

PR #2054, opened 2026-08-02
Launch session servers through RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context.

[P1] #2054's session-server address contract cannot represent multi-host placement and can abort startup; #2305 fixes it before safe merge checkpoint #2176, and the repair remains intact through the checked #2649 tip.

PR #2055, opened 2026-08-02
Support master-mode ports in RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
[P1][fixed before checkpoint] In #2055, the stale `self.cell_index` access would crash configured static Session Server ports during worker-manager initialization; #2156 fixes it and #2175 adds the multi-cell regression test before safe merge checkpoint #2176. I verified both remain intact through #2649; LGTM in stack context.

PR #2058, opened 2026-08-02
Restructure RayWorkerManager around group and cell managers

fzyzcjy:
Part of #1837


guapisolo (approved):
Looks good in stack context. Moving worker lifecycle ownership into each cell preserves the existing global launch/port/setup barriers, and no source finding requires follow-up at the #2176 safe-merge checkpoint.

PR #2059, opened 2026-08-02
Support stopping a cell in RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context through #2649. PR #2112 completes the cell-stop lifecycle introduced here before safe-merge checkpoint #2176, with no post-checkpoint follow-up remaining.

PR #2060, opened 2026-08-02
Keep one addr_info per cell and read the engine topology from RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
Codex stack review: [P1] #2060 leaves the Miles dashboard without engine topology because the reader moves before engine ownership and its lookup contract is mismatched. #2063 completes the ownership migration and #2167 aligns the lookup contract; both fixes land before safe-merge checkpoint #2176 and remain present through checked tip #2649.

PR #2061, opened 2026-08-02
Remove temporarily the engine env report

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context through #2176: this intentionally removes the actor-bound rollout env report before #2063 moves engine lifecycle to RayWorkerManager, and the engine-side replacement is implemented post-checkpoint in #2509. LGTM for this staging PR.

PR #2062, opened 2026-08-02
Remove server liveness check in health waiting

fzyzcjy:
Part of #1837


guapisolo:
[P1] In isolation, #2062 can block rollout initialization indefinitely when SGLang exits before readiness; #2098 replaces the production waiter with bounded, tick-driven probes before #2176, and the fix remains present through #2649, so no post-checkpoint follow-up is required.

PR #2063, opened 2026-08-02
Launch inference engines from RayWorkerManager and reduce cells to bookkeeping

fzyzcjy:
Part of #1837


guapisolo (approved):
[P1] The startup naming mismatches in #2063 blocked every rollout initialization, and #2084 removes both before safe merge checkpoint #2176. The repair remains intact through checked tip #2649; LGTM in stack context.

PR #2064, opened 2026-08-02
Remove rollout engine recovery to be rewritten later

fzyzcjy:
Part of #1837


guapisolo:
[P1] This PR removes `RolloutServer.recover()` while `EvalFleet.pin()` still awaits it, so every dedicated eval point reaching `pin()` is converted to `EvalSkip("unhealthy")` even with healthy engines. #2497 removes that call before checkpoint #2176, but the healthy real-server path still relies on the fake-only `.engines` interface, so the checkpoint remains broken. #2614 completes the repair after the checkpoint by using lock-guarded `.api_clients`; please include it before treating #2176 as safe (verified through #2649).

PR #2065, opened 2026-08-02
Remove the unnecessary debug_train_only guard when starting cells

fzyzcjy:
Part of #1837


guapisolo (approved):
The `debug_train_only` lifecycle gate remains at `InferenceController.init()`, so [PR #2065](https://github.com/radixark/miles/pull/2065/files) safely removes the duplicate `RolloutServer` guard. Verified at safe merge checkpoint #2176 and through checked tip #2649; LGTM.

PR #2066, opened 2026-08-02
Extract the immutable ServerCell metadata to avoid constructing ServerCell later

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed [PR #2066 changes](https://github.com/radixark/miles/pull/2066/files) in the #1837 stack through checked tip #2649. #2066 fixes the `ServerCell` metadata/lifecycle-state coupling, with no blocking issue found before safe merge checkpoint #2176.

PR #2067, opened 2026-08-02
Refactor start semantics to add semantics for cells

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed [PR #2067 changes](https://github.com/radixark/miles/pull/2067/files) in stack context. In isolation, its [P2] no-op `add_cell` fixture leaves three chunking tests with empty cell lists, while its [P3] eagerly created raw coroutines add unawaited-coroutine warnings on later startup validation failures; #2084 removes both before safe merge checkpoint #2176, and the repairs remain present through checked tip #2649.

PR #2068, opened 2026-08-02
Change stopping cells to dynamic removal

fzyzcjy:
Part of #1837


guapisolo (approved):
[P1] In isolation, #2068 can leave Megatron/FSDP targeting a removed engine because the cached weight-update topology is not invalidated. #2094 fixes this before #2176 via cell-hash snapshot reconciliation, and the fix remains present through checked tip #2649, so approving in stack context.

PR #2069, opened 2026-08-02
Remove the now-unused stopped cell state

fzyzcjy:
Part of #1837


guapisolo (approved):
[P2] #2069 removes `_mark_stopped()` while the eval health path can still call it after a failed probe, causing `AttributeError`; #2497 removes that path before #2176, and the repair remains through #2649. [P3] It also leaves `is_allocated` tautological and dead lifecycle branches; #2087 begins the cleanup and #2497 completes it before #2176, with no post-checkpoint follow-up.

PR #2070, opened 2026-08-02
Rename the remaining cell start and stop wording to add and dispose

fzyzcjy:
Part of #1837


guapisolo (approved):
Codex stack review: #2070 preserves behavior while aligning lifecycle names with ownership—`RayWorkerManager` owns worker processes, while `RolloutServer` and `ServerCell` manage logical membership and router registration. No #2070 follow-up is needed before safe-merge checkpoint #2176; checked tip is #2649.

PR #2075, opened 2026-08-03
session: apply the trained LoRA adapter to session-server rollouts

Shi-Dong:
## Summary

When LoRA training is enabled, session-server rollouts were served by the **base weights**: the session server forwarded `/v1/chat/completions` requests to SGLang without a `lora_path`, and SGLang only applies an adapter when the request names one. As a result, the adapter being trained never shaped the trajectories it was scored on — generation stayed on the frozen base policy while the trainer updated the adapter, so the rollouts drifted off-policy and the trained behavior never fed back into data collection.

This PR stamps `lora_path` with the trained adapter's registered name onto every request the session server forwards, whenever LoRA is enabled (`is_lora_enabled(args)`). Non-LoRA runs are unaffected.

## Test plan

- [x] Verified on a 4-node GLM-5.2 LoRA agentic run (colocate, 2 fp8 rollout engines): after this change the engine logs show the adapter applied on session requests, and `update_weights` load markers correspond to the adapter used at generation time.
- [x] Confirmed no behavior change when LoRA is disabled (`lora_path` is not added to the request body).


guapisolo (approved):
LGTM.

PR #2078, opened 2026-08-03
Skip optimizer step for adapter slots with non-finite grad norm

Shi-Dong:
## Summary

- `step_adapter_slots()` now checks the per-slot grad norm for finiteness before clipping, mirroring the `found_inf` gate that Megatron's regular optimizer step applies but the multi-LoRA chained path bypasses.
- Why: with the gate missing, one non-finite microbatch is unrecoverable — `clip_grad_by_total_norm_fp32` scales every grad by `clip/(nan + eps)` and `step_with_ready_grads` then writes NaN into every tensor of the slot, permanently destroying that adapter while the loss stays finite. The only symptom is `train/grad_norm = nan` on that step.
- On a non-finite norm the slot's step is skipped: the offending adapter tensors are named in an error log (the shared `misc_utils.report_nonfinite_grads`, fed by a new `adapter_slot_named_parameters` helper), the slot's accumulated grads are zeroed, and training continues. The norm is still recorded so the skip is visible in metrics.

Stacked on #2198, which introduces the shared `report_nonfinite_grads` reporter this PR reuses; the base is `shi/report-nonfinite-grads` and retargets to `main` automatically once #2198 merges.

## Test plan

- [x] Exercised on a 4-node GLM-5.2 LoRA agentic run: with an upstream kernel bug producing NaN grads, the ungated path bricked the adapter (all-NaN weights); with the gate the step is skipped, the log names the non-finite adapter tensors, and subsequent steps proceed with finite norms.
- [x] Normal path unchanged: same run after the kernel fix shows identical behavior to main (finite norms, clipping applied, adapters step every batch).

### DIFF
--- a/miles/backends/megatron_utils/multi_lora_optimizer.py
+++ b/miles/backends/megatron_utils/multi_lora_optimizer.py
@@ -15,25 +15,32 @@ from megatron.core.optimizer.optimizer import MegatronOptimizer
 from megatron.core.optimizer.optimizer_config import OptimizerConfig
 from megatron.core.process_groups_config import ProcessGroupCollection
 
+from miles.backends.megatron_utils.misc_utils import report_nonfinite_grads
+
 logger = logging.getLogger(__name__)
+
+
+def adapter_slot_named_parameters(model, slot: int) -> list[tuple[str, torch.nn.Parameter]]:
+    """One adapter slot's parameters with their qualified names, across model chunks."""
+    from megatron.bridge.peft.multi_lora_layers import MultiLoRALinear
+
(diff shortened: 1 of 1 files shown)


yushengsu-thu:
multi-lora is re-designing and refactoring now. Wait for the redesigned multi-lora being merged first.

yushengsu-thu (approved):

yushengsu-thu (commented):
@Shi-Dong  I think the lora will not use this `multi_lora_optimizer.py` 
Why it was modified.

PR #2081, opened 2026-08-03
Fix the redundant model_path check when deciding whether a cell takes weights

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context through #2176: #2081 preserves the supported recovery behavior while making the weight-resume step unconditional, and that sequence remains present at the checkpoint. LGTM.

PR #2082, opened 2026-08-03
Trim ServerCellMetadata down to the fields a cell actually needs

fzyzcjy:
Part of #1837


guapisolo (approved):
[P1] The initial #2082 schema migration left the affected fast tests failing at fixture construction; #2111 completes the migration before safe merge checkpoint #2176, and the fix remains through #2649. The metadata ownership refactor looks good.

PR #2083, opened 2026-08-03
Add cell watching to the worker provider

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/worker_provider/ray.py
+++ b/miles/utils/workers/worker_provider/ray.py
@@ -1,9 +1,11 @@
+import asyncio
 import logging
+from functools import partial
 
 import ray.actor
 
 from miles.utils.workers.ray_worker_manager import RayWorkerManager
-from miles.utils.workers.worker_provider.base import BaseWorkerProvider
+from miles.utils.workers.worker_provider.base import BaseWorkerProvider, CellInfo, ReconcileFn, StopWatchFn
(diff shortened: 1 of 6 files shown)


miles/utils/workers/worker_provider/ray.py
```diff
@@ -12,12 +14,65 @@
...
+    async def _poll_once(
+        self, reconcile: ReconcileFn, seen_infos: dict[str, CellInfo], *, pool_ids: list[str]
+    ) -> None:
+        all_infos = await self._worker_manager_handle.get_cell_infos.remote(pool_ids=pool_ids)
+        observed_infos: dict[str, CellInfo] = {cell_id: info for cell_id, info in all_infos.items() if info.alive}
+        for cell_id in sorted(set(seen_infos) | set(observed_infos)):
+            observed_info = observed_infos.get(cell_id)
+            if seen_infos.get(cell_id) == observed_info:
+                continue
+            await reconcile(cell_id, observed_info)
```
guapisolo, line 66: [P2] Could you confirm whether aborting the entire background polling pass when one cell reconcile keeps failing is intentional? Because the exception is only caught around `_poll_once`, a persistently failing earlier cell is retried first every cycle and prevents later cells from being reconciled, leaving unrelated cell updates stale indefinitely. If this cross-cell head-of-line blocking is an accepted design trade-off, please feel free to ignore this.

PR #2084, opened 2026-08-03
Drive the rollout server cells from reconcile against the worker provider

fzyzcjy:
Part of #1837


guapisolo:
[P2] Reject cross-node TP widths that do not tile the node

With `num_gpus=24`, `num_gpus_per_engine=12`, and `num_gpus_per_node=8`, this validation accepts the group and builds two one-worker/8-slot cells while each engine is launched with `tp_size=12` and `nnodes=1`. Each worker therefore exposes only 8 GPUs to a 12-rank engine, so startup can fail to become healthy and wait indefinitely. No later PR through checkpoint #2176 or checked tip #2649 repairs this gap. Please fail fast unless partial-node workers are explicitly supported.

PR #2085, opened 2026-08-03
Unify the memory release and resume performed when a colocated cell starts

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed [PR #2085 changes](https://github.com/radixark/miles/pull/2085/files). [P1] #2085 initially resumed colocated rollout weights before trainer initialization, risking startup OOM for HBM time-sharing jobs; #2099 defers cell initialization into the first weight-update window, fixing it before safe merge checkpoint #2176, and the ordering remains intact through checked tip #2649.

PR #2086, opened 2026-08-03
Inline the router registration helpers and hold the router client on the cell

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The production changes preserve router registration and removal behavior while moving the immutable router client onto each cell. No correctness or regression issue remains at deliver-1 within the requested scope.

PR #2087, opened 2026-08-03
Remove the always-true is_allocated flag from ServerCell

fzyzcjy:
Part of #1837


guapisolo (approved):
[P3] #2087 removes `ServerCell.is_allocated` but leaves a dangling read in `ServerCell.probe_and_mark_dead()`, so direct calls fail before probing; this is non-blocking because its only production caller already ends in the same inherited `EvalSkip("unhealthy")` outcome. [PR #2497 changes](https://github.com/radixark/miles/pull/2497/files) atomically removes the obsolete probe/recover chain before safe merge checkpoint #2176, the correction remains present through checked tip #2649, and I am approving in that stack context.

PR #2088, opened 2026-08-03
Restructure the rollout cell states while keeping the current behavior

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: the [P1] rollout-startup failure introduced in #2088 when `check_weight_update_equal` is enabled is fixed by #2098 before safe checkpoint #2176, and the fix remains present through the checked tip #2649.

PR #2089, opened 2026-08-03
Rename the update weights window entry points to start and end update weights

fzyzcjy:
Part of #1837


guapisolo (approved):
[P3] #2089 leaves the fault-tolerance `NotImplementedError` referring to deleted `get_updatable_engines`, so anyone diagnosing the guarded FT path is directed to a nonexistent API; #2111 removes that obsolete guard before #2176, and the repair remains present through checked tip #2649. Approved in stack context.

PR #2090, opened 2026-08-03
Mark cell weights ready only once the update weights window closes

fzyzcjy:
Part of #1837


guapisolo (approved):
Codex review: #2090 closes the stale-weight window by keeping restarted updatable cells pending until weight update and router registration both complete. Verified at current head dd82a110 in the #1837 stack before safe-merge checkpoint #2176.

PR #2091, opened 2026-08-03
Support disposing a rollout cell from any intermediate state

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No production correctness or regression issue introduced by this op remains at deliver-1. The temporary reuse of StateUnknown is replaced by an explicit terminal lifecycle state before the safe-deliver boundary.

PR #2092, opened 2026-08-03
Extract the weight update connection status into ConnStatusManager

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -744,12 +744,13 @@ class MegatronTrainRayActor(TrainRayActor):
         if process_groups_are_temporary:
             reload_process_groups()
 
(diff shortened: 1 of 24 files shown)


guapisolo (approved):
Approved in the #1837 stack: #2094 completes the snapshot-aware contract and #2150 normalizes the import ordering before #2176; both remain intact through #2649.

[P2] Non-blocking cross-stack conflict: #2756's own diff does not directly share changed paths, but its stacked base has replaced these updater files and still uses protocol-owned `_connection_stale`; replaying #2092 fails, and taking the #2756-base side wholesale would lose the post-success reconnect commit, so a failed trainer-stale reconnect could be treated as fresh. When cascading #2756, keep its unified updater but port `ConnStatusManager` into it and mark reconnected only after `protocol.connect()` plus `_registered_adapters.clear()` succeed.

PR #2093, opened 2026-08-03
Give the fsdp backend its own connection status manager

fzyzcjy:
Part of #1837


guapisolo (approved):
Stack-context LGTM for [PR #2093 changes](https://github.com/radixark/miles/pull/2093/files). The [P1] `needs_reconnect({})` signature mismatch would deterministically fail the added fast test, but #2094 fixes it before safe merge checkpoint #2176, and the repair remains present through checked tip #2649.

PR #2094, opened 2026-08-03
Detect engine reconnection from a snapshot of the rollout cell hashes

fzyzcjy:
Part of #1837


guapisolo (approved):
PR #2094 cleanly moves reconnection state from the controller-wide `has_new_engines` flag to trainer-local cell-hash snapshots, giving add/remove/restart changes a stable reconnect signal. Verified before checkpoint #2176 and retained through checked tip #2649.

PR #2095, opened 2026-08-03
Allocate an out-of-band gate port for every inference engine cell

fzyzcjy:
Part of #1837


guapisolo (approved):
[P2][fixed before checkpoint] [PR #2095 changes](https://github.com/radixark/miles/pull/2095/files) left three fast-test consumers on the pre-`CellAddrInfo` contract, causing 10 always-run `stage-a-cpu` tests to fail deterministically; #2096 and #2098 repaired them before safe merge checkpoint #2176. Verified through checked tip #2649; no follow-up is needed.

PR #2096, opened 2026-08-03
Launch sglang engines gated and release them once they are running

fzyzcjy:
Part of #1837


miles/backends/sglang_utils/sglang_engine.py
```diff
@@ -193,8 +197,6 @@ def _compute_server_args(
     if sglang_overrides:
         kwargs.update(sglang_overrides)
```
guapisolo, line 194: [P2] Please treat `gated_launch_port` as allocator-owned, like `host`/`port`. Under an explicit override, SGLang listens on that port while `ServerCell` activates the allocator port, so startup waits on the wrong endpoint; this is non-blocking for the intended configuration, but we should reject the override in `ServerGroupConfig.resolve()` and omit the corresponding CLI field.

guapisolo (approved):
Approved in stack context after reviewing [PR #2096 changes](https://github.com/radixark/miles/pull/2096/files). The upstream gate dependency is tracked in #1837, while #2098 removes the test hang before #2176 and the repair is retained through checked tip #2649; I left one non-blocking P2 about keeping the gate port allocator-owned.

PR #2097, opened 2026-08-03
Add a periodic tick sweep over the rollout cells

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/misc.py
+++ b/miles/utils/misc.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from collections.abc import Sequence
+from collections.abc import Awaitable, Callable, Sequence
 from typing import Any
 
 import ray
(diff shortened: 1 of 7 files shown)


miles/utils/misc.py
```diff
@@ -139,3 +139,25 @@ def filter_keys(d: dict[str, Any], interest_keys: Sequence[str]) -> dict[str, An
...
+    def __init__(self, fn: Callable[[], Awaitable[None]], *, interval_seconds: float):
+        self._fn = fn
+        self._interval_seconds = interval_seconds
+        self._task = asyncio.create_task(self._loop())
+
+    async def dispose(self) -> None:
+        self._task.cancel()
+        try:
+            await self._task
+        except asyncio.CancelledError:
```
guapisolo: [P2] Preserve cancellation from the disposing caller

`SimpleTicker.dispose()` catches every `asyncio.CancelledError` raised at this await, so it cannot distinguish the expected cancellation of `self._task` from cancellation of the task currently disposing the controller. If an outer shutdown deadline fires while this await is pending, that cancellation is swallowed and `InferenceController.dispose()` continues into watcher teardown instead of propagating it. This remains unchanged at safe-merge checkpoint #2176 and checked tip #2649. Please consume only the owned ticker task cancellation and add a deterministic test that cancels the dispose caller.

PR #2098, opened 2026-08-03
Rebuild the rollout cell around an explicit five-state machine

fzyzcjy:
Part of #1837


miles/ray/rollout/server_cell.py
```diff
@@ -82,42 +104,63 @@ async def add(self) -> None:
...
             await api_client.resume_memory_occupation(tags=[GPU_MEMORY_TYPE_WEIGHTS])
 
-        self._mark_pending_weights(addr_info)
+        serve_without_weight_update: bool = not self.meta.update_weights or self.args.debug_rollout_only
+        if serve_without_weight_update:
+            await self._register_with_router(addr_info=addr_info)
 
-        if not self.meta.update_weights or self.args.debug_rollout_only:
-            await self.mark_weights_ready()
+        self._change_state("mark_pending_weights", StateInitializing, StatePendingWeights(addr_info=addr_info))
```
guapisolo, line 111: [P2] Make initialization retries phase-aware

`_tick_when_initializing()` keeps the cell in `StateInitializing` across snapshot, release, and resume. If release succeeds server-side but its response is lost, resume fails, or the tick is cancelled after release, the next sweep starts over and snapshots weights that may already have been discarded. This can corrupt the weight-checker baseline or leave a colocated engine stuck after a transient HTTP failure. Please persist or reconcile the completed startup phase so retries resume from the first incomplete operation instead of replaying the whole sequence.

PR #2099, opened 2026-08-03
Initialize colocated engines only inside the weight update window

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/inference_controller.py
+++ b/miles/ray/rollout/inference_controller.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import time
 from dataclasses import dataclass
 
 from sglang.srt.constants import GPU_MEMORY_TYPE_CUDA_GRAPH, GPU_MEMORY_TYPE_KV_CACHE, GPU_MEMORY_TYPE_WEIGHTS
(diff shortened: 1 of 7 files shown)


miles/ray/rollout/inference_controller.py
```diff
@@ -122,6 +125,23 @@ async def end_update_weights(self, snapshot_cell_id_to_hashes: dict[str, str]):
             ]
         )
 
+    async def _ensure_cells_ready(self) -> None:
+        deadline = time.monotonic() + CELLS_READY_TIMEOUT_SECONDS
+        while True:
+            cells = [cell for srv in self.servers.values() for cell in srv.server_cells.values()]
```
guapisolo, line 144: [P1] Could we scope this recurring readiness barrier to the server(s) participating in the current training weight update? In the stack through safe-merge checkpoint #2176, generic rollout FT can replace a dedicated eval cell, which is non-colocated, uses separate GPU slots, and has `update_weights=False`. If that replacement remains `StateInitializing`, this all-server scan stalls or fails `start_update_weights()`, although eval failures are supposed to become `EvalSkip` and only skip that eval point. Initial all-model provisioning can remain a separate barrier.

PR #2100, opened 2026-08-03
Wait for the expected fleet before the first rollout starts

fzyzcjy:
Part of #1837


guapisolo (approved):
Stack-context approval: [P2] the init test double originally omitted the new startup barrier and would fail stage-a; #2175 adds the method and verifies the barrier blocking/order contract before safe-merge checkpoint #2176. The repair remains present through checked tip #2649, and the initial all-model provisioning barrier itself is intentional.

PR #2101, opened 2026-08-03
Add a context lock with a with_lock decorator

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in stack context: [P1] #2101’s inherited ContextVar grant could outlive the critical section and falsely authorize a child task, but #2306 replaces it with generation-scoped grants. The repair lands before and remains present through #2176; LGTM for this staging PR.

PR #2102, opened 2026-08-03
Add lock exemption and enforcement checkers to the context lock

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed [PR #2102 changes](https://github.com/radixark/miles/pull/2102/files) against the full #1837 stack through safe-merge checkpoint #2176. #2106 completes the planned missing-class-enforcement guard, and #2108's production consumers use the supported direct decorator forms; the remaining generic wrapper/descriptor limitations are non-blocking for this stack.

PR #2103, opened 2026-08-03
Add a requires_lock decorator to ensure lock is being held

fzyzcjy:
Part of #1837


guapisolo (approved):
Stack-context LGTM. The [P2] inherited ContextVar grant in #2103 can let an outliving child enter guarded code after release, but #2306 fixes this before production integration and safe merge checkpoint #2176; the repair remains through checked tip #2649.

PR #2104, opened 2026-08-03
Add acquires_lock and releases_lock for split lock windows

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved in stack context only. [P2] #2104's identity-only grant could let a stale child task bypass serialization or release a later owner's lock; #2306 replaces it with generation-scoped grants before checkpoint #2176, and the fix remains through checked tip #2649.

PR #2105, opened 2026-08-03
Remind periodically while a context lock acquisition is blocked

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/context_lock.py
+++ b/miles/utils/context_lock.py
@@ -2,16 +2,22 @@ import asyncio
 import contextvars
 import functools
 import inspect
+import logging
+import time
 from collections.abc import Callable
 from types import TracebackType
 from typing import Any
 
+logger = logging.getLogger(__name__)
+
+WAIT_LOG_INTERVAL_SECONDS: float = 5.0
 LOCK_ATTRIBUTE_NAME: str = "context_lock"
 
 _DISCIPLINE_MARKER_ATTRIBUTE_NAME: str = "_context_lock_discipline"
 
 # the annotation machinery (PEP 649) plants these in the class dict; they are not methods of the class
 _ANNOTATION_MEMBER_NAMES: frozenset[str] = frozenset({"__annotate__", "__annotate_func__"})
+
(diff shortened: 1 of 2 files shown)


miles/utils/context_lock.py
```diff
@@ -47,7 +53,11 @@ async def __aexit__(
 
     async def acquire(self) -> None:
         assert _held_lock.get() is None, f"Cannot acquire lock {self._name!r}: a context lock is already held"
-        await self._lock.acquire()
+        wait_reminder = asyncio.ensure_future(self._remind_while_waiting())
```
guapisolo, line 56: [P2][unresolved at safe checkpoint #2176 and checked tip #2649] Please avoid spawning a reminder task on every uncontended acquire. Because `asyncio.Lock.acquire()` can complete without yielding, a tight acquire/release sequence leaves one cancel-requested task and ready callback per call until the loop runs again; on this head, 100,000 calls retained about 91.8 MB before the next yield. Please preserve an allocation-free uncontended path and reliably drain the contended-path reminder without weakening lock fairness or cancellation safety.

tests/fast/utils/test_context_lock.py
```diff
@@ -307,6 +309,115 @@ async def _reattach_and_release(lock: ContextLock) -> None:
...
+    @pytest.mark.asyncio
+    async def test_a_blocked_acquirer_reminds_repeatedly(self, fast_reminder, caplog):
+        """A caller stuck on the lock keeps reporting that it is still waiting."""
+        lock = ContextLock("test")
+        holder = _HolderTask(lock)
+        await holder.start()
+
+        waiter = asyncio.create_task(lock.acquire())
+        with caplog.at_level(logging.INFO, logger="miles.utils.context_lock"):
+            await asyncio.sleep(0.08)
```
guapisolo: [P2][unresolved at safe checkpoint #2176 and checked tip #2649] Please wait on observed reminder events instead of assuming two 10 ms timers run within this 80 ms window. Under a busy or stalled CI event loop, the test sleep and first reminder can become runnable together, so the waiter is cancelled after only one log and this test flakes despite correct reminder behavior; a deterministic 100 ms event-loop stall reproduces it on this head.

PR #2106, opened 2026-08-03
Reject context lock decorators on classes that forgot enforce_lock_discipline

fzyzcjy:
Part of #1837


guapisolo:
Non-blocking design question: did we consider using [`wrapt`](https://wrapt.readthedocs.io/en/master/bundled.html#thread-synchronization) for the decorator and descriptor plumbing here? It already supports async locks and handles method binding and wrapper composition. Miles would still need its custom shared `ContextLock`, `requires_lock`, split acquire/release window, and class-level discipline audit, but building those policies on `wrapt` could reduce the owner-metadata and `__wrapped__` machinery we need to maintain.

PR #2107, opened 2026-08-03
Support temporarily releasing the context lock

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/context_lock.py
+++ b/miles/utils/context_lock.py
@@ -1,10 +1,11 @@
 import asyncio
+import contextlib
 import contextvars
 import functools
 import inspect
 import logging
 import time
-from collections.abc import Callable
+from collections.abc import AsyncIterator, Callable
 from types import TracebackType
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


miles/utils/context_lock.py
```diff
@@ -67,6 +68,14 @@ def release(self) -> None:
         _held_lock.set(None)
         self._lock.release()
 
+    @contextlib.asynccontextmanager
+    async def with_released(self) -> AsyncIterator[None]:
+        self.release()
+        try:
+            yield
+        finally:
+            await self.acquire()
```
guapisolo, line 77: [P1] Reacquire the context lock before propagating cancellation

If cancellation arrives while this await is queued behind another holder, `acquire()` exits before restoring the grant. The enclosing `@with_lock`/`@acquires_lock` cleanup then calls `release()` and raises `AssertionError`, masking the original `CancelledError` on a production path wired by #2108. I traced the later #1837 diffs through checkpoint #2176 and tip #2649; no fix lands. Please reacquire in the same task, propagate cancellation only after restoring the lock/grant, and add a deterministic blocked-reacquire cancellation test.

PR #2108, opened 2026-08-03
Guard the inference controller and its servers with ContextLock

fzyzcjy:
Part of #1837


miles/ray/rollout/inference_controller.py
```diff
@@ -142,11 +169,14 @@ async def _ensure_cells_ready(self) -> None:
                     f"{len(pending)}/{len(cells)} cells to become ready"
                 )
             logger.info(f"Waiting for {len(pending)}/{len(cells)} cells to become ready...")
-            await asyncio.sleep(CELLS_READY_POLL_INTERVAL_SECONDS)
+            async with self.context_lock.with_released():
```
guapisolo, line 191: [P1] Make the released-lock reacquire cancellation-safe before using it here

This is the first production use of `ContextLock.with_released()`. If cancellation arrives while its `finally` block is queued behind another lock holder, `acquire()` exits before restoring the grant; the enclosing `@acquires_lock` cleanup then calls `release()` and raises `AssertionError`, masking the original `CancelledError`. The issue remains present at checkpoint #2176 (`97bba455`) and tip #2649 (`48920692`). Please restore the lock/grant in the same task before propagating cancellation and add a deterministic contended-reacquire cancellation test.

PR #2109, opened 2026-08-03
Reuse SimpleHealthCheckerConfig for the rollout health check arguments

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: the previously noted Dumper + MilesRouter interaction is out of scope because MilesRouter is deprecated, and the supported-path refactor remains intact through checked tip #2649.

PR #2110, opened 2026-08-03
Drive health checking from a pulled activeness predicate

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/ft_utils/health_checker.py
+++ b/miles/utils/ft_utils/health_checker.py
@@ -5,9 +5,8 @@ import argparse
 import asyncio
 import logging
 from collections.abc import Callable, Coroutine
-from typing import Any
+from typing import Any, NamedTuple
 
-from miles.backends.sglang_utils.sglang_api_client import SGLangApiClient
 from miles.utils.ft_utils.api_server.models import TriState
 from miles.utils.pydantic_utils import StrictBaseModel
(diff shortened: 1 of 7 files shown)


miles/utils/ft_utils/health_checker.py
```diff
@@ -137,21 +150,28 @@ def stop(self) -> None:
...
         log_structured(logger.info, tag="ft", op="health", phase="resume", name=self._name)
-        self._paused = False
         self._need_first_wait = True
         self._status = TriState.UNKNOWN
         self._consecutive_failures = 0
 
     async def _loop(self) -> None:
         while True:
-            if self._need_first_wait:
+            activeness = self._get_activeness()
```
guapisolo, line 168: [P1] Invalidate the cached health status as soon as checking becomes inactive. `bump_active(False)` only updates the shared tracker, while `status` keeps returning the previous verdict until the polling loop wakes; during offload or reconfiguration, `/api/v1/cells` can therefore expose a stale `HealthCheckFailed` to the FT controller and cause a live cell to be suspended. Please return `UNKNOWN` whenever activeness is inactive or its epoch has not yet been observed, or invalidate the status synchronously. I traced the later stack: #2308 only discards probe results completed across an activeness transition, and this cached-status gap remains at checkpoint #2176 and through #2649.

PR #2111, opened 2026-08-03
Give each rollout cell a health checker driven by its own state

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/rollout_server.py
+++ b/miles/ray/rollout/rollout_server.py
@@ -1,6 +1,7 @@
 import asyncio
 import dataclasses
 import logging
+from collections.abc import Callable
 from typing import Any
 
 from miles.backends.sglang_utils.sglang_api_client import SGLangApiClient
@@ -9,6 +10,7 @@ from miles.backends.sglang_utils.sglang_router_api_client import SGLangRouterApi
 from miles.ray.rollout.router_manager import wait_router_ready
 from miles.ray.rollout.server_cell import ServerCell, ServerCellMetadata
 from miles.utils.context_lock import ContextLock, enforce_lock_discipline, lock_exempt, requires_lock
+from miles.utils.ft_utils.health_checker import ActiveAndEpoch
 from miles.utils.retry_utils import retry_until_deadline
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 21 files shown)


miles/ray/rollout/rollout_server.py
```diff
@@ -102,17 +110,27 @@ async def probe_and_mark_dead(self):
...
         cell_id = cell_meta.cell_id
         assert cell_id not in self.server_cells
-        cell = ServerCell(args=self.args, router_api_client=self._router_api_client, meta=cell_meta)
+        cell = ServerCell(
+            args=self.args,
+            router_api_client=self._router_api_client,
+            meta=cell_meta,
+            global_health_checker_activeness=self.global_health_checker_activeness,
+        )
+        self.server_cells[cell_id] = cell
```
guapisolo, line 119: [P1] Please make a failed cell initialization retryable. `add_cell` publishes the cell before awaiting `cell.init()`, so a transient failure leaves a same-hash `StateUninitialized` cell that subsequent reconciliation treats as converged. This permanently prevents a non-colocated rollout replacement from initializing. Please dispose the cell on failure and only insert it after successful initialization.

miles/utils/ft_utils/health_checker.py
```diff
@@ -148,8 +153,29 @@ def stop(self) -> None:
...
             self._task = None
+        if self._probe_task is not None:
+            self._probe_task.cancel()
+            self._probe_task = None
         self._status = TriState.UNKNOWN
 
+    async def cancel_inflight_probe(self) -> None:
+        probe_task = self._probe_task
+        if probe_task is None:
+            return
```
guapisolo, line 168: [P1] Please invalidate a published health verdict before the pause barrier returns. If the checker is sleeping after publishing `False`, there is no probe to cancel, so pause leaves that verdict visible until the next interval. The FT loop can then recycle a cell during offload or weight update. Please report `Unknown` immediately while inactive.

miles/ray/rollout/server_cell.py
```diff
@@ -211,3 +246,21 @@ async def check_weights(self, action: str, allow_quant_error: bool, selector: st
...
+    *,
+    args: Any,
+    name: str,
+    get_api_client: Callable[[], SGLangApiClient],
+    get_activeness: Callable[[], ActiveAndEpoch],
+) -> BaseHealthChecker:
+    if "rollout" not in args.ft_components:
+        return NoopHealthChecker()
+
+    config = SimpleHealthCheckerConfig.from_args(args, prefix="rollout_health_check")
```
guapisolo, line 261: [P2] Restore the rollout-specific failure threshold default. This checker reads the shared default of 3, so with the 30s rollout interval a dead engine is not reported unhealthy until roughly the third failed poll (~90s), instead of the previous first-failure contract (~30s). Please keep the trainer heartbeat default at 3, but make the rollout prefix default to 1.

PR #2112, opened 2026-08-03
Add start and stop of individual cells to RayWorkerManager

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed the current rebased head (`411f0017`) in the #1837 stack through safe merge checkpoint #2176 and checked tip #2649. The #2112 increment itself looks good; the normalized patch and supported production call graph remain unchanged.

PR #2113, opened 2026-08-03
Extract the health condition mapping onto CellCondition

fzyzcjy:
Part of #1837


guapisolo (approved):
PR #2113 cleanly centralizes the existing tri-state health-condition mapping on `CellCondition`; #2114 reuses that contract for rollout cells. The mapping remains unchanged through checkpoint #2176 and checked tip #2649.

PR #2114, opened 2026-08-03
Let the controller report cell status

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/inference_controller.py
+++ b/miles/ray/rollout/inference_controller.py
@@ -21,6 +21,7 @@ from miles.utils.context_lock import (
     requires_lock,
     with_lock,
 )
+from miles.utils.ft_utils.api_server.models import CellStatus
 from miles.utils.ft_utils.health_checker import ActivenessTracker
 from miles.utils.misc import SimpleTicker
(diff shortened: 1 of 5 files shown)


miles/ray/rollout/inference_controller.py
```diff
@@ -201,6 +202,14 @@ def _get_updatable_server(self) -> RolloutServer | None:
 
     # -------------------------- misc APIs -----------------------------
 
+    @lock_exempt
+    def get_cell_statuses(self) -> dict[str, CellStatus]:
```
guapisolo, line 225: **[P1][unresolved at safe checkpoint #2176 and checked tip #2649] Preserve the cell generation in this status snapshot**

This snapshot is consumed alongside worker-manager `CellInfo`, but it drops `cell.meta.workers_hash`. Restarts reuse the same `cell_id`, so the API can combine health from generation N with process state from generation N+1; mini-FT may then recycle a healthy replacement on stale `Healthy=False`, or leave a failed replacement alone on stale `Healthy=True`. Please return `workers_hash` with each `CellStatus` and only use the health verdict when it matches the current `CellInfo.workers_hash`; otherwise report Pending/Unknown or retry.

PR #2115, opened 2026-08-03
Serve rollout cells over the api server

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/ft_utils/api_server/handles.py
+++ b/miles/utils/ft_utils/api_server/handles.py
@@ -2,8 +2,11 @@ from __future__ import annotations
 
 import abc
 
+import ray
+
+from miles.ray.rollout.server_cell import compute_pending_rollout_cell_status
 from miles.ray.train.group import RayTrainGroup
-from miles.utils.ft_utils.api_server.models import Cell, CellCondition, CellMetadata, CellSpec, CellStatus
+from miles.utils.ft_utils.api_server.models import Cell, CellMetadata, CellSpec
 from miles.utils.test_utils.fault_injector import FailureMode
 
 
@@ -79,9 +82,15 @@ class _ActorCellHandle(_CellHandle):
         actors[sub_index].inject_fault.remote(mode.value)
 
 
-# TODO the code will NOT work before implementing rollout ft
 class _RolloutCellHandle(_CellHandle):
-    def __init__(self, *, inference_controller: object, rollout_cell_id: str) -> None:
+    def __init__(
+        self,
+        *,
+        worker_manager: ray.actor.ActorHandle,
+        inference_controller: object,
+        rollout_cell_id: str,
+    ) -> None:
+        self._worker_manager = worker_manager
         self._inference_controller = inference_controller
         self._rollout_cell_id = rollout_cell_id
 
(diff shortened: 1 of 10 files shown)


miles/utils/ft_utils/api_server/handles.py
```diff
@@ -105,15 +113,14 @@ async def get_cell(self) -> Cell:
...
-                phase=phase,
-                conditions=[CellCondition(**c) for c in conditions_raw],
-            ),
+            spec=CellSpec(suspend=status.phase == "Suspended"),
+            status=status,
         )
 
     async def suspend(self) -> None:
-        await self._inference_controller.stop_cell(self._rollout_cell_id)
+        await self._worker_manager.stop_cells.remote([self._rollout_cell_id])
```
fzyzcjy, line 121: [P1] Route suspension through the weight-update lock

When a Ray rollout repair reaches this call while the trainer is between `start_update_weights()` and `end_update_weights()`, the manager kills an engine rank in the fixed-size broadcast, so the trainer can wait forever for that rank. Deliver-1 still calls `stop_cells` directly. Deliver-2's minimum semantic fix is a controller method guarded by the existing context lock; forward-porting it here also requires a thread-safe hop onto the controller's owning event loop, which the later worker-handle/cell-operations seam provides.

miles/utils/workers/ray_worker_manager.py
```diff
@@ -58,26 +58,14 @@ async def init(self, specs: list[BaseWorkerSpec], pgs: dict[str, PlacementGroupI
...
-                await asyncio.gather(*[c.stop() for c in failed], return_exceptions=True)
-                cells = [
-                    c for c, outcome in zip(cells, outcomes, strict=True) if not isinstance(outcome, BaseException)
-                ]
-
-        if errors:
-            raise errors[0]
+        try:
+            await _gather_or_raise([c.launch_actors() for c in cells])
+            await _gather_or_raise([c.alloc_ports() for c in cells])
```
fzyzcjy, line 63: [P1] Keep restarting cells hidden until addresses are complete

When a non-colocated restart's launch and port allocation overlap the provider poll, `launch_actors()` has already made `actors` non-`None`, so `get_cell_infos()` advertises the generation before its addresses are usable. Reconcile then inserts a `ServerCell`, `init()` fails on the empty address map, and later same-hash observations no-op, leaving the engine Pending indefinitely. Deliver-1 still has this; Deliver-2's minimum coupled fix keeps `self_addrs` as `None` until a local map is complete and reports `CellInfo.alive` only after every worker has published addresses.

miles/utils/ft_utils/api_server/handles.py
```diff
@@ -105,15 +113,14 @@ async def get_cell(self) -> Cell:
...
+        self._inference_controller.notify_cell_suspended(self._rollout_cell_id)
 
     async def resume(self) -> None:
-        await self._inference_controller.start_cell(self._rollout_cell_id)
+        await self._worker_manager.start_cells.remote([self._rollout_cell_id])
```
fzyzcjy, line 125: [P2] Serialize worker-manager lifecycle mutations

When a retry sends a second resume while the first is still launching, Ray can interleave the async actor calls: the first has already set `actors`, so the second filters the cell out and returns success; if the first later fails and rolls back, the supposedly successful resume leaves the cell suspended. Deliver-1 still lacks serialization. Deliver-2's minimum forward-port is an `asyncio.Lock` created before `init()` invokes `start_cells` and held across both `start_cells` and `stop_cells`; its later liveness-scanner changes are not dependencies.

fzyzcjy (commented):
The patch exposes rollout lifecycle operations without coordinating manager concurrency or the weight-update critical section, and restarted cells can be observed before their endpoints exist. These defects remain in deliver-1, although deliver-2 contains separable fixes.

PR #2116, opened 2026-08-03
Resolve api server cells per request instead of at startup

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at deliver-1. Issues repaired before that safe deliver were treated as transient and excluded as requested.

PR #2117, opened 2026-08-03
Delete the pre-reconcile engine recovery

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The standalone op temporarily imports the absent `update_weights_window`, but deliver-1 deletes the legacy `miles.ray.actor_group` module. The removed recovery API and controller `rollout_id` state have no production callers at deliver-1, so no regression from this op survives the target safe deliver.

PR #2118, opened 2026-08-03
Let a rollout cell be crashed on demand

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/ft_utils/api_server/handles.py
+++ b/miles/utils/ft_utils/api_server/handles.py
@@ -134,3 +134,6 @@ class _RolloutCellHandler(_CellHandler):
     async def resume(self, cell_id: str) -> None:
         await self._worker_manager.start_cells.remote([cell_id])
         self._inference_controller.notify_cell_resumed(cell_id)
+
+    async def inject_fault(self, cell_id: str, *, mode: FailureMode, sub_index: int) -> None:
+        await self._worker_manager.inject_fault.remote(cell_id, mode=mode.value, worker_in_cell_index=sub_index)
--- a/miles/utils/workers/command_actor.py
+++ b/miles/utils/workers/command_actor.py
@@ -4,6 +4,7 @@ import subprocess
 import threading
 
 from miles.utils.misc import NodeProbeMixin
+from miles.utils.test_utils import fault_injector
 from miles.utils.workers import process_utils
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


miles/utils/workers/command_actor.py
```diff
@@ -33,6 +34,9 @@ def kill_subprocess(self) -> None:
         assert self._process is not None, "CommandActor has no subprocess to kill"
         process_utils.kill_process_tree(self._process)
 
+    def inject_fault(self, mode: str) -> None:
+        fault_injector.inject_fault(mode=mode)
```
fzyzcjy, line 38: [P1] Kill the rollout subprocess rather than its supervisor

When this targets a CommandWorker-based rollout cell, `fault_injector.inject_fault` runs inside the Ray `CommandActor` rather than the SGLang process it supervises. A `deadlock` leaves the engine serving, while fatal modes kill the supervisor and Linux PDEATHSIG covers only its direct child, so SGLang descendants can remain alive holding GPUs or ports and prevent recovery. This remains in deliver-1. Deliver-2 fixes the target in `6f725416d7`; the minimal forward change is to require `_process`, accept rollout `SIGKILL` only, reject other modes before this fire-and-forget RPC, and call the existing `process_utils.kill_process_tree`. Only the harness mode-selection change depends on the later fault-form split.

miles/utils/ft_utils/api_server/handles.py
```diff
@@ -134,3 +134,6 @@ async def suspend(self, cell_id: str) -> None:
     async def resume(self, cell_id: str) -> None:
         await self._worker_manager.start_cells.remote([cell_id])
         self._inference_controller.notify_cell_resumed(cell_id)
+
+    async def inject_fault(self, cell_id: str, *, mode: FailureMode, sub_index: int) -> None:
+        await self._worker_manager.inject_fault.remote(cell_id, mode=mode.value, worker_in_cell_index=sub_index)
```
fzyzcjy, line 139: [P1] Gate rollout faults on the controller lifecycle

When the background injector's GET/POST sequence overlaps `start_update_weights`–`end_update_weights`, or a colocated rollout is offloaded, this direct manager RPC bypasses `InferenceController.context_lock`; the engine can be killed mid-broadcast, and the recovery loop can relaunch it while trainer ranks own the shared GPUs. This remains in deliver-1. Deliver-2 fixes it in `403d4a3e13` by routing through a controller operation guarded by `@with_lock` and refusing inactive cells; moving that invariant forward requires a thread-safe controller/RPC seam, for which the later `RayCellOperations`/worker-handle layer is a dependency, so the commit is not directly cherry-pickable.

fzyzcjy (commented):
The patch faults the Ray supervisor instead of the rollout engine process group and bypasses rollout lifecycle serialization. Both correctness issues remain at deliver-1 and are corrected only later in deliver-2.

PR #2119, opened 2026-08-03
Generalize the ft soak harness beyond trainer cells

fzyzcjy:
Part of #1837

### DIFF
--- a/tests/e2e/ft/conftest_ft/fault_injection.py
+++ b/tests/e2e/ft/conftest_ft/fault_injection.py
@@ -50,6 +50,68 @@ class RecoveryGate:
         return [c for c in cells if cell_is_alive(c) and c["metadata"]["name"] not in self._state_of_cell_name]
 
 
+@dataclasses.dataclass(frozen=True)
+class InjectionOutcome:
+    cell_name: str
+    injection_index: int
+    recovered: bool
+    still_down: bool
+
+
+def _find_healing_end(phases: list[str], start: int) -> int | None:
+    remaining = ["Running", "Pending", "Running"]
+    for index in range(start, len(phases)):
+        if phases[index] == remaining[0]:
+            remaining.pop(0)
+            if not remaining:
+                return index
+    return None
+
+
+class PhaseHistory:
+    def __init__(self) -> None:
+        self.phases_of_cell_name: dict[str, list[str]] = {}
+        self._cell_type_of_name: dict[str, str] = {}
+        self._injection_marks_of_cell_name: dict[str, list[int]] = {}
+
+    def observe(self, cells: list[dict]) -> None:
+        for cell in cells:
+            name = cell["metadata"]["name"]
+            self._cell_type_of_name[name] = _cell_type_of(cell)
+            phases = self.phases_of_cell_name.setdefault(name, [])
+            phase = cell["status"]["phase"]
+            if not phases or phases[-1] != phase:
+                phases.append(phase)
+
+    def note_injected(self, cell_name: str) -> None:
+        phases = self.phases_of_cell_name.setdefault(cell_name, [])
+        marks = self._injection_marks_of_cell_name.setdefault(cell_name, [])
+        marks.append(max(len(phases) - 1, 0))
+
+    def injection_outcomes(self, *, cell_type: str | None = None) -> list[InjectionOutcome]:
+        outcomes: list[InjectionOutcome] = []
+        for name, marks in sorted(self._injection_marks_of_cell_name.items()):
+            if cell_type is not None and self._cell_type_of_name.get(name) != cell_type:
+                continue
+
+            phases = self.phases_of_cell_name[name]
+            next_search_start = 0
+            for injection_index, mark in enumerate(marks):
+                healing_end = _find_healing_end(phases, max(mark, next_search_start))
+                if healing_end is not None:
+                    next_search_start = healing_end
+                outcomes.append(
+                    InjectionOutcome(
+                        cell_name=name,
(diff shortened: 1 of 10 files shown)


tests/e2e/ft/conftest_ft/scenario_ft_random.py
```diff
@@ -58,19 +64,64 @@ def run_ci(
...
+
+    print(f"Random failure soak test PASSED (mode={mode}, seed={seed}, steps={num_steps})")
+
+
+def compute_injected_cell_type(ft_mode: FTTestMode) -> str | None:
+    match tuple(sorted(ft_mode.ft_components)):
+        case ("train",):
+            return "actor"
+        case ("rollout",):
+            return "rollout"
```
fzyzcjy, line 86: [P1] Crash the managed rollout engine, not its supervisor

In the new Ray `("rollout",)` mode, this sends `/inject-fault` to a rollout `CommandActor`; at deliver-1, `CommandActor.inject_fault()` crashes the supervising Ray actor while SGLang is its separately managed subprocess. EXIT/SEGFAULT therefore target the wrapper, and SIGKILL does not explicitly terminate the managed engine process group, so accepted injections do not reliably exercise engine loss and relaunch. This remains a deliver-1 blocker. Deliver-2 commit `6f725416d7` fixes it; minimally forward the subprocess-group SIGKILL behavior and restrict rollout faults to SIGKILL, without depending on the later fault-form package.

tests/e2e/ft/conftest_ft/modes.py
```diff
@@ -99,6 +109,19 @@ def total_node_gpus(self) -> int:
...
+    "colocate_dp2_cp2_rollout_ft": FTTestMode(
+        model_name=DENSE_MODEL_NAME,
+        model_hf_repo=DENSE_MODEL_HF_REPO,
+        megatron_model_type=DENSE_MODEL_TYPE,
+        num_cells=2,
+        train_gpus_per_node=4,
+        rollout_num_engines=4,
+        rollout_gpus_per_engine=1,
+        colocate=True,
+        ft_components=("rollout",),
```
fzyzcjy, line 122: [P1] Serialize colocated injection with lifecycle transitions

In this new colocated mode, the injector can observe a serving cell and then POST after `InferenceController` has begun offloading or updating weights, because deliver-1 routes fault injection directly to the worker manager outside the controller lock. Killing an engine during memory offload or shared-GPU weight transfer can fail or hang the run rather than exercise supported recovery. This remains a deliver-1 blocker. Deliver-2 commit `403d4a3e13` adds a lock-wrapped controller endpoint and rejects injection while health monitoring is paused; minimally forward those pieces through the current `_CellHandler`, without requiring the later cell-operations abstraction.

tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -71,29 +135,32 @@ def run_fault_injection_loop(
...
-        # on a later poll rather than killing the last live replica.
-        alive = gate.genuinely_alive(cells)
-        if len(alive) <= 1:
-            logger.info("Deferring injection: %d genuinely-alive cell(s), need >1 to keep a live replica", len(alive))
+        # Keep >=1 cell of each kind genuinely alive: if a prior injection has not recovered yet, wait
+        # and retry on a later poll rather than killing that kind's last live replica.
+        alive_of_type: dict[str, list[dict]] = {}
+        for cell in gate.genuinely_alive(cells):
+            alive_of_type.setdefault(_cell_type_of(cell), []).append(cell)
+        spare_types = sorted(kind for kind, kind_cells in alive_of_type.items() if len(kind_cells) > 1)
```
fzyzcjy, line 155: [P1] Require serving rollout replicas before injection

After a rollout cell relaunches, `StatePendingWeights` can report `Healthy=True` while `Serving=False`, but this eligibility map treats it as both a candidate and a spare. If the next exponential deadline is short, the injector can crash that recovering cell again or use it to justify killing the last router-serving sibling, leaving generation with no usable engine. This remains a deliver-1 blocker. Deliver-2 commit `548adf847d` requires quiescent Healthy-and-Serving replicas; the minimal forward-port can condition rollout eligibility on the `Serving` condition added by `c01380ecbf`, without taking the later event-log refactor.

fzyzcjy (commented):
The rollout-only soak neither targets nor schedules engine faults reliably, and its shutdown can race a final injection. These blockers remain at deliver-1, although deliver-2 contains fixes that can be adapted and moved forward with the dependencies described above.

tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -133,13 +224,22 @@ def start(self) -> None:
     def stop_and_join(self, *, timeout_seconds: float) -> None:
         self._stop_event.set()
         self._thread.join(timeout=timeout_seconds)
+        self._observe_final_snapshot()
```
fzyzcjy, line 227: [P2] Confirm the injector stopped before reading its witness

When training finishes while the injector's GET is in flight, the loop does not recheck `stop_event` after observing the response and may proceed to another POST; because `join(5)` is only one HTTP timeout while the remaining GET-plus-POST path can exceed it, this newly added final snapshot and the following assertions can race a thread that is still killing cells and mutating the witness. This remains in deliver-1. Deliver-2 commits `c176eedd15` and `234967e6f7` fix it; minimally recheck the stop signal after observation, use a bounded join budget longer than every in-flight operation, and assert the thread stopped before taking the snapshot—the later `PollingWorker` is optional.

PR #2120, opened 2026-08-03
Crash rollout engines from the soak harness instead of the controller

fzyzcjy:
Part of #1837


tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -171,7 +228,7 @@ def run_fault_injection_loop(
             )
             resp.raise_for_status()
             gate.note_injected(cell_name)
-            phase_history.note_injected(cell_name)
+            recovery_witness.note_injected(cell_name)
```
fzyzcjy, line 231: [P1] Serialize rollout faults with weight updates

When the new colocated mode is active, this POST can run while `start_update_weights()` has paused health checks and holds the controller lock. `_RolloutCellHandler.inject_fault()` bypasses that lock and calls `RayWorkerManager` directly, so an engine can die after its clients were snapshotted and every broadcast retry keeps using the dead client, aborting training instead of exercising recovery. Deliver-2 routes the call through a controller operation protected by the update lock and rejects offloaded cells; that implementation depends on the later independent-controller/cell-operations plumbing, so forward-port the locking contract rather than cherry-picking it.

tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -50,66 +52,121 @@ def genuinely_alive(self, cells: list[dict]) -> list[dict]:
...
+
+
+def compute_observed_cell_state(cell: dict) -> ObservedCellState:
+    phase = cell["status"]["phase"]
+    if phase == "Suspended":
+        return ObservedCellState.SUSPENDED
+    if phase == "Pending":
+        return ObservedCellState.PENDING
+    serving = any(cond["type"] == "Serving" and cond["status"] == "True" for cond in cell["status"]["conditions"])
+    return ObservedCellState.SERVING if serving else ObservedCellState.RUNNING_NOT_SERVING
```
fzyzcjy, line 72: [P1] Require Serving before counting rollout spares

When a replacement is `StatePendingWeights`, it may report `Healthy=True` while the newly exposed `Serving` condition remains false, but `RecoveryGate.genuinely_alive()` and `alive_of_type` still use only `cell_is_alive()`. The rollout soak can therefore re-admit it before the update window completes, count a non-serving engine as the spare, and eventually crash the last router-serving replica. Deliver-2 waits for stable all-Serving quiescence; the signal added here is already sufficient for a minimal port, so the later event-log refactor is not a dependency.

tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -50,66 +52,121 @@ def genuinely_alive(self, cells: list[dict]) -> list[dict]:
...
+    for event in events:
+        if event.kind == "injected":
+            pending.append(_RecoveryStage.AWAITING_RELAUNCH)
+            continue
+        if not pending:
+            continue
+        if pending[0] is _RecoveryStage.AWAITING_RELAUNCH and event.state in _RELAUNCH_STATES:
+            pending[0] = _RecoveryStage.AWAITING_SERVING
+        elif pending[0] is _RecoveryStage.AWAITING_SERVING and event.state is ObservedCellState.SERVING:
+            pending.pop(0)
```
fzyzcjy, line 159: [P2] Coalesce crashes before the next Serving state

With events `injected -> Pending -> injected -> Pending -> Serving`, only `pending[0]` is examined: the second Pending is ignored while the first debt awaits Serving, and the final Serving pops only the first debt. The cell is actually serving after its newest relaunch, yet the second injection remains unfinished and the strict assertion fails. Deliver-2 resets the stage on every injection and discharges all outstanding pre-serve crashes on the next valid Serving observation; that state-machine change can be applied directly without its later package split.

tests/e2e/ft/conftest_ft/fault_injection.py
```diff
@@ -230,7 +287,7 @@ def _observe_final_snapshot(self) -> None:
         cells = list_cells(base_url=self._base_url, cell_type=self._cell_type)
         if cells is None:
             return
-        self.phase_history.observe(cells)
+        self.recovery_witness.observe(cells)
```
fzyzcjy, line 290: [P2] Stop the injector before checking recovery

If shutdown is requested while the 5-second GET is in flight, the loop never rechecks `stop_event` after `list_cells()`, so it can record one final crash; this five-second join can also return while the POST or thread is still running. `_observe_final_snapshot()` and the unfinished-recovery check then race the writer or fail because no polling cycle remains to witness recovery. Deliver-2 rechecks after observation, waits longer than any in-flight injection, and asserts the worker stopped; none of those changes depends on the later injector package.

tests/e2e/ft/test_rollout_ft_random_colocate_dp2_cp2.py
```diff
@@ -0,0 +1,25 @@
...
+        "stage-c-8-gpu-h200 lane, then drop this argument -- nothing in the test itself is known broken. "
+        "Until then tests/fast/e2e/ft/test_rollout_ft_gated_recovery.py is the fast-layer stand-in for this entry "
+        "(suspend -> gated relaunch -> recovery, no GPU). See tests/e2e/ft/README.md."
+    ),
+)
+
+_MODE: str = "colocate_dp2_cp2_rollout_ft"
+
+if __name__ == "__main__":
+    run_ci(_MODE)
```
fzyzcjy, line 25: [P2] Crash the engine subprocess, not its supervisor

In this newly enabled rollout mode, the endpoint forwards SIGKILL, EXIT, or SEGFAULT to `CommandActor.inject_fault()`, which currently executes them in the Ray supervisor process; the SGLang child dies only as a consequence of parent death, so the soak skips the production-shaped child-death and babysitter path it claims to cover. Deliver-2 signals `CommandActor._process` instead and exposes only SIGKILL for rollout cells; the minimal port uses the existing process handle and `process_utils` and does not require the later fault-form package.

fzyzcjy (commented):
The external rollout injector is not coordinated with the weight-update critical section, and the harness can select unavailable replicas, mis-account recoveries, race shutdown, and crash the wrong process. These defects persist through deliver-1, although deliver-2 contains semantic fixes that require selective forward-porting.

PR #2123, opened 2026-08-03
refactor(session): extract helpers shared by v1 and v2

guapisolo:
## Summary

Extract the session primitives shared by linear v1 and tree-serving v2.

## Motivation

Session v2 needs the stable token, serialization, and record-building behavior already exercised by v1, while the default v1 path must keep the same observable behavior.

## Before / After

- **Before / After:** Shared behavior lived inside the linear implementation; it now lives in reusable session helpers while v1 retains its existing call flow.
- **What moved where:** Token and record primitives move into `miles/rollout/session/core.py`.
- Linear trajectory policy remains in `linear_trajectory.py`.

## Behavior Preservation

- **How we know:** `test_sessions_v1_pins.py` adds byte-exact HTTP characterization of the v1 session wire behavior.

## Verification

- **Existing test suite:** The verified CPU suite passed with 269 tests.

## Review Focus

- Scrutinize shared primitive ownership in `core.py`.
- Scrutinize retained v1 policy in `linear_trajectory.py`.
- Scrutinize `test_sessions_v1_pins.py` for exact coverage of the preserved v1 wire behavior.

PR #2124, opened 2026-08-03
feat(session): add the trajectory-tree data model

guapisolo:
> **Depends on:** #2123

## Summary

Add an append-only trajectory forest with deterministic attach-point search.

## Motivation

Session v2 must retain multiple complete trajectories instead of forcing retries and alternative continuations into one linear history.

## Usage

Use `SessionTree.create_node(parent, ...)` to append a root or child generation, then call `find_attach_point(request_messages)` to select the deepest matching path.

## Design Notes

- **Key choices:** Every node stores a full root-to-node token snapshot.
- Commit sequence `seq` is the deterministic ordering key.
- `MAX_NODES=1024` bounds one session forest.

## Verification

- **Tests added:** `test_tree_trajectory.py` covers the trajectory-tree data-model contract.

## Review Focus

- Scrutinize `SessionTree.find_attach_point` for deepest-prefix and latest-sequence tie-breaking.
- Scrutinize `TrajectoryNode.completion_span` as the loss-mask boundary carried into later layers.

### DIFF
--- a/miles/rollout/session/v2/tree_trajectory.py
+++ b/miles/rollout/session/v2/tree_trajectory.py
@@ -1,0 +1,139 @@
+"""Session-as-forest data model: trajectory nodes and attach-point search.
+
+The forest is append-only and ``seq`` (commit order) is the only ordering
+key. Everything here is synchronous pure data — serving policy,
+concurrency, and tokenization live one layer up in ``session_state``.
+"""
+
+from dataclasses import dataclass, field
+from typing import Any
+
+from miles.rollout.session.types import SessionRecord
+from miles.utils.chat_template_utils import message_matches
+
+MAX_NODES = 1024
+
+
+@dataclass
+class TrajectoryNode:
+    """End of one model generation (SessionRecord is 1:1 with the node)."""
+
+    delta_messages: list[dict[str, Any]]
+    token_ids: list[int]  # full root->node snapshot
+    completion_span: tuple[int, int]  # this node's sampled completion within token_ids
+    seq: int  # per-session logical commit order — THE ordering key
+    committed_at: float  # wall clock, decoration only (NTP-unsafe; never order by this)
+    response_id: str  # upstream response id: the agent-branch <-> leaf join key
+    record: SessionRecord
+    finish_reason: str
+    parent: "TrajectoryNode | None" = None
+    children: list["TrajectoryNode"] = field(default_factory=list, repr=False)
+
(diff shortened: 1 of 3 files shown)


miles/rollout/session/v2/tree_trajectory.py
```diff
@@ -0,0 +1,139 @@
...
+            ):
+                i += 1
+            best_overlap = max(best_overlap, offset + i)
+            if i < len(delta):
+                return  # partial delta: this node (and its subtree) is not a candidate
+            matched = offset + len(delta)
+            if matched > best_matched or (matched == best_matched and best is not None and node.seq > best.seq):
+                best, best_matched = node, matched
+            for child in node.children:
+                visit(child, matched)
```
Shi-Dong: Won't this result in a recursion-limit exceeded error if there are too many nodes?

miles/rollout/session/v2/tree_trajectory.py
```diff
@@ -0,0 +1,139 @@
...
+    ) -> TrajectoryNode:
+        if len(self.nodes) >= MAX_NODES:
+            raise ValueError(
+                f"node cap reached ({MAX_NODES}): the session cannot branch or extend "
+                f"further — this almost always means the harness is not replaying "
+                f"history verbatim"
+            )
+        node = TrajectoryNode(
+            delta_messages=list(delta_messages),
+            token_ids=token_ids,
```
Shi-Dong: Why not `list(token_ids)`?

miles/rollout/session/v2/tree_trajectory.py
guapisolo: Fixed in `ca5ab7e`. I replaced recursive attach-point traversal with an explicit stack, so a valid `MAX_NODES=1024` chain no longer depends on Python's recursion limit. I also added a regression test that lowers the recursion limit below the legal tree depth.

miles/rollout/session/v2/tree_trajectory.py
guapisolo: Fixed in `ca5ab7e`. `create_node` now stores `list(token_ids)`, so later mutations to the caller's list cannot change the node's root-to-node token snapshot. Since token IDs are integers, a shallow list copy is sufficient. I also added a regression test for this ownership boundary.

Shi-Dong (approved):
LGTM!

PR #2125, opened 2026-08-03
feat(session): add always-branch session state

guapisolo:
> **Depends on:** #2124

## Summary

Add session state that turns mismatched histories into branches instead of destructive retries.

## Motivation

Alternative agent continuations need to coexist as complete trajectories; positioning a new request must not delete or overwrite an earlier branch.

## Usage

Call `position_for_request(state, messages)` to select the deepest matching parent, then pass that parent to `commit_generation(...)` to append the new generation.

## Design Notes

- **Key choices:** `active_leaf` exposes one linear view over the retained forest.
- Commits append beneath the parent captured before generation.
- A truncated node rejects extension with `TruncatedGenerationError`.

## Verification

- **Tests added:** `test_session_state.py` covers the complete always-branch state contract.

## Review Focus

- Scrutinize `position_for_request` for branch selection without tree mutation.
- Scrutinize `commit_generation` for append-only concurrency semantics.
- Scrutinize `prepare_pretokenized` for byte-exact prefix reuse.

### DIFF
--- a/miles/rollout/session/v2/session_state.py
+++ b/miles/rollout/session/v2/session_state.py
@@ -1,0 +1,214 @@
+"""Session state over the trajectory tree: always-branch serving.
+
+This module is the serving policy on top. A request is never rejected for mismatching stored history and never
+destroys anything: it attaches at the deepest matching node, its unmatched
+suffix becomes the new branch's delta, and whatever is not a clean
+extension just grows a sibling or a new root. Whether a branch was a retry
+is decided later by the sample_picker, not here.
+
+Concurrency contract: single lock on the whole tree. A commit only appends a new
+node under the parent captured at positioning time, so concurrent
+generations from the same spot become sibling nodes instead of a conflict.
+"""
+
+import asyncio
+import logging
+import uuid
+from dataclasses import dataclass, field
+from typing import Any
+
+from miles.rollout.session.errors import MessageValidationError, TokenizationError, TruncatedGenerationError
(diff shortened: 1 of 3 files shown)


tests/fast/router/test_session_state.py
```diff
@@ -0,0 +1,934 @@
+"""Unit tests for SessionRegistryV2 and LinearTrajectory.
```
Shi-Dong: Are we still testing `LinearTrajectory` here?

miles/rollout/session/v2/session_state.py
```diff
@@ -0,0 +1,214 @@
...
+    _validate_suffix_roles(request_messages[len(stored) :], tito_tokenizer, assistant_exempt=True)
+    return tito_tokenizer.merge_tokens(
+        old_messages=stored,
+        new_messages=request_messages,
+        pretokenized_token_ids=parent.token_ids,
+        tools=tools,
+    )
+
+
+def _validate_suffix_roles(
```
Shi-Dong, line 113: Are the allowed roles checked again in `tito_tokenizer`?  What if the `assistant` role is exempted here, but not allowed in `tito_tokenizer` (this is an unlikely edge case but might be worth a guardrail)?

Shi-Dong (approved):
LGTM. Minor comments.

miles/rollout/session/v2/session_state.py
guapisolo: This is a deprecated impl. Will remove

PR #2131, opened 2026-08-03
ci(docker): rebuild scheduled images at least once every 24h

Zhichenzzz:
## What

Scheduled Docker builds (every 12h) only fire when sglang / Megatron-LM / miles-wheels move. When upstream is quiet the image is never rebuilt, so `radixark/miles:dev` silently drifts behind the miles repo itself (miles commits deliberately do not trigger rebuilds — that would rebuild far too often).

This adds a staleness fallback to `check-upstream`: if the last triggered build is **≥ 24h** old, `should_build` flips to true even with no upstream changes.

## How

The upstream-SHAs cache file gains a 6th line: the epoch when the build was last kicked off. The cache is already saved only when a build triggers, so the timestamp naturally tracks the last actual build decision — no extra API calls, no new state. Old 5-line caches (or a garbled line) read as "no timestamp" and trigger one rebuild, which migrates the file.

Upper bound discussed in the build channel; one day per @guapisolo.

## Behavior matrix

| upstream changed | last build age | builds? |
|---|---|---|
| yes | any | yes (unchanged) |
| no | < 24h | no (unchanged) |
| no | ≥ 24h | **yes (new)** |
| no | no timestamp (old cache) | **yes, once (migration)** |

Manual `workflow_dispatch` / push-to-main builds don't refresh the timestamp (they never touched this cache); worst case the next scheduled run rebuilds within a day of a manual build.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -106,12 +106,14 @@ jobs:
             LAST_WHEELS_CU13_X86=$(sed -n '3p' "$CACHE_FILE")
             LAST_WHEELS_CU13_ARM64=$(sed -n '4p' "$CACHE_FILE")
             LAST_WHEELS_CU12_X86=$(sed -n '5p' "$CACHE_FILE")
+            LAST_BUILD_EPOCH=$(sed -n '6p' "$CACHE_FILE")
           else
             LAST_SGLANG=""
             LAST_MEGATRON=""
             LAST_WHEELS_CU13_X86=""
             LAST_WHEELS_CU13_ARM64=""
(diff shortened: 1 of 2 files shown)


guapisolo:
use doc-dev to modify docs

Zhichenzzz:
Done — updated `docs/ci/02-docker-build.md` (the doc-dev doc governing the docker build): the check-upstream paragraph now describes the 24h staleness bound and the trigger table row reads "build if upstream moved or last build ≥ 24h ago".

guapisolo (approved):
LGTM

PR #1913, opened 2026-07-28
E2B sandbox backend (E2B Cloud / self-hosted AgentENV) + dedicated AgentENV recipe

nblintao:
## What

A third materialization of the per-task sandbox model (after the shared docker-mode server and the Daytona leg, #1710/#1711/#1675): `OPENENV_SANDBOX_BACKEND=e2b` runs every episode in its own sandbox on **any provider speaking the E2B API**, driven through the official `e2b` SDK.

The backend this leg is built for is **[AgentENV](https://github.com/kvcache-ai/AgentENV)** — the Kimi K3 team's self-hosted Firecracker microVM platform, whose native API *is* the E2B API (`OPENENV_SANDBOX_BACKEND=agentenv` is an accepted alias). Pointing `E2B_API_URL`/`E2B_SANDBOX_URL` at an AgentENV gateway gives snapshot-backed sub-second sandbox starts and removes the two Daytona scaling walls we've measured (org snapshot quota; org memory quota at ~256 concurrent). E2B Cloud works unmodified as the zero-infra default.

## How

- `tb2_sandbox_e2b.py` — E2B materialization of the (unchanged, provider-agnostic) `tb2_sandbox_recipe`. E2B separates template build from sandbox create, so unlike Daytona's declarative per-episode build: templates build once per task under a **recipe-digest alias** (`tb2-<task>-<sha256[:10]>` over base image + build commands — recipe edits re-bake automatically, identical inputs reuse), episodes warm-start from the template, and a bake CLI (`python tb2_sandbox_e2b.py --tasks-dir ... --all`) pre-builds a suite. TTL dead-man's switch = E2B kill-on-expiry timeout re-armed by a keepalive thread.
- `openenv_e2b_agent_function.py` — episode-dispatch twin of the Daytona agent function: same `_multi_turn` loop, `native_evaluate=True` scoring, create semaphore + jittered backoff, cancel-safe orphan reaper. `OPENENV_E2B_THROTTLE_PATTERNS` extends retryable capacity errors per deployment without code changes.
- Launcher: implicit "tasks_dir ⇒ Daytona" becomes an explicit `resolve_sandbox_backend()` (**unchanged default**: tasks_dir alone still selects Daytona); key-supply + SDK-preflight blocks parameterized per provider (`E2B_API_KEY`/`E2B_API_KEY_FILE`, same file-path-forwarding contract).
[...]

PR #2136, opened 2026-08-03
openenv/tbench2: drop the daytona bake CLI, which nothing can consume

nblintao:
## What

Deletes the named-snapshot bake CLI from `examples/experimental/openenv/tb2_sandbox_daytona.py` (`snapshot_name`, `bake`, `main`, the `__main__` guard, and the three imports that existed only for them). Nothing else changes; the daytona rollout leg is untouched.

## Why it can't be used

The per-task sandbox refactor made this leg declarative: an episode creates from an `Image` definition and the definition hash is the cache key. The bake CLI instead registers a **named snapshot** — and no create on this leg can reference one:

- `create_task_sandbox()` passes `CreateSandboxFromImageParams(image=...)`; consuming a snapshot needs the other param type, `CreateSandboxFromSnapshotParams(snapshot=...)` (daytona SDK 0.198.0).
- It always will, because dodging the org-level snapshot quota is the stated reason the declarative path exists — our shared org has sat within two snapshots of its 500 ceiling.

So the "optional warm cache" in its docstring warms nothing the trainer touches. It burns quota and needs a live API key to do it.

## Confirming nobody uses it

- No caller in the repo: the only importers of the module are `openenv_daytona_agent_function.py` (`make_daytona` + `create_task_sandbox`) and the module's own tests, which cover labels / API-key resolution / TTL-keepalive and never `bake`.
- Not documented: the openenv README covers the daytona leg's env vars and the *e2b* bake CLI, never this entry point — in any revision, on any branch.
- Not in any runbook: grepped every remote branch's markdown plus my TB2 notes. The only other `bake`-style tooling is harbor's own, unrelated `harbor/environments/daytona/snapshots.py`.
- Provenance: it landed in #1710 together with the declarative path — a carry-over of the pool era's named-snapshot habit, never wired up.

If we ever want to pre-warm this leg, it's one declarative create per task (the same path a rollout takes), not a snapshot registration. That's a separate change and not obviously worth it.

## Test

`pytest examples/experimental/openenv/tests/` → 21 passed, 1 skipped (this suite is outside the repo-level `testpaths`, so CI doesn't run it). ruff / black / isort clean.

## Interaction with #1913

Independent change, so it's based on `main`. #1913 (open) rewrites the deleted `main()`'s argument parsing to use a shared `recipe.add_task_selection_args`; whichever lands first, the other's rebase drops that hunk, and #1913's helper docstring then wants the singular ("the e2b bake CLI's task selection") since e2b becomes its only caller.

### DIFF
--- a/examples/experimental/openenv/tb2_sandbox_daytona.py
+++ b/examples/experimental/openenv/tb2_sandbox_daytona.py
@@ -15,16 +15,16 @@ turning that recipe into a running cloud sandbox:
       auto-stop+auto-delete TTL armed as a dead-man's switch: a keepalive
       thread beats the activity timer while the creating process lives, so
       a hard-killed caller's orphans are reclaimed instead of billing forever.
-  bake CLI (``python tb2_sandbox_daytona.py ...``)  optionally pre-register
-      named snapshots ``<prefix><task-id>`` as a warm cache.
+
+There is deliberately no bake step here: on this provider the image definition
+IS the cache key, so a create either hits the build cache or warms it, and
+nothing a create passes can name a pre-registered snapshot — registering one
+per task would only spend the org quota the declarative path exists to avoid.
 """
 
-import argparse
 import getpass
 import os
-import re
 import shlex
-import sys
 import threading
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Shi-Dong (approved):
LGTM!

PR #2139, opened 2026-08-04
[AMD] retune AMD 4-node dsv4 config

JessicaJiang-123:


guapisolo (approved):

XinyuJiangCMU:
With these changes, on 4 nodes (32x MI355X), a steady step at 8k takes ~150s for rollout and ~182s for actor training, with a full cycle time of ~7.9 min.

PR #2140, opened 2026-08-04
Fix Restore Frozen Weights

### LINKED ISSUE #1958: [OPD] Frozen SGLang teacher can resume without weights after colocated offload
We would be glad to collaborate with the MILES and SGLang maintainers on
testing and improving OPD/MOPD for large models, and to contribute fixes and
tests where useful. We are currently testing sampled-token OPD and MOPD at the
scale of
[GLM-5.2](https://huggingface.co/zai-org/GLM-5.2), a 753B-total-parameter MoE,
building on the recent MILES OPD work, including the
[OPD primitive](https://www.lmsys.org/blog/2026-07-18-opd-support-in-miles/),
[context-parallel alignment fix](https://github.com/radixark/miles/pull/1754),
and [single-teacher self-distillation example](https://github.com/radixark/miles/pull/1488).
We found a frozen-teacher lifecycle issue that is easy to miss because the
server and optimizer remain healthy.

## Observation

In a colocated setup with an updatable SGLang student and a frozen external
SGLang teacher, all 125,610 aligned teacher token log-probabilities collected
across 2,112 samples and 44 rollouts were exactly:

```text
-log(154880) ~= -11.950406
```

Their variance was zero. This is the log-probability produced by a uniform
distribution over the vocabulary. The teacher endpoint continued returning
successful responses and training updates remained finite, so a basic health
or finiteness check did not catch the problem. The invalid signal instead
pushed the student toward lower confidence, accompanied by longer responses,
late truncations, and quality degradation.

## Root cause

The colocated lifecycle releases SGLang weight memory before Megatron training
and resumes the SGLang engines afterward:

1. Student and frozen teacher initially have valid weights in GPU memory.
2. Memory saver releases both servers' weight allocations so Megatron can use
   the GPUs for the optimizer step.
3. MILES resumes the SGLang weight allocations.
4. Actor-to-SGLang synchronization repopulates the updatable student.
5. The frozen teacher is not an actor-sync target and is not otherwise
   restored.

Consequently, the teacher process and endpoint are live after resume, but its
model tensors have not been repopulated. In our reproduction this produces
zero-equivalent logits and the exact uniform-vocabulary score above.

Current MILES `main` has an
[`onload_weights_from_disk()` helper](https://github.com/radixark/miles/blob/d3aef3cf1751a89d2383258ba818e9222b060d57/miles/ray/rollout/server_group.py#L253-L261)
for non-updatable groups with a `model_path`, but the normal
[`RolloutManager.onload()` path](https://github.com/radixark/miles/blob/d3aef3cf1751a89d2383258ba818e9222b060d57/miles/ray/rollout/rollout_manager.py#L217-L229)
only resumes memory occupation. We could not find a caller for the disk-reload
helper in this lifecycle.

We reproduced the boundary deterministically with fixed token sequences:

| Teacher lifecycle | Result |
|---|---|
| Initial checkpoint load | Finite, aligned, nonuniform scores |
| Weight release + resume without backup | Exactly uniform `-log(V)` scores |
| Reload checkpoint from disk | Healthy scores restored |
| Release + resume with teacher-only CPU backup | Healthy scores preserved |

This also rules out forced-prefill scoring, token alignment, the exported
checkpoint, and the OPD loss as causes of the uniform signal.

## Temporary workaround

Setting `enable_weights_cpu_backup: true` only for the frozen teacher preserves
the weights correctly across release/resume. For a GLM-5.2-size model,
however, the additional host-memory copy exhausted available RAM and OOM-killed
teacher-footprint workers.

Our current workaround therefore keeps the same teacher-only backup marker but
redirects it to `torch_memory_saver` process-local disk backup:

```text
teacher scoring: GPU HBM
training phase:  process-local NVMe
next scoring:    local NVMe -> GPU HBM
recovery:        canonical checkpoint on shared storage
```

The student path is unchanged and still receives updated weights directly from
Megatron. The disk-backed teacher path has passed release/resume probes, a real
teacher-scored optimizer update, and longer quality runs with nonuniform scores.

We also added fail-fast checks before OPD post-processing for:

- sampled-token ID misalignment;
- missing or non-finite teacher scores;
- scores at the `-log(vocabulary_size)` signature;
- collapsed teacher-score variance.

## Suggested upstream behavior

It would be useful for MILES to make restoration explicit for every offloaded
server group:

- Updatable groups are repopulated by actor weight synchronization.
- Frozen groups must restore from a configured CPU/disk backup or reload their
  `model_path` before serving.
- Resume should fail closed when a frozen group has no valid restoration
  source, rather than exposing a live endpoint with missing weights.

An explicit per-group policy such as `weights_backup_mode = cpu | disk | reload`
would support large teachers where host RAM is insufficient. A colocated test
with one updatable student and one frozen teacher could compare fixed-sequence
scores before and after a complete offload/onload cycle.

We can contribute our lifecycle reproducer, fail-fast score checks, and the
disk-backed adapter as a starting point.

## References

- [MILES: On-Policy Distillation as a Primitive](https://www.lmsys.org/blog/2026-07-18-opd-support-in-miles/)
- [MOPD](https://arxiv.org/abs/2606.30406)
- [MILES OPD context-parallel alignment fix #1754](https://github.com/radixark/miles/pull/1754)
- [MILES single-teacher self-distillation example #1488](https://github.com/radixark/miles/pull/1488)
- [MILES teacher-scoring temperature change #1496](https://github.com/radixark/miles/pull/1496)

austin362667:
### Issue:

Fix https://github.com/radixark/miles/issues/1958

As @zhoutong-hai pointed out,
the frozen teacher weight didn't restore correctly and the health check didn't catch this.

### Solution:

Guard the restore behaviors by checking a per-server-group config `weights_backup_mode`.

This approach avoid putting fail-fast assertion (the heuristic `log(vocab_size)` observation) in the hot path (every `onload()` after `offload()`).

### Tests:

1. `TestFrozenGroupWeightsBackupMode` & `TestWeightsBackupModeResolveValidation`
    - run ```pytest tests.fast.ray.rollout.test_config_matrix```
    - screenshot: Screenshot 2026-08-03 at 11 48 00 PM

2. `TestSglangConfigUpdateWeights`
    - run ```pytest tests.fast.utils.test_sglang_config```
    - screenshot: Screenshot 2026-08-03 at 11 50 33 PM

### DIFF
--- a/miles/backends/sglang_utils/sglang_config.py
+++ b/miles/backends/sglang_utils/sglang_config.py
@@ -58,9 +58,7 @@ class ModelConfig:
                         pushed by training) is only valid for updatable models.
                         Frozen colocated models must pick ``cpu`` (host RAM
                         backup), ``disk`` (process-local backup), or ``reload``
-                        (re-read ``model_path``).  Required — there is no safe
-                        default, since each option trades host RAM against
-                        per-rollout restore latency.
+                        (re-read ``model_path``).
     """
 
     name: str
@@ -182,6 +180,7 @@ class SglangConfig:
                     num_gpus_per_engine=m.get("num_gpus_per_engine"),
                     server_groups=groups,
                     update_weights=m.get("update_weights"),
+                    weights_backup_mode=m.get("weights_backup_mode"),
                 )
             )
         return SglangConfig(models=models)
--- a/miles/backends/sglang_utils/sglang_config.py
+++ b/miles/backends/sglang_utils/sglang_config.py
@@ -35,7 +35,7 @@ class ServerGroupConfig:
         ), f"Invalid worker_type '{self.worker_type}', must be one of {valid_types}"
         assert self.num_gpus > 0, f"num_gpus must be > 0, got {self.num_gpus}"
 
-WEIGHTS_BACKUP_MODES = ("actor_sync", "cpu", "disk", "reload")
+WEIGHTS_BACKUP_MODES = ("actor_sync", "cpu", "reload")
 
 @dataclasses.dataclass
 class ModelConfig:
@@ -57,7 +57,7 @@ class ModelConfig:
                         colocated offload/onload cycle.  ``actor_sync`` (weights
                         pushed by training) is only valid for updatable models.
                         Frozen colocated models must pick ``cpu`` (host RAM
-                        backup), ``disk`` (process-local backup), or ``reload``
+                        backup), or ``reload``
                         (re-read ``model_path``).
     """
 
--- a/miles/backends/sglang_utils/sglang_config.py
+++ b/miles/backends/sglang_utils/sglang_config.py
@@ -35,6 +35,7 @@ class ServerGroupConfig:
         ), f"Invalid worker_type '{self.worker_type}', must be one of {valid_types}"
         assert self.num_gpus > 0, f"num_gpus must be > 0, got {self.num_gpus}"
 
+WEIGHTS_BACKUP_MODES = ("actor_sync", "cpu", "disk", "reload")
 
 @dataclasses.dataclass
 class ModelConfig:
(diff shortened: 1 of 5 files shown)


zhoutong-hai:
Thanks for picking this up—the explicit restoration policy is the right direction. One issue: weights_backup_mode: disk appears to be accepted by validation but does not currently activate a restoration mechanism. Startup only handles cpu, while ServerGroup only explicitly restores for reload; disk therefore seems to fall through after resume_memory_occupation.

This matters for GLM-5.2 from my previous run because CPU backup exhausts host memory, while repeatedly reloading the full checkpoint is expensive. Could we either implement the process-local disk/NVMe backup path end-to-end or temporarily remove disk from the accepted modes?

austin362667:
@zhoutong-hai Thanks for reviewing!

Totally agree, I just removed the `disk` option.
Let's add it back when we have `disk/NVMe` option supported.

We'd better work on a separate PR for complete disk-based/NVMe-based adapter.
(It might be bigger and require careful design, I'm thinking we should add disk-based first then NVMe-based in two phases)

I'd love to help! Feel free to let me know how you think.

PR #2143, opened 2026-08-04
Add a FIFO ticket lock over a torch.distributed store

fzyzcjy:
Part of #1837


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: #2144 retains the engine lock for the updater lifetime, resolving the rank-0 TCPStore lifetime concern on the production path; this remains true through the checked #2649 tip.

PR #2144, opened 2026-08-04
Guard the broadcast weight update with a store ticket lock to remove Ray based locks

fzyzcjy:
Part of #1837


guapisolo (approved):
Reviewed in the #1837 stack context. The failure-recovery concern is covered by the cell teardown path in [#2154](https://github.com/radixark/miles/pull/2154/files) and hardened in [#2173](https://github.com/radixark/miles/pull/2173/files); both land before the #2176 safe-merge checkpoint and remain intact through tip #2649. No blocking issues remain.

PR #2146, opened 2026-08-04
Restructure the v1 train group around one keyword dispatch path

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No regression from this op remains at deliver-1. The intermediate v1 `export_hf` positional-dispatch break is transient: deliver-1 removes v1 and provides a keyword-based v2 export path.

PR #2147, opened 2026-08-04
Dispatch to trainer actors with keyword arguments only

fzyzcjy:
Part of #1837


fzyzcjy (commented):
All production call sites in scope were converted consistently, and the forwarded keyword names match the trainer actor method signatures. No correctness or regression blocker persists at deliver-1.

PR #2148, opened 2026-08-04
Fail fast and preserve the cause when no trainer cell can recover

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/train/group.py
+++ b/miles/ray/train/group.py
@@ -26,7 +26,7 @@ from miles.utils.audit_utils.witness.allocator import WitnessIdAllocator, read_p
 from miles.utils.ft_utils.health_checker import ActivenessTracker, NoopHealthChecker, SimpleHealthCheckerConfig
 from miles.utils.ft_utils.indep_dp import IndepDPInfo
 from miles.utils.megatron_args_utils import compute_megatron_world_size_except_dp
-from miles.utils.retry_utils import retry
+from miles.utils.retry_utils import NonRetryableError, retry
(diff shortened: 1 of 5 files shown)


miles/ray/train/group.py
```diff
@@ -200,14 +200,15 @@ def _log_step_end_event(self, *, rollout_id: int, snapshot_alive_cells: list, re
...
+    def _check_train_one_attempt(self, snapshot_alive_cells, results):
         outcomes = RayTrainGroup._compute_attempt_outcomes(snapshot_alive_cells, results)
         if not outcomes["normal"] and not outcomes["discarded"]:
             log_structured(
                 logger.error, tag="ft", op="check", **outcomes, decision="retry", reason="all alive cells failed"
             )
-            raise RuntimeError("All cells failed in this training attempt")
+            raise self._make_all_cells_failed_error("All cells failed in this training attempt") from _first_exception(
+                results
+            )
```
fzyzcjy, line 211: [P2] Match the terminal failure log to the actual decision

When every live trainer cell fails and no recoverable cell remains, this helper returns `NonRetryableError`, which `retry()` re-raises immediately, but the branch has already emitted `decision=retry`. No subsequent retry or giving-up record is produced, so the structured FT log contradicts the terminal control flow. Derive the logged decision from the selected exception; this mismatch remains in deliver-1 (`deec6fce05`) and deliver-2 (`f5a18edf2a`).

fzyzcjy (commented):
The op's recoverability issue is transient and repaired before deliver-1. However, the terminal all-cell path still emits an incorrect retry decision in both safe delivers.

PR #2149, opened 2026-08-04
Report lifecycle failures instead of silently continuing without cells

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The production changes correctly report loss of the last live trainer cell while preserving partial-failure tolerance and the train-specific outcome handling. No blocker persists at the target safe deliver.

PR #2150, opened 2026-08-04
Return one train step output type from every trainer backend and scenario

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this commit remains at deliver-1. Subsequent return-value and external-data wiring is intentionally supplied by later operations and is outside this operation's scope.

PR #2151, opened 2026-08-04
Return one train result per worker from the cell based train group

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The production changes correctly return the successful attempt's per-worker results, and the generic retry return does not regress existing callers. No in-scope blocker remains at deliver-1.

PR #2152, opened 2026-08-04
Support external data in the cell based train group

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/train/cell.py
+++ b/miles/ray/train/cell.py
@@ -18,6 +18,7 @@ from miles.ray.train.cell_state import (
 from miles.utils.ft_utils.api_server.models import CellStatus
 from miles.utils.ft_utils.health_checker import BaseHealthChecker
 from miles.utils.ft_utils.indep_dp import IndepDPInfo
+from miles.utils.retry_utils import NonRetryableError
 from miles.utils.tracking_utils.structured_log import log_structured
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


miles/ray/train/cell.py
```diff
@@ -74,6 +75,29 @@ async def init(
...
+    async def train(
+        self,
+        *,
+        rollout_id: int,
+        rollout_data_ref,
+        witness_info,
+        attempt: int,
+        external_data: list | None = None,
+    ) -> list:
+        if external_data is not None and len(external_data) != len(self._get_actor_handles()):
```
fzyzcjy, line 87: [P2] Preserve non-list external-data broadcasts

When `external_data` is a single shared payload, which the existing v1 `RayTrainGroup.train` explicitly supports by broadcasting every non-list value, this `len()` call raises `TypeError` for objects such as `Box` or `TrainStepOutput`; because the cell remains alive, the group can retry the deterministic failure up to 30 times. Deliver-1 deletes v1 but retains this code, so handle non-list payloads as broadcasts before validating per-worker lists, consistent with `.claude/rules/general-code-style.md:14-15`.

fzyzcjy (commented):
Deliver-1 removes the v1 trainer while retaining the new list-only dispatch, so callers using the previously supported broadcast payload form fail or exhaust retries. This compatibility regression remains unresolved at the target deliver.

PR #2153, opened 2026-08-04
Support multi-LoRA adapter reconciliation in trainer v2

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Within the requested production scope, adapter reconciliation remains functional at deliver-1 and deliver-2, with no correctness or regression blocker identified.

PR #2154, opened 2026-08-04
Make the cell based train group the only trainer implementation

fzyzcjy:
Part of #1837


miles/ray/train/group.py
```diff
@@ -289,6 +289,13 @@ async def save_model(self, rollout_id: int, force_sync: bool = False):
             max_attempts=_RETRY_MAX_ATTEMPTS,
         )
 
+    async def export_hf(self, rollout_id: int, path: str):
+        """Export current weights as an HF checkpoint. Only cell 0 exports to avoid file write conflicts."""
+        await retry(
+            lambda _: self._execute_first_alive("export_hf", rollout_id=rollout_id, path=path),
```
fzyzcjy, line 295: [P1] Do not kill trainer cells on optional export failures

At a nonfinal staged snapshot eval, a converter or I/O error—or FSDP's unsupported `export_hf`—runs through `RayTrainCell.execute` with `kill_on_failure=True`, stopping the trainer before `EvalDispatcher` catches the error and skips the eval. The next training step then has no live cell; use non-destructive dispatch for this optional export. This remains in deliver-1 and deliver-2.

miles/utils/arguments.py
```diff
@@ -3236,11 +3236,6 @@ def miles_validate_args(args):
         assert (
             args.megatron_to_hf_mode != "bridge"
         ), "Critic models are not supported with --megatron-to-hf-mode bridge"
-        assert not enable_experimental_ft_trainer(), (
-            "Shared Actor/Critic PPO is not supported with MILES_EXPERIMENTAL_FT_TRAINER=1: the v2 "
-            "fault-tolerant train group cannot route critic values or lifecycle options yet. "
-            "Unset MILES_EXPERIMENTAL_FT_TRAINER or use a non-PPO advantage estimator."
-        )
         assert args.kl_coef == 0, (
```
fzyzcjy, line 3282: [P1] Reject multi-cell shared PPO during validation

With `--advantage-estimator ppo` plus `--indep-dp` or trainer fault tolerance, validation now succeeds even when multiple cells are created, but `train.py` passes critic results as `external_data` and `RayTrainGroup.train` immediately asserts that external data requires one cell. Preserve one-cell PPO support while rejecting the multi-cell case here; the missing guard and runtime assertion remain in deliver-1 and deliver-2.

fzyzcjy (commented):
The mandatory cell trainer leaves two configurations that pass validation or promise graceful degradation with runtime-fatal behavior. Both regressions persist through deliver-1 and deliver-2.

PR #2155, opened 2026-08-04
Assign the trainer master address from the driver instead of rank zero

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the deliver-1 safe point.

PR #2156, opened 2026-08-04
Extract the shared actor creation and port allocation in the worker manager

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The extraction preserves command-actor creation, port allocation, and shutdown behavior. The only direct semantic change corrects the invalid static-port cell-index access, and no regression introduced by this op survives deliver-1.

PR #2157, opened 2026-08-04
Give worker specs a launch context and serve mode scheduling options

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at deliver-1. Context-aware worker configuration preserves existing command-worker behavior, and serve-mode integration is completed before the target safe deliver.

PR #2158, opened 2026-08-04
Launch serve mode workers from RayWorkerManager

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/ray_worker_manager.py
+++ b/miles/utils/workers/ray_worker_manager.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Generic, TypeVar
 import ray
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
+from miles.utils.function_registry import load_function
 from miles.utils.http_utils import _wrap_ipv6
 from miles.utils.ray_utils import compute_ray_pin_head_options
 from miles.utils.workers.addr_allocator import PortAllocator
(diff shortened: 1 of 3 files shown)


miles/utils/workers/ray_worker_manager.py
```diff
@@ -343,6 +345,22 @@ async def _shutdown_gracefully(self) -> None:
...
+@dataclass
+class _ServeActorManager(_BaseActorManager[ServeWorkerSpec]):
+    def _compute_remote_options(self) -> dict:
+        groups = self.spec.concurrency_groups
+        return {} if groups is None else dict(concurrency_groups=groups)
+
+    async def launch_actor(self) -> None:
+        self.actor_handle = self._create_actor(
+            load_function(self.spec.worker_class),
+            **self.spec.ctor_kwargs(self.launch_context),
```
fzyzcjy, line 377: [P2] Inject manager probes into every served actor

When `ServeWorkerSpec.worker_class` is an ordinary class without `NodeProbeMixin`, launching it unchanged here makes the subsequent port allocation fail: it unconditionally invokes `_get_node_ip`, and the default dynamic RPC port also requires `_get_free_port_block`. Deliver-1 adds the mixin only to `TrainRayActor`, so the generic serve path remains broken; deliver-2's `dc98b17dd6` fixes this with a bootstrapped subclass, whose minimal forward dependencies are the wrapper and `NodeProbeMixin` import (`functools` only if its cache is retained), rather than the rest of that commit's constructor-local refactor.

fzyzcjy (commented):
The manager launches arbitrary serve classes without supplying probe methods that its own port-allocation lifecycle requires. Deliver-1 only compensates for trainer classes, leaving the generic serve contract broken.

PR #2159, opened 2026-08-04
Add worker specs for the trainer actor and critic

fzyzcjy:
Part of #1837


miles/ray/specs/train.py
```diff
@@ -0,0 +1,152 @@
...
+    num_cells = compute_trainer_num_cells(args, role=role)
+    assert total_gpus % num_cells == 0, f"{total_gpus=} must be divisible by {num_cells=}"
+    gpus_per_cell = total_gpus // num_cells
+
+    indep_dp_store_addr = _create_indep_dp_store_addr() if num_cells > 1 else None
+
+    return ServeWorkerSpec(
+        name=compute_trainer_pool_id(role),
+        port_infos=[PortInfo(name=MASTER_PORT_NAME, static_port=9000, mode="master", allow_dynamic=True)],
+        env_var=lambda ctx: compute_trainer_env_vars(args, ctx),
```
fzyzcjy: [P1] Precompute the Blackwell FP8 default before deferring env setup

For Blackwell jobs that leave `NVTE_FP8_BLOCK_SCALING_FP32_SCALES` unset, such as the B200/B300 path in `scripts/run_qwen3_30b_a3b.py`, deliver-1 evaluates this callback inside the zero-GPU `RayWorkerManager`. The `RAY_EXPERIMENTAL_NOSET_*` values it returns apply only to the child trainer, so `default_fp8_block_scaling_fp32_scales()` sees no CUDA and emits `"1"` instead of Blackwell's required `"0"`, launching TE with the wrong scaling mode. The deferred evaluation remains in deliver-2; capture this stable hardware-derived value while constructing the spec or derive it from the assigned worker, per [`.claude/rules/general-code-style.md:22-28`](.claude/rules/general-code-style.md#L22-L28).

fzyzcjy (commented):
The trainer spec defers a hardware-dependent FP8 default into a zero-GPU manager process, causing default Blackwell runs to launch with the wrong Transformer Engine scaling setting. This regression survives both designated safe delivers.

PR #2160, opened 2026-08-04
Inline the fault-tolerance concurrency groups onto the trainer actor

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/train/actor_factory.py
+++ b/miles/ray/train/actor_factory.py
@@ -3,7 +3,7 @@ from ray.util.placement_group import PlacementGroup
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
 from miles.ray.specs.train import compute_trainer_env_vars
-from miles.utils.ft_utils.heartbeat_utils import HeartbeatStatus
+from miles.ray.train_actor import TRAINER_CONCURRENCY_GROUPS
 from miles.utils.workers.worker_spec import WorkerLaunchContext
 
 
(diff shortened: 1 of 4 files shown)


miles/ray/train/actor_factory.py
```diff
@@ -33,11 +33,10 @@ def allocate_gpus_for_actor(
 
         actor_impl = FSDPTrainRayActor
 
-    ft = args.use_fault_tolerance
     TrainRayActor = ray.remote(
         num_gpus=1,
-        **(dict(concurrency_groups={"heartbeat_status": 1, "default": 1, "fault_injector": 1}) if ft else {}),
-    )(_with_ft_concurrency_groups(actor_impl) if ft else actor_impl)
+        concurrency_groups=TRAINER_CONCURRENCY_GROUPS,
```
fzyzcjy, line 38: [P1] Restore the non-FT concurrency-group gate

When `args.use_fault_tolerance` is false, this unconditional `concurrency_groups` option turns the trainer into a threaded Ray actor, so `init` and NCCL setup/collectives run on an executor thread. This exact configuration previously deadlocked the four-rank non-FT LoRA run on scitix-72 and was deliberately gated in the pre-change code; both deliver-1 (`deec6fce`) and deliver-2 (`f5a18edf`) still carry the unconditional spec. Preserve a plain actor for non-FT runs and apply the method annotations and group declaration only on the FT launch path.

fzyzcjy (commented):
The patch removes an intentional guard and reintroduces a known multi-GPU NCCL deadlock for default non-FT training. The regression remains present at deliver-1, so the patch is not safe to deliver.

PR #2161, opened 2026-08-04
Launch trainer actors through RayWorkerManager

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Under the requested safe-deliver criterion, no correctness or regression blocker introduced by this op remains in deliver-1. The intermediate issues are repaired before the target deliver, and no forward-port from deliver-2 is needed.

PR #2162, opened 2026-08-04
Delete the trainer actor factory and the placement group plumbing

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The patch removes dead actor-factory and placement-group plumbing while updating all production callers and correcting the FSDP actor path. No correctness regression introduced here persists at deliver-1.

PR #2163, opened 2026-08-04
Restart trainer cells through RayWorkerManager

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness regression introduced by this op remains at deliver-1. The isolated compute_cell_id keyword mismatch is removed by the provider-driven trainer-cell rewrite before that safe delivery, so it is transient under the requested policy.

PR #2164, opened 2026-08-04
Drive trainer cell bookkeeping from the worker provider

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blockers introduced by this op remain at deliver-1. Breakages confined to the intermediate commit were repaired by the specified safe deliver and treated as transient.

PR #2165, opened 2026-08-04
Wait for the expected number of trainer cells at startup

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No production correctness regression introduced by this op remains at deliver-1. Intermediate cell-bookkeeping issues are repaired before that safe deliver and are therefore transient under the requested review policy.

PR #2166, opened 2026-08-04
Serve trainer cells over the api server like the rollout cells

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at deliver-1. The trainer-cell API integration is consistent with the worker-manager cell identity and lifecycle.

PR #2167, opened 2026-08-04
Key the trainer cells by cell id instead of cell index

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the target safe deliver. The cell-ID plumbing is consistent across the trainer controller, API handler, worker manager/provider, actor actions, and dashboard.

PR #2168, opened 2026-08-04
Suspend trainer cells through the worker manager like the rollout cells

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/test_utils/ft_test_actions.py
+++ b/miles/utils/test_utils/ft_test_actions.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Literal
 from pydantic import TypeAdapter
 
 from miles.utils.pydantic_utils import FrozenStrictBaseModel
+from miles.utils.workers.ray_worker_manager import RayWorkerManager
 
 if TYPE_CHECKING:
     from miles.ray.train.controller import RayTrainGroup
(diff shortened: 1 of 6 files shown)


fzyzcjy (commented):
The resume path introduces an observer race that survives deliver-1 and can prevent the intended next-rollout healing. Deliver-2 contains the minimal bounded observation wait, but deliver-1 remains blocked without moving it forward with its dependencies.

miles/utils/test_utils/ft_test_actions.py
```diff
@@ -55,10 +56,11 @@ async def run_after_step(self, rollout_id: int) -> None:
                 cell_id = action.resolve_cell_id(self._controller.cell_ids)
                 logger.info("FT test action: %s cell %s after rollout %d", action.action, cell_id, rollout_id)
 
+                worker_manager = RayWorkerManager.get_handle()
                 if action.action == "stop_cell_at_end":
-                    await self._controller.stop_cell(cell_id)
+                    await worker_manager.stop_cells.remote([cell_id])
                 elif action.action == "start_cell_at_end":
-                    self._controller.start_cell(cell_id)
+                    await worker_manager.start_cells.remote([cell_id])
```
fzyzcjy, line 63: [P1] Wait until the controller observes a resumed cell

Once deliver-1 makes trainer membership provider-driven, `start_cells` can return before the provider's next 5-second poll reconciles the resumed generation, so the following rollout may snapshot stale membership and omit the scheduled heal required by [the FT timeline](tests/e2e/ft/README.md#L200-L204). This race remains in deliver-1; deliver-2 fixes it in `da47fa6f66` with a bounded `retry_until_deadline` wait for the cell to reappear in `controller.cell_ids`. Forward only that wait after the `82d9398c7c`/`276dc69688` removal-and-explicit-ID dependencies, rather than assuming the broader later refactor can be cherry-picked.

PR #2169, opened 2026-08-04
Remove trainer cells that the provider stops reporting

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the deliver-1 boundary. The intermediate trainer-cell healing mismatch is repaired before that safe deliver.

PR #2170, opened 2026-08-04
Remove the now-unused stopped trainer cell state

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/train/group.py
+++ b/miles/ray/train/group.py
@@ -244,19 +244,28 @@ class RayTrainGroup:
     def _check_train_one_attempt(self, snapshot_alive_cells, results):
         outcomes = RayTrainGroup._compute_attempt_outcomes(snapshot_alive_cells, results)
         if not outcomes["normal"] and not outcomes["discarded"]:
-            log_structured(
-                logger.error, tag="ft", op="check", **outcomes, decision="retry", reason="all alive cells failed"
-            )
             cause = _first_exception(results)
-            raise self._make_all_cells_failed_error(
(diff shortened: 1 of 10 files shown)


miles/ray/train/group.py
```diff
@@ -385,7 +385,7 @@ def get_cell_statuses(self) -> dict[str, CellStatus]:
     # ------------------------ utils to forward calls to cells ------------------------
 
     def _is_recoverable(self) -> bool:
-        return any(cell.is_alive for cell in self._cells)
+        return any(cell.is_alive or cell.is_uninitialized for cell in self._cells)
```
fzyzcjy: [P1] Do not count uninitialized cells as recoverable

The added `is_uninitialized` branch only changes the result when no cell is alive, but `_refresh_cells` rejects exactly that state because a replacement needs a live checkpoint source. If the last alive cell fails while another cell has just been reconciled as uninitialized, `_execute_all_alive_and_catch` suppresses the failure, so `onload`, `offload`, or `clear_memory` can return successfully with zero usable cells; first-alive operations also retry only to fail with `No alive cells`. Count only alive cells unless this path actually drives healing. This line is unchanged in deliver-1 (`deec6fce05fa45684c14c38cf242f15a7038e478`), so it blocks that deliver.

fzyzcjy (commented):
The patch treats an uninitialized replacement as a recovery source precisely when no live trainer remains, even though healing requires a live source. This can mask a terminal all-cells-lost failure and remains present at deliver-1.

PR #2171, opened 2026-08-04
Remove the pending state in trainer cells

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The removed pending state was constructor-local and unobservable, and all reachable trainer-cell status behavior remains equivalent through deliver-1.

PR #2172, opened 2026-08-04
Merge the now duplicated actor and rollout cell handlers

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/ft_utils/api_server/handles.py
+++ b/miles/utils/ft_utils/api_server/handles.py
@@ -1,40 +1,36 @@
 from __future__ import annotations
 
-import abc
-import asyncio
+from typing import Protocol
 
 import ray
 
 from miles.ray.rollout.server_cell import compute_pending_rollout_cell_status
-from miles.ray.train.group import RayTrainGroup
 from miles.utils.ft_utils.api_server.models import Cell, CellCondition, CellMetadata, CellSpec, CellStatus, TriState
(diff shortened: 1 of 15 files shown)


miles/utils/ft_utils/api_server/handles.py
```diff
@@ -76,7 +55,7 @@ async def get_cell(self, cell_id: str) -> Cell:
         return self._compute_cell(
             cell_id,
             cell_infos=await self._get_cell_infos(),
-            statuses=self._group.get_cell_statuses(),
+            statuses=self._controller.get_cell_statuses(),
```
fzyzcjy, line 58: [P2] Fence rollout statuses to the current worker generation

When a rollout cell is suspended and resumed before `RayWorkerProvider`'s 5-second poll observes the stop—for example, with back-to-back API patches or `--mini-ft-controller-resume-delay < 5`—`_get_cell_infos()` describes the new live generation while this call can still return the previous generation's status. The unified handler therefore publishes a stale healthy or unhealthy verdict, causing the mini FT loop to skip or repeat recovery until reconciliation; the rollout path needs a generation fence or synchronous status invalidation. This remains unchanged in deliver-1, and deliver-2 retains the same unfenced merge.

fzyzcjy (commented):
The rollout handler now combines worker-manager liveness with an unfenced controller status, allowing a prior generation's health verdict to be attached to a newly resumed cell. The regression survives the target safe deliver.

PR #2173, opened 2026-08-04
Kill failed trainer workers through their own RPC channel instead of the manager

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No qualifying production correctness or regression issue remains at deliver-1 after excluding unit-test-only and transient issues as requested.

PR #2174, opened 2026-08-04
Add a ray worker handle with death confirmation

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this change remains at the specified deliver-1 commit. Deliver-2's positional-call and RPC death-confirmation changes support later production paths rather than repairing deliver-1 behavior.

PR #2175, opened 2026-08-04
Route all trainer worker calls through the worker handle instead of Ray handles

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The change preserves existing Ray call and worker-death semantics while migrating all production WorkerInfo consumers. No correctness regression survives at deliver-1, and no deliver-2 repair needs to be moved forward.

PR #2176, opened 2026-08-04
Rename the trainer group and cell to TrainerController and TrainerCell

fzyzcjy:
Part of #1837

ci-sglang-pr: tom/refactor-miles-repo-sglang/deliver-1


fzyzcjy (commented):
The rename is applied consistently across production call sites and type references, and the accompanying keyword change corrects an existing constructor mismatch. No correctness or regression blocker remains at the target safe deliver.

PR #2177, opened 2026-08-04
Add a self-provisioned Kubernetes apiserver test environment

fzyzcjy:
Part of #1837

PR #2178, opened 2026-08-04
Add a kind cluster test environment for Kubernetes testing

fzyzcjy:
Part of #1837

### DIFF
--- a/tests/e2e/k8s_kind/kind_cluster.py
+++ b/tests/e2e/k8s_kind/kind_cluster.py
@@ -1,0 +1,70 @@
+# doc-dev: docs/developer/reconcile-loop.md
+from __future__ import annotations
+
+import logging
+import platform
+import shutil
+import sys
+import urllib.request
+from dataclasses import dataclass
+from pathlib import Path
+
+from miles.utils.misc import exec_command
+
+logger = logging.getLogger(__name__)
+
+_READY_TIMEOUT = "180s"
+
+_KIND_VERSION = "v0.32.0"
+_KIND_CACHE_DIR = Path.home() / ".cache" / "miles" / "bin"
+
+
+@dataclass(frozen=True)
+class KindCluster:
+    name: str
+    kubeconfig: Path
+
+
+def create_cluster(*, run_id: str, kubeconfig: Path) -> KindCluster:
+    kind = _resolve_kind_binary()
+    exec_command(f"{kind} create cluster --name {run_id} --kubeconfig {kubeconfig} --wait {_READY_TIMEOUT}")
+    return KindCluster(name=run_id, kubeconfig=kubeconfig)
+
+
+def delete_cluster(cluster: KindCluster) -> None:
+    kind = _resolve_kind_binary()
+    exec_command(f"{kind} delete cluster --name {cluster.name} --kubeconfig {cluster.kubeconfig}")
+
+
+def _resolve_kind_binary() -> str:
+    installed = shutil.which("kind")
+    if installed is not None:
+        return installed
+
+    cached = _KIND_CACHE_DIR / f"kind-{_KIND_VERSION}"
+    if not cached.exists():
+        _download_kind(cached)
+    return str(cached)
+
+
+def _download_kind(destination: Path) -> None:
+    url = f"https://kind.sigs.k8s.io/dl/{_KIND_VERSION}/kind-{_host_os()}-{_host_arch()}"
+    logger.warning(f"kind is not installed, downloading it once {url=} {destination=}")
+
+    destination.parent.mkdir(parents=True, exist_ok=True)
+    partial = destination.with_suffix(".partial")
+    urllib.request.urlretrieve(url, partial)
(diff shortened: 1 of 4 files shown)


tests/e2e/k8s_kind/kind_cluster.py
```diff
@@ -0,0 +1,70 @@
...
+
+def _download_kind(destination: Path) -> None:
+    url = f"https://kind.sigs.k8s.io/dl/{_KIND_VERSION}/kind-{_host_os()}-{_host_arch()}"
+    logger.warning(f"kind is not installed, downloading it once {url=} {destination=}")
+
+    destination.parent.mkdir(parents=True, exist_ok=True)
+    partial = destination.with_suffix(".partial")
+    urllib.request.urlretrieve(url, partial)
+    partial.chmod(0o755)
+    partial.replace(destination)
```
fzyzcjy, line 71: [P1] Verify the kind binary before execution

When `kind` is absent from `PATH`, the downloaded bytes are made executable without being checked against a pinned digest. A substituted release response can therefore execute arbitrary code with the test process's credentials; pin and verify the per-platform SHA-256 before moving the file into the executable cache, as analogous downloads do in `docker/install-kube-tools.sh:10-17`.

tests/e2e/k8s_kind/kind_cluster.py
```diff
@@ -0,0 +1,70 @@
...
+
+def delete_cluster(cluster: KindCluster) -> None:
+    kind = _resolve_kind_binary()
+    exec_command(f"{kind} delete cluster --name {cluster.name} --kubeconfig {cluster.kubeconfig}")
+
+
+def _resolve_kind_binary() -> str:
+    installed = shutil.which("kind")
+    if installed is not None:
+        return installed
```
fzyzcjy, line 49: [P2] Honor the pinned kind version on PATH

When a developer or hosted runner already has any `kind` on `PATH`, this early return bypasses `_KIND_VERSION`, so the suite silently uses an arbitrary CLI and default Kubernetes image rather than v0.32.0. This makes results host-dependent and still violates the safe-deliver contract of a pinned binary (`docs/developer/reconcile-loop.md:120`); verify the installed version or always resolve the versioned cache.

tests/e2e/k8s_kind/kind_cluster.py
```diff
@@ -0,0 +1,70 @@
...
+
+@dataclass(frozen=True)
+class KindCluster:
+    name: str
+    kubeconfig: Path
+
+
+def create_cluster(*, run_id: str, kubeconfig: Path) -> KindCluster:
+    kind = _resolve_kind_binary()
+    exec_command(f"{kind} create cluster --name {run_id} --kubeconfig {kubeconfig} --wait {_READY_TIMEOUT}")
```
fzyzcjy, line 37: [P2] Quote kind's shell arguments

When the resolved executable or pytest temporary path contains whitespace—for example, `pytest --basetemp '/tmp/k8s run'`—the command runner tokenizes this formatted string, so `kind` receives split arguments and cluster creation fails; the delete command has the same problem. Pass an argv sequence or shell-quote every interpolated value.

fzyzcjy (commented):
The kind bootstrap still accepts unverified or arbitrarily versioned executables and constructs commands that fail for legitimate paths containing whitespace. These issues remain present at the target safe deliver.

guapisolo (approved):
codex cmt

Review scope: [PR #2178 changes](https://github.com/radixark/miles/pull/2178/files), frozen source head `67a80d5300e3bc2c50bd15abacf0410d782d6037` (incremental base `07850b03a629928a065e8c00772128d8495fea67`). Stack repair checkpoint: [PR #3122](https://github.com/radixark/miles/pull/3122) at `d28546aca8111ae3d6f4d477f70898725d814d54`, inclusive. The recorded source review is reused after verifying its binding; this is not a new full-stack review.

Design: The kind test layer complements the bare apiserver layer by providing the kubelet and scheduler needed to observe actual Pod lifecycle behavior.

- [P1] The new `tests/e2e/k8s_kind/kind_cluster.py` independently imports nonexistent `miles.utils.misc.exec_command`, so the kind test harness cannot collect even if the earlier apiserver import is repaired. Fixed later by [PR #3050](https://github.com/radixark/miles/pull/3050) (`eb9208b7f8b9487a561e222724ebffcd61c11357`): the harness imports the existing `run_process` helper and routes cluster commands through `_run_kind`. The repair remains present through the stated checkpoint; the source head is still broken without that later repair.

Verification: The recorded native and separate code-quality reviews match the unchanged source increment. The checkpoint kind harness imported successfully in a separate fresh Python process using real dependencies. No real kind cluster, Docker workload, or Kubernetes E2E was run. The 343 PR bindings in the #2177-through-#3122 chain and its issue-list snapshot were revalidated unchanged.

Approved in the reviewed stack context, requiring #3050 and its helper prerequisites to land with the delivery. This does not claim that the source head independently provides a working harness or that the entire delivery is approved.

PR #2179, opened 2026-08-04
Add a level-triggered reconcile loop over a source stream

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/reconcile/object_store.py
+++ b/miles/utils/workers/reconcile/object_store.py
@@ -1,0 +1,75 @@
+# doc-dev: docs/developer/reconcile-loop.md
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any
+
+from miles.utils.workers.reconcile.source_event import DeleteEvent, ObjectKey, ParentKey, SourceEvent, UpsertEvent
+
+logger = logging.getLogger(__name__)
+
+KeyMapFn = Callable[[Any], ParentKey | None]
+
+
+@dataclass(frozen=True)
+class _CachedObject:
+    obj: Any
+    parent: ParentKey
+
+
+class ObjectStore:
+    def __init__(self, *, key_map: KeyMapFn | None) -> None:
+        self._key_map = key_map
+        self._cache: dict[ObjectKey, _CachedObject] = {}
+
+    def get_by_parent(self, parent_key: ParentKey) -> list[Any]:
+        return [self._cache[key].obj for key in sorted(self._cache) if self._cache[key].parent == parent_key]
+
+    def __contains__(self, key: ObjectKey) -> bool:
+        return key in self._cache
+
+    def handle_event(self, event: SourceEvent) -> set[ParentKey]:
+        match event:
+            case UpsertEvent():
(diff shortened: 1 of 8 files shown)


miles/utils/workers/reconcile/object_store.py
```diff
@@ -0,0 +1,75 @@
...
+        self._cache[key] = _CachedObject(obj=obj, parent=parent)
+        return {parent} if previous is None else {parent, previous.parent}
+
+    def _apply_delete(self, *, key: ObjectKey, last_obj: Any) -> set[ParentKey]:
+        removed = self._cache.pop(key, None)
+        parent = removed.parent if removed is not None else None
+        if parent is None:
+            if last_obj is None:
+                logger.warning(f"ObjectStore dropping a delete it cannot attribute to a parent {key=}")
+                return set()
```
fzyzcjy, line 64: [P2] Reconcile direct-key deletes without tombstones

When `key_map=None`, an uncached `DeleteEvent` with `last_obj=None` reaches this return even though its parent is unambiguously the event key—the same class maps objects directly to `key` in this mode. The delete is therefore never enqueued, so a direct-key reconciler can miss the transition to an empty snapshot; derive the parent from `key` before requiring a tombstone.

fzyzcjy (commented):
The direct-key mode can still silently drop a tombstone-free delete, and this behavior remains present at the target safe deliver.

guapisolo (approved):
codex cmt

Review scope: [PR #2179 changes](https://github.com/radixark/miles/pull/2179/files), frozen source head `73a452e9b35c26ed1cd31b890259207e140e7282` (incremental base `67a80d5300e3bc2c50bd15abacf0410d782d6037`). Stack repair checkpoint: [PR #3122](https://github.com/radixark/miles/pull/3122) at `d28546aca8111ae3d6f4d477f70898725d814d54`, inclusive. The recorded source review is reused after verifying its binding; this is not a new full-stack review.

Design: Source events update the current object cache and deduplicated parent-key queue; reconciliation reads the latest snapshot rather than replaying every historical event.

- [P2] The new `TestStop.test_a_second_stop_is_rejected` requires `AssertionError`, while `ReconcileLoop.stop()` is idempotent. This makes the focused test suite fail deterministically for correct repeated-stop behavior; it is a test-contract defect, not evidence that idempotent teardown itself is broken. Fixed later by [PR #2330](https://github.com/radixark/miles/pull/2330) (`3f1d4b9a8f7f8d165ef913f131be765540275ecf`): the test becomes `test_a_second_stop_is_a_no_op` and checks that the source closes exactly once. The repair remains present through the stated checkpoint.

Verification: The unchanged source head's recorded focused run had 41 passed and 1 failed, with this test the only failure. The corrected second-stop test and the explicit-stop-after-aborted-start test both passed at the checkpoint in this verification. These isolated tests excluded ancestor conftest fixtures; no real Kubernetes E2E was run. The 343 PR bindings in the #2177-through-#3122 chain and its issue-list snapshot were revalidated unchanged.

Approved for this increment in the stack context, with #2330 required. This does not approve later lifecycle changes, erase the source-head test failure, or assert that the whole delivery is defect-free.

PR #2181, opened 2026-08-04
Reopen a dying source stream and gate startup on the first listing

fzyzcjy:
Part of #1837

PR #2182, opened 2026-08-04
Retry a failed key with per-key exponential backoff

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/reconcile/retry_scheduler.py
+++ b/miles/utils/workers/reconcile/retry_scheduler.py
@@ -1,0 +1,74 @@
+# doc-dev: docs/developer/reconcile-loop.md
+from __future__ import annotations
+
+import asyncio
+import math
+from collections.abc import Callable, Hashable
+from dataclasses import dataclass
+from typing import Generic, TypeVar
+
+from miles.utils.test_utils.clock import Clock
+
+KeyT = TypeVar("KeyT", bound=Hashable)
+
+POLL_INTERVAL = 1.0
+
+
+@dataclass
+class _RetryInfo:
+    failures: int
+    retry_at: float | None
+
+
+class RetryScheduler(Generic[KeyT]):
+    def __init__(
+        self,
+        *,
+        on_retry: Callable[[KeyT], None],
+        failure_base_delay: float,
+        failure_max_delay: float,
+        clock: Clock,
+    ) -> None:
+        assert failure_base_delay > 0, f"{failure_base_delay=} must be positive"
+        assert failure_max_delay >= failure_base_delay, f"{failure_max_delay=} must be >= {failure_base_delay=}"
+
+        self._on_retry = on_retry
+        self._failure_base_delay = failure_base_delay
+        self._failure_max_delay = failure_max_delay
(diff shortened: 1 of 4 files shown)


fzyzcjy (commented):
The constructor creates an unowned background task that survives cancellation during initial synchronization. This lifecycle regression persists through the stated safe deliver.

miles/utils/workers/reconcile/retry_scheduler.py
```diff
@@ -0,0 +1,74 @@
...
+        assert failure_max_delay >= failure_base_delay, f"{failure_max_delay=} must be >= {failure_base_delay=}"
+
+        self._on_retry = on_retry
+        self._failure_base_delay = failure_base_delay
+        self._failure_max_delay = failure_max_delay
+        self._max_backoff_exponent = max(0, math.ceil(math.log2(failure_max_delay / failure_base_delay)))
+        self._clock = clock
+
+        self._infos: dict[KeyT, _RetryInfo] = {}
+        self._poller = asyncio.create_task(self._poll())
```
fzyzcjy, line 42: [P2] Cancel the poller when loop startup is aborted

`RetryScheduler` starts `_poll()` before its owner can guarantee shutdown. If `ReconcileLoop.start()` is cancelled while awaiting the initial `ReplaceEvent`—for example, cancellation of `KubernetesWorkerProvider.watch_cells()` inside `AsyncExitStack.enter_async_context()`—`__aenter__` never succeeds, so `stop()` is never invoked; the cancellation handler only cancels the driver task, leaving this poller alive for the lifetime of the event loop. Start it as part of successful loop startup or shut it down on aborted startup. This remains unchanged in safe deliver `f5a18edf`.

guapisolo: codex cmt

[P2] Startup cancellation still leaves the retry poller running at #3122

Confirmed on this PR's current head (`a1671d04821e`) and through #3122 (`d28546aca811`). If the caller cancels or times out context entry before initial synchronization completes, [the startup cancellation handler](https://github.com/radixark/miles/blob/a1671d04821ed7fe3f25172e988619f898ffa027/miles/utils/workers/reconcile/loop.py#L67-L84) only cancels and joins the source driver. Failed `__aenter__()` does not invoke `__aexit__()`, so the new retry poller remains alive, retains the scheduler and queue, and wakes once per second with the default clock. Repeated attempts with new loop instances accumulate background tasks unless the caller explicitly calls `stop()` after failed entry. [The same cleanup path remains at #3122](https://github.com/radixark/miles/blob/d28546aca8111ae3d6f4d477f70898725d814d54/miles/utils/workers/reconcile/loop.py#L79-L89).

A focused probe using a source that never emits its initial listing and `await asyncio.wait_for(loop.__aenter__(), timeout=0.02)` reports `pending_after_failed_enter=["RetryScheduler._poll"]` on both heads; the preceding #2181 reports no pending tasks.

Please shut down the retry scheduler in startup cancellation cleanup before re-raising, e.g. `await self._retry.shutdown()`, and cover failed context entry without an explicit `stop()` in the test. The aborted-start test added in #2330 explicitly calls `stop()`, so it does not cover this failure path's automatic cleanup.

PR #2186, opened 2026-08-04
Document how the reconcile loop maps onto the Kubernetes Go stack

fzyzcjy:
Part of #1837

### DIFF
--- a/docs/developer/reconcile-loop.md
+++ b/docs/developer/reconcile-loop.md
@@ -1,0 +1,116 @@
+---
+title: "Reconcile Loop"
+description: "A minimal, level-triggered controller runtime for Miles, and how it maps onto the Kubernetes Go stack."
+---
+
+## What this is
+
+- A deliberately small port of the [client-go informer stack](https://github.com/kubernetes/client-go/tree/master/tools/cache) plus [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime).
+- Tracks pools of processes — SGLang engine cells, trainer cells — that appear, disappear and restart mid-run.
+- The alignment record: **every deviation from Go must appear below with a reason.**
+
+## Modules
+
+All under `miles/utils/workers/reconcile/`.
+
+| Ours | Does | Go | Not 1:1 |
+| --- | --- | --- | --- |
+| `k8s_api.py` | LIST/WATCH calls; the only module importing `kubernetes_asyncio`, lazily so the loop stays importable without a Kubernetes backend | [typed client `List` / `Watch`](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod.go) | |
+| `k8s_reflector.py` | Cursor bookkeeping, relist on cursor rejection | [`cache.Reflector`](https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go) | |
+| `source_event.py` | Reflector-to-loop wire format: `UpsertEvent`, `DeleteEvent`, and a whole-world `ReplaceEvent` | [`watch.Event`](https://github.com/kubernetes/apimachinery/blob/master/pkg/watch/watch.go) + [`Store`'s write methods](https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go) | Go writes to a store by method call; we send the same operations as values |
+| `object_store.py` | Cache, parent index, replace with deletion synthesis | [`cache.Store`](https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go) + [`DeltaFIFO.Replace()`](https://github.com/kubernetes/client-go/blob/master/tools/cache/delta_fifo.go) | Absorbs `Replace()`, the only part of `DeltaFIFO` we keep |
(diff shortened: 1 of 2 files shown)


docs/developer/reconcile-loop.md
```diff
@@ -0,0 +1,116 @@
...
+
+### `loop.py`
+
+| Upstream | Solves | Decision | Reason |
+| --- | --- | --- | --- |
+| Resync period | Level-triggered backstop | **Kept**, on by default at Go's 10 h, re-enqueues parents that still have members | Go defaults it on for the same reason we need it: insurance against this controller, or something it depends on, failing to requeue. Go's own limit is kept too, in that a parent which lost its last member is not re-driven. The 10 percent jitter is **dropped**: it spreads the LIST storm of many controllers, and there is one loop here whose resync only adds keys to a dedup set |
+| `MaxConcurrentReconciles` | Overlap reconcile I/O | **Dropped**; one worker | controller-runtime's default. No I/O to overlap |
+| Predicates | Save reconciles at scale | **Dropped** | Reconcile is cheap |
+| Lifecycle | Never leak a running loop | **`async with loop:`** | `start()` and `stop()` stay public, because `stop()` sometimes has to be scheduled from inside reconcile. Every caller that owns both ends writes the `async with` instead, and a caller that hands teardown back to someone else keeps the pairing with `AsyncExitStack.enter_async_context(loop)` |
```
fzyzcjy, line 91: [P2] Clean up the retry poller on failed context entry

If the owner is cancelled while `__aenter__()` is waiting for the initial `ReplaceEvent`, Python never invokes `__aexit__()`. `start()` cancels only the driver task, leaving the constructor-created `RetryScheduler` poller alive; the documented `async with`/`AsyncExitStack.enter_async_context()` pattern therefore leaks on aborted startup, including the target deliver's provider caller. Ensure failed entry shuts down the loop before re-raising, as the sentinel-bound code and central document must agree ([doc-dev](.claude/skills/doc-dev/SKILL.md#L30-L32)).

docs/developer/reconcile-loop.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Test layers
+
+All three run on every PR. `pytest tests/e2e/k8s_apiserver tests/e2e/k8s_kind` self-provisions everything below from a Docker daemon, and the modules that provision it depend on nothing from the reconcile package, so an environment can be verified before anything uses it.
+
+| Layer | Where | Proves | Provisions |
+| --- | --- | --- | --- |
+| Fakes + fake clock | `tests/fast/utils/workers/reconcile/` | Control flow, timing, shutdown | Nothing |
+| Real apiserver, no kubelet | `tests/e2e/k8s_apiserver/` | API semantics: cursors, watch timeouts, real 410, relist | etcd and the apiserver as containers, plus a second apiserver with `--watch-cache=false` — with the cache on, a compacted `resourceVersion` is still served, so nothing can invalidate a live cursor |
+| kind cluster | `tests/e2e/k8s_kind/` | What only a kubelet produces: Running, restarts, graceful deletion, bookmarks | A pinned kind binary, downloaded on demand |
```
fzyzcjy: [P2] Enforce the documented kind version pin

On a host with any `kind` executable on `PATH`, `_resolve_kind_binary()` returns it without checking `_KIND_VERSION`; only the fallback download is fixed at v0.32.0. This makes the E2E layer host-version-dependent despite the stated pin, and the mismatch remains in the target deliver; either enforce the pinned resolver or document the PATH override, per [doc-dev](.claude/skills/doc-dev/SKILL.md#L30-L32).

fzyzcjy (commented):
Two documented guarantees do not match the governed implementation: aborted async-context entry can leak the retry poller, and the kind binary is not always pinned. Both issues persist at the target safe deliver.

docs/developer/reconcile-loop.md
guapisolo: codex cmt

[P2] Qualify the failed-start cleanup guarantee

On this PR's current head (`f9b2ce203569`), [the shutdown row](https://github.com/radixark/miles/blob/f9b2ce203569bc8949f497a582cd94abaaff2261/docs/developer/reconcile-loop.md#L62) says that a `start()` that raised leaves no tasks, and [the lifecycle row](https://github.com/radixark/miles/blob/f9b2ce203569bc8949f497a582cd94abaaff2261/docs/developer/reconcile-loop.md#L91) presents `async with` as preventing leaks. If the caller cancels or times out context entry during initial synchronization, however, `__aenter__()` fails and `__aexit__()` is never invoked. Startup cancellation only joins the driver, leaving the retry poller running. A caller relying on the documented guarantee can therefore omit the explicit cleanup still required after failed entry.

The same claim remains at [#3122 (`d28546aca811`)](https://github.com/radixark/miles/blob/d28546aca8111ae3d6f4d477f70898725d814d54/docs/developer/reconcile-loop.md#L72), where the cancellation probe still reports `pending_after_failed_enter=["RetryScheduler._poll"]`. This PR introduces the documentation guarantee; the underlying lifecycle defect is tracked in [#2182's discussion](https://github.com/radixark/miles/pull/2182#discussion_r3984562539).

Until startup performs that cleanup itself, please document that cancellation during initial synchronization requires an explicit `await loop.stop()` in the caller's exception cleanup, and scope the `async with` cleanup guarantee to successful entry.

PR #2187, opened 2026-08-04
Add a process supervisor for running several ranks in one container

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/process_supervisor.py
+++ b/miles/utils/workers/process_supervisor.py
@@ -1,0 +1,152 @@
+from __future__ import annotations
+
+import argparse
+import logging
+import math
+import os
+import signal
+import sys
+from types import FrameType
+
+from torch.distributed.elastic.multiprocessing import DefaultLogsSpecs, Std, start_processes
+from torch.distributed.elastic.multiprocessing.api import RunProcsResult, SignalException
+
+logger = logging.getLogger(__name__)
+
+SUBPROCESS_INDEX_ENV_VAR = "MILES_SUPERVISOR_SUBPROCESS_INDEX"
+
+_DEFAULT_TERMINATION_GRACE_PERIOD_SECONDS = 20.0
+_FORWARDED_SIGNALS = (signal.SIGTERM, signal.SIGINT)
+_SIGNALS_TO_HANDLE_ENV_VAR = "TORCHELASTIC_SIGNALS_TO_HANDLE"
+_SUPERVISED_SUBPROCESS_NAME = "miles_supervised_subprocess"
+
+
+def main(argv: list[str] | None = None) -> int:
+    argv = sys.argv[1:] if argv is None else argv
+    args, command = _parse_args(argv)
+
+    logging.basicConfig(
+        level=logging.INFO,
+        format="[%(asctime)s] [process_supervisor] %(levelname)s %(message)s",
+    )
+
+    return supervise(
+        command=command,
+        num_subprocesses=args.num_subprocesses,
+        termination_grace_period_seconds=args.termination_grace_period_seconds,
+    )
+
+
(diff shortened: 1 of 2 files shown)


miles/utils/workers/process_supervisor.py
```diff
@@ -0,0 +1,152 @@
...
+
+import argparse
+import logging
+import math
+import os
+import signal
+import sys
+from types import FrameType
+
+from torch.distributed.elastic.multiprocessing import DefaultLogsSpecs, Std, start_processes
```
fzyzcjy, line 11: [P2] Install SIGTERM handling before importing Torch

For multi-worker pools at deliver-2, `pool_entry._serve_command()` makes this module the container command, so it is PID 1. Linux PID 1 ignores SIGTERM while its disposition is default, but no handler is installed until `start_processes()` after the top-level Torch import, argument parsing, and temporary-log setup. A pod deleted during that startup window therefore continues until kubelet sends SIGKILL; install an early handler or place a real init process in front. The target deliver leaves this path unchanged.

miles/utils/workers/process_supervisor.py
```diff
@@ -0,0 +1,152 @@
...
+    termination_grace_period_seconds: float = _DEFAULT_TERMINATION_GRACE_PERIOD_SECONDS,
+) -> int:
+    if num_subprocesses < 1:
+        raise ValueError(f"num_subprocesses must be at least 1, got {num_subprocesses}")
+
+    os.environ[_SIGNALS_TO_HANDLE_ENV_VAR] = ",".join(forwarded.name for forwarded in _FORWARDED_SIGNALS)
+
+    indices = range(num_subprocesses)
+    try:
+        context = start_processes(
```
fzyzcjy, line 94: [P2] Preserve the received signal during spawn cleanup

When SIGINT or SIGTERM arrives after Torch has populated the process table but while `start_processes()` is still starting the tee threads, Torch catches its own `SignalException` and calls `context.close()` with the default SIGTERM and 30-second timeout before re-raising. This outer handler therefore cannot preserve SIGINT or the configured grace period: partial ranks can receive the wrong signal, and a slow SIGTERM shutdown can consume Kubernetes' entire default termination window. The target deliver leaves this call unchanged.

fzyzcjy (commented):
The supervisor mishandles termination in two startup windows, and both defects remain at the target safe deliver. There is no subsequent safe deliver containing a fix.

guapisolo (approved):
codex cmt

Review scope: [PR #2187 changes](https://github.com/radixark/miles/pull/2187/files), frozen source head `34a885a0c818eca2bc6279bc4b29d4ccf1540ea4` (incremental base `f9b2ce203569bc8949f497a582cd94abaaff2261`). Stack repair checkpoint: [PR #3122](https://github.com/radixark/miles/pull/3122) at `d28546aca8111ae3d6f4d477f70898725d814d54`, inclusive. The recorded source review is reused after verifying its binding; this is not a new full-stack review.

Design: The process-supervisor entry point delegates rank launch/wait and group cleanup to torchelastic, propagates failure exit status, and provides per-rank environment/log identity. This step adds an entry point, not a complete Kubernetes scheduler integration.

No actionable findings were identified in this increment by the recorded native or separate code-quality review. There is no verified source finding to relabel as fixed later.

Verification: The implementation, signal/exit paths, tests, and callers were statically reviewed on this unchanged increment. The targeted suite could not collect in the original review environment because torch was unavailable; Linux PID 1/process-group behavior has not been executed in this verification. No runtime pass is claimed. The 343 PR bindings in the #2177-through-#3122 chain and its issue-list snapshot were revalidated unchanged.

Approved for this process-supervisor increment based on the recorded code review and current binding verification, with the stated runtime evidence gaps. This does not approve unrelated reconcile lifecycle issues or the entire delivery.

PR #2191, opened 2026-08-04
Install kubectl and helm into the training image

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The documentation matches the existing Dockerfiles and shared installer. No in-scope correctness or regression issue remains at deliver-2.

PR #2192, opened 2026-08-04
Lint every Helm chart from pre-commit and CI

fzyzcjy:
Part of #1837

### DIFF
--- a/tools/kubernetes/lint_charts.py
+++ b/tools/kubernetes/lint_charts.py
@@ -1,0 +1,134 @@
+from __future__ import annotations
+
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+from collections.abc import Sequence
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+CHARTS_DIR = REPO_ROOT / "charts"
+
+BASE_VALUES: dict[str, list[str]] = {
+    "miles-workbench": ["--set", "objectName=lint-miles-workbench"],
+}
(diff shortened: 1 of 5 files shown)


fzyzcjy (commented):
The rejected-variant check can mistake an unrelated Helm rendering failure for proof that a forbidden value was rejected. This false-positive path remains in the target safe deliver.

tools/kubernetes/lint_charts.py
```diff
@@ -0,0 +1,134 @@
...
+
+    ok = True
+    base = BASE_VALUES.get(chart.name, [])
+    for extra in [[], *VARIANTS.get(chart.name, [])]:
+        result = run(["helm", "lint", str(chart), *base, *extra])
+        if result.returncode != 0:
+            print(result.stdout + result.stderr, file=sys.stderr)
+            ok = False
+    for extra in REJECTED_VARIANTS.get(chart.name, []):
+        result = run(["helm", "lint", str(chart), *extra])
```
fzyzcjy: [P2] Apply base values when checking rejected variants

`miles-workbench` defaults `objectName` to an empty string, and `BASE_VALUES` supplies a valid name for ordinary lint runs. This invocation omits `base`, so the rejected run already fails on invalid empty object names; if the schema starts accepting `infra.env.PYTHONPATH`, that unrelated failure is still treated as successful rejection and CI remains green. This omission remains at deliver-2, so pass `*base` to the rejected invocation as well.

PR #2194, opened 2026-08-04
Add the miles-workbench Helm chart

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-workbench/templates/_helpers.tpl
+++ b/charts/miles-workbench/templates/_helpers.tpl
@@ -1,0 +1,44 @@
+{{- define "miles-workbench.fullname" -}}
+{{- .Values.objectName }}
+{{- end }}
+
+{{- define "miles-workbench.labels" -}}
+{{- include "miles-common.labels" (dict "context" . "component" "workbench") }}
+{{- end }}
+
+{{- define "miles-workbench.selectorLabels" -}}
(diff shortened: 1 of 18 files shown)


charts/miles-workbench/templates/_helpers.tpl
```diff
@@ -0,0 +1,44 @@
...
+{{- define "miles-workbench.roleRules" -}}
+- apiGroups: [""]
+  resources: ["configmaps", "secrets", "serviceaccounts", "services"]
+  verbs: ["create", "delete", "get", "list", "patch", "update", "watch"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["delete", "get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["pods/exec"]
+  verbs: ["create"]
```
fzyzcjy: [P2] Grant GET for WebSocket exec requests

`docker/install-kube-tools.sh` pins kubectl v1.36, where remote-command WebSockets are enabled by default and send `GET` requests to `pods/exec`; Kubernetes authorizes those as `get`. With only `create` here, `kubectl exec` from inside the workbench receives 403 rather than reaching a run pod, and deliver-2 retains this rule and its matching preflight contract. Grant `get` alongside `create`.

fzyzcjy (commented):
The workbench Role does not authorize the default WebSocket transport used by the bundled kubectl for pod exec, and the issue remains in the target deliver.

PR #2196, opened 2026-08-04
[DO NOT MERGE] CI

fzyzcjy:
ci-sglang-pr: tom/refactor-miles-repo-sglang/deliver-1
ci-megatron-pr: #72

### DIFF
--- a/docs/developer/migration.md
+++ b/docs/developer/migration.md
@@ -9,7 +9,7 @@ launch scripts.
 
 ### What changed
 
-`train.py`, `train_async.py`, and `RayTrainGroup` now use Python `async`/`await`
+`train.py`, `train_async.py`, and `TrainerController` now use Python `async`/`await`
 instead of `ray.get()`-style sync calls.
 
 ### Why
--- a/miles/ray/placement_group.py
+++ b/miles/ray/placement_group.py
@@ -125,7 +125,7 @@ def create_placement_groups(args) -> dict[str, PlacementGroupInfo]:
 
 
 async def create_training_models(args, inference_controller, rollout_executor):
-    actor_model = RayTrainGroup(
+    actor_model = TrainerController(
         args=args,
         role="actor",
         with_ref=args.kl_coef != 0 or args.use_kl_loss,
@@ -136,7 +136,7 @@ async def create_training_models(args, inference_controller, rollout_executor):
     actor_start_rollout_ids = await actor_model.init()
 
     if args.use_critic:
-        critic_model = RayTrainGroup(
+        critic_model = TrainerController(
             args=compute_critic_args(args),
             role="critic",
             with_ref=False,
--- a/miles/ray/train/cell.py
+++ b/miles/ray/train/cell.py
@@ -26,7 +26,7 @@ KILL_RPC_TIMEOUT_S = 10.0
 CONFIRM_DEAD_TIMEOUT_S = 120.0
 
 
-class RayTrainCell:
+class TrainerCell:
     def __init__(
         self,
         *,
--- a/miles/ray/train/cell_monitor.py
+++ b/miles/ray/train/cell_monitor.py
@@ -11,12 +11,12 @@ from miles.utils.ft_utils.api_server.models import CellCondition, CellStatus, Tr
 from miles.utils.ft_utils.health_checker import ActiveAndEpoch, SimpleHealthChecker, SimpleHealthCheckerConfig
 
 if TYPE_CHECKING:
-    from miles.ray.train.cell import RayTrainCell
+    from miles.ray.train.cell import TrainerCell
 
 
 def create_trainer_cell_health_checker(
     *,
-    cell: "RayTrainCell",
+    cell: "TrainerCell",
     config: SimpleHealthCheckerConfig,
     get_activeness: Callable[[], ActiveAndEpoch],
 ) -> SimpleHealthChecker:
--- a/miles/ray/train/group.py
+++ b/miles/ray/train/group.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 from miles.backends.megatron_utils.ft.types import TrainStepOutcome
 from miles.ray.specs.train import compute_trainer_num_cells, compute_trainer_pool_id
-from miles.ray.train.cell import RayTrainCell
+from miles.ray.train.cell import TrainerCell
 from miles.ray.train.cell_monitor import create_trainer_cell_health_checker
 from miles.utils.async_utils import AsyncioGatherUtils
 from miles.utils.audit_utils.checksum_utils import flatten_inference_engine_checksums
@@ -36,7 +36,7 @@ _RETRY_MAX_ATTEMPTS = 30
 _CELLS_READY_TIMEOUT_SECONDS = 3600.0
 
 
-class RayTrainGroup:
+class TrainerController:
     def __init__(
         self,
         args,
@@ -66,7 +66,7 @@ class RayTrainGroup:
 
         self._health_checker_activeness = ActivenessTracker(active=True)
 
-        self._cells_by_id: dict[str, RayTrainCell] = {}
+        self._cells_by_id: dict[str, TrainerCell] = {}
 
         self._witness_allocator: WitnessIdAllocator | None = (
             WitnessIdAllocator(buffer_size=args.witness_buffer_size) if args.enable_witness else None
@@ -74,7 +74,7 @@ class RayTrainGroup:
         if self._witness_allocator is not None and args.save_debug_event_data is not None:
             self._witness_allocator.resume(read_persisted_witness_counter(Path(args.save_debug_event_data)))
 
-        self._test_action_executor = FTTestActionControllerExecutor.from_args(args, group=self)
+        self._test_action_executor = FTTestActionControllerExecutor.from_args(args, controller=self)
 
     @property
     def pool_id(self) -> str:
@@ -89,7 +89,7 @@ class RayTrainGroup:
         return compute_trainer_num_cells(self.args, role=self._role)
 
     @property
-    def _cells(self) -> list[RayTrainCell]:
+    def _cells(self) -> list[TrainerCell]:
         return sorted(self._cells_by_id.values(), key=lambda cell: cell.cell_index)
 
     @property
@@ -130,8 +130,8 @@ class RayTrainGroup:
         cell = self._cells_by_id.pop(cell_id)
         cell.health_checker.stop()
 
-    def _create_cell(self, cell_id: str, *, cell_index: int, workers_hash: str) -> RayTrainCell:
-        cell = RayTrainCell(
+    def _create_cell(self, cell_id: str, *, cell_index: int, workers_hash: str) -> TrainerCell:
+        cell = TrainerCell(
             args=self.args,
             role=self._role,
             with_ref=self._with_ref,
@@ -242,7 +242,7 @@ class RayTrainGroup:
             )
 
     def _check_train_one_attempt(self, snapshot_alive_cells, results):
-        outcomes = RayTrainGroup._compute_attempt_outcomes(snapshot_alive_cells, results)
+        outcomes = TrainerController._compute_attempt_outcomes(snapshot_alive_cells, results)
         if not outcomes["normal"] and not outcomes["discarded"]:
             log_structured(
                 logger.error, tag="ft", op="check", **outcomes, decision="retry", reason="all alive cells failed"
--- a/miles/utils/audit_utils/event_analyzer/rules/witness.py
+++ b/miles/utils/audit_utils/event_analyzer/rules/witness.py
@@ -39,7 +39,7 @@ def check(events: list[Event]) -> list[WitnessIssue]:
     * WitnessAllocateIdEvent: when allocating `witness_id` to `sample_index`
     * WitnessSnapshotParamEvent: near the end of each train() step in MegatronTrainRayActor
         * If a witness_id appears in the weight, it means the corresponding data is consumed at least once.
-    * TrainGroupStepEndEvent: after each train() step in RayTrainGroup
+    * TrainGroupStepEndEvent: after each train() step in TrainerController
 
     Check:
     1. For each (rollout_id, cell_index),
--- a/miles/utils/ft_utils/api_server/server.py
+++ b/miles/utils/ft_utils/api_server/server.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 def start_api_server(
     *,
     args,
-    actor_model: RayTrainGroup,
+    actor_model: TrainerController,
     inference_controller: object,
     port: int,
     ft_components: list[str],
--- a/miles/utils/test_utils/ft_test_actions.py
+++ b/miles/utils/test_utils/ft_test_actions.py
@@ -9,7 +9,7 @@ from miles.utils.workers.naming import parse_cell_id
 from miles.utils.workers.ray_worker_manager import RayWorkerManager
 
 if TYPE_CHECKING:
-    from miles.ray.train.controller import RayTrainGroup
+    from miles.ray.train.group import TrainerController
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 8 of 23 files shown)


fzyzcjy:
### CI went fully green at `63c943bac`

[Run 31062593296](https://github.com/radixark/miles/actions/runs/31062593296) — **15/15 SUCCESS**, with only the ROCm lane failing (see the caveat below).

| Stage | Result |
| --- | --- |
| `stage-a-cpu` ×4, `stage-b-cpu` | SUCCESS |
| `stage-b-2-gpu-h200`, `stage-c-2-gpu-h200` ×2 | SUCCESS |
| `stage-c-4-gpu-h200` ×3 | SUCCESS |
| `stage-c-8-gpu-h100` ×2, `stage-c-8-gpu-h200` ×2 | SUCCESS |
| `stage-c-4-gpu-mi300x` ×2 | FAILURE — not caused by this PR, see below |

The `bypass-fastfail` label is on, so each shard runs every test rather than stopping at the first failure. A shard reporting SUCCESS therefore means every test in it passed.

#### The three long-running failures this closes

- **`test_qwen3_4B_offload_disk`** — failed deterministically 4× at ~43s. Cause: nothing cleaned up accelerator memory between CI test files. The sglang scheduler is a *grandchild* of the test process, so per-test teardown never reached it, and the workflow's `pkill` only runs once at job start. The CUDA suite runner now reaps between files. Two earlier explanations were refuted by evidence and dropped: disk exhaustion (zero `ENOSPC`/`No space left` hits across two runs) and stale dependency pins (a 4th failure reproduced on freshly pinned sglang).
- **`test_qwen3_4B_disk_delta`** — `Weight version mismatch! Engine: default, Updater: 0`, reproduced on two separate rounds. Only rank 0 publishes the engine weight version in `_capture_baseline`, while the `ci_test` check runs on every rank with no synchronisation between them. The broadcast and p2p paths already close with `dist.barrier(group=get_gloo_group())`; the delta path did not. Added.
- **`test_deepseek_v4_flash_4layer_ci`** — all four ranks died together inside `wake_up` with `connection error code 2`, 119/139 GB of device memory still free, i.e. host-RAM pressure rather than anything in this branch. Not reproduced on the rerun.

#### Also fixed here

- The sglang launch command is now rendered through the parser's own argparse actions instead of hand-spelled option strings, which fixes `--no-` flags, JSON-typed fields, and fields whose flag differs from their dest. The dict input is kept deliberately: `ServerArgs.__post_init__` is not idempotent (DP attention divides `chunked_prefill_size` by `dp_size` on every run), so rendering from an already-post-inited instance breaks the roundtrip assertion when `sglang_dp_size > 1`.
- A disposed rollout cell now unregisters from the router based on whether it actually registered, not on which state it is in. Both registration sites await `add_worker` *before* the state advances, so a cell can own a live router entry while still initializing or pending weights — and, in the other direction, can reach either state having never registered at all.
- Around 20 tests the base left stale, two infinite hangs, and an `examples/` breakage that would have failed with `unrecognized arguments`.

#### Caveat: the ROCm lane is failing for a reason outside this PR

`stage-c-4-gpu-mi300x` fails with `ValueError: Unknown suite stage-c-4-gpu-mi300x for backend ROCM`.

`pr-test-rocm.yml` is a `pull_request_target` workflow, so its definition comes from `main`, where `tests/ci/run_suite.py` registers `stage-c-4-gpu-mi300x`. This PR's base branch does not have that registration yet, so **any** PR based on it fails this lane identically. On `main` the mi300x support is a whole feature — the `amd` label, `suite="stage-c-4-gpu-mi300x"` on nine e2e files, and an assertion in `test_run_suite.py` — so pulling it into this stack just to turn the lane green would be the wrong move. Left as is, per @fzyzcjy.

#### Verification beyond CI

The full `tests/fast` suite was also run on a separate CPU devbox: **6989 passed**, with two failures that are environment-only and pass in CI — the baked sglang in the image predates `gated_launch_port`, and the router `--help` subprocess takes 10.05s against its own `timeout=10`. That run is what caught the dispose regression above before CI did.

fzyzcjy:
### Heads-up: no checks are being created for this PR right now

Actions is not creating any workflow run for this PR — not `PR Test`, not `PR Test (ROCm)`, not `pre-commit`. `GET /actions/runs?head_sha=<head>` returns `total_count: 0` for the last two pushes. Other PRs in this repo are getting runs normally in the same window, so this is specific to #2196.

**Cause.** Run [31108219821](https://github.com/radixark/miles/actions/runs/31108219821) is stuck: it reports `status: queued` while every job in its latest attempt is already completed. It got there because I re-ran its jobs repeatedly while the GPU fleet had no free runners with the requested labels, and one of those re-runs was never actually queued:

```
POST /actions/runs/31108219821/cancel       → 409 Cannot cancel a workflow re-run that has not yet queued
POST /actions/runs/31108219821/force-cancel → 409 (same)
POST /actions/runs/31108219821/rerun        → 403 This workflow is already running
```

**What I tried, all with no effect:** `rerun --failed`, `rerun --job`, full `rerun`, `gh run cancel`, the REST cancel and force-cancel endpoints, cycling a label (`labeled` is one of the trigger types), pushing an empty commit, pushing a real commit, and close/reopen.

**What unblocks it:** cancelling or deleting that run from the web UI, or waiting for GitHub to expire it (queued runs age out after about 24h).

**The lesson, for the record:** a job that no runner picks up reports `cancelled` with an empty `runner_name` and zero steps. That is capacity, not a test failure, and re-running it does not conjure a runner — it only risks wedging the run. Both facts are now written down in `docs/ci/00-stage.md`.

#### Where the branch actually stands

Verified at `1c4fc1ada` (the two commits after it are a poke and a docs note):

| Scope | Result |
| --- | --- |
| Full `tests/fast` on a CPU devbox | 6996 passed; 2 environment-only failures that pass in CI |
| CI CPU | `stage-a-cpu (0)(1)(2)(3)`, `stage-b-cpu` — all SUCCESS |
| CI GPU | `stage-b-2-gpu-h200`, `stage-c-2-gpu-h200 (0)(1)`, `stage-c-8-gpu-h200 (0)(1)`, `stage-c-8-gpu-h100 (1)`, `stage-c-4-gpu-h200 (2)` — all SUCCESS |

Still unverified, all from the interrupted round:

- `stage-c-8-gpu-h100 (0)` — 8 of its 9 files passed; `tests/e2e/short/test_dumper.py` failed with `No engines/engine_* subdirs`. I ruled out this PR's `dumper_utils` change by reading `retry_until_deadline` (it does not sleep before the first attempt) and confirming the log shows zero retries, which makes the new code byte-identical in behaviour to the old single call. **Why a dumper that configured successfully produced no dumps is still unexplained** and needs a reproduction.
- `stage-c-4-gpu-h200 (0)` and `(1)` — cancelled mid-run, not test failures.
- `stage-c-4-gpu-mi300x (0)(1)` — fails for a reason outside this PR, see the previous comment and the new note in `docs/ci/00-stage.md`.

One thing worth a decision: `abs(train/ppo_kl) < 1e-9` in `miles/backends/training_utils/ci_utils.py` is a marginal tolerance, not a property of this change. `test_qwen3_4B_disk_delta` produced `0.0`, then `-1.374e-9` (over the bound, failed), then `2.841e-10` across three rounds — it was never deterministically zero. Loosening it would weaken a check this PR does not own, so I left it alone.

fzyzcjy:
## CI green on `dea17123a`

Every stage passes except the two ROCm ones, which fail for a reason unrelated to this branch (the base's stage list predates the current ROCm lane) and are being ignored by agreement.

| Stage | Result |
|---|---|
| `stage-a-cpu (0)(1)(2)(3)` | SUCCESS |
| `stage-b-cpu` | SUCCESS |
| `stage-c-2-gpu-h200 (0)(1)` | SUCCESS |
| `stage-c-4-gpu-h200 (0)(1)(2)` | SUCCESS |
| `stage-c-8-gpu-h100 (0)(1)` | SUCCESS |
| `stage-c-8-gpu-h200 (0)(1)` | SUCCESS |
| `stage-c-4-gpu-mi300x (0)(1)` | FAILURE — unrelated, exempt |

A full `tests/fast` run on a devbox is green as well: **7065 passed, 20 skipped, 6 xfailed, 0 failed, 0 errors** (39m34s, `h200-sci-k8s`).

### Two flakes worth naming

- `stage-c-4-gpu-h200 (0)` `test_r3_baseline.py` needed three attempts; `(1)` `test_int4_rollout.py` needed two. Both died the same way: a trainer worker vanished with `SYSTEM_ERROR ... connection error code 2. End of file`, with no CUDA OOM traceback. That config runs a 30B MoE on 4 GPUs with `--colocate --optimizer-cpu-offload`, so a host OOM-kill fits better than a CUDA one. The e2e itself is untouched by this branch apart from a mechanical rename, and the seed logic is identical to the commit these stages passed on, so this is a margin problem rather than a regression — but 1-pass-in-3 is not a clean flake signal. If it keeps failing on main, the first thing to check is whether this stack raised peak memory on the colocate path.
- `test_deepseekv4`'s TITO threshold (`13/64 = 0.203 > 0.2`) failed on an earlier commit and passes now. That ratio is decided by generated text, and this branch had left sglang's `random_seed` unset, so it was redrawn every launch and sat right on the line. The seed is derived from `--seed` again, so the ratio is deterministic now. If it starts drifting again, that means the seed stopped reaching the engines.

### Weakening audit

Per the review agreement, subagents audited the whole `deliver-1-base..HEAD` diff (production code, tests, and everything outside `miles/`) for behaviour or tests that got weaker. Fourteen real weakenings were found and fixed, plus three tests that had been rewritten to assert the weakened behaviour. The most consequential:

- `--use-fault-tolerance` alone no longer healed anything: the health checkers only publish a status, and the only thing that acts on it needed `--mini-ft-controller-enable` **and** a non-zero `--api-server-port`, both off by default, while sixteen launch scripts pass neither. It now drives the cell registry in-process and follows `--ft-components`.
- A pinned port that someone else already held was no longer refused, so the new process would fail to bind and die while whoever waited for the port connected to the stale occupant instead.
- The sglang engine seed was commented out with no way to put it back, so `--seed` stopped reaching the engines.
- Multi-LoRA on an Inkling checkpoint silently widened the adapter to every compatible module.
- Dumper mode stopped silencing the heartbeats it says it silences (`ft_components` is resolved before the override runs).
- `--rollout-external` raised `NotImplementedError`; it works again, and so does pinning the router port.

A second, adversarial pass over those fixes found four bugs in the fixes themselves — the worst being that the restored seed keyed off the node-local GPU id, which gives every engine of a one-engine-per-node fleet the same seed. All four are fixed.

Seven rollout test modules had decayed into `pytest.skip` husks kept parseable by a `noqa` for symbols that no longer exist. Their 61 tests were mapped against the current suite; the uncovered guarantees are covered by new tests and the husks are gone. `RolloutServer`'s three fan-outs had no behavioural test at all, which is how the `needs_offload` filter, the tags pass-through, and the reach across every cell could each have been dropped silently.

### Before merge

`15ac046ba` is an empty poke commit from unwedging a stuck CI run; worth squashing out.

fzyzcjy:
> 🤖 **Posted autonomously by Codex acting on the user's behalf.** I diagnosed the current CI failures, committed the validated fixes, and pushed a new CI run. Please push back if any conclusion is off.

The previous head `7bbce621ad` failed two CPU tests: the server-argument round-trip failed on a missing `disaggregation_mode` baseline value in [stage-a-cpu shard 3](https://github.com/radixark/miles/actions/runs/31361311561/job/93376496970), and the PPO terminal-reward test aborted in [stage-a-cpu shard 1](https://github.com/radixark/miles/actions/runs/31361311561/job/93376497026). The first failure and the earlier missing `wt` parameterization are fixed by [`3cd6376e0b`](https://github.com/radixark/miles/commit/3cd6376e0b); the full `stage-a-cpu` suite passed in the matching remote CPU environment before this commit was pushed.

The ROCm failures are independent base drift: both [shard 0](https://github.com/radixark/miles/actions/runs/31366299130/job/93385343262) and [shard 1](https://github.com/radixark/miles/actions/runs/31366299130/job/93385343232) request `stage-c-4-gpu-mi300x`, while this old PR base still registers the pre-rename ROCm suite names. I am treating those as pre-existing for this delivery branch and will track the new commit through the CPU and H100/H200 lanes on the [PR checks page](https://github.com/radixark/miles/pull/2196/checks).

fzyzcjy:
> 🤖 **Posted autonomously by Codex acting on the user's behalf.** I inspected this GPU job's complete log and classified the failure. Please push back if any conclusion is off.\n\n[stage-c-8-gpu-h200 (1)](https://github.com/radixark/miles/actions/runs/31366917741/job/93388919377) failed while generating an SGLang launch command for . The root fingerprint is  in [](https://github.com/radixark/miles/blob/3cd6376e0b3f2fe64e763acdf0da38b64134e191/miles/utils/workers/argv_utils.py#L122), the same raw-versus-normalized renderer bug seen in the sibling H200 lane. I am adding a renderer regression fix as a follow-up commit.

fzyzcjy:
CI status for head ece8ee727654ffdc1b63fe2201e8c3fe9cfb772a:

- 21 checks have passed, including pre-commit, all CPU shards, both 8-GPU H200 shards, all 4-GPU H200 shards, both 2-GPU H200 shards, and H100 shard 0.
- H100 shard 1 failed on tests/e2e/short/test_qwen2.5_0.5B_gsm8k_short.py after the 1800-second timeout. The original attempt and the failed-job rerun both hit the same startup failure: a Ray CommandActor running SGLang could not bind the torch.distributed TCPStore port (EADDRINUSE; port 15108 in the rerun).
- The two MI300X shards are still queued.

I am treating the repeated H100 failure as a deterministic port-allocation/startup bug rather than a flake. I will reproduce and validate the fix on a tom-remote GPU devbox before pushing a new head and rerunning CI.

fzyzcjy:
CI status summary for head `0b14d5f94d06fcac55e04c4c18ca367260186df6`:

### Branch-related failure and fix

- The reproducible failure was `stage-c-8-gpu-h100 (1)`: two runs failed because Miles allocated dynamic service ports inside Ray's default worker-port range (`10002-19999`). The observed collisions were `15210` (SGLang bind) and `15108` (TCPStore bind).
- Commit `0b14d5f94d` moves the Miles dynamic-port start from `15000` to `20000`, immediately above Ray's worker range and below the Linux ephemeral-port range.
- Validation of this fix:
  - targeted allocator tests: 5/5 passed;
  - the exact previously failing 8-GPU E2E passed once, then passed three additional consecutive stability runs (4/4 total) on an 8xH200 devbox;
  - full `pre-commit run --all-files` passed;
  - both 8-GPU H100 CI shards now pass.

### Remaining red H200 check

- The only red check is `stage-c-4-gpu-h200 (2)`.
- Attempt 1 failed late in `test_qwen3_30B_A3B/test_baseline.py`, after rollout 0, training, and weight update had completed. Several Ray actors disappeared nearly simultaneously with `SYSTEM_ERROR`/EOF and no Python or CUDA traceback. The exact same shard passed on the preceding head (`ece8ee`, job `93758246522`), whose only relevant delta from this head is the port-range fix. This is consistent with a transient runner/node failure, not the port change.
- Attempts 2, 3, and 4 did not reproduce that failure. Instead, six independent tests failed immediately at W&B initialization with `wandb: ERROR user is not logged in`. Attempt 1 had successfully authenticated the same CI identity earlier. These retries therefore fail on the repository `WANDB_API_KEY`/external W&B authentication path before exercising the affected code.
- Repeated immediate reruns were stopped to avoid wasting H200 capacity until the repository W&B credential is repaired.

### Current rollup

- Passed: pre-commit, image/build setup, all CPU shards, both H100 8-GPU shards, all other H200 shards, and both MI300X shards.
- Skipped as expected: `approve-held-runs` and Mintlify.
- Red only because of the H200 infrastructure/authentication history described above.

From a code-regression perspective, I consider this head fully validated: every branch-attributable failure has been fixed and the relevant test has passed repeatedly. GitHub is not literally all-green because `stage-c-4-gpu-h200 (2)` remains recorded as failed; clearing that check requires restoring the W&B CI credential and rerunning the shard, or treating the documented failure as infrastructure.

fzyzcjy:
agent wrong operation, no real merge to main

PR #2197, opened 2026-08-04
fix: fail at launch when static micro-batching cannot align the rollout-side DP schedule

guapisolo:
> **Depends on:** #1927

## Summary
Fail at launch, not mid-run, when static micro-batching cannot satisfy the DP schedule alignment.

## Symptom & Reproduction
- **Symptom:** a static-batch run under `--use-dynamic-global-batch-size` trains normally for many steps, then dies in `build_dp_schedule` with `static path: micro-batch count ... is not a multiple of dp_size * mb_group`; the message's "adjust the config" advice is unsatisfiable because `step_size` is data, not config.
- **Reproduction:** `dp_size=2`, `micro_batch_size=2`, one step whose rollouts yield 5 samples: `ceil(5/2)=3` micro-batches, `3 % 2 != 0`, raise.

## Root Cause
1. `_pack_micro_batches_static` emits `ceil(step_size / micro_batch_size)` micro-batches.
2. `step_size` varies per step under compact rollouts and `--use-dynamic-global-batch-size`.
3. `build_dp_schedule` requires the count to be a multiple of `dp_size * mb_group`.
4. The static path has no counterpart of `expand_bins_by_splitting`, so misalignment is fatal.

## Fix
`miles_validate_args` rejects `--use-dynamic-global-batch-size` without `--use-dynamic-batch-size`, turning the data-dependent mid-run death into a launch-time error. The runtime raise stays as defense in depth; the launch guard removes the only config combination that makes it fire nondeterministically. Multi-LoRA is exempt: it auto-enables dynamic global batch size but `can_schedule_on_rollout_side` excludes it, so it keeps the legacy training-side schedule. Compact rollouts with a fixed global batch size stay guarded at runtime only, because compactness is a property of the generate fn and invisible at argument-parse time. The two refactor-tail commits make each packing path own its alignment handling so the static raise sits next to the packing it protects.

## Verification
- **New / updated test:** `test_dynamic_global_batch_size_requires_dynamic_batch_size` asserts the launch-time rejection.
- **Existing suites:** `tests/fast/utils/test_arguments.py`, `test_dp_schedule.py`, `test_lora_arguments.py` — 84 passed.

## Review Focus
- Scrutinize the `not args.multi_lora` exemption in `miles_validate_args`: multi-LoRA legitimately runs static under its auto-enabled dynamic global batch size.
- Scrutinize the residual static exposure: with a fixed global batch size `step_size` is constant, so the runtime raise fires at the first schedule build or never, except when compact rollouts first appear mid-run.

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -2944,20 +2944,17 @@ def miles_validate_args(args):
             args.log_probs_max_tokens_per_gpu = args.max_tokens_per_gpu
 
     # --use-dynamic-global-batch-size has two motivations:
-    # 1. compaction/subagent rollouts: the number of samples per rollout is unpredictable
-    #    at runtime, so steps must size themselves to the collected data. Assembling such
-    #    steps with static micro-batching is not supported yet -- no config keeps
-    #    ceil(step_size / micro_batch_size) aligned to dp_size * mb_group -- so
-    #    --use-dynamic-batch-size is required.
+    # 1. compaction/subagent rollouts: static micro-batching cannot guarantee alignment
+    #    when the physical sample count is data-dependent, so --use-dynamic-batch-size
+    #    is required.
     # 2. multi-LoRA (auto-enabled, no compaction/subagent): the per-round sample count is
     #    a config-shaped multiple of dp_size trained as exactly one step on the legacy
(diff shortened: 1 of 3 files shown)


Zhichenzzz (approved):
a great corner case fix! thanks

PR #2198, opened 2026-08-04
megatron: name the tensors behind a non-finite grad norm

Shi-Dong:
## What

When a training step is skipped because the grad norm is non-finite, log which trainable gradients are non-finite (and how many of those are NaN), plus the list of tensors whose gradients stayed finite. No behavior change on healthy steps.

The walk-and-log logic lives in `misc_utils.report_nonfinite_grads` so the multi-LoRA per-slot gate (#2078, stacked on this PR) can reuse it instead of carrying its own copy; `model.py` keeps a thin wrapper that feeds it every model chunk's trainable params.

## Why

Today the only symptom of a non-finite gradient is `train/grad_norm = nan` next to a perfectly finite loss, which gives no clue where the corruption entered the network.

Gradients flow from the loss downward, so the shallowest tensor with a *clean* gradient sits just above the layer that first produced a non-finite value. Printing the finite list alongside the non-finite one brackets the faulty layer immediately; the offender list alone cannot do that because it is emitted in parameter-name order.

The walk only runs on steps that are already invalid and skipped, so there is no steady-state cost.

This report is how a NaN gradient on 4-node H200 GLM-5.2 744B-A40B LoRA runs was localised to the sparse-MLA backward kernel (fixed separately in #2079).

### DIFF
--- a/miles/backends/megatron_utils/misc_utils.py
+++ b/miles/backends/megatron_utils/misc_utils.py
@@ -1,3 +1,11 @@
+import logging
+from collections.abc import Iterable
+
+import torch
+
+logger = logging.getLogger(__name__)
+
+
 def strip_param_name_prefix(name: str | None) -> str | None:
     if name is None:
         return None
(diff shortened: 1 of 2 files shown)


yushengsu-thu (approved):

yushengsu-thu (changes requested):
run lora 
in ci-megatron

PR #2199, opened 2026-08-04
Cap train-actor torch memory via MILES_TRAIN_MEMORY_FRACTION

Shi-Dong:
## What

Add an opt-in `MILES_TRAIN_MEMORY_FRACTION` environment variable. When set to a value > 0, the Megatron train actor calls `torch.cuda.set_per_process_memory_fraction(...)` at startup. Unset (the default) changes nothing.

## Why

NCCL allocates its buffers with `cudaMalloc`, outside torch's caching allocator, so torch's own allocator knobs cannot reliably keep memory free for it: `PYTORCH_CUDA_ALLOC_CONF=garbage_collection_threshold` is relative to the per-process cap, which defaults to the whole device. The caching allocator therefore keeps growing its reserved pool until NCCL finds 0 bytes free mid-backward, and the step dies with an OOM raised from inside NCCL even though torch's *allocated* memory is far below the device capacity.

A hard per-process cap makes the allocator release cached blocks once it reaches the cap instead of growing reserved memory without bound. On a 4-node H200 GLM-5.2 744B-A40B LoRA run this was the difference between reserved memory growing to ~113 GiB against ~57 GiB actually allocated (then NCCL OOMing during backward) and a stable run.

An env var rather than a CLI flag keeps this out of the argument surface: it is a workaround knob for tight-memory configurations, not a tuning parameter most runs should touch.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -149,6 +149,15 @@ class MegatronTrainRayActor(TrainRayActor):
         self.train_parallel_config = {} if args.indep_dp else {"dp_size": get_parallel_state().intra_dp.size}
         dist.barrier(group=get_gloo_group())
 
+        # NCCL allocates outside torch's caching allocator, so allocator knobs alone
+        # cannot keep memory free for it: gc_threshold is relative to the per-process
+        # cap, which defaults to the whole device. A hard cap forces the allocator to
+        # release cached blocks instead of growing reserved until NCCL's cudaMalloc
+        # during backward finds 0 bytes free.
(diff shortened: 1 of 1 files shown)


Shi-Dong:
Closing after discussion: we'd rather not grow a dedicated memory-cap knob for this. For the record: torch >= 2.10 exposes the same cap natively via PYTORCH_ALLOC_CONF=per_process_memory_fraction:<f>, but miles launchers export env vars into the job-wide ray runtime env, so under --colocate the rollout engines inherit the cap too and OOM whenever it sits below --sglang-mem-fraction-static. If this resurfaces, the cleaner shape is a --train-memory-fraction argument (next to the existing --train-memory-margin-bytes), applied inside the train actor.

PR #2137, opened 2026-08-04
[multi-lora] change multi-LoRA training backend design 

yushengsu-thu:
- Replaces the Option 2 multi-LoRA path with Option 1

(option1)
image (18)

(option2)
image (17)

- Whole child batches stay atomic; persistent round-robin selection; `BatchPlan` is the sole rollout→train control plane.
- Registration-scoped serving identity (anti-ABA); unbound registrations queue; transactional bind-at-selection via `SlotPool`; optimizer-inclusive swap checkpoints.
- NaN/found-inf gate vetoes per-adapter steps; per-slot schedulers tick on optimizer steps.
- End-of-collection aborts scoped per tenant (one adapter can no longer cancel another's in-flight requests).

Validated: fast suites green; GPU E2E Qwen3-4B DP2 + GLM-5.2 MoE (7-phase service smoke).

Requires 
- sglang https://github.com/sgl-project/sglang/pull/33299 
- Megatron-Bridge https://github.com/radixark/Megatron-Bridge/pull/26.

Code:

```text
run_multi_lora.py (CLI) ──▶ train_multi_lora_async.py :: main()   # THE DRIVER
                              │  parse_args() → validate_multi_lora_args()   # wires MultiLoRARolloutFn
                              │
        ┌─────────────────────┴──────────────────────────────────────────┐
        │            startup: rollout engines + controller + trainers    │
        │                                                                │
        │  MultiLoRAController (Ray actor + HTTP /adapter_runs)          │
        │    └─ MultiLoRABackend ─ AdapterRegistry ─ SlotPool            │
        │       # register: PENDING + registration_id, slot or queued    │
        │       # truth for: lifecycle, steps, versions, slot tenancy    │
        └────────────────────────────────────────────────────────────────┘
                              │
                              ▼  loop per rollout_id:
  ┌──(1) reconcile_adapters ──▶ trainer ranks: retire/cleanup, load new slots,
  │                             free slots, bootstrap queued registrations
  ├──(2) update_weights ──────▶ push stale adapters to sglang as
  │                             serving_lora_name(name, reg_id)   # PENDING→ACTIVE
  ├──(3) generate ────────────▶ MultiLoRARolloutFn:
  │                               one child rollout per adapter (own data + reward)
  │                               → samples stamped AdapterRef; requests routed via
  │                                 apply_adapter_routing (lora_path / rid / KV key)
  │                               → round-robin select whole batches
  │                               → controller.plan_bind (reserve slots, txn)
[...]

PR #2208, opened 2026-08-05
[multi-lora] 1/4: fix optimizer builder wrapping and slot-scoped master reloads

yushengsu-thu:
First of a 4-PR series splitting #2137 into reviewable pieces (the stack's final tree is byte-identical to #2137, which stays open as the integrated reference).

Two correctness fixes for the CURRENT multi-LoRA path, independent of the redesign:

- **Optimizer builder**: `build_multi_lora_optimizer` handed pre-wrapped Megatron children to `LayerWiseDistributedOptimizer`; the Megatron-Core pinned by Megatron-Bridge rejects that with a `TypeError`. Hand it the raw torch optimizers — `param_groups` proxy to the raw groups, so per-slot tags survive.
- **Slot-scoped master reloads**: deregistration/adapter-loading called the global `optimizer.reload_model_params()`, re-deriving EVERY resident slot's fp32 master from bf16 and silently dropping accumulated low mantissa bits. `reload_adapter_slot_model_params` refreshes one slot only.

Series: 1/4 (this) → #2209 2/4 identity → #2210 3/4 SlotPool/sidecars → #2211 4/4 Option 1 swap.

### DIFF
--- a/miles/backends/megatron_utils/multi_lora_optimizer.py
+++ b/miles/backends/megatron_utils/multi_lora_optimizer.py
@@ -109,7 +109,11 @@ def build_multi_lora_optimizer(
             for child in children:
                 for group in child.param_groups:
                     group["miles_multi_lora_slot"] = slot
-                base_optimizers.append(child)
+                # LayerWiseDistributedOptimizer wraps raw torch optimizers itself; the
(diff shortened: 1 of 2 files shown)


yushengsu-thu:

PR #2209, opened 2026-08-05
[multi-lora] 2/4: registration-scoped serving identity and request routing

yushengsu-thu:
Part 2/4 of the #2137 split (stacked on #2208; diff shows only this layer).

Serving names, request ids, and KV-cache keys carry the registration id (anti-ABA): a re-registered name is a new tenant with a fresh namespace, so a stale abort, weight push, or cache entry can never cross tenants.

- `serving_lora_name` / `make_rid` / `rid_prefix` / `cache_extra_key` derive every engine-facing identity from `(name, registration_id)`.
- `apply_adapter_routing` is the single injection point for generation AND prefill scoring, with a `serving_version` override for the live-upsert path.
- `AdapterRef` carries `registration_id`/`serving_version`; weight pushes register under the registration-scoped name; retire/free_slot aborts target one registration's rid prefix; the end-of-rollout abort under multi-LoRA never `abort_all`s shared engines.
- The existing rollout path keeps working unchanged apart from the wider identity; new adapter-routing unit tests pin the scheme.

Validated: full fast suite green on this tree (4045 tests).

Series: #2208 1/4 → 2/4 (this) → 3/4 → 4/4.

### DIFF
--- a/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
+++ b/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
@@ -277,7 +277,7 @@ class DistBucketedWeightUpdateMixin:
         rank transmits."""
         from megatron.bridge.peft.multi_lora_layers import expose_adapter_slot
 
-        from miles.utils.multi_lora import slot_lora_name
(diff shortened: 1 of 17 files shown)


yushengsu-thu:

PR #2210, opened 2026-08-05
[multi-lora] 3/4: SlotPool bind transactions and optimizer-inclusive slot sidecars

yushengsu-thu:
Part 3/4 of the #2137 split (stacked on #2209; diff shows only this layer).

Two self-contained, purely additive components for slot oversubscription — nothing is wired into the trainer loop yet (4/4 does the switch), so this layer is zero-risk to the running path:

- **SlotPool**: trainer-slot tenancy as rented containers, with transactional `plan_bind`/`commit_bind`/`abort_bind` (reservations prevent one plan from handing the same free slot to two adapters), keep-warm LRU eviction, pins, and selection-order admission.
- **Slot sidecars**: one checkpoint format for swap-out/swap-in and per-adapter saves — bf16 slot weights, fp32 masters (not re-derivable from bf16), Adam moments with both step counters, rank/alpha — in per-rank atomic shards behind a rank-0 manifest. A step-0 sidecar is a legitimate state, distinguished from a missing one.

23 unit tests pin the transaction, pin-lifecycle, naming, and sentinel semantics.

Series: #2208 → #2209 → 3/4 (this) → 4/4.

### DIFF
--- a/miles/backends/megatron_utils/multi_lora_checkpoint.py
+++ b/miles/backends/megatron_utils/multi_lora_checkpoint.py
@@ -1,0 +1,275 @@
+"""Optimizer-inclusive per-slot checkpoints: the slot swap primitive.
+
+One sidecar format serves slot swap-out/swap-in, periodic per-adapter saves,
+and the future Tinker ``save_state``/load-with-optimizer. The payload must be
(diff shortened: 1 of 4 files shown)


yushengsu-thu:

PR #2211, opened 2026-08-05
[multi-lora] 4/4: Option 1 — whole-batch atomic scheduling with slot oversubscription

yushengsu-thu:
Part 4/4 of the #2137 split (stacked on #2210; diff shows only this layer). The tree after this PR is byte-identical to #2137.

Replaces the accumulated-groups collector with the Option 1 design and wires the dormant components in:

- **MultiLoRARolloutFn**: one child RolloutFn per registration; whole child batches atomic (overshoot, never split); persistent round-robin with soft target and separate empty/coalesce clocks; per-adapter failure isolation; child aborts scoped to their own rid namespace; the `BatchPlan` is the only rollout→train control channel.
- **Variable-GBS integration**: the whole selection trains as one step on the rollout-side DP schedule; micro-batches re-order by slot after packing; nothing trims; dp-divisibility is not required; step normalization counts rollout executions (agentic siblings count once).
- **Oversubscription**: registry tenancy moves onto the SlotPool; registrations beyond the pool queue unbound and bootstrap when a slot frees; `commit_bind` keeps records and pool in sync.
- **Trainer**: found-inf/NaN veto per slot; schedulers tick on optimizer steps; swaps use the sidecars; driver executes bind plans fail-stop.
- Deletes the Option 2 path; rewrites the examples around the service workflow (README, `register_and_train.py` client).

Validated on this exact tree: fast suites green; GPU E2E — Qwen3-4B DP2 (7-phase smoke), oversubscription queue/bootstrap, and 4 adapters × 50 steps training quality (all four rewards improved, avg 0.070 → 0.185; `train_rollout_logprob_abs_diff` mean 0.0084 over 400 steps).

Series: #2208 → #2209 → #2210 → 4/4 (this). Supersedes #2137 as the merge vehicle if the series is preferred; #2137 remains the integrated reference.


yushengsu-thu:

PR #2213, opened 2026-08-05
fix(fsdp): apply the GDN packing patch to dense qwen3_5 and fix patch

Rockdu:
## What

- Add dense `qwen3_5` to the GDN packing patch's module loop (which fixes **GDN sequence packing leakage**). The dense arch lives in its own transformers module (`transformers.models.qwen3_5`), separate from `qwen3_5_moe`/`qwen3_next`, so it was silently skipped — while the "packing fix applied" log still fired, satisfied by the moe/next classes.
- Assign the per-forward packing boundaries unconditionally in the decoder wrapper: a packed forward writes fresh `cu_seqlens`/`seq_idx`, a non-packed forward clears them to `None`.

## Why

Under THD packing, dense Qwen3.5's linear-attention layers ran with **no per-document state reset**: the delta-rule recurrence and causal-conv state leaked across packed documents. Full-attention layers were never affected (`create_causal_mask` derives document boundaries from `position_ids` on its own).

The boundary attributes are a mailbox between the decoder-layer wrapper (producer) and the GDN wrapper (consumer). Writing them only on packed forwards leaves stale geometry behind for the next single-document row — such rows occur whenever a lone sample's length is an exact pad multiple, so no pad-zero pseudo-documents appear. A length mismatch then crashes the conv kernel (`seq_idx must have shape (batch_size, seqlen)`); an equal-length row would instead silently apply the previous row's boundaries.

## Validation

Live GRPO, `train/train_rollout_logprob_abs_diff` (Qwen3.5-4B, dapo-math-17k, 8xH200, 100 rollouts, identical seeds/args; wandb project `miles_fsdp_eval`). The leak-present runs climb to 2-4e-2; the fixed runs hold the ~6e-3 kernel-noise floor for all 100 rollouts:

train-vs-rollout logprob gap, with and without the fix

| run | code | packing fix | tail (last 5 rollouts) |
|---|---|---|---|
| `bdpw3y29` | main | no | 1.1e-2 – 2.0e-2 |
| `udjqh61a` | **main + this PR only** | yes | **5.7e-3 – 7.1e-3** |
| `knf6b9u1` | this PR + an unrelated precision experiment | yes | 5.1e-3 – 6.3e-3 |
| `14awwrd3` | `knf6b9u1` with the fix reverted (single-variable ablation) | no | 2.7e-2 – 3.3e-2 |

`udjqh61a` vs `bdpw3y29` is the exact single-variable comparison for this PR; `udjqh61a` overlapping `knf6b9u1` shows the fix accounts for the whole improvement on this metric.

Behavioral isolation test (Qwen3.5-4B, eager, two 96-token docs packed into one row):

| variant | doc-B packed vs doc-B alone | doc-B under different doc-A contents |
|---|---|---|
| without fix | up to 0.90 rel diff on logits | content-dependent (leak) |
| with fix | kernel tiling noise only | **bit-identical** |

## Files

- `miles/backends/experimental/fsdp_utils/models/qwen3_5_moe.py` — module loop + unconditional boundary assignment (+5/−6).

## Checklist

- [ ] `pre-commit run --all-files` passes — **not run locally**
- [ ] Added/updated tests for new behaviour — **no tests added** (removed per review; behavior covered by the measurements above)
- [ ] `pytest -x` is green — **not run on this branch**
- [x] If launch flags changed, `python3 train.py --help` still parses — no flags changed
- [x] If a public flag was added, it appears in the CLI reference docs — none added
- [x] If an example was added, it has a real walkthrough — none added


Zhichenzzz (commented):
rename the file qwen3_5_moe to qwen?

Rockdu:
> rename the file qwen3_5_moe to qwen?

Renamed qwen3_5_moe -> qwen3_5 for all related files

Zhichenzzz (approved):

PR #2219, opened 2026-08-05
[feat] Add training log-prob reuse to skip the redundant forward-only pass

guapisolo:
## Summary

Skip the redundant actor scorer forward for deterministic single-step Megatron training.

## Motivation

A single-step rollout evaluates the same actor weights twice: once to produce training log-probs and again inside the training forward. The standalone pass consumes model time without changing the old-policy baseline, and its cost repeats for every rollout.

## Usage

```bash
python train.py ... \
  --skip-actor-forward-only
```

For a supported single-step configuration, actor training omits standalone `compute_log_prob`. Policy loss uses detached training log-probs as the old policy unless `--use-rollout-logprobs` supplies it.

## Design Notes

- `validate_skip_actor_forward_only` permits only Megatron policy loss with one deterministic optimizer step; passive dumper observation and `--dump-details` are allowed because they do not recreate a skipped forward, while training source patching remains blocked.
- `policy_loss_function` selects detached current-training log-probs when `--skip-actor-forward-only` is set, otherwise reads optional precomputed batch log-probs; it selects the PPO old-policy source independently through `--use-rollout-logprobs` and keeps custom TIS inputs distinct.
- `MegatronTrainRayActor.train_actor` leaves the training forward responsible for consuming prefilled replay queues after standalone scoring is skipped.

## Verification

- `pytest -q tests/fast/utils/test_arguments.py tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py tests/fast/utils/test_dumper_utils.py`: 166 passed, covering dumper gates plus reference, teacher, and training forward preservation.
- `pytest -q tests/fast/utils/test_arguments.py tests/fast/dashboard/test_dump_reader_views.py::test_summary_and_tokens_survive_dump_without_log_probs tests/fast/dashboard/test_dump_reader_join.py::test_optional_columns_absent`: 131 passed, covering the direct and derived dump-details configurations plus dump readers without actor `log_probs`.
- `pre-commit run black --files miles/utils/arguments.py tests/fast/utils/test_arguments.py tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py`: passed.
- `python -m compileall -q miles/utils/arguments.py tests/fast/utils/test_arguments.py tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py`: passed.
- `git diff --check`: passed.

## Review Focus

- `validate_skip_actor_forward_only`: boundary between passive dumper observation, optional train-data `log_probs`, and training source patching.
- `MegatronTrainRayActor.train_actor`: skipped actor scoring while reference, teacher, and training forwards remain intact.
- `compute_advantages_and_returns` / `policy_loss_function`: fixed-score detach and old-policy source boundaries.


miles/backends/training_utils/loss.py
```diff
@@ -60,18 +64,24 @@ def compute_advantages_and_returns(args: Namespace, rollout_data: RolloutBatch)
...
-    # the policy update are fixed training data and must not retain a graph.
-    _detach_rollout_tensor_list(rollout_data, "log_probs")
-    _detach_rollout_tensor_list(rollout_data, "rollout_log_probs")
-    _detach_rollout_tensor_list(rollout_data, "ref_log_probs")
-    _detach_rollout_tensor_list(rollout_data, "teacher_log_probs")
-    log_probs = rollout_data.get(log_probs_key)
-    ref_log_probs = rollout_data.get("ref_log_probs")
+        if not (allow_missing_log_probs and get_parallel_state().is_pp_last_stage):
+            return
+    else:
```
yueming-yuan: Reported by my agent: when using KL loss and `allow_missing_log_prob`, the _detach will be skipped, so the ref_log_probs will not be detached, and the bug in #1966 exists again

miles/utils/arguments.py
```diff
@@ -3469,9 +3480,80 @@ def miles_validate_args(args):
...
+                args.custom_megatron_before_train_step_hook_path is not None,
+            ),
+            ("--custom-model-provider-path", args.custom_model_provider_path is not None),
+            ("--dumper-enable", args.dumper_enable),
+            ("--dumper-fwd-only", args.dumper_fwd_only is not None),
+            (
+                "--dumper-source-patcher-config-train",
+                args.dumper_source_patcher_config_train is not None,
+            ),
+            ("--dump-details", args.dump_details is not None),
```
yueming-yuan: qq: Why are (1) `--dump-details` and (2) the dumper args not allowed? For (1), I think it should not be blocked theoretically, just some field in the dump details might be missing; for (2), intuitively, some dumper passes might not be supported, but the rollout/training should still be able to dump?

(correct me if this is incorrect)

guapisolo: Good point. I removed the `--dump-details` incompatibility in `12cc66f1`. With `--skip-actor-forward-only`, the train-data dump can omit the precomputed actor `log_probs`; `DumpReader` already treats it as optional, while the policy-loss debug dump still records the training-forward and selected old-policy log-probs. I kept standalone `--save-debug-train-data` gated; values derived from `--dump-details` remain allowed. The dumper arguments were enabled in `7edbd954`.

yueming-yuan (approved):

PR #2226, opened 2026-08-06
fix: preserve routing-replay state around MTP spec creation

Shi-Dong:
## Motivation

Running PPO (`--advantage-estimator ppo`) with `--use-rollout-routing-replay` on a MoE model that has MTP layers (e.g. GLM-4.7-Flash with `mtp_num_layers=1`) crashes the critic on its first training step:

```
IndexError: pop_forward out of range: forward_index=0, len(top_indices_list)=0
```

Root cause: `routing_replay_manager` is a process-global singleton. The critic role deliberately leaves it disabled (`MegatronTrainRayActor.init` only enables replay managers for non-critic roles), because routing replay exists to align the actor's recomputed log-probs with the rollout's expert choices — the critic computes values fresh and has no log-probs to align.

However, `get_model_provider_func` brackets the MTP block-spec creation by setting `routing_replay_manager.enabled = False` and then unconditionally setting it back to `True`. That re-enable assumed the manager was enabled beforehand, which is only true for the actor. When the critic builds its model, the bracket flips the manager on as a side effect; the critic's MoE routers then register replay buffers, every training forward runs in replay mode, and the pop fails because replay data is only ever filled on the actor path.

The crash is not covered by CI: the PPO CI model (Qwen3-4B) is dense and has no MTP layers, and the bug needs the full combination of critic + MoE + MTP + `--use-rollout-routing-replay`.

## Modification

Save the manager's previous `enabled` state before the MTP bracket and restore that state afterwards, instead of forcing `True`. Behavior for the actor is unchanged (its previous state is `True`); the critic now stays disabled and computes routing fresh, which is the intended behavior.

### DIFF
--- a/miles/backends/megatron_utils/model_provider.py
+++ b/miles/backends/megatron_utils/model_provider.py
@@ -306,6 +306,7 @@ def get_model_provider_func(
             # hard code here to skip r3 registration for mtp layers
             # getattr is required to avoid ckpt conversion errors
             if getattr(args, "use_rollout_routing_replay", False):
+                prev_routing_replay_enabled = routing_replay_manager.enabled
                 routing_replay_manager.enabled = False
                 logger.warning(
                     "Rollout routing replay is not applicable for MTP modules, so skipped replay registration"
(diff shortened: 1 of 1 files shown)


guapisolo (approved):

guapisolo:
good fix!

yueming-yuan (approved):

PR #2229, opened 2026-08-06
[DO NOT MERGE] test CI: is test_fully_async flaky on main?

yueming-yuan:
**Do not merge.** Diagnostic PR only — a comment-only change on top of unmodified `main`.

## Why

`tests/e2e/megatron/test_qwen3_30B_A3B/test_fully_async.py` has failed twice on `stage-c-8-gpu-h100 (1)` with the same signature, on two unrelated PRs (#1572, #2223), both times with `ncclUnhandledCudaError` and never with a Python-level assertion. Neither PR changes anything on the fully-async path (`colocate=False`).

This PR isolates the variable: run the same test against clean `main` and see whether it fails there too.

Labelled `run-ci-fully-async`. Will be closed once the answer is in; any fix lands in its own PR.

### DIFF
--- a/tests/e2e/megatron/test_qwen3_30B_A3B/test_fully_async.py
+++ b/tests/e2e/megatron/test_qwen3_30B_A3B/test_fully_async.py
@@ -41,3 +41,5 @@ if __name__ == "__main__":
         os.environ.pop(proxy_var, None)
     prepare(CASE, need_fp8=CASE.use_fp8_rollout, need_int4=CASE.use_int4_rollout, all_bridge=CASE.use_bridge)
     execute(CASE, wandb_file=__file__)
+
+# CI probe: does test_fully_async fail on an unmodified main? Do not merge.


yueming-yuan:
## Answer: not a code problem — it is `novita-host2`

Clean `main` runs `test_fully_async.py` green (838s, 28 pass / 0 fail overall). The discriminator turned out to be the runner:

| run | runner | result |
|---|---|---|
| #1572 `8-gpu-h100 (1)` | `novita-host2` | FAIL `ncclUnhandledCudaError` / `Failed to CUDA calloc 2097152 bytes` |
| #2223 `8-gpu-h100 (1)` | `novita-host2` | FAIL, byte-for-byte same signature |
| this probe `8-gpu-h100 (0)` | `novita-host4` | PASS |

Both failures are on the same host, the only pass is on a different one.

Consistent with the rest of the evidence:

- The 2 MB `cudaMalloc` failed with **40+ GB free** on the trainer GPUs, so capacity was never the issue.
- The signature is byte-identical across two unrelated PRs (same 5 ranks, same size) — deterministic machine state, not a race.
- `test_fully_async` is the only test that runs the trainer and SGLang **concurrently** across one node's 8 GPUs, so it is the most sensitive to leftover GPU state (leaked handles / VA space / a sticky context error), which is why it is the one that surfaces this.

Neither #1572 nor #2223 touches the fully-async path (`colocate=False`), and no flag/coverage assertion fires in either failure log.

### Suggested action (ops, not a PR)

On `novita-host2`: `nvidia-smi -q | grep -i xid`, `dmesg | grep -i xid`, and check for leftover processes holding GPU handles; drain or reboot the host if confirmed.

Closing — diagnostic only, nothing here should merge.

PR #2235, opened 2026-08-07
feat(fsdp): support rollout routing replay (R3) for fsdp backend

Zhichenzzz:
Adds `--use-rollout-routing-replay` (R3, [arXiv 2510.11370](https://arxiv.org/abs/2510.11370)) to the experimental FSDP backend for `qwen3_moe`, `qwen3_5_moe`, and `glm4_moe_lite`.

## Design

R3's core is already backend agnostic and untouched here (`replay_base.py`, `replay_data.py`), as is the rollout side that emits `routed_experts`. Megatron installs the hook inside its fork's `TopKRouter`; FSDP trains stock HF modeling, so the hook is installed per model instance instead.

**Per arch hooks** (`models/replay_routers.py`) rebind the expert selection topk with a callable wrapped by `get_topk_fn`:

| arch | module owning the topk |
|---|---|
| `qwen3_moe` | `Qwen3MoeTopKRouter` (the block's `gate`) |
| `qwen3_5_moe` | `Qwen3_5MoeTopKRouter`, identical apart from always renormalizing, so it shares one forward |
| `glm4_moe_lite` | `Glm4MoeLiteMoE.route_tokens_to_experts` |

For GLM only the final expert selection topk is replaced; the intra group `topk(2)` and group selection are left intact, since replaying those would substitute expert ids into a group index space. Its `topk_weights = router_logits.gather(1, topk_indices)` already matches R3 semantics, so replaying indices keeps routing differentiable.

**Streams are keyed by global decoder layer index**, not by an ordinal over MoE blocks: the rollout tensor is `[tokens, num_layers, topk]` indexed by global layer id, so an arch with leading dense layers would be off by one on every layer. The index is parsed from the module path rather than walked to, because Qwen3.5 puts decoder layers at `model.model.layers` (text only) or `model.model.language_model.layers` (multimodal). Discovery matches by class name, so GLM's dense layer 0 is skipped without a config lookup.

**Stage orchestration** mirrors Megatron: ref log probs run `fallthrough`, the actor log prob pass runs `replay_forward` (or `record` for the non rollout `--use-routing-replay` variant, whose queues are still empty) then rewinds the forward cursors, and training runs as `replay_backward` while `_train_step` flips only its forward to `replay_forward`. That nesting is what makes gradient checkpointing correct, since HF re-runs each layer's forward during backward and that recompute draws from the independent backward cursor.

**The ref model deliberately does not register**: a second registration would double `manager.replays` and invalidate every `stream_idx`. `install` returns 0 without touching the model when replay is off.

Two pre-existing defects on FSDP reachable paths are fixed: `args.num_layers` did not exist under FSDP though `sglang_rollout` reshapes the routing buffer with it, and `fill_replay_data` read `args.sequence_parallel` directly and would raise `AttributeError`.

## Verification

8xH200, real checkpoints, `--colocate --gradient-checkpointing --ci-test`, 3 rollouts per run. Each arch was run twice, identical except for the flag.

`train/train_rollout_logprob_abs_diff`, per step:

| arch | model | R3 off | R3 on | reduction |
|---|---|---|---|---|
| `qwen3_moe` | Qwen3-30B-A3B | 0.0177, 0.0177, 0.0164 | 0.0084, 0.0087, 0.0093 | **~2.0x** |
| `glm4_moe_lite` | GLM-4.7-Flash | 0.0321, 0.0341, 0.0298 | 0.0274, 0.0286, 0.0258 | ~1.17x |
| `qwen3_5_moe` | Qwen3.5-35B-A3B | 0.0153, 0.0149, 0.0141 | 0.0074, 0.0076, 0.0072 | **~2.0x** |

Both qwen3 architectures land at roughly 2x. GLM's gain is much smaller, and worth stating plainly rather than averaging away: its absolute diff is roughly 3x higher than qwen3_moe's under *both* settings, so a large part of its train/infer gap is not routing related and R3 cannot remove it.

Two caveats on how to read this. Rollouts are sampled at `--rollout-temperature 1`, so the two runs of a pair see different rollout batches; this is a cross batch comparison, not a same data A/B. The per step spread within each run is under 10%, which is what makes the between condition gap meaningful. Runs also had `--ci-test` on, which enables `check_replay_result`: it compares the training engine's recomputed topk against the replayed indices per token and raises above a 1e-2 mismatched fraction. It never fired on any run.

Layer registration matched the expected count on every arch: 48 for `qwen3_moe`, 40 for `qwen3_5_moe`, and 46 at global indices 1..46 for `glm4_moe_lite`. That last one is the strongest single signal, since it rules out numbering by MoE block ordinal (`0..45`) and failing to skip the dense layer (`0..46`) at once. A fourth run with `--use-kl-loss --ref-load` confirmed the ref path: streams registered once per rank rather than doubled.

### On the tokens that did mismatch

A small number of zero overlap tokens were logged without ever crossing the threshold: 1 distinct (rank, token) pair for `qwen3_moe`, 20 for `glm4_moe_lite`, 43 for `qwen3_5_moe`. Only fully disjoint selections count, since `replay_check_min_overlap_ratio` is 0.0.

These are tie noise, not misalignment. Neighbouring tokens match essentially perfectly, for example on GLM:

```
token 754: orig=[29,42,57, 5]  replay=[29,57,42, 5]   same set, reordered
token 755: orig=[49,52,53,42]  replay=[60,62,57,44]   disjoint  <<<
token 756: orig=[29,38,45,18]  replay=[29,38,45,18]   identical
```

An alignment defect (off by one, boundary drift, wrong `stream_idx`) would be systematic and corrupt whole spans; an isolated token surrounded by exact matches cannot be produced that way. The spread across archs tracks routing structure: `qwen3_moe` has 128 experts and no grouping so ties flip one expert at a time, `qwen3_5_moe` has 256 experts, and `glm4_moe_lite` uses group limited routing with topk 4 so a group flip changes all four at once. That last case is exactly what R3 corrects, and is covered by `test_replay_can_select_experts_outside_the_recomputed_groups`.

One measurement caveat: Ray collapses repeated log lines, so a precise global mismatch rate cannot be recovered from the logs. What is certain is that the threshold assertion never fired, so mismatched tokens stayed under 1% in every invocation.

## Tests

Two unit files, one per module under test, plus shared fixtures:

* `test_fsdp_routing_replay.py` covers `adaptations/routing_replay.py`: stream index mapping, dense layer skipping, discovery behind an extra wrapper, adapter resolution per model type, no-op when disabled, stage set/restore/nesting including on exception, `log_prob_stage` picking record over replay for the non rollout variant, and a guard that `_create_ref_model` never installs replay.
* `test_fsdp_replay_routers.py` covers `models/replay_routers.py` for both archs: fallthrough matches the stock forward, replay overrides expert selection, forward and backward cursors stay independent, weights stay differentiable, replayed weights gather from the right tensor, and replay can select experts the training side group mask would have excluded.

These suites are `stage-a-cpu` so they use a device neutral `CpuReplay` (`tests/fast/fixtures/replay_fixtures.py`), following `tests/fast/utils/test_replay_base.py`; the real `Replay` pins its host buffer and returns indices on CUDA. Full `tests/fast/backends/` passes at 435 tests.

One e2e case per arch in `tests/e2e/fsdp/r3/`, all with `--ci-test`. These are the first MoE e2e cases on the FSDP backend; the existing FSDP e2e tests all use dense models, so there was nothing to extend.

## Known gap

`qwen3_5_moe_text` is not added to `VERIFIED_MODEL_TYPES`, since that list records archs with a completed FSDP RL run or e2e coverage that has actually run in CI.


miles/backends/experimental/fsdp_utils/actor.py
```diff
@@ -442,24 +446,30 @@ def train(
...
         ), f"Invalid num_microbatches {num_microbatches} for micro_batch_size {self.args.micro_batch_size} and global_batch_size {self.args.global_batch_size}"
 
         if self.ref_model is not None:
-            ref_results = self._compute_log_prob("ref", data_iterator, num_microbatches, store_prefix="ref_")
+            with routing_replay.stage(routing_replay.FALLTHROUGH):
+                ref_results = self._compute_log_prob("ref", data_iterator, num_microbatches, store_prefix="ref_")
             rollout_data.update(ref_results)
 
-        actor_results = self._compute_log_prob("actor", data_iterator, num_microbatches)
+        with routing_replay.stage(routing_replay.log_prob_stage(self.args)):
```
Rockdu, line 463: Not about this PR specifically, but I just realized we're forcing FSDP to use the recomputed logprob; is it as expected?

Rockdu:
LGTM! Also tested Qwen35 35B-A3B on DAPO with 8k context; the curve goes well
image
image

Rockdu:
Since it touches utils, which is outside the fsdp backend, cc @guapisolo for approval

guapisolo (approved):
approve to unblock

yueming-yuan (approved):

PR #2237, opened 2026-08-07
fix(openenv): build E2B task templates as root

nblintao:
E2B Cloud runs template-build commands as a non-root user, so every layer of the per-task TB2 recipe fails there — `apt-get` exits 100 and `/opt` is not writable. A self-hosted AgentENV builds as root, which is why the e2b backend passed its bring-up without this surfacing.

`set_user("root")` before the first build command fixes it. The build user also joins the template-alias digest, so changing it re-bakes rather than serving whatever the first builder produced.

The server exec now pins the same user. **That one is not a fix** — both endpoints were measured to already default to root at runtime. It states the contract, so that a provider default which changed later would not silently narrow what the agent may do inside a task image that expects root (TB2 solutions and tests `apt-install` freely).

## Verification

- Offline tests: 60 passed, 2 skipped (`pytest examples/experimental/openenv/tests/ -q`)
- ruff + black clean at the repo-pinned versions
- Live: template builds and episodes verified against E2B Cloud and a self-hosted AgentENV
- Since merging: all four templates of a training subset rebuild clean (2m57s) and
  a GRPO run on the E2B backend trains through them — 3 rollouts, non-zero rewards,
  no build or permission failures. Details in #2276.

### DIFF
--- a/examples/experimental/openenv/tb2_sandbox_e2b.py
+++ b/examples/experimental/openenv/tb2_sandbox_e2b.py
@@ -55,20 +55,33 @@ from tb2_sandbox_recipe import (
 )
 
 
+# The user every build command and the env server run as. The TB2 task images
+# are built for a root agent (their solutions and tests apt-install freely), so
+# anything less would change the task environment, not just the build.
+#
+# The BUILD is where this is load-bearing: E2B Cloud runs template-build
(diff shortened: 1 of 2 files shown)


Shi-Dong (approved):
LGTM!

PR #2238, opened 2026-08-07
refactor(openenv): unify the sandbox backends onto one SandboxBackend, and add Modal

nblintao:
> Stacked on #2237 (base is `tao/openenv-e2b-build-as-root`). Review that one first; this PR's diff is only the refactor and Modal.

Adds **Modal** as a third per-episode sandbox provider for the tbench2 recipe, and extracts what the providers had been duplicating so that adding one is two files plus a registry entry rather than a third copy of the orchestration.

## The abstraction

`openenv_sandbox_common.SandboxBackend` now owns everything provider-blind: throttled cancel-safe creates, the per-episode env client, and the training / direct-drive entry points. A backend module supplies only what is genuinely its own — how ONE sandbox comes into being, which of its SDK's errors are retryable throttling, and its knobs.

The registry (`AGENT_MODULES` / `resolve_backend` / `load_backend`) is the single place naming providers, so the launcher and the sibling tools cannot drift. `scan_golden` and `eval_tbench2_via_api` now reach a provider by naming it instead of growing an import branch each. That drift had already happened once — their help text still listed `daytona`/`e2b` after modal existed — so `backend_names()` generates it now and a test guards it.

## Modal specifics worth reviewing

- The image is **one layer per recipe command**, so a recipe edit re-runs only the layers below it.
- The env server is the sandbox **entrypoint**, not a follow-up exec, so the task image's own `CMD` cannot decide what runs.
- **No keepalive thread**, unlike the other two backends: a Modal timeout cannot be extended. Orphans are reclaimed by `idle_timeout` instead — the episode's WebSocket counts as activity, so a live episode of any length is safe and a hard-killed caller's sandbox disappears. This makes `OPENENV_MODAL_SANDBOX_TTL_S` a **hard ceiling** that must exceed `OPENENV_MAX_ROLLOUT_TIME_SECONDS`; the docs say so.

## Also here

- **Knob cleanup, 30 → 20.** Everything describing ONE sandbox moved next to the create that uses it; the backend modules keep only the fan-out knobs, read through one shared `backend_knobs(prefix)`. Adding Modal therefore cost 4 knobs, not 14.
- **A launcher guard** against a forwarded endpoint that smuggles a credential in its userinfo. `forward` values ride ray's `runtime_env`, which is echoed into driver logs in plaintext; the existing test could only check the variable's *name*, which a URL defeats.
- **`tb2_sandbox_recipe`** gains `run_with_deadline` and `task_env_resources`, which the materializations had copied. The resource floors are recipe policy and were drifting — Daytona applied them in GB, the others in MB.
- **Docs:** per-episode sandboxes are now the recipe's primary path with one shared env server as the alternative, each backend gets its own heading, and the sanity checks become their own (optional) step.

## Verification

Offline: **100 passed, 3 skipped**; ruff + black clean at the repo-pinned versions (0.14.7 / 24.3.0).

**One episode end to end, per backend** — create → connect → `reset` over the WebSocket → `exec` → `evaluate` → close, same task (`adaptive-rejection-sampler`) everywhere:

| Backend | Endpoint | Create → connected | Total |
|---|---|---|---|
| Modal | Modal Cloud | 9 s | 23 s |
| E2B | E2B Cloud | 13 s | 26 s |
| E2B (`agentenv`) | self-hosted AgentENV | 8 s | 17 s |
| Daytona | Daytona Cloud | — | 60 s |

All four: `whoami` → `root`, the marker exec round-trips, and `evaluate` returns a verdict from `harness='tests/test.sh'` with no error. **Daytona had unit-test coverage only when this PR was opened; this is its first live run on the shared backend**, so all three providers are now confirmed on the same code path rather than two confirmed and one inferred.

**Scoring is correct, not merely present.** The single episodes above all score `0.0` — nothing solved the task — which proves a verdict comes back but not that a *correct* solution would be recognized. So each task's official solution was replayed through the full sandbox and scoring path, 5 tasks at concurrency 5:

| | Modal | E2B Cloud |
|---|---|---|
| `chess-best-move` | 1.0 | 1.0 |
| `cancel-async-tasks` | 1.0 | 1.0 |
| `count-dataset-tokens` | 1.0 | 1.0 |
| `distribution-search` | 1.0 | 1.0 |
| `bn-fit-modify` | 1.0 | 1.0 |
| **golden pass** | **5/5** (0 errored) | **5/5** (0 errored) |

Running 5 at once also exercises what a single episode never touches and what this PR moved: the process-wide create semaphore, the throttle/backoff path, and E2B's per-alias build lock. Concurrent sandboxes were observed live under the recipe's ownership tags.

**The multi-turn agent loop**, with a real policy (DeepSeek-v4-flash) driving `eval_tbench2_via_api` on E2B — this is the path through the entry points the refactor renamed:

```
[0] chess-best-move       turns=15  tool_calls=15  eval=6.5s
[0] cancel-async-tasks    turns=15  tool_calls=15  eval=17.2s
```

A turn count cannot distinguish "the model did not solve it" from "the loop burned its turns on env errors", so `cancel-async-tasks` was re-run with the full trajectory captured. **It scored 1.0** — same backend, same code path, same model, sampled at temperature 0.8. A run that reaches a passing verdict rules out the loop being broken in the run that did not.

The trajectory shows why: `reset` delivers the task's real instruction; turn 1 writes `/app/run.py`; turn 2 `cat`s it back and receives the file's actual contents (so sandbox state persists across turns and the outputs fed back are genuine, not canned); later turns run Python harnesses and get their real stdout (`started [1, 2] / cleaned [1, 2] / started==cleaned True`). The zeros above are the model, not the infrastructure.

**Nothing leaked.** After every run above, on all four endpoints:

```
modal   leftovers: []
e2b     leftovers: []
daytona leftovers: []
AgentENV leftovers: []
```

### DIFF
--- a/examples/experimental/openenv/tests/test_tb2_sandbox_modal.py
+++ b/examples/experimental/openenv/tests/test_tb2_sandbox_modal.py
@@ -1,0 +1,246 @@
+"""Tests for the Modal sandbox half: image layering, sandbox sizing, the
+tunnel URL, and the create path's lifetime/diagnostic contract.
+
+Not collected by the repo-level pytest run (testpaths = ./tests); run manually
+when touching the recipe:
+
+    pytest examples/experimental/openenv/tests/ -q
+"""
+
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+import tb2_sandbox_modal as sandbox  # noqa: E402
+
+_COMMANDS = ("apt-get install -y curl", "curl -LsSf https://astral.sh/uv/install.sh | sh", "mkdir -p /opt/tb2-tasks")
+
+
+# --- fake modal SDK ---------------------------------------------------------
+
+
+class _FakeImage:
+    """Records layering: every run_commands call is one layer."""
+
+    def __init__(self, tag, layers=()):
+        self.tag = tag
+        self.layers = list(layers)
+
+    def run_commands(self, *commands):
+        return _FakeImage(self.tag, [*self.layers, commands])
+
+    @staticmethod
+    def from_registry(tag, **kwargs):
+        _FakeImage.from_registry_kwargs = kwargs
+        return _FakeImage(tag)
+
+
+class _FakeTunnel:
+    def __init__(self, url):
+        self.url = url
+
+
+class _FakeSandbox:
+    """Stands in for one created sandbox."""
+
+    def __init__(self, exit_code=None, output="", tunnel_url="https://abc.r5.modal.host"):
+        self._exit_code = exit_code
+        self.stdout = types.SimpleNamespace(read=lambda: output)
+        self.terminated = False
+        self.tunnel_timeout = None
+        self._tunnel_url = tunnel_url
+
+    def tunnels(self, timeout=50):
+        self.tunnel_timeout = timeout
+        return {8000: _FakeTunnel(self._tunnel_url)}
+
+    def terminate(self):
+        self.terminated = True
+
+    def poll(self):
+        return self._exit_code
+
+
+class _FakeSandboxCls:
+    created = None
+    instance = None
+
+    @staticmethod
+    def create(*args, **kwargs):
+        _FakeSandboxCls.created = {"args": args, "kwargs": kwargs}
+        return _FakeSandboxCls.instance
+
+
+class _FakeApp:
+    lookups = None
+
+    @staticmethod
+    def lookup(name, **kwargs):
+        _FakeApp.lookups.append((name, kwargs))
+        return f"app:{name}"
+
+
+@pytest.fixture
+def fake_modal(monkeypatch):
+    """Install a fake `modal` module and reset the backend's process-wide app cache."""
+    _FakeApp.lookups = []
+    _FakeSandboxCls.instance = _FakeSandbox()
+    mod = types.ModuleType("modal")
+    mod.Image = _FakeImage
+    mod.Sandbox = _FakeSandboxCls
+    mod.App = _FakeApp
+    monkeypatch.setitem(sys.modules, "modal", mod)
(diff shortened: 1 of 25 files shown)


examples/experimental/openenv/tests/test_tb2_sandbox_modal.py
```diff
@@ -0,0 +1,246 @@
...
+    with pytest.raises(RuntimeError, match="health check timed out"):
+        sandbox.create_task_sandbox(Path("/tasks/t"))
+    assert _FakeSandboxCls.instance.terminated
+
+
+def test_create_enforces_the_build_wall_clock(fake_modal, monkeypatch):
+    """The first create for a task builds the image, which Modal does not
+    deadline itself; a hung build must give the slot back rather than block the
+    episode forever."""
+    import threading
```
Shi-Dong, line 223: please hoist this import to the beginning of the script

nblintao:
Closing in favour of a three-PR stack, per @Shi-Dong's suggestion — thanks, it split cleanly along exactly the lines you proposed:

- **1/3** #2274 — extract the duplicated TB2 helpers (~130 lines, no new behaviour; removes duplication that exists on `main` today)
- **2/3** #2275 — the shared `SandboxBackend` and the registry
- **3/3** #2276 — Modal, the launcher credential table, and the docs

Two notes on how the pieces were assigned. The launcher credential work went into 3/3 rather than 2/3 because it is Modal-motivated: Modal has no single API key, so the daytona/e2b if-else could not express its token pair. The README restructure went there too, for the same reason — the old "2b. Alternative" framing broke down at three backends.

The three layers stacked reconstruct this PR's tree exactly, with one intentional difference: the `import threading` you flagged is now hoisted. Each layer is independently green (61 / 70 / 100 tests, ruff + black clean). Live verification against all four endpoints is reported on 3/3, since that is the state that was actually run.

PR #2259, opened 2026-08-07
feat(session): translate Anthropic message requests

Shi-Dong:
## Summary

- add a dependency-free Anthropic Messages request translator for the session server
- preserve system prompts, multimodal text/image content, reasoning history, tool calls, and ordered tool results
- translate Anthropic sampling and tool-choice fields into SGLang's OpenAI-compatible request shape
- reject unsupported or lossy inputs with explicit protocol errors
- keep pure protocol coverage under `tests/fast/rollout/session/`
- fix multi-part user content being cleared after conversion by copying the emitted parts before resetting the accumulator

## Why

Claude Code sends Anthropic `/v1/messages` requests, while Miles records OpenAI chat-completion requests for TITO and sample assembly. This isolates the request-side wire conversion before any session-server behavior changes.

The expanded tests cover the common Claude Code history shapes identified by the coverage review: list and empty tool results, omitted tool choice, plain assistant text, base64 and URL images, mixed content, and malformed protocol inputs.

## Stack

PR 1 of 4. The remaining PRs add response/SSE conversion, the recorded session route, and Miles-owned sampling overrides.

## Tests

`python -m pytest tests/fast/rollout/session/test_anthropic_protocol.py`

- 38 tests pass
- isolated branch coverage for `miles/rollout/session/anthropic.py`: 180/180 statements and 114/114 branches (100%)

`uvx pre-commit run --files miles/rollout/session/anthropic.py tests/fast/rollout/session/test_anthropic_protocol.py`

PR #2261, opened 2026-08-07
feat(session): render Anthropic message responses

Shi-Dong:
## Summary

- convert complete OpenAI chat-completion responses into Anthropic Messages responses
- preserve reasoning, text, tool calls, stop reasons, and token usage
- synthesize protocol-valid Anthropic SSE streams from Miles's complete recorded response
- reject aborted generations with a dedicated `AnthropicGenerationAbortedError` instead of disguising them as successful `end_turn` responses
- guarantee `tool_use` when a completed response contains tool-use blocks, even if upstream reports `finish_reason: stop`
- sanitize malformed and non-finite tool arguments before SSE serialization
- translate upstream errors into Anthropic error envelopes, including Anthropic-specific 413 `request_too_large` and 529 `overloaded_error` types
- use assignment expressions for response fields that are fetched and immediately validated

## Why

Miles deliberately keeps the SGLang request non-streaming so TITO can record complete token and log-probability metadata. Claude Code still expects Anthropic streaming events. This layer renders a complete recorded response as a valid `message_start` / content blocks / `message_delta` / `message_stop` stream.

An aborted Miles generation is incomplete training data. The converter raises the dedicated `AnthropicGenerationAbortedError` for `finish_reason: abort`, allowing the session route to make only that condition retryable without catching unrelated protocol failures.

## Protocol decisions

- a successful upstream response with no representable content remains a protocol-valid empty text response; HTTP failure detection and non-2xx propagation belong to the session route
- Miles receives reasoning text but no Anthropic thinking signature, so it cannot emit `signature_delta`
- OpenAI finish reasons do not identify which stop sequence matched, so `stop_sequence` remains `null`

## Stack

PR 2 of 4, stacked on #2259. The remaining PRs add the `/v1/messages` session route and Miles-owned sampling overrides.

## Tests

`python -m pytest tests/fast/rollout/session/test_anthropic_protocol.py`

`uvx pre-commit run --files miles/rollout/session/anthropic.py tests/fast/rollout/session/test_anthropic_protocol.py`

- all 62 protocol tests pass locally
- isolated coverage for `miles/rollout/session/anthropic.py`: 285/285 statements and 174/174 branches (100%)
- regression coverage includes the dedicated abort exception, `stop` plus tool calls, and non-finite JSON arguments rendered through SSE

### DIFF
--- a/miles/rollout/session/anthropic.py
+++ b/miles/rollout/session/anthropic.py
@@ -286,8 +286,8 @@ def _response_text_blocks(content: Any) -> list[dict[str, Any]]:
         return [{"type": "text", "text": str(content)}]
     blocks: list[dict[str, Any]] = []
     for part in content:
-        if isinstance(part, Mapping) and part.get("type") == "text" and isinstance(part.get("text"), str):
-            blocks.append({"type": "text", "text": part["text"]})
+        if isinstance(part, Mapping) and part.get("type") == "text" and isinstance(text := part.get("text"), str):
+            blocks.append({"type": "text", "text": text})
     return blocks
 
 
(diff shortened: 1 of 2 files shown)


guapisolo:
CC told me there is a bug that tool call serialize / deserialize error. I have another PR #2240 to fix this issue.

PR #2272, opened 2026-08-08
refactor: move the FSDP backend out of experimental

Zhichenzzz:
Moves the FSDP backend out of `experimental`: `miles/backends/experimental/fsdp_utils/` becomes `miles/backends/fsdp_utils/`, a sibling of `megatron_utils`, `sglang_utils` and `training_utils`. `experimental/` held nothing else, so it is removed entirely.

The package loses one level of nesting, so the seven cross-package relative imports inside it are converted to absolute imports rather than having their dot counts adjusted. That matches the repo's code style and makes the package position independent.

Beyond the Python imports, three references would have broken the ROCm and NPU builds and are updated too:

* `docker/Dockerfile.rocm` creates and copies `sglang_attn_bridge` into the package path
* `docker/amd_patch/latest/miles.patch` and `docker/npu_patch/miles.patch` apply to files inside it

`.github/CODEOWNERS` is updated in place; the entry keeps its position after `/miles/backends/` so the more specific rule still wins.

`setup.py` uses `find_packages(include=["miles*"])`, and no CI workflow filters on the old path, so neither needed changes.

## Verification

* `pre-commit run --all-files` clean
* `tests/fast/backends/` 447 passed
* Full `tests/fast/` 4547 passed with one failure, `test_tracking_backend_end_to_end_local_ray`, which fails identically on unmodified `main` and is unrelated to this move

## Not included

`docs/developer/experimental-features.md` and `docs/developer/architecture.md` still describe the old location. Left for a follow-up so this stays a code-only move.


fzyzcjy (approved):
LGTM if it is pure move

PR #2274, opened 2026-08-08
refactor(openenv): move duplicated sandbox helpers into the TB2 recipe

nblintao:
> 1/3 of a stack that adds a Modal sandbox backend. This one stands alone: it removes duplication that exists on `main` today, with no new behaviour.
> **2/3** #2275 · **3/3** #2276

The Daytona and E2B materializations had each grown their own copy of two things that are neither provider-specific nor harmless when the copies drift. Both now come from `tb2_sandbox_recipe`, the module that already owns what is common to every provider.

**`task_env_resources`** reads the task's CPU/memory/disk from `task.toml` and applies the floors the env server needs to run alongside the task. Those floors are recipe policy, and the copies had already diverged — Daytona applied them in GB (`max(2, memory_mb // 1024)`), the others in MB — so the same `task.toml` could be honored differently depending on where it ran.

**`run_with_deadline`** bounds a provider call that blocks across many requests with no deadline of its own. Both copies carried the same comment explaining why it cannot be a `with` block (its `__exit__` joins the worker and waits out the very call the deadline exists to abandon), which is the kind of reasoning that should live in one place.

Also folded in:

- Two knobs that describe **one sandbox** move next to the create that uses them: `TB2_COMMAND_TIMEOUT_S` (read where the server command line is built, so every provider gets the same value instead of three `getenv` copies) and a per-provider ready timeout.
- A failed create now guards its own cleanup, so a `kill`/`delete` that itself fails cannot replace the exception explaining why the create failed. The provider's TTL reclaims the sandbox either way.

## Verification

Offline tests: **61 passed, 2 skipped**; ruff + black clean at the repo-pinned versions (0.14.7 / 24.3.0). Live verification for the stack is on 3/3, which is the state that was actually run against the providers.

### DIFF
--- a/examples/experimental/openenv/tb2_sandbox_daytona.py
+++ b/examples/experimental/openenv/tb2_sandbox_daytona.py
@@ -28,13 +28,20 @@ from pathlib import Path
 
 import tb2_sandbox_recipe as recipe
 from tb2_sandbox_recipe import (
-    read_task_config,
+    COMMAND_TIMEOUT_S,
     resolve_docker_image,
     sandbox_labels,
     server_cmd,
     server_layer_commands,
+    task_env_resources,
     wait_server_ready,
 )
+
+
+# Every knob describing ONE Daytona sandbox lives here, next to the create
+# that uses it; the backend module keeps only the fan-out knobs. Read at import:
+# a rollout worker is a fresh process per run.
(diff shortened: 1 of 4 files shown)


Shi-Dong (approved):
LGTM!

PR #2275, opened 2026-08-08
refactor(openenv): give the sandbox backends one SandboxBackend to share

nblintao:
> 2/3 of a stack that adds a Modal sandbox backend. Base is #2274, so review that first; the diff here is only the abstraction.
> **1/3** #2274 · **3/3** #2276

The Daytona and E2B agent functions were two copies of the same orchestration around two genuinely different holes. Everything not provider-specific now lives once, in `openenv_sandbox_common.SandboxBackend`: throttled cancel-safe creates, the env client bound to one episode's sandbox, and the training / direct-drive entry points.

A backend module keeps only what is its own — how ONE sandbox comes into being, which of its SDK's errors are worth retrying, and its knobs — and shrinks to about a hundred lines, most of it documentation.

That matters because the copies were already diverging in ways nobody would choose deliberately: the cancel-mid-create reaper and the backoff constants were maintained per module. Anything the two must agree on is now impossible to disagree on.

**The registry** (`AGENT_MODULES` / `resolve_backend` / `load_backend` / `backend_names`) becomes the single place that names providers. The sibling tools (`scan_golden`, `eval_tbench2_via_api`) reach a backend by naming it rather than growing an import branch each, and their operator-facing help text is now generated from the registry instead of written by hand — with a test that fails if a registered backend goes unmentioned, because that drift is easy to miss in review.

**Knobs** are read through one shared `backend_knobs(prefix)`, so every backend gets the same names under its own prefix and can be tuned without touching another's.

**Four names in `openenv_agent_function`** go from protected to public (`multi_turn`, `run_for_training`, `MESSAGE_TIMEOUT_S`, `load_tbench2`). They now have a consumer outside their own module, so the leading underscore was claiming an encapsulation that no longer held.

`leg` is kept where it still means the shared-server-vs-sandbox distinction, and dropped where it had come to mean "backend".

### Why now, with only two backends

Fair question for an abstraction over two implementations. The answer is 3/3 in this stack: adding a provider on top of this is two files plus a registry entry, and none of the behaviour above is restated. Reviewing them separately is the point — this PR is the shape, that one is the payoff.

The recipe README still uses the older wording in places; it is rewritten in 3/3, alongside the third backend that motivates the restructure.

## Verification

Offline tests: **70 passed, 2 skipped**; ruff + black clean at the repo-pinned versions. Live verification for the stack is on 3/3, which is the state that was actually run against the providers.

Since merging, a GRPO run on the E2B backend has exercised this abstraction end
to end (3 rollouts, all `outcome=NORMAL`, non-zero rewards, weights broadcast
back each step) — i.e. `SandboxBackend` carries a real training run and not just
the per-episode paths. Numbers are in #2276, alongside the same run on Modal.

### DIFF
--- a/examples/experimental/openenv/eval_tbench2_via_api.py
+++ b/examples/experimental/openenv/eval_tbench2_via_api.py
@@ -2,7 +2,7 @@
 per-episode sandboxes as the env. No GPU, no miles training pipeline.
 
 This reuses the exact agent-env loop miles runs during training
-(``openenv_agent_function._multi_turn``: reset -> {policy emits a shell command
+(``openenv_agent_function.multi_turn``: reset -> {policy emits a shell command
 -> exec -> feed output back} -> canonical tests/test.sh -> binary reward) but
 swaps miles' session-server policy for a plain API client. The machine running
 this only orchestrates; the policy runs in the cloud (e.g. DeepSeek) and each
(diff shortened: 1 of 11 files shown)


Shi-Dong (approved):
LGTM!

PR #2276, opened 2026-08-08
feat(openenv): add Modal as a third per-episode sandbox backend

nblintao:
> 3/3 of a stack. Base is #2275, so review **1/3** #2274 and **2/3** #2275 first; the diff here is Modal plus the docs.
> This is the state that was verified live against all four endpoints — see below.

With the shared `SandboxBackend` from 2/3 in place, a provider is its two genuinely specific parts plus a registry entry: how ONE sandbox comes into being (`tb2_sandbox_modal`) and which of its SDK's errors are retryable throttling (`openenv_modal_agent_function`). Nothing about the rollout-facing behaviour is restated — which is what the two commits below this one bought.

## Three things are Modal's own

- **One image layer per recipe command**, so editing the recipe re-runs only the layers below the edit. Layers are *not* shared across tasks — TB2 pins a different official image per task, so every chain hashes differently from the first layer down. That is a property of the benchmark, not of this backend, and it applies to Daytona and E2B equally.
- **The env server is the sandbox ENTRYPOINT**, not a follow-up exec. Modal would otherwise run the task image's own `CMD`, and what starts in the sandbox has to be the recipe's decision.
- **No keepalive thread**, unlike the other two backends, because a Modal timeout *cannot be extended*. Orphans are reclaimed by `idle_timeout` instead: an open tunnel connection counts as activity and the episode holds its WebSocket for its whole life, so a live episode of any length is safe while a hard-killed caller's sandbox disappears. The consequence worth knowing when sizing a run: `OPENENV_MODAL_SANDBOX_TTL_S` is a **hard ceiling** that must exceed `OPENENV_MAX_ROLLOUT_TIME_SECONDS`, not merely a heartbeat interval.

## Launcher credentials

This backend broke the launcher's assumption: Modal has no single API key, so the daytona/e2b if-else could not express it, and the wiring becomes a table. A partially-supplied token pair is now treated as **missing** rather than usable — half a credential authenticates nothing and would fail every episode.

Forwarded values are also checked for a credential smuggled in a URL's userinfo (`https://user:token@host`). `forward` may ride ray's `runtime_env` only because it names *addresses*, and `runtime_env` is echoed into driver logs in plaintext; the existing test checks the variable's **name**, which a URL defeats.

## Docs

Per-episode sandboxes become the recipe's primary path with one shared env server as the alternative, each backend gets its own heading, and the sanity checks become their own optional step. The old "2b. Alternative" framing dated from when there was one provider.

## Verification

Offline: **100 passed, 3 skipped**; ruff + black clean at the repo-pinned versions (0.14.7 / 24.3.0).

**One episode end to end, per backend** — create → connect → `reset` over the WebSocket → `exec` → `evaluate` → close, same task everywhere:

| Backend | Endpoint | Create → connected | Total |
|---|---|---|---|
| Modal | Modal Cloud | 9 s | 23 s |
| E2B | E2B Cloud | 13 s | 26 s |
| E2B (`agentenv`) | self-hosted AgentENV | 8 s | 17 s |
| Daytona | Daytona Cloud | — | 60 s |

All four: `whoami` → `root`, the marker exec round-trips, `evaluate` returns a verdict from `harness='tests/test.sh'` with no error.

**Scoring is correct, not merely present.** Those single episodes all score `0.0` — nothing solved the task — which proves a verdict comes back but not that a *correct* solution is recognized. So each task's official solution was replayed through the full sandbox and scoring path, 5 tasks at concurrency 5: **5/5 at reward 1.0 on Modal and 5/5 on E2B Cloud**, 0 errored. Running five at once also exercises what a single episode never touches: the process-wide create semaphore, the throttle/backoff path, and E2B's per-alias build lock.

**The multi-turn agent loop**, with a real policy (DeepSeek-v4-flash) on E2B, reached `reward=1.0` on `cancel-async-tasks`. The trajectory shows `reset` delivering the real instruction, turn 1 writing `/app/run.py`, turn 2 `cat`-ing it back and receiving the file's actual contents (so sandbox state persists across turns and the fed-back output is genuine), and later turns running Python harnesses against their real stdout.

**Nothing leaked** — after every run above, on all four endpoints:

```
modal   leftovers: []
e2b     leftovers: []
daytona leftovers: []
AgentENV leftovers: []
```

## GRPO training on merged `main`

The verification above is per-episode. This is the same code inside a real GRPO
run: 8×H200 single node, GLM-4.7-Flash, TP=4 / EP=2 colocated, on `d2e3ad9f1` —
the *merged* form of this stack rather than the branch, since the devbox image
already carried it.

Aperture: `--rollout-batch-size 2 --n-samples-per-prompt 4` (8 episodes per step),
`OPENENV_MAX_TURNS=15`, over a 4-task subset the base policy is known to
sometimes solve, so GRPO sees advantage variance instead of a flat batch.
`--num-rollout` was left at the launcher's hardcoded 40 and the run stopped once
the loop had repeated several times; nothing in the recipe was edited to run it.

Per-episode rewards, read out of `--dump-details`
(`dump/rollout_data/<rollout_id>.pt`) rather than the driver log:

| Step | Modal | E2B |
|---|---|---|
| rollout 0 | 0.000 `[0,0,0,0,0,0,0,0]` | 0.000 `[0,0,0,0,0,0,0,0]` |
| rollout 1 | 0.125 `[0,0,0,0,0,0,1,0]` | 0.250 `[0,0,1,0,0,0,0,1]` |
| rollout 2 | 0.250 `[0,0,0,0,1,0,0,1]` | 0.250 `[0,0,0,0,0,1,1,0]` |
| rollout 3 | 0.250 `[0,0,0,0,0,1,1,0]` | — |
| rollout 4 | 0.125 `[0,0,0,0,0,0,1,0]` | — |

Every train step reported `outcome=NORMAL`, and each was followed by
`update_weights_from_tensor` broadcasting 118 tensors back into SGLang — so the
loop closes: rollout → reward → train step → weight update → next rollout. That
the means are non-zero *and* varying matters more than their absolute value: an
all-zero batch yields zero advantage and would have proven nothing about the
training path.

Golden replay was also re-run per backend on this box: **Modal 4/4 at reward 1.0**
(1m46s, layer cache warm), and all four E2B templates rebuilt clean (2m57s).

**Reclamation, observed against the providers rather than reasoned about.**
Mid-run, E2B had **6 live sandboxes** during rollout 3 — the in-flight batch, not
the ~24 accumulated over all steps — each carrying its `openenv-launcher` /
`openenv-tbench2-task` labels, so per-episode cleanup works. Killing a run
mid-episode is the case that cleanup *cannot* cover, and it is the one each
provider's backstop exists for: Modal left 1 sandbox behind and it was gone after
the idle-timeout window; E2B left 3, reclaimed once the keepalive thread stopped
re-arming the TTL. Both ended at **0 live, nothing left behind**.

The credential contract shows up in the run's own logs, too. The `EXEC:` line
echoes the whole `ray job submit`, and its `--runtime-env-json` contains
`"E2B_API_KEY_FILE": "/root/.config/e2b/api_key"` — the path, never the value.
That is at once the evidence that `runtime_env` does reach plaintext logs, and
that forwarding only the path keeps the credential out of them.

### DIFF
--- a/docs/user-guide/environments.md
+++ b/docs/user-guide/environments.md
@@ -19,11 +19,11 @@ where the environment itself comes from:
 | Integration | Plugs in at | Guide |
 |---|---|---|
 | [Harbor](https://github.com/harbor-framework/harbor) | agent function | [guide](/user-guide/harbor) |
-| [OpenEnv](https://github.com/huggingface/openenv) | agent function | [guide](/user-guide/openenv) |
 | [NeMo-Gym](https://github.com/NVIDIA-NeMo/Gym) | agent function | [guide](/user-guide/nemo-gym) |
+| [OpenEnv](https://github.com/huggingface/openenv) | agent function | [guide](/user-guide/openenv) |
 | [Strands Agents](https://strandsagents.com/) | generate function | [example](https://github.com/radixark/miles/tree/main/examples/experimental/strands_sglang) |
-| [τ-bench](https://github.com/sierra-research/tau-bench) | generate function | [example](https://github.com/radixark/miles/tree/main/examples/experimental/tau-bench) |
 | [Verifiers (Prime Intellect)](https://github.com/PrimeIntellect-ai/verifiers) | rollout function | [guide](/user-guide/verifiers) |
+| [τ-bench](https://github.com/sierra-research/tau-bench) | generate function | [example](https://github.com/radixark/miles/tree/main/examples/experimental/tau-bench) |
 
 Sandbox providers are a different axis: they provision the task containers
 *inside* a connector rather than occupying a rollout layer.
@@ -33,8 +33,9 @@ Sandbox providers are a different axis: they provision the task containers
 | [AgentENV](https://github.com/kvcache-ai/AgentENV) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/agentenv) |
 | [Daytona](https://www.daytona.io/) | OpenEnv, Harbor, NeMo-Gym | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 | [E2B](https://e2b.dev/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
+| [Modal](https://modal.com/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 
-All external ecosystem support is experimental.
+Everything above is experimental, and listed alphabetically.
 
 ## Integration shapes
 
--- a/docs/user-guide/openenv.md
+++ b/docs/user-guide/openenv.md
@@ -19,11 +19,13 @@ hook.
 
 The maintained end-to-end recipe is **Terminal-Bench-2 GRPO** in
 [`examples/experimental/openenv`](https://github.com/radixark/miles/tree/main/examples/experimental/openenv).
-It runs against a shared Docker env server (full per-task image fidelity) or
-per-episode cloud sandboxes built from each task's official image (no resident
-infrastructure) on a choice of providers: [Daytona](https://www.daytona.io/),
-[E2B](https://e2b.dev/), or a self-hosted, E2B-compatible
-[AgentENV](https://github.com/kvcache-ai/AgentENV) deployment. Follow the
+It gives every episode its own cloud sandbox, built from that task's official
+image so no resident infrastructure is left behind, on a choice of providers —
+[AgentENV](https://github.com/kvcache-ai/AgentENV) (self-hosted, E2B-compatible),
+[Daytona](https://www.daytona.io/), [E2B](https://e2b.dev/), or
+[Modal](https://modal.com/). One shared Docker env server is supported as well,
+for running without any sandbox platform.
+Follow the
 [recipe README](https://github.com/radixark/miles/blob/main/examples/experimental/openenv/README.md)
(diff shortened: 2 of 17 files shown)


nblintao:
@Shi-Dong the `import threading` is hoisted, and I swept the rest of the recipe for the same pattern rather than fixing just that one. Result across the stack:

**One more genuinely unjustified case, now fixed** — `eval_tbench2_via_api.py` imported `openenv_sandbox_common` inside `main()`. It is a sibling module that pulls in nothing but stdlib, so there was no reason to defer it; hoisted in #2275, which is where that file changes anyway.

**Three that are load-bearing, now documented instead of moved.** `tb2_sandbox_recipe` defers `requests` and `tbench2_env`, and the reason turns out to be concrete: neither package is installed in the repo venv the offline tests run in, and the module still imports there *because* those imports are deferred. Hoisting either would break `pytest examples/experimental/openenv/tests/`. They had no comment saying so, which is exactly the gap your review is about, so each now states it (in #2274).

**The provider SDK imports stay deferred**, as before — `e2b` / `modal` / `daytona` are optional dependencies the launcher preflights, and the offline tests import these modules with a fake SDK in `sys.modules`. Those are covered by a statement in each module's docstring rather than repeated at every call site.

One non-finding worth mentioning so it does not look overlooked: `tb2_sandbox_modal.py:40` shows `import asyncio, modal` inside the module docstring — it is example code in the sandbox-sweep snippet, not a real import.

Shi-Dong (approved):
LGTM!

PR #2278, opened 2026-08-08
feat(session): request and assemble additional R3 rows under in-place weight updates

guapisolo:
## Summary
TITO sessions return only new R3 rows in `in_place` mode while preserving the full tensor.

## Motivation
TITO sessions resend the full token sequence every turn, so with routing replay enabled SGLang returns routed-expert rows for the whole prefix again and the cumulative R3 payload grows with the sum of all turn lengths.

## Usage
```bash
--use-session-server --use-rollout-routing-replay --pause-generation-mode in_place
```
`SessionServer` then sends `routed_experts_start_len = max(0, len(checkpoint_token_ids) - 1)` on each routing-replay chat request. `retract` continues requesting full R3, while `--use-session-server --pause-generation-mode abort` fails argument validation.

## Design Notes
- `_maybe_request_addition_r3` uses the checkpoint length as the strict append boundary; LCP is only a sanity assertion that those retained rows remain a causal prefix of the new prompt.
- `merge_samples_with_addition_r3` runs the ordinary sample merge first, requires each consumed record to start at the current covered row, and concatenates only enough patches to materialize the selected `(len(tokens) - 1, num_layers, topk)` prefix.
- `SessionCoreV2` uses the same request logic and assembles each root-to-leaf record path independently; no user-facing `use_addition_r3` option is added.

## Verification
- Six focused fast suites on `7ee93a878c4ae78ece58a9f71a646a5f229309b7`: 206 passed.
- `TestAdditionR3Assembly` covers strict patch assembly plus malformed-payload rejection.
- The samples-op tests cover v1/v2 collection plus truncation.
- `TestAdditionR3RequestOffset` covers append boundaries under v1 rollback.
- Its v2 twin covers branch positioning.
- The `retract` cases confirm that the offset field is omitted.
- `TestSessionServerPauseGenerationMode` covers the session-server plus `abort` assertion.
- Pre-rebase GPU check on `d58f2f017a81b1c5458dc6b705e547454607ba21`: `CUDA_VISIBLE_DEVICES=0,1 SGLANG_SESSION_ADDITION_R3_MODEL_PATH=/root/models/Qwen3-30B-A3B PYTHONPATH=. python3 tests/e2e/sglang/test_session_server_addition_r3.py` passed on 2×H200; ten real TITO turns matched every additional patch and the final `/samples` tensor byte-for-byte against the same-trajectory full-R3 oracle.
- `python -m tests.ci.run_suite --hw cuda --suite stage-c-2-gpu-h200 --cadence regular --list-only --labels run-ci-sglang` on `7ee93a878c4ae78ece58a9f71a646a5f229309b7`: passed and selected `test_session_server_addition_r3.py`.

## Review Focus
- `_maybe_request_addition_r3` in `miles/rollout/session/core.py`: the offset must bind to the post-positioning checkpoint and remain a strict append boundary.
- `merge_samples_with_addition_r3` in `miles/rollout/session/samples/merge.py`: record coverage must be contiguous while ordinary merge/truncation decides the required final prefix.
- `test_tito_session_addition_r3_matches_full_r3`: the one-token oracle must reproduce the exact final token trajectory before any R3 bytes are compared.


Shi-Dong (approved):
LGTM

PR #2282, opened 2026-08-08
fix(inkling): preserve ordered blocks through TITO

Michaelsqj:
## Motivation

Inkling can emit repeated or interleaved thinking/text model blocks. SGLang's standard OpenAI fields flatten those blocks, so a TITO session cannot reconstruct the original role/control-token boundaries from `reasoning_content` and `content` alone. This produced persistent `special_token_count` mismatches even after the ordinary assistant-text framing bug was fixed.

Companion SGLang PR: https://github.com/sgl-project/sglang/pull/34119

## Modifications

- Retain SGLang's server-owned ordered `content_blocks` side channel in the stored assistant checkpoint.
- Preserve it across a client's flattened history replay only after the existing `message_matches` equivalence check succeeds.
- Strip the side channel from the Harbor/client response without mutating the response retained by the session server.
- Render Inkling assistant messages from ordered blocks when present, including repeated, empty, and interleaved blocks; preserve the existing flattened fallback.

The change remains backward compatible when SGLang does not return `content_blocks`; the complete runtime repair requires the companion SGLang PR.

## Validation

- 169 Miles session, trajectory, and TITO-template tests passed.
- Focused real-tokenizer tests reproduce the legacy flattened mismatch and prove exact ordered-block reconstruction.
- Black, isort, and `git diff --check` pass.
- A two-rollout Inkling/Harbor smoke reported zero overall, assistant-text, special-token-count, special-token-type, and non-assistant-text TITO mismatches across 16 accepted samples. The pre-fix parser-on baseline had assistant-text mismatches in 115/128 samples and special-token-count mismatches in 13/128.

## Runtime impact

No model, optimizer, or inference-kernel changes. The additional state is retained only for assistant responses that provide the optional ordered-block side channel.

### DIFF
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -59,11 +59,16 @@ def _samples_response(payload: bytes) -> Response:
 
 
 _CLIENT_STRIPPED_META_KEYS = ("routed_experts", "indexer_topk")
+_CLIENT_STRIPPED_MESSAGE_KEYS = ("content_blocks",)
 
 
 def _strip_replay_payloads(response: dict) -> dict:
     stripped_choices = []
     for choice in response.get("choices", []):
+        message = choice.get("message")
+        if isinstance(message, dict) and any(k in message for k in _CLIENT_STRIPPED_MESSAGE_KEYS):
+            message = {k: v for k, v in message.items() if k not in _CLIENT_STRIPPED_MESSAGE_KEYS}
+            choice = {**choice, "message": message}
         meta = choice.get("meta_info")
         if isinstance(meta, dict) and any(k in meta for k in _CLIENT_STRIPPED_META_KEYS):
             meta = {k: v for k, v in meta.items() if k not in _CLIENT_STRIPPED_META_KEYS}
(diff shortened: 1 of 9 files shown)


guapisolo (dismissed):
Good fix. LGTM.

miles/rollout/session/core.py
```diff
@@ -59,11 +59,16 @@ def _samples_response(payload: bytes) -> Response:
 
 
 _CLIENT_STRIPPED_META_KEYS = ("routed_experts", "indexer_topk")
+_CLIENT_STRIPPED_MESSAGE_KEYS = ("content_blocks",)
 
 
 def _strip_replay_payloads(response: dict) -> dict:
     stripped_choices = []
     for choice in response.get("choices", []):
+        message = choice.get("message")
```
guapisolo, line 68: good fix.

miles/rollout/session/linear_trajectory.py
```diff
@@ -17,6 +17,34 @@
 MAX_ASSISTANT_ROLLBACK_STEPS = 1
 
 
+def _preserve_ordered_content_blocks(
```
guapisolo: I suggest we add a more general func `fix_request_msg_by_stored` to describe "fixing request messages by stored messages", and it call `_preserve_ordered_content_blocks`, and put them in template.py.

guapisolo (approved):
LGTM now. We can merge after https://github.com/sgl-project/sglang/pull/34119 merge.

Michaelsqj:
Superseded by #2302, which parses Inkling directly from raw completion token IDs in the Miles session server. That covers native invocation recipients and marker-looking ordinary text without requiring a paired SGLang side channel.

PR #2294, opened 2026-08-09
session server: assemble collect_samples off the event loop

Shi-Dong:
## Problem

The session server is a single process with a single event loop (`miles/rollout/session/server.py`), and every route shares it — chat completions, `DELETE`, and `collect_samples`. But `collect_samples` reconciled, truncated, merged and encoded the entire trajectory **inline on that loop**:

```python
samples = compute_samples_from_openai_records(...)   # walks every record, every output token
samples = truncate_samples_by_total_tokens(...)
samples = [merge_samples(samples, tokenizer)]
```

There is no `run_in_executor` / `to_thread` / thread pool anywhere in `miles/rollout/session/`. On a long agentic episode this is seconds of straight-line Python, and for its whole duration the server answers nothing else.

## Why it is worse than "slow"

The stall is self-amplifying. A client-side timeout on collect marks the sample `ABORTED` (`generate_hub/agentic_tool_call.py`), and the `check_no_aborted` dynamic-sampling filter then rejects **that sample's entire group** — so the rollout relaunches a whole group of episodes onto a server that is already behind.

One blocked assembly therefore costs 8 new episodes (at `n_samples_per_prompt=8`), which produce more collects, which block the loop further.

## Evidence

Observed on an 8-node GLM-5.2 744B Terminal-Bench-2 run once EAGLE speculative decoding raised episode-completion rate ~4×. Two runs died identically during the step-2 rollout:

| run | sessions created | expected | collect timeouts | failed deletes |
|---|---|---|---|---|
| A | 623 | 232 | 60 | 59 |
| B | 639 | 232 | 86 | 66 |

A control run with speculative decoding **off** held sessions to 1.1× expected with zero timeouts and zero failed deletes.

The excess matches the amplifier: 391 extra sessions ÷ 8 per group ≈ 49 rejected groups, against 60 observed collect timeouts. In both failures generation collapsed to ~1 token/s while fresh sandboxes churned — the cluster busy *starting* work it had already lost the ability to collect.

## Change

Move the CPU phase into `_assemble_samples_payload` and `await asyncio.to_thread(...)` it. `session.records` is snapshotted so the worker cannot race concurrent appends.

Response contracts are unchanged:
- `no_records` — unchanged, still on the loop (trivial)
- `all_truncated` — now signalled by a `None` sentinel returned from the worker
- `AssertionError` / `ValueError` → 422 — still holds, since `to_thread` re-raises in the awaiting coroutine

Also corrects the module docstring, which explicitly documented the old behaviour (*"synchronously on the loop like the lock-free `get_session`"*).

## Test

`tests/fast/rollout/session/test_collect_samples_nonblocking.py` counts heartbeat ticks on a concurrent coroutine while a 0.5s assembly runs. ~50 ticks are available; it requires ≥25. The inline version yields essentially zero.

Verified as a genuine regression test rather than a decorative one:

```
fixed code  → 3 passed
inline code → 1 failed, 2 passed     # the blocking test catches the bug
```

The two semantic tests (`all_truncated`, 422) pass either way, which is correct — they don't depend on threading.

Full session suite: **45 passed**.

## Notes for reviewers

- HuggingFace fast tokenizers release the GIL in Rust, so the worker thread genuinely runs in parallel rather than just deferring the stall.
- This does not change how much CPU the assembly costs, only where it is paid. A server under sustained overload will still fall behind; this removes the pathological coupling where one slow assembly times out *unrelated* requests.
- Worth considering separately: `check_no_aborted` discarding an entire group for one infrastructure-caused abort is what converts a hiccup into an avalanche. That seemed out of scope here.

### DIFF
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -5,9 +5,10 @@ HTTP-agnostic: the FastAPI adapter (``sessions.py`` + ``server.py``) turns each 
 - ``chat_completions`` strips the R3 replay payloads (``routed_experts`` / ``indexer_topk``) from the client reply copy-on-write; the ``SessionRecord`` keeps the full response for the training path (``GET /sessions/{id}``).
 - ``chat_completions`` holds the per-session lock for prep and state update but not across the proxy call; ``closing`` re-checks and the ``num_assistant`` check gate concurrent DELETE/chat.
 - ``stream: true`` is served as fake streaming: the backend call stays non-streaming (TITO needs the complete message + meta_info) and the full response is re-rendered as a single SSE chunk plus ``data: [DONE]``. Errors all happen before the SSE body is built, so they keep their real status codes as JSON.
-- ``collect_samples`` assembles training Samples from the session's records on the server (compute -> truncate -> merge, synchronously on the loop like the lock-free ``get_session``); deterministic assembly failures return 422 with the assertion text.
+- ``collect_samples`` assembles training Samples from the session's records on the server (compute -> truncate -> merge -> encode) in a worker thread via ``asyncio.to_thread``, because that phase is seconds of CPU on a long agentic trajectory and this process serves every session route from a single event loop; deterministic assembly failures return 422 with the assertion text.
 """
 
+import asyncio
 import json
 import logging
 import time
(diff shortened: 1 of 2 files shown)


Shi-Dong:
Closing: v1 is being deprecated, so fixing it is not worth the review cycle.

The same inline-assembly pattern exists in the v2 core (`miles/rollout/session/v2/core.py`, `collect_samples`), where the docstring states it explicitly:

> Synchronous on the server loop (no await), so the session read cannot interleave with chat commits.

There is no `to_thread` / `run_in_executor` anywhere under `miles/rollout/session/v2/`. I am investigating whether that synchronous read is a real invariant for v2's tree model — if it can be preserved (snapshot or lock), I will open a separate PR against v2 with the same evidence.

Original diagnosis, which applies to either core: assembling a long agentic trajectory is seconds of CPU on a single shared event loop; a client-side timeout on collect marks the sample ABORTED, and `check_no_aborted` then rejects the whole group, relaunching 8 episodes onto the already-behind server. Measured on 8-node GLM-5.2 TB2 runs: sessions created reached 2.7-2.8x expected with 60-86 collect timeouts, versus 1.1x and zero on a control without speculative decoding.

PR #2302, opened 2026-08-10
fix(inkling): parse raw completions in the session server

Michaelsqj:
## Motivation

Miles already treats generated token IDs as the source of truth for TITO, but
Inkling's agent-facing message was previously parsed in SGLang and then rendered
again by the Miles session server. That creates two independent parsers around a
model protocol whose role boundaries, reasoning blocks, and invocation headers
are encoded by special token IDs. Flattening the response into OpenAI fields is
not sufficient to reconstruct the original token stream.

The problematic pipeline was:

```text
raw Inkling token IDs -> SGLang parser -> flattened OpenAI fields
                      -> Miles chat-template replay -> guessed token structure
```

This PR makes the session-server boundary authoritative:

```text
raw Inkling token IDs -> model-specific Miles parser -> OpenAI projection for agent
                                                 \-> exact raw-ID replay for TITO
```

Concrete failures observed in real rollouts:

1. **Invocation recipient differs from callable name.** Inkling emitted:

   ```text
   <|message_model|>bash<|content_invoke_tool_json|>
   {"name":"bash_command","args":{"command":"pwd"}}<|end_message|>
   ```

   The OpenAI tool call correctly names `bash_command`, but the raw invocation
   recipient is `bash`. A flattened representation loses that distinction, so
   canonical replay changes assistant text.

2. **Marker-looking payload text is not a control token.** A reasoning span
   spelled the literal text `<|endoftext|>` using ordinary BPE IDs. A
   decoded-string parser interpreted it as control syntax and removed it. Parsing
   token IDs preserves the text because only the tokenizer's actual special ID is
   treated as framing.

3. **Repeated and interleaved blocks lose boundaries when flattened.** Concatenated
   `reasoning_content` and `content` preserve visible text but not the original
   sequence of model-role boundaries. Exact completion-ID checkpoints preserve
   every boundary without requiring SGLang and Miles to agree on a side-channel
   schema.

These cases produced nonzero `assistant_text` or `special_token_count` TITO
mismatch even though the sampled token IDs themselves were valid training data.

## Changes

- Add default-no-op model-family response hooks to `TITOTokenizer`; models that do
  not opt in retain the current SGLang response and replay behavior.
- Call the hook once in `SessionCore` after a successful non-streaming completion.
- Add a stateless Inkling parser that scans completion token IDs once and projects
  ordered thinking, text, and native JSON invocation blocks into standard OpenAI
  `reasoning_content`, `content`, and `tool_calls` fields.
- Return parser identity/errors in response metadata so the agent harness can
  decide whether to retry; Miles does not embed retry prompts or tool-execution
  policy.
- Preserve ordered blocks and exact raw completion IDs in the server-owned
  assistant checkpoint, even when the next client request replays only flattened
  OpenAI messages.
- During tokenized Inkling replay, splice the original completion IDs at generated
  assistant checkpoints and use the existing incremental merge path for later
  user/tool messages.
- Remove Inkling's SGLang reasoning/tool parser bindings. SGLang can serve raw
  completion text and token IDs; no paired SGLang parser patch is required.

Malformed JSON, a true special token inside payload, or a length-cut/incomplete
block produces a parse error and never exposes a structured tool call for
execution. Retry/discard behavior remains the responsibility of the agent
harness.

## Why this is preferable to the ordered-block side channel

This supersedes #2282 and its companion SGLang PR
https://github.com/sgl-project/sglang/pull/34119. The earlier design preserved
ordered thinking/text blocks but could not represent native invocation headers
losslessly and still depended on SGLang's decoded-string parser. Keeping the raw
completion IDs at the existing TITO boundary:

- supports all observed Inkling block kinds, including native tool calls;
- distinguishes real control IDs from marker-looking ordinary text;
- removes cross-repository parser/schema coupling;
- preserves exact replay across multiple session-server processes; and
- leaves every other model family unchanged.

## Validation

Rebased cleanly onto upstream `main` at
`d2e3ad9f197d86c138347063fdd9750dbd3d296e`.

- 104 tokenizer and Inkling parser tests passed.
- 42 rollout session sample/codec tests passed.
- 81 session concurrency, state, and request tests passed.
- Black, isort, and `git diff --check` pass.
- Focused cases cover ordered/interleaved blocks, native invocation recipient,
  marker-looking payload text, real special tokens, malformed JSON, incomplete
  length-cut blocks, and independent parser instances.
- A prior CAMEL native-tool smoke using SGLang parsing had 2/16 accepted samples
  with `assistant_text` mismatch, exactly matching the recipient and literal-marker
  examples above.
- The session-parser smoke then completed 11 optimizer steps with every audited
  TITO mismatch category at zero. The first two saved rollouts of the subsequent
  full run also contain 0/256 accepted samples with stored TITO mismatch.

## Runtime impact

No model, optimizer, kernel, routing, or agent-harness changes. Inkling completion
parsing is linear in completion length, parser state is local to each tokenizer
instance, and raw completion IDs are retained only in session-server state used
for exact TITO replay.


guapisolo:
I think this path is not safe actually. 

```
raw Inkling token IDs -> model-specific Miles parser -> OpenAI projection for agent
                                                 \-> exact raw-ID replay for TITO
```

Though it can resolve the mismatch issue, but just a workaround.
Inkling don't have boundary issue, but some models have, like qwen and glm, according to https://www.lmsys.org/blog/2026-05-13-no-token-left-behind. 

So you shouldn't hardcode apply chat template behavior as merge token

guapisolo:
I think the motivation of you close https://github.com/sgl-project/sglang/pull/34119 and open this PR is to move the hacky parser to miles instead of sglang. This is correct.
But I am confused that why you close #2282 ? 
- In that PR, you actually modify chat template correctly for multi content block.
- This PR will make TITO data flow very strange.

guapisolo:
I think you can keep decouple the parser behavior in this PR:
```
input: tokens returned from sglang
output: parsed message, including content blocks
```

and when a new message coming in, you still:
- apply chat template to additional tokens
- call merge tokens naively to merge prefix and new tokens.

Then add 2282's chat template change to this PR. (no need to reopen 2282)

One thing to double confirm, does 2282 chat template correct ?

guapisolo:
Modified. Need code author prod run, thanks

guapisolo:
/rerun-test tests/e2e/megatron/model_scripts/test_inkling_small_4layer_ci.py

guapisolo (approved):

guapisolo:
/rerun-test tests/e2e/megatron/model_scripts/test_inkling_small_4layer_ci.py

guapisolo:
/rerun-test tests/e2e/sglang/test_session_server_multi_role/test_inkling.py

PR #2304, opened 2026-08-10
Render the sglang server and router argv through one renderer

fzyzcjy:
Part of #1837


guapisolo (approved):
Looks good in stack context. The raw CLI input / normalized object split preserves the SGLang argv roundtrip, and this change is present at safe merge checkpoint #2176.

PR #2305, opened 2026-08-10
Let the session servers sit on more than one host

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -48,23 +48,18 @@ async def wait_session_server_ready(args):
         await provider.get_addr(worker_name=compute_worker_name(pool_id="session-server", cell_index=i))
         for i in range(args.session_server_workers)
     ]
-    ips = [x.host for x in addrs]
-    assert len(set(ips)) == 1, f"{ips=}"
-    ip = ips[0]
-    ports = [x.port for x in addrs]
-
-    # The canonical driver-side value; rollout code picks from this list.
-    args.session_server_ip = ip
-    args.session_server_ports = ports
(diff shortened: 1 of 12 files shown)


guapisolo:
[P2] One multi-host compatibility boundary needs confirmation: #2305 allows session servers such as `10.0.0.1:5005` and `10.0.0.2:5005`, but the off-cluster NeMo Gym, OpenEnv, and SWE-Agent adapters replace both hosts with the single `MILES_ROUTER_EXTERNAL_HOST`, so both destinations become `external-host:5005` and the external agent can lose session affinity. @nblintao @Shi-Dong, should multi-host session servers be supported with these off-cluster adapters? If yes, we need a per-instance external address mapping; otherwise this combination should fail fast or be documented as unsupported.

nblintao:
> [P2] One multi-host compatibility boundary needs confirmation: #2305 allows session servers such as `10.0.0.1:5005` and `10.0.0.2:5005`, but the off-cluster NeMo Gym, OpenEnv, and SWE-Agent adapters replace both hosts with the single `MILES_ROUTER_EXTERNAL_HOST`, so both destinations become `external-host:5005` and the external agent can lose session affinity. @nblintao @Shi-Dong, should multi-host session servers be supported with these off-cluster adapters? If yes, we need a per-instance external address mapping; otherwise this combination should fail fast or be documented as unsupported.

Yes. That will be where most session server calls happen, so we should definitely support it.

Instead of "a per-instance external address mapping", I think we should get rid of MILES_ROUTER_EXTERNAL_HOST and resolve_session_url entirely.

Here is what I imagine it will look like: https://github.com/radixark/miles/pull/2851

cc @fzyzcjy @guapisolo @Shi-Dong

PR #2306, opened 2026-08-10
Track the context lock grant by generation instead of identity

fzyzcjy:
Part of #1837


guapisolo (approved):
Generation-scoped grants now expire child-task authorization at release/detach while preserving in-section fan-out and cross-task reattach. Verified in PR #2306 before safe merge checkpoint #2176.

PR #2307, opened 2026-08-10
Open the api server and the healing loop with --use-fault-tolerance

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -99,6 +99,7 @@ def reset_arg(parser, name, **kwargs):
 
 
 _FT_CHOICES = ["rollout", "train"]
+_DEFAULT_FT_API_SERVER_PORT = 18080
 
 
 def get_miles_extra_args_provider(add_custom_arguments=None):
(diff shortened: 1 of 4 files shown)


miles/utils/arguments.py
```diff
@@ -2858,15 +2863,24 @@ def _validate_rematerialize_param_from_master_weight(args):
...
+def _resolve_api_server_port(args: argparse.Namespace) -> int:
+    if (port := args.api_server_port) is not None:
+        return port
+    return _DEFAULT_FT_API_SERVER_PORT if args.ft_components else 0
+
+
+def _resolve_mini_ft_controller_enable(args: argparse.Namespace) -> bool:
+    if (enable := args.mini_ft_controller_enable) is not None:
+        return enable
+    return bool(args.ft_components) and args.api_server_port != 0
```
fzyzcjy, line 2905: [P1] Gate implicit healing on the post-config FT flag

When a launcher passes `--use-fault-tolerance` but its `--custom-config-path` sets `use_fault_tolerance: false`, `_resolve_ft_components` has already materialized `['rollout']` before YAML is loaded. These helpers therefore still select port 18080 and enable the mini controller, which can suspend or resume cells despite the documented rule that [YAML overrides CLI](docs/diffusion/user-guide/cli-reference.md#L14-L16). Gate the implicit defaults on the post-config `args.use_fault_tolerance` value; the same ordering remains in deliver-1 and deliver-2, so this blocks deliver-1.

fzyzcjy (commented):
The new implicit server and controller defaults can ignore a supported YAML override that disables fault tolerance. This defect remains present at both deliver-1 and deliver-2.

PR #2308, opened 2026-08-10
Discard stale health probe results instead of cancelling them

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this commit remains at deliver-1. The epoch check correctly discards probes spanning either a pause or a complete pause/resume window, and no additional relevant fix appears in deliver-2.

PR #2309, opened 2026-08-10
Address FT test action targets by explicit cell id

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -2272,8 +2272,8 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
                 default=None,
                 help="JSON array of fault injection actions. Each action: "
                 '{"at_rollout": N, "action": "stop_cell_at_end"|"start_cell_at_end"|"crash_before_allreduce", '
-                '"cell_index": I, "rank": 0, "attempt": 0}. '
-                "cell_index -1 means last cell.",
+                '"cell_id": "trainer-actor-2", "rank": 0, "attempt": 0}. '
+                "cell_id is the full cell id (spec name plus cell index) of the target cell.",
             )
             parser.add_argument(
                 "--ci-inject-rollout-data-path",
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -2237,8 +2237,8 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
                 default=None,
                 help="JSON array of fault injection actions. Each action: "
                 '{"at_rollout": N, "action": "stop_cell_at_end"|"start_cell_at_end"|"crash_before_allreduce", '
-                '"cell_index": I, "rank": 0, "attempt": 0}. '
-                "cell_index -1 means last cell.",
+                '"cell_id": "trainer-actor-2", "rank": 0, "attempt": 0}. '
+                "cell_id is the full cell id (spec name plus cell index) of the target cell.",
             )
             parser.add_argument(
                 "--ci-inject-rollout-data-path",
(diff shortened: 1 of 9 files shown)


miles/utils/test_utils/ft_test_actions.py
```diff
@@ -55,37 +58,43 @@ def from_args(args: object, *, controller: "RayTrainGroup") -> "FTTestActionCont
...
                 elif action.action == "start_cell_at_end":
-                    await worker_manager.start_cells.remote([cell_id])
+                    await worker_manager.start_cells.remote([action.cell_id])
+
+    def _check_action_target(self, action: FTTestAction) -> None:
+        parsed = parse_cell_id(action.cell_id)
+        assert parsed.pool_id == self._controller.pool_id, (
+            f"FT test action targets pool_id {parsed.pool_id!r} but this controller drives {self._controller.pool_id!r} "
+            f"(action={action})"
+        )
```
fzyzcjy, line 75: [P1] Route controller actions only to their owning pool

With `--use-critic`, `compute_critic_args` copies the same actions into both trainer groups, and the critic trains first. A valid `trainer-actor-*` stop/start action therefore reaches the critic executor and this assertion aborts the run instead of allowing only the actor controller to dispatch it; targeting the critic fails symmetrically. This remains at deliver-1 and blocks it, so actions must be routed or filtered by owning pool while target existence is validated globally.

miles/utils/test_utils/ft_test_actions.py
```diff
@@ -37,6 +33,13 @@ def _load_actions(args: object, action_filter: set[str]) -> list[FTTestAction]:
     if not raw:
         return []
     all_actions = _ACTION_LIST_ADAPTER.validate_json(raw)
+
+    for action in all_actions:
+        try:
+            parse_cell_id(action.cell_id)
+        except ValueError as e:
+            raise ValueError(f"FT test action has malformed cell_id {action.cell_id!r} (action={action})") from e
```
fzyzcjy, line 41: [P2] Reject actor targets that no configured cell owns

For a standalone `crash_before_allreduce`, a syntactically valid typo such as `trainer-actro-1` or noncanonical `trainer-actor-01` passes this parse-only validation. Controller filtering removes actor actions before target checks, and every actor then misses the exact string comparison, so the requested crash is silently skipped; the former index scheme constructed the canonical pool ID itself. This remains at deliver-1 and blocks it, so validate actor-action IDs against the configured cells before execution.

fzyzcjy (commented):
Explicit controller targets fail in supported actor-critic runs, while malformed-but-parseable actor targets can silently suppress the requested fault. The documented interface is also left on the rejected schema, and these issues remain at deliver-1.

miles/utils/arguments.py
```diff
@@ -2244,8 +2244,8 @@ def add_debug_arguments(parser):
                 default=None,
                 help="JSON array of fault injection actions. Each action: "
                 '{"at_rollout": N, "action": "stop_cell_at_end"|"start_cell_at_end"|"crash_before_allreduce", '
-                '"cell_index": I, "rank": 0, "attempt": 0}. '
-                "cell_index -1 means last cell.",
+                '"cell_id": "trainer-actor-2", "rank": 0, "attempt": 0}. '
+                "cell_id is the full cell id (spec name plus cell index) of the target cell.",
```
fzyzcjy, line 2276: [P2] Migrate the developer fault-action example

The flag now rejects `cell_index` and requires `cell_id`, but `docs/developer/debug.md:104-108` still publishes the old `cell_index: -1` command, which now fails during action loading. This persists at deliver-1 and blocks the interface migration. Deliver-2 later rewrites the paragraph; the minimal forward move here is only the `cell_id` example and explanation using the current `trainer-actor-*` naming, since its `trainer-engine` naming and `sleep_forever_at_end` text depend on later ops. See `.claude/rules/general-code-style.md:41-43`.

PR #2311, opened 2026-08-10
Stop the fast suite from starting its own ray clusters and leaking accelerators

fzyzcjy:
Part of #1837

### DIFF
--- a/tests/ci/run_suite.py
+++ b/tests/ci/run_suite.py
@@ -17,6 +17,7 @@ from tests.ci.ci_utils import (
     CI_GATE_RECORD_DIR_ENV,
     build_store_from_env,
     gate_provenance_from_env,
+    reaping_is_isolated,
     run_unittest_files,
 )
 from tests.ci.labels import KNOWN_LABELS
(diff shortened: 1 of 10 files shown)


tests/ci/run_suite.py
```diff
@@ -257,6 +257,7 @@ def run_a_suite(args):
         gate_store=gate_store,
         gate_write_baseline=policy.write_baseline,
         gate_provenance=gate_provenance,
+        reap_leftovers=True,
```
fzyzcjy: [P1] Gate process-wide reaping to isolated CI runs

When this CLI is used through its supported local path (`--match-all-labels` is explicitly described as a local override at `tests/ci/run_suite.py:357-360`), this unconditional opt-in runs `ray stop --force` and process-wide `pkill -9 -f` commands before every file. The safety assumption only holds in a per-job PID namespace (`tests/ci/ci_utils.py:676-680`); on a host or with `MILES_SCRIPT_EXTERNAL_RAY=1`, it terminates unrelated or required Ray/SGLang processes, potentially destroying another workload and breaking the suite. This remains unchanged in target deliver `d3507a9`.

fzyzcjy (commented):
The patch introduces destructive process-wide cleanup for supported local and externally managed Ray-cluster runs. The issue persists at the target safe deliver.

PR #2313, opened 2026-08-10
Let worker handles forward positional arguments

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the target safe deliver.

PR #2314, opened 2026-08-10
Implement RpcWorkerHandle.wait_dead

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/rpc/client/handle.py
+++ b/miles/utils/workers/rpc/client/handle.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Awaitable, Callable
 from typing import Any
 
@@ -10,8 +11,10 @@ from miles.utils.workers.rpc.client.call import RpcCall
 from miles.utils.workers.rpc.client.misc import (
     RETRY_INITIAL_DELAY_SECONDS,
     RETRYABLE_ERRORS,
+    WORKER_IS_GONE_ERRORS,
     BootUuidPin,
     RpcTransport,
+    ServerRestartedError,
 )
 from miles.utils.workers.rpc.common.metadata import (
     RpcMethodSpec,
@@ -25,6 +28,8 @@ DEFAULT_CALL_TIMEOUT_SECONDS = 3600.0
 DEFAULT_READY_TIMEOUT_SECONDS = 600.0
 
 _HEALTH_TIMEOUT_SECONDS = 5.0
+
+logger = logging.getLogger(__name__)
 
 
 class RpcWorkerHandle(BaseWorkerHandle):
@@ -50,6 +55,9 @@ class RpcWorkerHandle(BaseWorkerHandle):
         self._transport = RpcTransport(
             server_url=server_url, http_client=http_client, boot_uuid_pin=self._boot_uuid_pin
         )
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self._worker_cls_name})"
 
     def __getattr__(self, name: str) -> Callable[..., Awaitable[Any]]:
         if name.startswith("_"):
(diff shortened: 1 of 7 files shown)


miles/utils/workers/rpc/client/handle.py
```diff
@@ -84,8 +92,16 @@ async def attempt(remaining: float) -> None:
...
 
-    async def wait_dead(self, *, timeout: float) -> None:
-        raise NotImplementedError("RpcWorkerHandle cannot confirm worker death yet")
+    async def _probe_is_dead(self) -> bool:
+        try:
+            await self._transport.request(
+                "GET", HEALTH_PATH, seconds=_HEALTH_TIMEOUT_SECONDS, response_model=HealthResponse
+            )
+        except (ServerRestartedError, *WORKER_IS_GONE_ERRORS):
+            return True
```
fzyzcjy, line 101: [P1] Ensure restart errors do not skip death confirmation

For the stable-pinned RPC handles used in deliver-2, a worker that has already restarted makes the preceding `handle.kill_self()` call return HTTP 412 and raise `ServerRestartedError`. `TrainerCell._kill_workers_and_confirm_dead` awaits all kill calls before invoking `wait_dead`, while `_kill_worker` suppresses only `WorkerUnreachableError` and timeouts, so that first gather raises and this restart-aware branch is never reached for any rank. Recovery can therefore skip death confirmation while other ranks remain alive; `f5a18edf` retains this sequence. Treat the kill-time restart as already gone, or otherwise guarantee the probe phase runs.

fzyzcjy (commented):
The restart-aware probe is bypassed by the production kill-before-probe sequence when that same restart raises `ServerRestartedError`. This persists at deliver-2 and can leave multi-rank teardown unconfirmed.

guapisolo:
codex cmt

Frozen review: [#2314 @ adab57999540](https://github.com/radixark/miles/commit/adab57999540b2a0ce4902adcec24e07ba3a5c0a) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

### [P2] Require evidence of death before accepting a connection error

Source: [miles/utils/workers/rpc/client/handle.py:100](https://github.com/radixark/miles/blob/adab57999540b2a0ce4902adcec24e07ba3a5c0a/miles/utils/workers/rpc/client/handle.py#L100).

The new RPC death probe treats every `httpx.ConnectError` as proof that the worker is gone. A live worker can produce the same exception when DNS resolution fails or the network is unreachable; `RpcTransport` does not restrict it to a refused connection. Consequently, `wait_dead()` can report success without confirming the old worker has exited or been replaced, skipping further confirmation and the deadline diagnostic. This is a defect in the newly implemented RPC API contract: the reviewed source head still uses the Ray trainer provider, so this finding does not establish a failure in the active training path.

**Status at #3122: not fixed.** The recorded trace found no repair through [#3122](https://github.com/radixark/miles/pull/3122) ([198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69)). Connection failures that do not establish death must remain retryable; return success only with reliable evidence that the worker is dead or that the old boot identity has been replaced. No real network fault-injection or RPC training run was performed for this finding.

- [P2] Five test handles still implemented the old `wait_dead` contract, so the new abstract probe requirement prevented instantiation. [Source](https://github.com/radixark/miles/blob/adab57999540b2a0ce4902adcec24e07ba3a5c0a/miles/utils/workers/worker_handle.py#L31). Fixed in [#2622](https://github.com/radixark/miles/pull/2622) ([704c8e3392](https://github.com/radixark/miles/commit/704c8e3392c5f090de7bdc0bcd7f3e7641537241)): Removed two obsolete test handles and implemented the probe method on the three remaining dashboard handles.

- [P2] The persistent-unavailability test patched the clock in its former module, raising `AttributeError` before checking the deadline. [Source](https://github.com/radixark/miles/blob/adab57999540b2a0ce4902adcec24e07ba3a5c0a/miles/utils/workers/worker_handle.py#L22). Fixed in [#2979](https://github.com/radixark/miles/pull/2979) ([fcbfeea0be](https://github.com/radixark/miles/commit/fcbfeea0bee16d67b8eebad2599de548b4fdd775)): Moved the clock patches and logger lookup to `worker_handle_module`, which owns the shared wait loop.

PR #2315, opened 2026-08-10
Annotate the controllers' and rollout executor's public methods

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The added annotations match the existing method contracts and introduce no behavioral regression. Potential RPC-surface gaps are resolved by the target safe deliver and are therefore transient.

PR #2319, opened 2026-08-10
Make the inference controller an independent worker

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/placement_group.py
+++ b/miles/ray/placement_group.py
@@ -7,10 +7,12 @@ from ray.util.placement_group import PlacementGroup, placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
 from miles.ray.rollout.router_manager import resolve_router_addrs, wait_session_server_ready
+from miles.ray.specs.inference import create_inference_controller_handle
 from miles.ray.specs.train import compute_critic_args
 from miles.ray.train.group import TrainerController
+from miles.utils.workers.worker_handle import BaseWorkerHandle
+
 from ..utils.ray_utils import compute_ray_pin_head_options
-from .rollout.inference_controller import InferenceController
 from .rollout.rollout_executor import RolloutExecutor
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


miles/ray/placement_group.py
```diff
@@ -174,7 +176,7 @@ async def create_rollout_components(args) -> RolloutComponents:
         await resolve_router_addrs(args)
         await wait_session_server_ready(args)
 
-    inference_controller = InferenceController(args)
+    inference_controller = create_inference_controller_handle()
```
fzyzcjy, line 179: [P1] Make the detached update lock recoverable

When hot restart replaces orchestration after remote `start_update_weights()` detaches the controller lock but before `end_update_weights()` is sent, the surviving actor remains locked. Deliver-2 still has this sequence: `wait_idle()` sees no in-flight controller RPC, then `init_or_reset_inference_controller()` blocks in `abort_all()` until the takeover timeout. Collapse the update window into one RPC or add a takeover-safe way to abort and release it.

fzyzcjy (commented):
The independent controller can retain a detached update lock after its orchestration process is replaced. This remains unresolved at deliver-2 and can prevent hot-restart takeover.

guapisolo:
codex cmt

Frozen review: [#2319 @ ad5cbfe9c16f](https://github.com/radixark/miles/commit/ad5cbfe9c16f5d318ff59b9954a0cee51b446a83) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

- [P1] Reading `eval_fleet` as a property on the remote controller handle supplied a function, causing snapshot evaluation to fail on `.pin()`. [Source](https://github.com/radixark/miles/blob/ad5cbfe9c16f5d318ff59b9954a0cee51b446a83/miles/ray/placement_group.py#L179). Fixed in [#2614](https://github.com/radixark/miles/pull/2614) ([03d3e50ba2](https://github.com/radixark/miles/commit/03d3e50ba2a7a819c67232bcd0527b54166a0bc5)): Transferred fleet metadata through RPC and introduced an executor-owned fleet that pins through the controller RPC, preserving `None` when no fleet exists.

PR #2320, opened 2026-08-10
Ship the actor-critic values handoff through the object store

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -17,7 +17,7 @@ from miles.backends.megatron_utils.rematerialize_utils import build_main_cast_co
 from miles.dashboard import hooks as dashboard_hooks
 from miles.ray.specs.train import compute_trainer_pool_id
 from miles.ray.train_actor import TrainRayActor
-from miles.utils import async_utils, train_dump_utils
+from miles.utils import async_utils, object_store, train_dump_utils
 from miles.utils.argparse_utils import inplace_modify_args
 from miles.utils.audit_utils.event_logger.logger import event_logger_context
 from miles.utils.audit_utils.witness.allocator import WitnessInfo
@@ -27,6 +27,7 @@ from miles.utils.ft_utils.indep_dp import IndepDPInfo
 from miles.utils.hf_config import load_hf_config
 from miles.utils.memory_utils import clear_memory, print_memory
 from miles.utils.multi_lora import is_multi_lora_enabled
+from miles.utils.object_store import ValueSpec
 from miles.utils.processing_utils import load_tokenizer
 from miles.utils.ray_utils import Box
 from miles.utils.reloadable_process_group import destroy_process_groups, monkey_patch_torch_dist, reload_process_groups
@@ -71,6 +72,8 @@ if TYPE_CHECKING:
 logging.getLogger("megatron").setLevel(logging.WARNING)
 
 logger = logging.getLogger(__name__)
+
+CRITIC_VALUES_VALUE_SPEC: dict[str, ValueSpec] = {"values": ValueSpec(codec="auto")}
 
 
 def _setup_disk_offload_reclaim(disk_dir: str) -> None:
@@ -457,7 +460,10 @@ class MegatronTrainRayActor(TrainRayActor):
         values = None
         if get_parallel_state().is_pp_last_stage and "values" in rollout_data:
             # Ship by object reference
-            values = Box(ray.put([value.detach().cpu() for value in rollout_data["values"]]))
+            values = object_store.get_instance().put(
+                value={"values": [value.detach().cpu() for value in rollout_data["values"]]},
+                value_spec=CRITIC_VALUES_VALUE_SPEC,
+            )
         return TrainStepOutput(outcome=train_step_outcome, values=values)
 
     def _use_rollout_replay(self, m) -> bool:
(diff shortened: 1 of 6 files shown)


miles/backends/megatron_utils/actor.py
```diff
@@ -72,6 +73,8 @@
 
 logger = logging.getLogger(__name__)
 
+CRITIC_VALUES_VALUE_SPEC: dict[str, ValueSpec] = {"values": ValueSpec(codec="auto")}
```
fzyzcjy: [P1] Store critic tensors outside Mooncake metadata

When shared PPO (`--advantage-estimator ppo`) runs with `--object-store-backend mooncake`, `_field_schemas_for_value` maps this `auto` codec to `meta_info`, while `train_critic` passes a list of `torch.Tensor`s. Mooncake serializes `meta_info` into its JSON-safe exported reference, so `export_ref` raises on the first tensor and the first critic step never reaches the actor; declare `values` as a non-metadata ragged-tensor field instead. The same spec remains in deliver-2 (`f5a18edf2a`), so this is a blocker.

fzyzcjy (commented):
The newly added handoff cannot serialize critic values through the Mooncake backend it is intended to support, and this defect remains in deliver-2.

guapisolo:
codex cmt

Worker-platform review findings for #2320, source increment [`ad5cbfe9c16f5d318ff59b9954a0cee51b446a83`](https://github.com/radixark/miles/commit/ad5cbfe9c16f5d318ff59b9954a0cee51b446a83) → [`f2bb10f17abdb328da66e32da98b1497f7a16a59`](https://github.com/radixark/miles/commit/f2bb10f17abdb328da66e32da98b1497f7a16a59). Repair status is measured through [#3122](https://github.com/radixark/miles/pull/3122) at [`198e14c872e4c5f6dd8e7dd04571f743cb18fb69`](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

These source findings are retained separately from later stack repairs. The source locations and repair history were rechecked after the stack was rebased.

### [P1] Restore tensors after decoding critic values with typed_ragged

Source: [miles/backends/megatron_utils/actor.py:616](https://github.com/radixark/miles/blob/f2bb10f17abdb328da66e32da98b1497f7a16a59/miles/backends/megatron_utils/actor.py#L616).

With `--use-critic --object-store-backend mooncake`, the first actor step that is not critic-only reads critic values through the newly selected `typed_ragged` codec. The decoder in the Mooncake revision selected by the CUDA 13 image converts each row with `.tolist()`, but the actor still calls `.to()` on each decoded value. Those values are Python lists, so the actor raises `AttributeError: 'list' object has no attribute 'to'` before computing advantages or completing training. The Ray backend retains tensors, so Ray or fake-store tests do not exercise this regression.

**Status at #3122: not fixed.** The recorded trace found no tensor-restoration repair through [#3122](https://github.com/radixark/miles/pull/3122) ([198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69)). Restore each decoded value with `torch.as_tensor(value, dtype=torch.float32, device=torch.cuda.current_device())` before computing advantages; the critic's `forward_only` output uses FP32 by default. Verification must cover the real `typed_ragged` critic-to-store-to-actor path, which was not run in this review.

### [P2] Release critic references on exceptional consumer exits

Source: [train.py:131](https://github.com/radixark/miles/blob/f2bb10f17abdb328da66e32da98b1497f7a16a59/train.py#L131).

The new Mooncake critic-value objects require explicit `cleanup_dataproto`, but both drivers call `remove_train_output_refs` only after the actor train/offload steps succeed. An exception in either step bypasses removal. At this source PR, the controller also loses successful workers' references when it rejects or retries the enclosing attempt, before the driver ever receives those references. Both paths retain producer-owned persistent objects; the base Ray-only `Box` did not introduce this explicit persistent-store cleanup obligation.

**Status at #3122: not fixed.** [#2321](https://github.com/radixark/miles/pull/2321) ([8f555dadf9cb](https://github.com/radixark/miles/commit/8f555dadf9cb0e04a820bad2479e7de91e3cbfd2)) partially repaired the finding by collecting successful worker outputs and releasing them when `_check_train_one_attempt` raises. [#2324](https://github.com/radixark/miles/pull/2324) ([7b3018a56bfb](https://github.com/radixark/miles/commit/7b3018a56bfbac09f3fc4ce37a1c7efbfe6499cd)) regressed that controller cleanup when it moved the controller into a process without an initialized object-store singleton; [#2619](https://github.com/radixark/miles/pull/2619) ([cdbe5677568b](https://github.com/radixark/miles/commit/cdbe5677568b8bbe8e488fe3d91203254a8a86f1)) restored it with controller-local initialization. The recorded [#3122](https://github.com/radixark/miles/pull/3122) ([198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69)) state therefore still requires `try/finally` cleanup around critic-reference ownership and subsequent consumption in both drivers, while the discarded-controller-attempt cleanup is already present. Neither driver's exceptional-exit gap was repaired, so there is no complete fix for this finding.

Verification: the earlier frozen source review was carried forward by comparing the rebased source and relevant repair paths, with static inspection of the intersecting changes. The cited source-file blobs are unchanged. Tests were not rerun against these rebased heads; this publication does not claim a new runtime or full E2E pass.

Later repairs describe the stated stack checkpoint and do not remove the findings from this source increment in isolation.

PR #2322, opened 2026-08-10
Give the rpc layer a codec for an argparse Namespace

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The lossy Namespace JSON codec is transient: c944aba845 replaces it with the marked Pickled codec before deliver-2. No blocker introduced by this op remains at f5a18edf2a.

PR #2326, opened 2026-08-10
Make the multi-LoRA async data source async

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The coroutine conversion is propagated to all affected production call sites, and no correctness or regression blocker persists at deliver-2.

PR #2327, opened 2026-08-10
Make the multi-lora controller an independent worker

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the target deliver. Controller initialization, discovery, and production call sites consistently use the worker-handle API.

PR #2329, opened 2026-08-10
Give a worker's constructor the backend it runs on

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Within the requested production correctness and regression scope, no blocker introduced by this op remains at the deliver-2 safe point.

PR #2333, opened 2026-08-10
Mint the independent-DP quorum store in the trainer controller

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/fsdp_utils/actor.py
+++ b/miles/backends/fsdp_utils/actor.py
@@ -69,12 +69,14 @@ class FSDPTrainRayActor(TrainRayActor):
         with_opd_teacher: bool = False,
         recv_ckpt_src_rank: int | None = None,
         indep_dp_info: IndepDPInfo,
+        indep_dp_store_addr: str | None,
     ) -> int | None:  # type: ignore[override]
         super().init(args, role, with_ref, with_opd_teacher=with_opd_teacher)
 
         # Unsupported
         assert recv_ckpt_src_rank is None
         assert indep_dp_info.quorum_id == 0
+        assert indep_dp_store_addr is None
 
         if args.dumper_enable:
             from sglang.srt.debug_utils.dumper import dumper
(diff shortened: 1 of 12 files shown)


fzyzcjy (commented):
No production correctness or regression blocker introduced by this op remains at deliver-2. Store ownership and address propagation are consistent across initialization, healing, and quorum reconfiguration.

guapisolo:
codex cmt

Frozen review: [#2333 @ 7a9f7559061b](https://github.com/radixark/miles/commit/7a9f7559061bdd613e5c5f22d95bf2919d469a14) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

- [P2] Two existing tests retained the old store argument and deleted helper patch, causing `TypeError` and `AttributeError` before their assertions. [Source](https://github.com/radixark/miles/blob/7a9f7559061bdd613e5c5f22d95bf2919d469a14/miles/backends/megatron_utils/actor.py#L949). Fixed in [#3077](https://github.com/radixark/miles/pull/3077) ([c7992d55f2](https://github.com/radixark/miles/commit/c7992d55f244228b2698c5a92d1572dbdb52bc23)): Passed the store address explicitly in the lifecycle test and removed the critic-spec patch of the deleted helper.

- [P2] The new cleanup assertion compared identity-only `Box` instances across Ray serialization, failing even when the correct reference contents were released. [Source](https://github.com/radixark/miles/blob/7a9f7559061bdd613e5c5f22d95bf2919d469a14/tests/fast/ray/train/test_group.py#L999). Fixed in [#3067](https://github.com/radixark/miles/pull/3067) ([d46bdaa842](https://github.com/radixark/miles/commit/d46bdaa84231c69ed32aa02eef6ca4df40b59762)): Restored comparisons of `ref.inner` contents while retaining the Ray actor round trip.

PR #2334, opened 2026-08-10
Derive a cell's pod packing from the gpus a node has

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/specs/inference.py
+++ b/miles/ray/specs/inference.py
@@ -236,6 +236,7 @@ def _compute_spec_inference_engine(
         # TODO: may need real num for k8s native mode
         num_gpus_per_worker=0.2,
         num_gpu_slots_per_worker=min(server_group_config.num_gpus_per_engine, args.num_gpus_per_node),
+        num_gpus_per_node=args.num_gpus_per_node,
         pg_name="rollout",
         pg_slot_offset=server_group_config.gpu_offset,
     )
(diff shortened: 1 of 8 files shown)


miles/ray/specs/inference.py
```diff
@@ -236,6 +236,7 @@ def _compute_launch_command(ctx: LaunchCommandContext) -> str:
         # TODO: may need real num for k8s native mode
         num_gpus_per_worker=0.2,
         num_gpu_slots_per_worker=min(server_group_config.num_gpus_per_engine, args.num_gpus_per_node),
+        num_gpus_per_node=args.num_gpus_per_node,
```
fzyzcjy, line 266: [P2] Reject partial-node engines before deriving pod packing

When `num_gpus_per_engine` exceeds but is not divisible by `args.num_gpus_per_node` (for example, 12 GPUs on 8-GPU nodes), the floor division above produces one worker with eight slots. Consequently, `pods_per_cell()` sees an 8/8 cell and skips `exact_div`; in deliver-2, `build_entry()` emits one 8-GPU pod while `compute_engine_launch_cmd()` still starts SGLang with TP=12, so the engine cannot start. Validate the declared engine width or derive the worker count exactly before enabling node-based packing.

fzyzcjy (commented):
The new packing metadata accepts a partially filled multi-node inference engine while describing fewer GPUs than its declared TP size. This defect remains in the target deliver and causes invalid Kubernetes resource sizing.

PR #2336, opened 2026-08-10
Read an observed pod into a ParsedPod

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker remains at the specified safe deliver. The terminating-pod observation gap introduced here is fixed by 6901211c83 before f5a18edf2a, so it is transient under the requested review policy.

PR #2337, opened 2026-08-10
Project a cell's pods into workers and addresses

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue attributable to this op remains at the target safe deliver.

PR #2338, opened 2026-08-10
Implement worker provider for dynamic Kubernetes scenarios

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/worker_provider/kubernetes/core/provider.py
+++ b/miles/utils/workers/worker_provider/kubernetes/core/provider.py
@@ -1,8 +1,21 @@
 from __future__ import annotations
 
+import logging
+from collections.abc import AsyncIterator, Awaitable
+from contextlib import AsyncExitStack, asynccontextmanager
+
 from miles.utils.pydantic_utils import FrozenStrictBaseModel
+from miles.utils.workers.k8s_types import Pod
+from miles.utils.workers.reconcile.k8s_api import KubernetesAsyncioPodApi
+from miles.utils.workers.reconcile.k8s_reflector import KubernetesReflector
+from miles.utils.workers.reconcile.loop import ReconcileLoop
+from miles.utils.workers.worker_info import WorkerInfo
+from miles.utils.workers.worker_provider.base import BaseWorkerProvider, CellInfo, ReconcileFn, StopWatchFn
+from miles.utils.workers.worker_provider.kubernetes.core import cell_view, pod_view
 from miles.utils.workers.worker_provider.kubernetes.core.pod_view import CellLabelKeys
-from miles.utils.workers.worker_spec import BaseWorkerSpec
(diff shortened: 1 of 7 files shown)


miles/utils/workers/worker_provider/kubernetes/core/provider.py
```diff
@@ -1,12 +1,116 @@
...
+    def get_worker_infos(self, *, cell_ids: list[str]) -> list[list[WorkerInfo]]:
+        return [
+            cell_view.compute_worker_infos(cell_id, pods=self._pods_of_cell(cell_id), run=self._run)
+            for cell_id in cell_ids
+        ]
+
+    async def watch_cells(self, reconcile: ReconcileFn) -> StopWatchFn:
+        def notify_cell(cell_id: str) -> Awaitable[None]:
+            info = self.cell_info(cell_id)
+            return reconcile(cell_id, info if info is not None and info.alive else None)
```
fzyzcjy, line 51: [P1] Keep consumer I/O off the single reconcile worker

When a gated engine cannot answer activation, `InferenceController._reconcile` awaits `activate_launch_gate()` for up to 1,800 seconds. Returning that coroutine directly makes `ReconcileLoop`'s sole worker await it, so ready/deleted events for every other cell—and even deletion of the stuck cell—cannot be delivered during that interval, despite the loop contract forbidding I/O in its callback (`miles/utils/workers/reconcile/loop.py:29-31`). This wiring remains in deliver-2; dispatch consumer reconciliation outside the store worker while preserving per-cell ordering.

miles/utils/workers/worker_provider/kubernetes/core/provider.py
```diff
@@ -1,12 +1,116 @@
...
+class KubernetesWorkerProvider(BaseWorkerProvider):
+    def __init__(self, *, run: KubernetesRunInfo, pool_ids: list[str], resync_period: float | None) -> None:
+        self._run = run
+        self._pool_ids = pool_ids
+        self._resync_period = resync_period
+        self._loop: ReconcileLoop | None = None
+
+    async def get_addrs(self, worker_name: str) -> NamedHostAndPorts:
+        for cell_id in self.cell_ids():
+            for worker in self._workers_of_cell(cell_id):
```
fzyzcjy, line 37: [P2] Restrict address lookup to the requested cell

When an earlier observed cell cannot be projected—for example, one pod advertises three GPUs for two workers—`_workers_of_cell()` raises before this scan reaches an otherwise healthy requested worker, so one bad cell prevents unrelated cells from resolving addresses. This violates the bad-pod isolation invariant in `docs/developer/reconcile-loop.md:64` and remains unchanged in deliver-2; derive the cell ID from `worker_name` and inspect only that cell.

fzyzcjy (commented):
The provider can stall all Kubernetes cell updates behind one long-running consumer reconciliation, and an unrelated malformed cell can poison address lookup. Both issues remain in the target safe deliver.

PR #2339, opened 2026-08-10
Support cell operations in Kubernetes

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/cell_operations/kubernetes.py
+++ b/miles/utils/workers/cell_operations/kubernetes.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+import asyncio
+
+from miles.utils.test_utils.fault_injector import FailureMode
+from miles.utils.workers.cell_operations.base import BaseCellOperations
+from miles.utils.workers.worker_provider.base import CellInfo, StopWatchFn
+from miles.utils.workers.worker_provider.kubernetes.core.provider import KubernetesWorkerProvider
+
+
+class KubernetesCellOperations(BaseCellOperations):
+    def __init__(self, *, provider: KubernetesWorkerProvider, namespace: str) -> None:
+        self._provider = provider
+        self._namespace = namespace
+        self._watching: asyncio.Task[StopWatchFn] | None = None
+
+    async def cell_infos(self, *, pool_ids: list[str]) -> dict[str, CellInfo]:
+        await self._ensure_watching()
+
+        wanted = set(pool_ids)
+        infos = (self._provider.cell_info(cell_id) for cell_id in self._provider.cell_ids())
+        return {info.cell_id: info for info in infos if info is not None and info.pool_id in wanted}
+
+    async def suspend(self, *, cell_id: str) -> None:
+        await self._ensure_watching()
+
+        pods = self._provider.pod_names_of_cell(cell_id)
+        assert pods, f"cannot suspend {cell_id}, which has no pods"
+        await _delete_pods(namespace=self._namespace, pod_names=pods)
+
+    async def resume(self, *, cell_id: str) -> None:
+        raise NotImplementedError(
+            "a deleted cell comes back when its workload recreates it, so resume has no moment to return at"
+        )
+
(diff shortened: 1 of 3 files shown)


miles/utils/workers/cell_operations/kubernetes.py
```diff
@@ -0,0 +1,62 @@
...
+    async def inject_fault(self, *, cell_id: str, mode: FailureMode, sub_index: int) -> None:
+        raise NotImplementedError("fault injection reaches into a worker process, which needs the rpc layer")
+
+    async def _ensure_watching(self) -> None:
+        if self._watching is None:
+            self._watching = asyncio.ensure_future(self._provider.watch_cells(_ignore_cell))
+        try:
+            await asyncio.shield(self._watching)
+        except BaseException:
+            self._watching = None
```
fzyzcjy, line 45: [P2] Preserve the shielded watch on caller cancellation

When the first `cell_infos()` or `suspend()` caller is cancelled while the pod watch is starting, `asyncio.shield()` leaves the underlying task running, but this block clears its tracked reference. The next call can launch a second reflector and overwrite the provider's `_loop` while the first watch and API client remain live, leaking resources and violating the one-watch invariant. Clear `_watching` only when the underlying task itself fails or is cancelled, consistent with the single lazy-initialization rule ([.claude/rules/general-code-style.md:22-28](.claude/rules/general-code-style.md#L22-L28)).

fzyzcjy (commented):
The watch-initialization cancellation bug remains present at deliver-2 and can create duplicate live reflectors and leaked Kubernetes clients. It is therefore a persistent blocker rather than a transient chain issue.

guapisolo:
codex cmt

Frozen review: [#2339 @ e1fe4588952b](https://github.com/radixark/miles/commit/e1fe4588952bda4fa18f3af31a8b75a038ae7bd6) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

### [P2] Isolate cancellation between shared-watch waiters

Source: [miles/utils/workers/cell_operations/kubernetes.py:43](https://github.com/radixark/miles/blob/e1fe4588952bda4fa18f3af31a8b75a038ae7bd6/miles/utils/workers/cell_operations/kubernetes.py#L43).

Two concurrent first calls to `cell_infos()` directly await the same pending `_watching` task. Cancelling one waiter propagates cancellation to that shared task, so the other waiter also receives `CancelledError` instead of its cell information. The newly added concurrent-cancellation test specifies precisely this sequence. This is an initialization defect in the new component; the source head has no production construction site for `KubernetesCellOperations`, so deployed Kubernetes API failure is not claimed.

**Status at #3122: not fixed.** Not fixed at the reviewed [#3122](https://github.com/radixark/miles/pull/3122) ([198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69)) checkpoint. Isolate each waiter's cancellation and clean up pending startup when the last owner leaves, clearing the cache only if it still refers to that task. Preserve the single-waiter cancellation cleanup behavior as well. [#2905](https://github.com/radixark/miles/pull/2905) ([26bf8a6f529f](https://github.com/radixark/miles/commit/26bf8a6f529f483799c6d1bfbc5962223bd437da)) deleted the concurrent regression test without changing `_ensure_watching`; removing that test was not an implementation repair.

- [P2] Six existing controller tests lacked the new `cells_auto_resume` and `cluster_backend` fixture contracts, failing before their intended assertions. [Source](https://github.com/radixark/miles/blob/e1fe4588952bda4fa18f3af31a8b75a038ae7bd6/miles/utils/ft_utils/mini_ft_controller.py#L54). Fixed in [#3077](https://github.com/radixark/miles/pull/3077) ([c7992d55f2](https://github.com/radixark/miles/commit/c7992d55f244228b2698c5a92d1572dbdb52bc23)): Supplied `cells_auto_resume` in the runner helper and fake, and explicitly selected Kubernetes in the entrypoint fixture.

PR #2340, opened 2026-08-10
Treat a pod under deletion as gone from its cell

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The patch correctly propagates Kubernetes deletion state into the parsed pod, marks affected cells unavailable, and changes the membership hash. The behavior remains intact at the target safe deliver.

PR #2344, opened 2026-08-10
Read a worker's own identity out of its pod

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/serving/worker_identity.py
+++ b/miles/utils/workers/serving/worker_identity.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+import os
+from collections.abc import Mapping
+from dataclasses import dataclass
+
+from miles.utils.workers.backend_capability.base import BackendCapability
+from miles.utils.workers.process_supervisor import SUBPROCESS_INDEX_ENV_VAR
+from miles.utils.workers.worker_spec import SchedulingSpec, WorkerCtorContext
+
+CELL_INDEX_ENV_VAR = "MILES_CELL_INDEX"
+POD_INDEX_ENV_VAR = "MILES_POD_INDEX"
+
+
+@dataclass(frozen=True)
+class KubernetesWorkerIdentity:
+    cell_index: int
+    pod_in_cell_index: int
+    worker_in_pod_index: int
+    workers_per_pod: int
+    gpu_slots_per_worker: int
+
+    @property
+    def worker_in_cell_index(self) -> int:
+        return self.pod_in_cell_index * self.workers_per_pod + self.worker_in_pod_index
+
+    @property
+    def gpu_ids(self) -> list[int]:
+        first = self.worker_in_pod_index * self.gpu_slots_per_worker
+        return list(range(first, first + self.gpu_slots_per_worker))
+
+    def ctor_context(self, *, capability: BackendCapability) -> WorkerCtorContext:
+        return WorkerCtorContext(
+            cell_index=self.cell_index,
+            worker_in_cell_index=self.worker_in_cell_index,
+            gpu_ids=self.gpu_ids,
+            capability=capability,
+        )
+
+
+def read_worker_identity(
+    *, scheduling: SchedulingSpec, environ: Mapping[str, str] | None = None
+) -> KubernetesWorkerIdentity:
+    environ = os.environ if environ is None else environ
+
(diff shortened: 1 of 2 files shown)


miles/utils/workers/serving/worker_identity.py
```diff
@@ -0,0 +1,94 @@
...
+
+def read_worker_identity(
+    *, scheduling: SchedulingSpec, environ: Mapping[str, str] | None = None
+) -> KubernetesWorkerIdentity:
+    environ = os.environ if environ is None else environ
+
+    workers_per_pod = scheduling.workers_per_pod()
+    pods_per_cell = scheduling.pods_per_cell()
+
+    worker_in_pod_index = read_worker_in_pod_index(environ)
```
fzyzcjy, line 49: [P2] Require the subprocess index in multi-worker pods

When `workers_per_pod > 1` but `MILES_SUPERVISOR_SUBPROCESS_INDEX` is absent—for example, an external launcher bypasses the supervisor—this call silently selects index 0. The pod can then serve only worker 0 while the Kubernetes provider synthesizes all `workers_per_pod` endpoints, leaving the rest unreachable instead of failing fast; deliver-2 (`f5a18edf2a`) retains this default. Require the variable in the multi-worker case.

miles/utils/workers/serving/worker_identity.py
```diff
@@ -0,0 +1,94 @@
...
+        f"{POD_INDEX_ENV_VAR} is {pod_in_cell_index} in a cell spread over {pods_per_cell} pods; "
+        f"the workers this pod reports would belong to no cell of this pool"
+    )
+
+    return KubernetesWorkerIdentity(
+        cell_index=_index_from(
+            environ,
+            CELL_INDEX_ENV_VAR,
+            required_because="nothing else tells this pod which cell of its pool it belongs to",
+        ),
```
fzyzcjy, line 75: [P2] Reject cell indices outside the scheduled pool

When a numeric `MILES_CELL_INDEX` is `>= scheduling.num_cells`, this path accepts it because only pod and subprocess indices are bounded. Schedulers and address builders enumerate cells from `range(scheduling.num_cells)`, so the process constructs itself—including trainer and checkpoint identity—as a cell the run did not schedule; deliver-2 (`f5a18edf2a`) retains this path. Reject values outside the declared cell range before constructing the identity.

fzyzcjy (commented):
The identity reader accepts both an incomplete multi-worker identity and an out-of-range cell identity. Both behaviors remain in deliver-2 and can start workers under identities that the scheduler and address book do not own.

guapisolo:
codex cmt

Frozen review: [#2344 @ 35bcaac3132a](https://github.com/radixark/miles/commit/35bcaac3132a9e4e6c9c33d09ce7f67b09af5ddf) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

Updated disposition of the earlier [subprocess-index](https://github.com/radixark/miles/pull/2344#discussion_r3888580495) and [cell-bound](https://github.com/radixark/miles/pull/2344#discussion_r3888580500) reports.

- [P2] The new identity parser accepted `cell_index == num_cells`, contradicting its rejection test and the provider addressing contract. [Source](https://github.com/radixark/miles/blob/35bcaac3132a9e4e6c9c33d09ce7f67b09af5ddf/miles/utils/workers/serving/worker_identity.py#L71). Fixed in [#2345](https://github.com/radixark/miles/pull/2345) ([c92568ca40](https://github.com/radixark/miles/commit/c92568ca400c48ceb35fab7ef21ce8ed59cb7e2a)): Required `cell_index < scheduling.num_cells` before constructing `KubernetesWorkerIdentity`.

- [P2] The new identity parser defaulted a missing subprocess index to zero for multi-worker pods, returning an ambiguous identity contrary to its rejection test. [Source](https://github.com/radixark/miles/blob/35bcaac3132a9e4e6c9c33d09ce7f67b09af5ddf/miles/utils/workers/serving/worker_identity.py#L49). Fixed in [#2345](https://github.com/radixark/miles/pull/2345) ([c92568ca40](https://github.com/radixark/miles/commit/c92568ca400c48ceb35fab7ef21ce8ed59cb7e2a)): Required the supervisor subprocess index when `workers_per_pod > 1`, retaining the zero default for single-worker pods.

PR #2345, opened 2026-08-10
Build a pod's worker from the run its own argv describes

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The new server entrypoint removes the configurable bind host and always listens on IPv4, while the Kubernetes provider can publish IPv6 worker addresses. IPv6-only runs therefore cannot reach any served worker.

miles/utils/workers/serving/serve_inner.py
```diff
@@ -1,35 +1,65 @@
...
-    worker = factory(worker_argv)
+    spec = compute_serve_worker_spec(specs_fn=args.specs, pool_id=args.pool_id, worker_argv=worker_argv)
+    worker = create_worker(spec, specs_fn=args.specs, worker_argv=worker_argv)
+    _log(f"pool_id={args.pool_id} worker_class={spec.worker_class}")
 
+    port = _rpc_port_of(spec) + read_worker_in_pod_index(os.environ)
     app = create_rpc_app(worker)
-    uvicorn.run(app, host=args.host, port=args.port)
+    _log(f"serve host={DEFAULT_HOST} port={port}")
+    uvicorn.run(app, host=DEFAULT_HOST, port=port)
```
fzyzcjy: [P1] Support IPv6 when binding served workers

When a Kubernetes run is IPv6-only, including the repository's `MILES_PREFER_IPV6` strict mode, `cell_view._host_of_pod()` publishes the pod's IPv6 address, but this call always opens an IPv4 `0.0.0.0` listener. RPC clients therefore dial `[pod-ip]:port` with no matching socket and every served worker is unreachable; preserve the former configurable host or select a wildcard matching the pod's address family.

guapisolo:
codex cmt

Frozen review: [#2345 @ c92568ca400c](https://github.com/radixark/miles/commit/c92568ca400c48ceb35fab7ef21ce8ed59cb7e2a) · checkpoint [#3122 @ 198e14c872e4](https://github.com/radixark/miles/commit/198e14c872e4c5f6dd8e7dd04571f743cb18fb69).

- [P1] The new native serve path constructed the real trainer without connecting platform GPU identity, causing Ray-dependent GPU lookup to fail before RPC startup. [Source](https://github.com/radixark/miles/blob/c92568ca400c48ceb35fab7ef21ce8ed59cb7e2a/miles/utils/workers/serving/serve_inner.py#L60). Fixed in [#3041](https://github.com/radixark/miles/pull/3041) ([fdbed01e06](https://github.com/radixark/miles/commit/fdbed01e06de787faa48753ae0f9c057c79090fe)): Made native GPU lookup use platform cell identity and the supervisor subprocess index for the one-GPU-per-worker trainer spec, preserving the Ray fallback.

- [P2] The new debug-train-only test referenced undefined `compute_router_providers` and `_unusable_capability`, failing before its provider assertion. [Source](https://github.com/radixark/miles/blob/c92568ca400c48ceb35fab7ef21ce8ed59cb7e2a/tests/fast/ray/specs/test_entrypoint.py#L50). Fixed in [#2953](https://github.com/radixark/miles/pull/2953) ([c6bf72fc22](https://github.com/radixark/miles/commit/c6bf72fc22553c24d0bad7427b7bc053aacedfec)): Restored the production helper import and a valid capability fixture together, retaining the debug-train-only assertions.

PR #2353, opened 2026-08-10
dashboard: report model FLOPs utilization

Zhichenzzz:
## Motivation

`perf/actor_train_tflops` already measures per-GPU throughput. MFU is that divided by the device's peak — and nothing in miles knew what any device's peak was, so the ratio could not be formed. This adds the missing denominator and surfaces the result.

Separately, the FSDP backend passed `compute_total_fwd_flops=None`, so FSDP runs had **no** throughput metrics at all. That is fixed here too.

## Changes

- **`miles/utils/device_flops.py`** (new) maps a device name to its peak **dense** BF16 throughput. Datasheets headline the 2:4-sparsity figure, which is exactly twice the dense one, so quoting it would halve every MFU reported. `--mfu-peak-tflops` overrides the table; an unrecognised device logs no MFU rather than a percentage against an assumed peak.
- **`log_perf_data_raw`** emits `perf/actor_train_mfu` and `perf/mfu_peak_tflops`. The denominator ships alongside the ratio so the number is checkable by whoever reads it.
- **`flops_args_from_hf_config`** adapts an HF config to the shape `calculate_fwd_flops` reads, so the FSDP backend reports the same FLOPs as megatron for the same checkpoint — asserted by cross-backend parity tests. The text config is unwrapped first, so a multimodal checkpoint is sized by its text tower, not its vision tower.
- **`fwd_tflops_per_gpu`** names the `/ world_size / 1e12` convention both backends previously spelled out inline.
- **Compute Utilization** gains an MFU tile; the advisory panel gains a sustained-low-MFU rule. Both read one summary computed server-side, so the tile and the rule cannot disagree about what "mean MFU" means.
- **`docs/user-guide/dashboard.md`** gains a `Model FLOPs utilization` section covering the formula, the conventions, the caveats and the measured baselines.

## Measured

8xH200, three rollout steps each, all with gradient checkpointing:

| | colocate | fully-async |
|---|---|---|
| Qwen3-4B dense, FSDP | 25.3 / 24.6% | 26.0 / 26.2% |
| Qwen3.5-35B-A3B MoE, EP, `--optimizer-cpu-offload`, MTP | 5.0 / 5.8% | 5.4% |

MFU barely moves between colocate and fully-async while `perf/wait_time_ratio` collapses (0.708 → 0.060 dense, 0.760 → 0.027 MoE). The two answer different questions — how efficiently the training step computes, versus how much of the step waited for rollout data — which is why this stayed a train-step ratio rather than an end-to-end one. A blended number would have read as "async made training 3.5x more efficient", which is false.

## For reviewers

**`LOW_MFU = 0.15` is a placeholder, not a calibrated threshold.** The two rows above are both healthy and differ five-fold, so no single absolute number separates "slow by configuration" from "slow because something broke". At 0.15 this rule fires on any offload- or recompute-heavy MoE run. It is kept as a starting point; a drop-vs-own-baseline check would be the better shape once enough real runs are on record. Happy to drop the rule from this PR and ship only the metric and the tile if reviewers prefer.

**Known gap:** `calculate_fwd_flops` models the base model only, so a run with `--enable-mtp-training` does real work that never reaches the numerator. This affects the pre-existing `perf/actor_train_tflops` identically; it is documented rather than fixed here.

**Conflict warning:** #2027 rewrites `compute_advisories` in the same file. The new rule is a standalone function, so it should move into that PR's alarm tier with a one-line change, but the two will need reconciling.

## Test

- `tests/fast/dashboard` + `tests/fast/utils` on 8xH200: 3028 passed, 3 skipped
- End to end against a real MoE run's dump: the summary correctly excludes step 0, and the rule declines to judge below its step floor
- Dashboard rendered and verified in a browser; the tile is absent for runs with no MFU key


yueming-yuan (approved):

miles/utils/device_flops.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+from functools import cache
+
+_PEAK_BF16_TFLOPS: dict[str, float] = {
+    "A100": 312.0,
+    "H100": 989.0,
```
yueming-yuan, line 8: This might not be accurate for all specs, e.g. H100 PCIe & H100 NVL

miles/dashboard/advisory.py
```diff
@@ -18,6 +18,12 @@
 LOW_CACHE_HIT_RATE = 0.10
 HIGH_TOKEN_USAGE = 0.95
 
+LOW_MFU = 0.15
```
yueming-yuan: I'm not sure if this can be too high for large models.. Previously, I remember I profiled long-context k2.5 training on GB300 (which is MLA arch, and mature in megatron, and it can only reach ~17-18% mfu at best config.)

miles/utils/flops_utils.py
```diff
@@ -125,3 +128,61 @@ def calculate_fwd_flops(
...
+    return SimpleNamespace(
+        hidden_size=hidden_size,
+        num_attention_heads=num_attention_heads,
+        num_query_groups=_first(config, "num_key_value_heads", default=num_attention_heads),
+        vocab_size=config.vocab_size,
+        num_layers=num_layers,
+        ffn_hidden_size=config.intermediate_size,
+        kv_channels=_first(config, "head_dim", default=hidden_size // num_attention_heads),
+        num_experts=num_experts,
+        moe_ffn_hidden_size=getattr(config, "moe_intermediate_size", None),
```
yueming-yuan: this will crash at `moe_ffn = args.moe_ffn_hidden_size * args.moe_router_topk + shared_expert_ffn` in flops_utils.py when it is None

miles/dashboard/advisory.py
yueming-yuan: (this one is not change request, just fyi)

Zhichenzzz: will add one choice for users, where they could set the boundary by themselves workload and expected MFU

PR #2354, opened 2026-08-10
Delete the glm4-9B, mimo-7B, moonlight-16B and deepseek-r1 launch scripts

yueming-yuan:
First of three chained PRs cleaning up `scripts/`. Stack: **#2354 (this)** → #2355 → #2356.

These four models are being dropped, so their launchers go, along with the recorded command snapshots and the doc pages built around them.

## What is deleted

| | |
|---|---|
| launchers | `run-glm4-9B.sh`, `run-glm4-9B-4xgpu-radixtree.sh`, `run-mimo-7B-rl-eagle.sh`, `run-moonlight-16B-A3B.sh`, `run-deepseek-r1.sh` |
| snapshots | the matching `tests/snapshots/launch_scripts/sh/scripts/*.txt` |
| doc pages | `docs/models/glm/glm4.md`, `docs/models/mimo/mimo.md`, `docs/models/kimi/moonlight.md` |

## What deliberately stays

Three of the four Megatron model definitions under `scripts/models/` are **not** deleted, because something that still runs references them:

- `mimo-7B-rl.py` — `tests/e2e/megatron/test_mimo_7B_mtp_only_grad.py`, which is registered on `stage-c-4-gpu-h200` and `stage-c-4-gpu-mi350` with no `disabled=`
- `moonlight.py` — `examples/infra_features/low_precision/run-moonlight-16B-A3B-int4.sh` and the p2p profile
- `deepseek-v3.py` — `scripts/run_deepseek.py` and the V3.2 launchers

`glm4-9B.py` **is** deleted, in the second commit. It initially looked referenced, but neither reference held up: `tests/e2e/megatron/test_quick_start_glm4_9B.py` registers itself `disabled="Not worthy for testing, too naive."` and `run_suite.py` runs only entries whose `disabled` is None, so it has never executed; and the p2p weight-transfer profile was the only real consumer, so it goes with the model. The disabled test, the p2p prepare/run configs, the `GLM-Z1-9B.sh` wrapper, both recorded p2p argv snapshots and three rows of `docs/advanced/p2p-weight-transfer.md` are removed with it. Dropping the measured RDMA-vs-NCCL row is deliberate — the configuration behind that number no longer exists.

## `deepseek.md` is rewritten rather than deleted

It was the only page covering a model that survives: DeepSeek-V3 is still supported through `scripts/run_deepseek.py`. The page's whole recipe section was transcribed from the deleted shell script, so it is rewritten around the python launcher's actual values (TP/PP/CP/EP ladder keyed off `--num-nodes`, the `deepep low_latency` sglang block, the `--max-tokens-per-gpu 2048` placeholder flagged as such). R1-specific content is gone; index pages and `docs.json` are updated.

`docs/user-guide/training-script-walkthrough.md` used `run-glm4-9B.sh` as its reference script and is re-pointed at a surviving one.

## Testing

`pytest tests/manual/launch_scripts tests/fast/launch_scripts` — 524 passed. The discovery test that asserts snapshot-set == script-set is what proves nothing was left dangling.


guapisolo (approved):
LGTM.

docs/models/deepseek/deepseek.md
```diff
@@ -1,57 +1,89 @@
...
 
-### 3.1 Required env vars
+## 3. Launch
+
+`scripts/run_deepseek.py` drives the whole pipeline: HF download, FP8 → BF16 cast, HF → Megatron
+`torch_dist` conversion, an rsync of both directories to node-local storage, and the `train.py`
+submission.
+
+```bash
+python scripts/run_deepseek.py train --num-nodes 16 --num-gpus-per-node 8
```
guapisolo, line 32: [P1] Start and preserve Ray before preprocessing

This documented command fails on a clean host before the launcher reaches its own `ray start --head`. `train()` calls `_prepare_megatron_ckpt()` and `_prepare_cp()` first; both eventually call `exec_command_multi_node()`, whose first operation is `ray.init(address="auto")`, so they require an existing Ray head. The multi-node workaround is also lost by default because `execute_train()` stops that cluster unless `MILES_SCRIPT_EXTERNAL_RAY=1`, then starts only a new head. Since this PR makes `run_deepseek.py` the documented replacement, please either establish and preserve Ray before these preprocessing steps or document and wire an external-Ray workflow.

yueming-yuan: solved in next PR

PR #2356, opened 2026-08-10
Replace all the `.sh` launch scripts with `.py` launch script

yueming-yuan:
Last of three chained PRs cleaning up `scripts/`. Stack: #2354 → #2355 → **this**. **Base is `yueming/script-bugfix`, review the other two first.**

After this, `scripts/` contains no `.sh` launcher. Part of #1837, which already moved the model definitions to python (#1910); this finishes the job for the launchers themselves.

## What replaces what

Near-duplicate recipes are merged behind a per-variant table instead of being copied per file:

| new launcher | replaces |
|---|---|
| `run_qwen3_dense.py` | `run-qwen3-{4B,4B_4xgpu,32B}.sh`, `run-qwen3.5-{4B,9B,27B}.sh`, `run-qwen3.6-27B.sh` — 6 recipes behind `--model-name` |
| `run_qwen3_sft.py` | `run-qwen3-4B-base-sft.sh`, `run-qwen3-235B-A22B-sft.sh` |
| `run_qwen3_235b_a22b.py` | `run-qwen3-235B-A22B.sh` |
| `run_qwen3_next_80b_a3b.py` | `run-qwen3-next-80B-A3B{,-8gpus}.sh` — `--topology` |
| `run_kimi_k2.py` | `run-kimi-k2-{Instruct,Thinking}.sh` |
| `run_nemotron_3_nano.py` | `run-nemotron-3-nano-{4b,30b-a3b}.sh` |
| `run_nemotron_3_super_120b_a12b.py` | `run-nemotron-3-super-120b-a12b.sh` — `train` / `worker` commands for the head/worker role split |
| `run_gpt_oss_20b.py` | `run-gpt-oss-20b-bf16.sh` |
| `run_glm45_355b_a32b_8node.py` | `run-glm4.5-355B-A32B.sh` |
| `amd/run_qwen3_4b.py` | `amd/run-qwen3-4B-amd.sh` |

Three shell recipes were absorbed by python launchers that already shadowed them, which gained the fields their recipes needed: `run-qwen3.5-35B-A3B-mtp.sh` → `run_qwen3_5_35b_a3b_mtp.py` (renamed from `..._cp2_ep8.py`, since the parallelism is now `--parallelism`), `run-glm4.7-flash.sh` → `run_glm47_flash.py`, `run-kimi-k25.sh` → `run_kimi_k25.py`.

## How semantic equivalence was checked

Not by eye. For each deleted launcher, its recorded snapshot from the base branch was diffed against a fresh recording of the replacement at equivalent settings, comparing the ray runtime env plus the `train.py` argv as a flag → values multiset — so ordering is ignored and every flag is accounted for. Merged launchers were recorded once per variant, since the snapshot suite only records default args.

**Result: no parallelism, batch-size, response-length, algorithm-constant, optimizer or sglang value differs anywhere, for any of the 23 recipes.**

The intended differences are uniform and deliberate:

- **Launcher infrastructure.** `execute_train` owns the preamble, so every script gains `PYTHONUNBUFFERED` / `no_proxy` / `MASTER_ADDR` and a repo-root-prefixed `PYTHONPATH` in the runtime env, kills `miles`/`redis` rather than `python`, drops `--dashboard-host/--dashboard-port` from `ray start` (submission still targets the default 8265, so only remote dashboard access changes), and uses an absolute `train.py` path.
- **Directories.** `--model-dir` (default `/root/models`), `--data-dir` (default `/root/datasets`) and `--output-dir` replace hardcoded `/root/<Model>`, `${BASE_DIR}`, `${BASE_FOLDER}`, `${MODEL_DIR}`. Every old path is reachable by setting the fields, and the msc miles flavor already injects the cluster values through `MILES_SCRIPT_*`.
- **wandb** comes from `get_default_wandb_args`, so it engages only when `WANDB_API_KEY` is set. Two scripts had hardcoded project/group names; those are gone.
- **`--num-gpus-per-node` is now always passed**, so the GPU-count field is actually honoured under `--colocate` on sub-8-GPU nodes.
- `--rollout-num-gpus` is dropped where `--colocate` is set, since `arguments.py` documents and implements it as ignored there.
- Two shell quirks were dropped rather than ported, both on review feedback: the GLM-4.5 and
  Kimi-K2-Instruct recipes read the eval set from `<data-dir>/rl_data/`, a layout nothing
  produces (`hf_download_dataset` writes `<data-dir>/aime-2024`), so they now use the standard
  one; and the gpt-oss launcher no longer injects `HF_HOME=/workspace/hf_cache` into the ray
  runtime env — the shell script exported it for the submitting process only, so the workers
  never saw it.

Two per-script notes: the three Kimi-K2 recipes never ran `ray start` at all, which is expressed with `MILES_SCRIPT_EXTERNAL_RAY=1` — that knob also skips the `ray stop --force` / `pkill -9 ray` the shell scripts did perform, which is unavoidable today. And `run-qwen3-4B_4xgpu.sh` omitted `--sglang-mem-fraction-static` while its sibling set `0.7`; the merged launcher applies the family value.

## Judgement calls worth a look

- **`run_glm45_355b_a32b_8node.py` is a new file rather than a mode of the existing `run_glm45_355b_a32b.py`.** The two are different experiments for the same model — gspo vs grpo, TP8/EP16 vs TP4/EP8, 8 GPUs per node vs 4, EAGLE MTP and DeepEP and a 20-var NCCL/IB tuning block on one side only, router-health and fault-tolerance flags on the other. A recorded diff between them shows 77 differences, and the existing script asserts against H100. Merging would have bloated it and risked a recipe in active use.
- **`run_qwen3_4b_npu.sh` is deleted with no replacement.** It exported `SLIME_SCRIPT_*` names nothing in the repo reads, duplicated env vars `execute_train_npu` already sets, and invoked `run_qwen3_4b.py` — the CUDA launcher — not the NPU one. Its allocator and PYTHONPATH exports therefore never reached the NPU launcher, so folding them into `run_qwen3_4b_npu.py` would mean guessing at a configuration that has never run.

## Also here

- `command_utils.ssh_start_ray_workers` — the four recipes that fan out to an MPI-style hostfile over ssh before submitting shared that loop; it is now one helper wired in through `before_ray_job_submit`.
- `MLP_SOCKET_IFNAME` added to the harness's `CLEARED_ENV`, so a developer with it exported cannot unfreeze the glm4.5 recording.
- `tests/fast/launch_scripts/test_sh_harness.py` used `scripts/run-qwen3-4B.sh` as its own fixture and is re-pointed at a surviving `examples/` launcher; the six `scripts/` entries in `_SCRIPTS_REFUSING_TO_RUN_WITHOUT_EXPLICIT_INPUTS` are removed and the discovery lower bound adjusted for the 39 remaining `examples/` snapshots.
- ~21 doc pages updated to the python invocations, including re-keying the per-variant knob tables by `--model-name` and updating `hf download --local-dir` destinations to the new directory defaults.

## Testing

`pytest tests/manual/launch_scripts tests/fast/launch_scripts tests/fast/utils/test_command_utils.py` — 570 passed under random ordering. `pre-commit run --all-files` clean.

## Follow-ups not in scope

- `scripts/run_qwen3_4b_npu.py` has its own defects (a `/root/model` vs `/root/models` typo, a dataset `prepare()` downloads but `execute()` never reads, `TRAIN_BACKEND` asserted then ignored) and cannot be imported without the NPU docker patch, so it needs someone who can verify on NPU hardware.
- `docs/user-guide/training-script-walkthrough.md` is still framed around bash `XXX_ARGS` arrays, which remains accurate for `examples/`; re-centring it on the python launchers is a separate docs pass.


docs/models/glm/glm4-5.md
```diff
@@ -26,26 +26,25 @@ The 106B-A12B variant has no launcher under `scripts/`; the canonical recipe is
...
 
 ### 3.2 Download model + datasets
 
 ```bash
-hf download zai-org/GLM-4.5 --local-dir $BASE_DIR/GLM-4.5-355B-A32B
-hf download --repo-type dataset zhuzilin/dapo-math-17k --local-dir $BASE_DIR/dapo-math-17k
-hf download --repo-type dataset zhuzilin/aime-2024     --local-dir $BASE_DIR/rl_data/aime-2024
+hf download zai-org/GLM-4.5 --local-dir /root/models/GLM-4.5-355B-A32B
+hf download --repo-type dataset zhuzilin/dapo-math-17k --local-dir /root/datasets/dapo-math-17k
+hf download --repo-type dataset zhuzilin/aime-2024     --local-dir /root/datasets/rl_data
```
Zhichenzzz: we should unify the `/root/datasets/rl_data` and `/root/datasets/`

scripts/run_qwen3_next_80b_a3b.py
```diff
@@ -0,0 +1,282 @@
...
+        f"{grpo_args} "
+        f"{U.get_default_wandb_args(__file__, run_id=args.run_id)} "
+        f"{perf_args} "
+        f"{eval_args} "
+        f"{sglang_args} "
+        f"{misc_args} "
+        f"{args.extra_args} "
+    )
+
+    start_workers = args.ssh_start_workers and args.recipe.actor_num_nodes > 1
```
Zhichenzzz: only in this script using this gate

scripts/run_gpt_oss_20b.py
```diff
@@ -0,0 +1,159 @@
...
+        f"{args.extra_args} "
+    )
+
+    U.execute_train(
+        train_args=train_args,
+        config=args,
+        num_gpus_per_node=args.num_gpus_per_node,
+        megatron_model_type="gpt-oss-20b",
+        # the ray workers download the tokenizer / config, so the cache belongs in their env
+        extra_env_vars={"HF_HOME": "/workspace/hf_cache"},
```
Zhichenzzz: the unique one?

tests/fast/launch_scripts/py_harness.py
```diff
@@ -35,6 +35,7 @@
     "GLOO_SOCKET_IFNAME",
     "KEEP_MOE_LORA",
     "MILES_SCRIPT_EXTERNAL_RAY",
+    "MLP_SOCKET_IFNAME",
```
guapisolo, line 38: [P2] Clear `MLP_WORKER_0_HOST` when freezing launcher snapshots

`freeze_environment()` inherits the caller's environment, but this list only clears `MLP_SOCKET_IFNAME`; four new launchers also read `MLP_WORKER_0_HOST` when building the SSH worker command. With `MLP_WORKER_0_HOST=10.9.8.7`, the recorded command skips that host while the committed snapshot expects `127.0.0.1`, so the snapshot test is host-dependent and fails in MLP environments. Please add `MLP_WORKER_0_HOST` to `CLEARED_ENV`.

guapisolo (commented):
One snapshot-isolation issue.

Zhichenzzz (approved):
LGTM if they could smoothly run

PR #2367, opened 2026-08-11
Add a PPO example for the shared actor/critic setup

Shi-Dong:
## What this adds

A runnable PPO recipe under `examples/ppo/`: a single-node Qwen3-4B launch script plus a README, and one line in the examples index.

PPO has been supported in miles for a while (`--advantage-estimator ppo`), but there was no example for it. Every other advantage estimator a user is likely to reach for has one. The only place the working flag combination existed was `tests/e2e/megatron/test_qwen3_4B_ppo.py`, which is not somewhere people look for a recipe, and which is written as a 3-step smoke test rather than as something you would train with.

## Why now

That e2e test used to be disabled (`PPO placement group has conflict on port`). It is enabled again now that the actor and critic share a placement group, so single-node PPO actually works and is covered by CI. That makes this a good moment to document it.

## What the README covers

- The mental model for why PPO needs a critic at all, and how that differs from GRPO's group baseline.
- Every `--critic-*` flag and what it falls back to when unset. Several are derived rather than defaulted, which is not obvious: `--critic-lr` falls back to `--lr`, `--critic-load` to `--load`, and `--critic-save` to `--save` with a `_critic` suffix so the two models do not share an iteration tracker.
- The constraints that are enforced at argument validation, so a reader hits them in the docs rather than in a stack trace. The two non-obvious ones: the critic is colocated on the actor's GPUs and inherits its parallelism (which is *why* `--offload-train` is forced on, and why scaling only ever means changing the actor's placement); and `--kl-coef` must be zero, because the critic trains before the actor and never sees ref log probs, so reward-level KL would silently leave the penalty out of its value targets.
- Which numbers are CI-verified and which are not.

## Deviations from the CI test

The parallelism and the PPO flag set follow the e2e test. Three values intentionally do not, and the README says so explicitly:

- `--eps-clip` is `0.2` here against `4e-4` in CI. `4e-4` pins the actor nearly in place, which is what you want for a fast deterministic test and not what you want for training.
- `--num-rollout` is `300` here against `3` in CI.
- `--rollout-num-gpus-per-engine` is `1` here against `2` in CI. Qwen3-4B fits comfortably on one GPU, so one engine per GPU avoids paying tensor-parallel communication for no capacity gain.

I would rather ship an example that trains and flag where it diverges from the tested config than ship the smoke-test values under a name that suggests they are tuned.

## Testing

The launcher was run **unmodified** on a single 8xH200 node, straight from this branch, exactly as the README says to (`python examples/ppo/run_qwen3_4b_ppo.py`). It went through its own `prepare` step — model download, dataset download, `torch_dist` conversion — and then **20 rollout steps (0 through 19) with zero errors**.

What that confirms, against the claims the README makes:

- **The critic-only warmup gate works.** `rollout_id=0` ran only `train_critic`; from `rollout_id=1` onward both `train_critic` and `train_actor` ran, which is `--num-critic-only-steps 1` behaving as documented.
- **The critic really is colocated and inherits the actor's placement.** `critic_num_gpus_per_node` resolved to 4, same as the actor, and the whole job (actor, critic and rollout engines) ran inside a 4-GPU Ray cluster.
- **One engine per GPU took effect.** `rollout_num_gpus_per_engine = 1` with four distinct SGLang engine actors, rather than two engines at TP=2.
- **Both checkpoints are written, to separate directories.** At rollout 19 the run produced `<save>/checkpoints/iter_0000019` (57 GB) and `<save>/checkpoints_critic/iter_0000019` (53 GB), so the derived `--critic-save` path is not just resolved, it is actually written.
- **The value function learns.** `train/critic-value_loss` fell from 0.594 to 0.0364 across the 20 rollouts.
- **The actor trains sanely.** Final step reported `train/pg_loss -0.0573`, `pg_clipfrac 0.0054`, `ess_ratio 0.998`.
- Median `perf/step_time` was 117s (min 97, max 259) with median `train_wait_time` 99s, i.e. rollout-dominated, as expected at `--rollout-max-response-len 8192`.

Also `bash`/`py_compile` syntax checks, `black` 24.3.0, `ruff` and `isort` all clean, and every flag was checked against `miles/utils/arguments.py`; the remainder are Megatron/SGLang-owned and are used identically in the e2e test and the existing on-policy-distillation example.

This branch is currently behind main, but none of the commits main has moved ahead touch the critic or advantage code paths (they are FSDP, dashboard and session changes), so the result above still applies on merge.

No docs nav entry: only a subset of examples have Mintlify pages, and this follows `lora` and `on_policy_distillation` in not adding one.


examples/ppo/run-qwen3-4b-ppo.sh
```diff
@@ -0,0 +1,163 @@
+#!/bin/bash
```
guapisolo: Make it as an python file?

examples/ppo/run-qwen3-4b-ppo.sh
```diff
@@ -0,0 +1,163 @@
...
+
+WANDB_ARGS=(
+   # --use-wandb
+   # --wandb-project miles-dev
+   # --wandb-group qwen3-4B-ppo
+   # --wandb-key ${WANDB_KEY}
+)
+
+SGLANG_ARGS=(
+   --rollout-num-gpus-per-engine 2
```
guapisolo: 1 will be faster for 4b model

examples/ppo/README.md
```diff
@@ -0,0 +1,91 @@
+# PPO Example
```
guapisolo, line 1: Just confirm. Do you intentionally put PPO doc here b/c it's not fully verified?

examples/ppo/README.md
```diff
@@ -0,0 +1,91 @@
...
+These are enforced at argument validation, so you get an error rather than a silent wrong result:
+
+* **Megatron only.** PPO raises with any other train backend, and is unsupported with
+  `--megatron-to-hf-mode bridge`.
+* **`--kl-coef` must be 0.** Reward-level KL is rejected because the critic trains *before* the
+  actor and never sees ref log probs, so its value targets would silently exclude the KL penalty
+  applied to the actor's rewards. Use loss-level `--use-kl-loss` / `--kl-loss-coef` instead.
+* **Not compatible with `MILES_EXPERIMENTAL_FT_TRAINER=1`.** The v2 fault-tolerant train group
+  cannot route critic values yet.
+* **`--offload-train` is forced on.** Actor and critic share the train GPUs, so both resident at
```
guapisolo: Actor and critic trainer are collocated.  
Then offload train is forced on.
And currently they share the same paralism setting.

examples/ppo/README.md
```diff
@@ -0,0 +1,91 @@
...
+  for a fast deterministic test and wrong for actual training. `0.2` is the standard PPO value.
+* `--num-rollout 300` here vs. `3` in CI.
+
+Treat the rest — learning rates, `--kl-loss-coef`, `--entropy-coef` — as starting points to tune,
+not as tuned values.
+
+## Scaling up
+
+The actor world size is `--actor-num-nodes` × `--actor-num-gpus-per-node`, and `TP × PP × CP` must
+divide it. The critic inherits the same shape automatically, so when you change the actor's
```
guapisolo: I think we can remove this, and mention in constrains ?

examples/ppo/README.md
Shi-Dong: The example *is* fully verified. We will add this example to the Miles docs in a separate PR.

guapisolo (approved):

PR #2368, opened 2026-08-11
fix(rollout): group session v2 leaf samples

Shi-Dong:
## Summary

- stamp every leaf sample returned by one session-server-v2 agentic rollout with the same `rollout_id`
- preserve an explicit input `rollout_id`, with the globally assigned input `Sample.index` as the fallback
- fail early if a v2 rollout has neither identity instead of emitting samples that cannot be grouped safely

## Why

Session server v2 may return multiple trainable leaves for one environment rollout, including leaves created by compaction or subagents. Those leaves need a shared rollout identity so Miles can validate nested output, apply rollout-level masking, and treat them as siblings during downstream reward processing.

This change is independent of the reward-normalization companion PR #2369 and targets `main` directly.

## Tests

- repository-pinned pre-commit hooks on both changed files
- `uvx ruff check miles/rollout/generate_hub/agentic_tool_call.py tests/fast/rollout/generate_hub/test_agentic_v2.py`
- focused macOS run: `pytest -q --noconftest tests/fast/rollout/generate_hub/test_agentic_v2.py` (7 passed)

No README update is needed because this only fills the existing `Sample.rollout_id` contract; it adds no user-facing option.

### DIFF
--- a/miles/rollout/generate_hub/agentic_tool_call.py
+++ b/miles/rollout/generate_hub/agentic_tool_call.py
@@ -115,6 +115,7 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
 
     samples = result.samples
     if use_v2:
+        # FIXME: handle sample index issues.
         rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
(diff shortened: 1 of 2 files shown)


guapisolo (approved):
good fix.

PR #2371, opened 2026-08-11
HUD computer-use RL: train a VLM on HUD tasksets through a browser

nblintao:
Trains a VLM to operate a real GUI through [HUD](https://hud.ai) environments, using
a HUD taskset row as the task definition. Nothing under `miles/` changes.

**Result** — Qwen3-VL-4B-Instruct, GRPO, 8×H200, 32 episodes per step, 2048 in a
browser: over 16 steps mean game score went 200 → 865, and the *worst* episode of a
batch went 0 → 540. The second number is the one that matters — a lucky episode was
already worth ~800 points before training, so what training bought is reliability,
not a higher ceiling. Episodes reaching tile 64+ went 22% → 97%.

## Where it plugs in

| seam | file |
|---|---|
| `--custom-generate-function-path` | `rollout.py` — multi-turn episode loop |
| `--custom-rm-path` | `rollout.py` `reward_func` |
| `--custom-config-path` | `hud2048_config.yaml` |
| interaction env | `hud_task_env.py` |

A HUD row carries the image (`mcp_config`'s `Mcp-Image`), the setup call and the
grading call, so the adapter has no per-task code: the env boots that image, makes
that setup call, and grades with that evaluate call. Anything named `hud2048_*` or
`run_hud2048` is this one taskset; the rest is not supposed to know which taskset it
is running.

The model acts by writing `Action: press("Left","Down")` / `click(x, y)` /
`type("...")` / `done()` — plain text rather than provider tool calls, because the
loss is computed on the tokens the policy generated, so an action has to *be* those
tokens.

## Format coverage

This supports the HUD taskset row format the `hud-evals/*` datasets publish —
`mcp_config`, `setup_tool`, `evaluate_tool` — which is what their CUA tasksets ship
and what the matching environment images serve.

HUD's current spec (v6) defines a task differently: an `@env.template()` generator
inside an environment package, reached over HUD's own control channel. Supporting it
is the natural next step, and the seam for it is already in the right place — a v6
leg replaces `mcp_client.py`, the fetch in `make_hud_data.py` and the two tool calls
in `hud_task_env.py`, while the rollout, reward wiring and sandbox lifecycle, which
is most of this example, stay as they are. This PR lands that format-independent
part, plus a verified training result to carry onto v6.

## What the environment needed

Four ceilings, each found the expensive way and each annotated where it is set:

- **Move budget.** A random policy and a corner strategy both cap at tile 8 by ~12
  moves and only separate from ~20 on, so one action carries 4 keys and 24 turns buy
  ~92 moves.
- **ViT prefill.** Every turn re-prefills the screenshot history through the vision
  encoder; at 960px × 14 turns a cache-missing episode re-prefilled ~5.5k tokens at
  14 tok/s, slow enough to stall a batch behind one engine.
- **A step-function reward.** The taskset grades `log2(max_tile)/log2(target)`, five
  levels. Twice a population converged on one tile, within-group variance went to
  zero and GRPO had no gradient. Training moved onto the same environment's score
  (continuous); the taskset's own grade is still recorded every episode as
  `task_reward`.
- **Saturation.** At a mean of 0.71, 59% of a batch sat at exactly 1.0.

Diagnostic worth knowing: a mean reward on an *exact* fraction (0.428571, 0.714286)
means zero within-group variance. `rollout/advantages ≈ 0` does not reveal it — GRPO
centres advantages within each group either way.

## Sandboxes

One episode is one Daytona sandbox: a confirmed delete, a per-process reaper, and
Daytona's TTL behind that. The fire-and-forget default silently left a step's worth
of sandboxes running every step.

## Tests

25 offline cases against a fake MCP server — no GPU, no network, no Daytona, under
two seconds:

```
python -m pytest examples/experimental/hud/tests/ -q
```

They cover the three contracts this example translates between: the task row, the
model's text actions, and MCP calls. `smoke_episode.py` runs one real episode against
one real sandbox for the parts a fake cannot cover — that is the check to run when
pointing this at a different taskset, and it costs no GPUs.

## Known rough edge

`rollout.py` imports the multi-turn VLM engine from `examples/geo3k_vlm/multi_turn/`
as private helpers. Sharing one implementation of token accumulation and per-turn loss
masking beats copying it, but that engine has **no test coverage** — nothing under
`tests/` imports it, and the only VLM e2e test runs single-turn — and no API contract,
so a refactor there could break this leg silently. Unit-testing those helpers is worth
more than moving them; after that they belong in `miles/rollout/generate_hub/`.

### DIFF
--- a/docs/user-guide/environments.md
+++ b/docs/user-guide/environments.md
@@ -19,6 +19,7 @@ where the environment itself comes from:
 | Integration | Plugs in at | Guide |
 |---|---|---|
 | [Harbor](https://github.com/harbor-framework/harbor) | agent function | [guide](/user-guide/harbor) |
+| [HUD](https://hud.ai) | generate function³ | [example](https://github.com/radixark/miles/tree/main/examples/experimental/hud) |
 | [NeMo Gym](https://github.com/NVIDIA-NeMo/Gym) | agent function | [guide](/user-guide/nemo-gym) |
 | [OpenEnv](https://github.com/huggingface/openenv) | agent function | [guide](/user-guide/openenv) |
 | [Strands Agents](https://strandsagents.com/) | generate function | [example](https://github.com/radixark/miles/tree/main/examples/experimental/strands_sglang) |
@@ -31,7 +32,7 @@ Sandbox providers are a different axis: they provision the task containers
 | Sandbox provider | Used within | Guide |
 |---|---|---|
 | [AgentENV](https://github.com/kvcache-ai/AgentENV) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/agentenv) |
-| [Daytona](https://www.daytona.io/) | OpenEnv, Harbor, NeMo Gym | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
+| [Daytona](https://www.daytona.io/) | OpenEnv, Harbor, HUD, NeMo Gym | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 | [E2B](https://e2b.dev/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 | [Modal](https://modal.com/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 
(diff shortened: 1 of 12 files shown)


nblintao:
closing in favor of hud v6 api: https://github.com/radixark/miles/pull/2479

PR #2373, opened 2026-08-11
docs: NeMo-Gym server no longer needs a fork branch

nblintao:
The NeMo-Gym connector needs a per-request `policy_base_url` override in NeMo-Gym's responses-API agents — that is how each episode is pointed at the training session's URL. It was still a proposal when this integration landed, so the guide, the recipe README and the adapter docstring told readers to clone a fork branch (`nblintao/Gym@mini-swe-agent-per-request-policy-url`) "until it merges".

It is upstream now, so all three say `main` (>= `fcca3a8`) and the clone command points at `NVIDIA-NeMo/Gym`.

Docs and comments only.


Shi-Dong (approved):
LGTM!

PR #2376, opened 2026-08-11
docs(developer): rewrite the developer guide against the code

Zhichenzzz:
Based on `main`. The Developer Guide had drifted from the code in ways that actively mislead a new contributor: it documented a formatter setting we do not use, a package layout with a directory that no longer exists, and a set of debugging flags that were never real. It also said nothing about the two things a contributor most needs, which is how to make CI run their test and how the dependency versions fit together.

Every statement below was read out of the code, the workflows, or `.claude`.

## `developer/contributor-guide.md`: rewritten

**Repository layout** regenerated from the tree. `backends/experimental/` is gone, `fsdp_utils/` sits beside `megatron_utils/`, and `dashboard/`, `true_on_policy/`, `miles_plugins/optimizers/` now appear.

**Code style** now describes `.pre-commit-config.yaml` instead of prose. The old page said "black, line length 100"; `[tool.black]` says 119. Each hook is listed with what it enforces, including that `E402` and `E501` are deliberately ignored. The three repo-local pygrep bans are documented for the first time, each with the API to use instead and why: `mpu.get_*` → `get_parallel_state()` (the two backends share one `ParallelState`), bare `Auto{Config,Tokenizer}.from_pretrained` → `load_hf_config` / `load_tokenizer`, `huggingface-cli` → `hf`.

**`.claude`** is explained: `rules/` are path-scoped conventions (`general-code-style.md`, which `AGENTS.md` also points Codex at), `skills/` are procedures, and the `doc-dev:` sentinel is spelled out with the current sentinel table, because a contributor editing `.github/workflows/pr-test.yml` or `docker/build.py` walks into one whether or not they run an agent.

**Running CI** covers what a PR triggers, how registration works (`tests/fast/` auto-registers as `stage-a-cpu`; everywhere else needs one top-level `register_*_ci(...)`, parsed from the AST, with `est_time` / `suite` / `labels` / `nightly` / `disabled`), how to verify pickup with `run_suite.py --list-only` before trusting it, and what each label does (`run-ci-<x>`, `run-ci-all`, `nightly`, `run-ci-image`, `bypass-fastfail`), including that a `run-ci-*` label doubles as fork-PR CI approval.

**PR-description CI tags** are documented for the first time. All three are parsed with `grep -oP '^<key>:\s+\K\S+'` against the PR body:

| Line | Effect |
|---|---|
| `ci-image-tag: <tag>` | GPU suites run on `radixark/miles:<tag>`; must be a bare tag |
| `ci-sglang-pr: <ref>` | SGLang checked out at `<ref>` instead of `sglang-miles` |
| `ci-megatron-pr: <ref>` | Megatron-LM checked out at `<ref>` instead of `miles-main` |

Including the `#1234` shorthand that resolves to `refs/pull/1234/head` (this is how you test an unmerged upstream PR), the precedence (dispatch input > PR body > default), and the exception that a freshly built `pr-<n>` image outranks `ci-image-tag:`.

Also new: `/run-lint` as the way out of a red `pre-commit` job, and the `ci-infra` / `flaky` issue labels.

## `developer/debug.md`: corrected and completed

The old "Useful debugging knobs" block listed `--debug-determinism`, `--debug-weight-sync` and `--debug-rollout-print-every`. None of those flags exist in `arguments.py`. The "grep for these signal phrases" list was in the same state: `forward_logp_max_diff`, `Continuous async rollout worker started` and `Loaded checkpoint from` appear nowhere in the codebase. Both sections are replaced with the real surface.

Documented from the code, with what each is designed to isolate:

- **Cutting the loop in half**: `--debug-rollout-only` / `--debug-train-only` (mutually exclusive, asserted), `--save-debug-rollout-data` / `--load-debug-rollout-data` (the latter implies train-only).
- **`--ci-inject-rollout-data-path`**, which the old page never mentioned: generation still runs and is compared before being discarded, so unlike `--load-debug-rollout-data` the engines stay alive and `--ci-inject-rollout-data-min-match-ratio` (default 0.9) catches wrong engine weights.
- **Cutting one step into pieces**: `--debug-skip-weight-update` (keeps the offload schedule, drops the sync), `--debug-disable-optimizer`, `--debug-exit-after-rollout`.
- **`--debug-deterministic-collective`**, with what it actually does: the `det_nccl` backend folds order-sensitive SUM/AVG reductions in a fixed tree order, which is what makes two different reduction topologies bitwise comparable.
- **The `--ci-test` assertion harness**, which is why an e2e test is mostly a normal run with flags. Each checker is documented with its real thresholds and its escape hatch: the KL checker (`1e-9`, `1e-8` for MLA, relaxed for LoRA), the log-probs checker (`1e-8`, `5e-3` under R3, `0.03` trainer-versus-engine, entropy in `(0, 0.7)`, exact equality under true-on-policy), the weight-update checker, `--ci-metric-checker-key/threshold` (passes if any eval clears the bar, verdict at dispose), `--ci-save/load-grad-norm` (`{role}`/`{rollout_id}`/`{step_id}` templates, tolerance 0.03, rank 0 only), `--ci-save/check-model-hash` (per-layer SHA256, one JSON per rank), and `--ci-ft-test-actions`.

The precision-alignment and kernel-pitfall triage was kept, with the log section rebuilt around the metric keys the checkers actually assert on.

## `developer/versions.md`: new page

The three trees (miles, the `sglang-miles` branch of `sgl-project/sglang`, the `miles-main` branch of `radixark/Megatron-LM`), why they are branches rather than releases, and why all three being editable installs is what lets CI move them without reinstalling.

Then: where every pin is written (Dockerfile build-args with their current defaults, `build.py`'s variant table, `requirements.txt`, the inline commit pins, the TE `2.17.0` triplet assertion and the `docker/patch/cu13/` patches that fail the build if they do not apply), the published images and tags, and the part that decides whether a change needs a rebuild:

| Change | New image? |
|---|---|
| `requirements.txt` | No, the next CI run installs it |
| A Dockerfile layer | Yes |
| SGLang / Megatron-LM code | No, point CI at a ref |

The **bumping principle** is stated as rules with the evidence behind each: bump where the pin lives exactly once; prefer moving the branch over pinning a commit (`SGLANG_COMMIT` is empty by default); validate an image-affecting bump on the PR that makes it (the docker path filter builds `pr-<n>` and runs the GPU matrix inside it, and that outranks `ci-image-tag:`); validate a ref bump with the PR-description directives instead of a rebuild; expect `dev` to move on its own within a bound (`check-upstream` polls sglang / Megatron / the wheels fingerprint, deliberately not miles, and forces a build at 24h, so `dev` trails `main` by at most a day); pin a timestamped tag when you need it to stop moving; bump the ROCm images by hand because they have no automatic path.

Closing table of post-bump failure modes: the cuDNN downgrade that shows up as `CUDNN_STATUS_BAD_PARAM` in fused-attention backward (and why CI restores the image's pin after `pip install -r requirements.txt`), a TE patch that no longer applies, the TE triplet assertion, and Mooncake's structured object store being cu13-only.

## `developer/architecture.md`: file structure fixed

The package tree said `backends/experimental/fsdp_utils/  # FSDP-flavoured trainer (in progress)`, which has not been true since #2272. Regenerated, and the sibling `miles_plugins/` tree is now described. "Where common changes go" gains an FSDP row and points weight-sync changes at the right directory per backend; "Where to look first" pointed at `miles/utils/distributed_utils.py` for weight sync, which holds process-group helpers, so it now points at `megatron_utils/update_weight/`. The tests tree gains `manual/` and `snapshots/`.

## `developer/migration.md`: deleted

It documented the sync-to-async `train.py` migration, which is long done, and "breaking flag changes between releases" that stop at that one entry. Removed from the nav and from the Developer Guide cards.

## Notes for the reviewer

- The published Developer Guide and `docs/ci/contributor-guide.md` now overlap on the CI-selection material. I left `docs/ci/*` untouched: it is `doc-dev`-governed and owned by CI work, and the new page links to it for the long form. Folding the two is a reasonable follow-up.
- No redirect was added for `/developer/migration`. #2375 introduces the `redirects` block in `docs.json`; adding a second one here would conflict for no benefit. Worth adding `/developer/migration` → `/developer/index` once both land.
- This PR and #2375 both touch `docs.json` and `docs/developer/index.md`, and #2375 also touches `docs/developer/architecture.md`, so whichever lands second needs a small conflict resolution.


docs/developer/architecture.md
```diff
@@ -43,24 +43,31 @@ flowchart TB
...
 ├── router/               # FastAPI proxy + worker load-balancer (router.py)
-└── utils/                # async, types, IO, distributed helpers, arguments.py
+├── dashboard/            # run dashboard: collector, backend, dump reader
+├── true_on_policy/       # true-on-policy contracts and per-model profiles
+└── utils/                # arguments.py, async / IO / distributed helpers, audit_utils/
 ```
 
-`train.py` and `train_async.py` are the two entry points. They're thin: ~200 lines
-each. Most logic lives in the modules above.
+The `miles_plugins/` tree sits beside it and is only ever reached by name from a flag:
```
Shi-Dong: I'm not sure what "reached by name from a flag" means.

docs/developer/debug.md
```diff
@@ -1,132 +1,198 @@
 ---
 title: Debugging
-description: Aligning precision, separate train/rollout debugging, common kernel pitfalls.
+description: Isolating rollout from training, the debug and CI assertion flags, aligning precision, and reading logs.
```
Shi-Dong: I think this description is overly detailed.  Probably just replace it with "Useful tips for debugging".

docs/developer/debug.md
```diff
@@ -1,132 +1,198 @@
...
+flags on this page exist to answer it by cutting the loop into pieces you can run one at a
+time. Once you know which side is wrong, it becomes an ordinary debugging session.
 
-## Aligning precision
+## Cut the loop in half
 
-The single most common class of bug. Walk through these checks before anything else.
+| Flag | What it does |
+|---|---|
+| `--debug-rollout-only` | Run generation only. Megatron is never initialized. |
```
Shi-Dong: Add FSDP alongside Megatron?

docs/developer/versions.md
```diff
@@ -0,0 +1,179 @@
+---
+title: Versions and Images
+description: How the miles, SGLang and Megatron-LM trees fit together, what the Docker images pin, and the principle for bumping any of them.
```
Shi-Dong: "Miles, SGLang, and Megatron-LM"

docs/developer/versions.md
```diff
@@ -0,0 +1,179 @@
...
+---
+A Miles run is three Python source trees on one `PYTHONPATH`. Understanding which tree owns
+a given behavior, and which file pins its version, is most of what you need to debug a
+version problem or land a bump.
+
+## The three trees
+
+| Tree | Comes from | Installed as | Lives at |
+|---|---|---|---|
+| miles | this repository | `pip install -e . --no-deps` | `/root/miles` |
```
Shi-Dong: miles -> Miles

Shi-Dong (approved):
LGTM!

PR #2379, opened 2026-08-11
docs: point the TB2 agent-server eval at the public harbor branch

nblintao:
`examples/experimental/eval/terminal_bench_via_agent_server` told readers to run `miles_agent_server` from a branch of `radixark/harbor-private` — a repo nobody outside the org can clone, in a public example. The Miles integration is on `harbor-miles-v0.20.0` in the public `harbor-framework/harbor`, which is the branch [`examples/swe-agent-harbor-docker`](https://github.com/radixark/miles/tree/main/examples/swe-agent-harbor-docker) and `swe-agent-harbor-daytona` already point at, so this example now points there too.

Checked against that branch before swapping the pointer — everything the example leans on is present:

- `miles_agent_server.py` at the repo root, with the `--port` / `--dashboard-port` / `--max-concurrent` / `--trials-dir` flags the setup snippet passes;
- the `poll_steps()` / `max_seq_len` truncation wrapper the `--max-seq-len 65536` note describes;
- `/api/sessions` in the dashboard app, referenced by the tuning notes;
- a root `registry.json` whose `terminal-bench` 2.0 entry holds exactly 89 tasks, in the `[{name, tasks: [{name}]}]` shape `load_task_names()` parses — so `--registry-path` works verbatim against it;
- `harbor datasets download ... --export`, used to populate `HARBOR_TASKS_DIR`.

One thing this does not change: the DeepSeek-V4-Pro numbers in the README were measured against the older private branch. The commands are compatible with `harbor-miles-v0.20.0`, but the run itself was not repeated on it.

Docs and docstrings only.

### DIFF
--- a/examples/experimental/eval/terminal_bench_via_agent_server/README.md
+++ b/examples/experimental/eval/terminal_bench_via_agent_server/README.md
@@ -16,7 +16,7 @@ training-loop rollouts interact with the same backend.
 ## Prerequisites
 
 1. **A running `miles_agent_server`** from the
-   [`shi/rebase-on-upstream-v0.7.0` branch of harbor-private][branch], with
+   [`harbor-miles-v0.20.0` branch of `harbor-framework/harbor`][branch], with
    the server's `$HARBOR_TASKS_DIR` populated with the 89 Terminal-Bench 2.0
    tasks. The server's README explains setup; the short version is:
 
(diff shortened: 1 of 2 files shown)


Shi-Dong (approved):
LGTM, thanks!

PR #2386, opened 2026-08-11
fix: drop context parallelism from the FSDP backend

Zhichenzzz:
## What

The FSDP backend has never supported context parallelism. `miles/utils/arguments.py:2663` asserts `context_parallel_size == 1` on the FSDP path, so every CP branch underneath it was unreachable. One of them had already rotted: `_get_model_inputs_args` dereferenced a `self.cp_group` that is never assigned anywhere, so it would have raised `AttributeError` if it had ever been entered.

This removes the machinery.

- `build_fsdp_meshes` builds a 1-D `dp` mesh instead of a `(dp, cp)` mesh, and no longer returns the `dp_cp` and `cp` views that nobody read.
- `create_fsdp_parallel_state` drops the `cp_rank` / `cp_size` arithmetic and the conditional `ring_flash_attn.substitute_hf_flash_attn` import.
- `_get_model_inputs_args` drops the sequence-chunking branch.
- `validate_hybrid_shard_args` drops the CP divisibility checks, since `data_parallel_size` is now just `world_size`.

Two things deliberately stay:

- **`cp.group` is still a real single-rank group, not `None`.** Shared code passes it straight to `dist.nn.all_reduce` (`training_utils/cp_utils.py:390`), and `group=None` there means WORLD, which would silently turn a no-op into a cross-rank reduction. `cp` and `tp` now share one `dist.new_group([rank])`, which is one fewer process group than before.
- **`context_parallel_size` stays on `FSDPArgs`.** Shared validation reads `args.context_parallel_size` unconditionally, and keeping the flag is what lets the existing assert reject a context-parallel FSDP run with a clear message instead of `unrecognized arguments`.

No behavior change on any reachable path.

## Verification

On 4x H200:

- `tests/fast-gpu/test_fsdp_hybrid_shard.py` passes, so FSDP2 gradients still match across `r1s4`, `r2s2` and `r4s1`. This is the real guard on the mesh rewrite, since it also covers the `_unflatten` call now made on a root mesh rather than on a slice.
- A `ParallelState` probe confirms `meshes == {dp, fsdp}`, the `dp` group spans the world, `fsdp` has the expected `ndim` for `dp_replicate_size` 1 and 2, and an `all_reduce` on `cp.group` leaves the tensor untouched.
- Two rollouts of `scripts/run_qwen3_0_6b_fsdp.py` (Qwen3-0.6B, dapo-math-17k, 4 GPUs, colocated) train end to end with `grad_norm` 0.135 and 0.130, `raw_reward` 0.113, and `train_rollout_logprob_abs_diff` around 0.018.


Rockdu (approved):
LGTM

PR #2387, opened 2026-08-11
docs: rewrite the launch script page around the python launchers

yueming-yuan:
Docs follow-up to the `scripts/` cleanup stack. **Base is `yueming/script-rewrite` (#2356), review that one first.**

#2356 left this as an explicit follow-up: `training-script-walkthrough.md` was framed around the bash `XXX_ARGS` arrays, which after that PR no longer exist anywhere under `scripts/`. This rewrites the page around the python launchers.

## The page

Renamed `user-guide/training-script-walkthrough` → `user-guide/launch-script`, and grouped in the nav with Argument Groups under a "Launch Script" group, ahead of Argument Groups rather than after it — a reader needs to know what a launcher does before a page about where a flag belongs is useful.

Section flow, all of it keyed off `scripts/run_qwen3_dense.py` as the reference launcher:

| Section | Covers |
|---|---|
| How a launch script starts a training job | The four layers — launcher, `scripts/models/`, `command_utils`, `train.py` — and that the launcher submits a Ray job rather than training |
| The structure of a launch script | `_RECIPES` and `--model-name`; `ScriptArgs`; `execute()`'s flag blocks |
| Three ways to override a recipe | `--extra-args`, a script option (flag or env var), editing the recipe |
| What execute_train runs on your machine | The `EXEC:` sequence in order, and the two env vars that skip parts of it |
| Multi-step and multi-node launchers | `prepare-*` / `full-train`; per-role subcommands and ssh fan-out |
| Model architecture definitions in scripts/models/ | Resolution by file name, `load_sibling_model_args`, the config.json warning |

## What is new rather than moved

`ScriptArgs` had no user-facing documentation. The page now states the mechanism `@U.dataclass_cli` implements — every field is both a `--kebab-case` option and a `MILES_SCRIPT_<FIELD>` env var, CLI beating env beating default — and tables the options every launcher inherits from `ExecuteTrainConfig`. Same for `MILES_SCRIPT_EXTERNAL_RAY` and `MILES_SCRIPT_ENABLE_RAY_SUBMIT`, which until now were only discoverable by reading `command_utils.py`.

## What was dropped from the page

Everything the old page carried beyond the launcher itself: the eight per-group flag walkthroughs (Argument Groups owns these), sync vs async rollout, colocation, dynamic sampling, partial rollout, and BF16 train + FP8 inference. Those belong on `fully-async`, `usage`, `customization` and `advanced/fp8-low-precision` respectively; **this PR deletes them from this page without relocating them**, so that relocation is a follow-up and is the one thing worth pushing back on here.

Two links pointed into anchors on those deleted sections and are re-pointed as a stopgap: `rollout-endpoints.md` → `cli-reference#async--partial-rollout`, and `cli-reference.md` → `usage`. They will need updating again once the content lands in its new home.

## Also here

- Index and cross-reference updates so no page still links to the old slug: `docs.json`, `docs/index.md`, `user-guide/index.md`, `user-guide/concepts.md`, `user-guide/usage.md`, `getting-started/quick-start.md`.
- No redirect was added for the old URL — `docs.json` has no `redirects` block today. Say the word if the published site needs one.

## Verification

Every claim was read out of this branch's code: `command_utils.execute_train` for the `EXEC:` sequence and the runtime env, `dataclass_cli` for the flag/env-var mechanism, `run_qwen3_dense.py` for the snippets, `run_deepseek_v4.py` / `run_nemotron_3_super_120b_a12b.py` / `run_qwen3_sft.py` for the launcher shapes. The Argument Groups anchors were checked against the ids in the rendered HTML, not guessed. Rendered locally with `mint dev`.

## Known gap

`argument-groups.md`, now the second page in this group, still opens with "Miles launch scripts are bash arrays" and keys its eight sections to `MODEL_ARGS` / `PERF_ARGS` / etc. That is stale on this branch for `scripts/` and needs its own pass — not in this PR.

A flow diagram placeholder is left in the page where the launcher → ray → `train.py` path is described; I did not fabricate one.


docs/user-guide/cli-reference.md
```diff
@@ -27,8 +27,7 @@ This page has two passes.
 | `--rollout-num-gpus-per-engine` | `1` | TP size of each SGLang engine. |
 | `--colocate` | off | Share GPUs between actor and rollout. |
 
-See [Training Script Walkthrough: Colocation](/user-guide/training-script-walkthrough#colocation-share-gpus-or-dont)
-for what `--colocate` flips on under the hood.
+See [Training Backend](/user-guide/usage) for what `--colocate` flips on under the hood.
```
Zhichenzzz, line 30: it could be merged first, and then fix it under https://github.com/radixark/miles/pull/2375

Zhichenzzz (approved):

PR #2388, opened 2026-08-11
fix(fsdp): make --config actually apply, and reject keys it does not know

Zhichenzzz:
## The problem

`--config <yaml>` is documented as the FSDP backend's config mechanism (`docs/developer/experimental-features.md:59`). It did not work, in both directions at once.

`load_fsdp_args` applied a config entry only `if not hasattr(args, k)`. Since the parser registers **every** `FSDPArgs` field with argparse, that condition is never true for a declared field. Measured on `main`:

```
yaml: lr: 5e-05                    ->  args.lr = 2e-05                     ignored
yaml: gradient_checkpointing: true ->  args.gradient_checkpointing = False ignored
yaml: brand_new_key: 42            ->  args.brand_new_key = 42             accepted
```

So the config file silently dropped every key it was supposed to honor, and silently absorbed every key it was supposed to reject. A misspelled `weigth_decay` became a new attribute nobody reads, and the run continued with the default.

## Why the one-character fix is wrong

Dropping the `not` inverts one failure into another. Argparse cannot distinguish "the user typed `--lr`" from "this is the default", so an unconditional `setattr` would let the config file override an explicit command-line flag.

## The change

Split parser construction out of `parse_fsdp_cli` into `build_fsdp_parser`, then feed the YAML entries back as defaults and reparse:

```python
parser = build_fsdp_parser(extra_args_provider)
args = parser.parse_args()
if args.config:
    reject_unknown_config_keys(data, set(vars(args)))
    parser.set_defaults(**data)
    args = parser.parse_args()
```

Precedence becomes **CLI > YAML > dataclass**. Explicit flags are still genuinely parsed on the second pass, so they win.

Unknown keys now raise, with a did-you-mean from `difflib`:

```
ValueError: unknown key(s) in the YAML config: 'weigth_decay' (did you mean 'weight_decay'?)
```

The known-key set is taken from `vars(args)` after the first parse, so it is exactly one entry per registered action. That covers whatever `extra_args_provider` contributed, not just `FSDPArgs`, which matters because the real RL entrypoint passes `add_miles_arguments`.

`parse_fsdp_cli` keeps its name, signature and behavior; its two test callers are untouched.

## Breaking change worth a second look

Unknown keys used to land on the namespace. Since overrides never worked for known keys, that passthrough was the **only** thing `--config` ever actually did — so anyone using `--config` today is using precisely the path this now rejects.

Nothing in the tree passes `--config`, and the docs describe it only as a table entry. But a config carried outside the repo will now fail loudly instead of being half-ignored. That is the intent: a key the program does not know is a typo, not a feature. A genuine plugin knob should be registered through `extra_args_provider`.

## Testing

`tests/fast/backends/test_fsdp_config_precedence.py` (new) pins every tier and the rejection:

- config beats the dataclass default, for a scalar and for bools in both directions
- an explicit CLI flag beats the config
- an unknown key raises; a misspelled key names the field it almost matched
- a valid config is not tripped by the check, fields the config omits keep their defaults, and a run with no `--config` is unaffected

Two of these fail against the old logic, so they are load-bearing rather than decorative.

Verified on an H200 box against three trees, since this touches the same function as #2384: today's `main` (467 passed for all of `tests/fast/backends/`), `main` with the old YAML logic restored (the two precedence tests go red as expected), and rebased on top of #2384. The two PRs touch different parts of `build_fsdp_parser` and merge cleanly in either order.

### DIFF
--- a/miles/backends/fsdp_utils/arguments.py
+++ b/miles/backends/fsdp_utils/arguments.py
@@ -1,5 +1,6 @@
 import argparse
 import dataclasses
+import difflib
 from dataclasses import dataclass
 
 import yaml
@@ -64,7 +65,7 @@ class FSDPArgs:
     config: str | None = None
 
 
-def parse_fsdp_cli(extra_args_provider=None):
+def build_fsdp_parser(extra_args_provider=None) -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser("FSDP SFT Training (miles)")
     parser.add_argument("--config", type=str, default=None, help="YAML config path")
     for f in dataclasses.fields(FSDPArgs):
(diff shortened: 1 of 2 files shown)


Rockdu (approved):
LGTM

PR #2397, opened 2026-08-11
fix(fsdp): move the AMD Triton attention bridge in-tree

Arist12:
## Problem

`import miles.backends.fsdp_utils` fails on every published ROCm image, so the FSDP backend is unusable there. On an unmodified `rocm/sgl-dev:miles-rocm720-mi35x-20260811`:

```console
$ python3 -c "import miles.backends.fsdp_utils"
  File "/root/miles/miles/backends/fsdp_utils/actor.py", line 227
    def _build_model_with_attn_bridge(self, checkpoint_path: str, init_context):
IndentationError: expected an indented block after 'if' statement on line 226
```

`docker/amd_patch/latest/miles.patch` has landed in the wrong place — its new method sits between `if native_cls_name is not None:` and the body it belongs to, in the middle of `get_model_cls`:

```python
            native_cls_name = MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.get(getattr(self.hf_config, "model_type", ""))
            if native_cls_name is not None:
    def _build_model_with_attn_bridge(self, checkpoint_path: str, init_context):
```

## Cause

The patch has **zero context lines**. Its two replacement hunks still anchor on the lines they delete and land correctly, but the hunk that inserts the helper (`@@ -229,0 +227,22 @@`) has nothing to match on and is placed by line number alone — and `actor.py` has moved since the patch was generated. `git apply` normally refuses a zero-context diff for exactly this reason, but the build passes `--unidiff-zero` to switch that check off, and the `--3way` fallback cannot engage because the pre-image blob named by the patch is not in the repo:

```console
$ git apply --3way --unidiff-zero docker/amd_patch/latest/miles.patch; echo "exit=$?"
error: repository lacks the necessary blob to perform 3-way merge.
Falling back to direct application...
exit=0
```

The build step cannot fail, whatever the hunks hit. It stayed hidden because ROCm CI installs the PR checkout over the image (`_run-ci-rocm.yml`: `cd $GITHUB_WORKSPACE && pip install -e .`), so it never imports the patched `/root/miles` that users get.

## Fix

Regenerating the patch would fix today and drift again the next time `actor.py` moves. This removes the overlay instead, which is what `docker/Dockerfile.rocm` already asks for — *"TODO: remove these patches once the changes are merged into the main codebase."*

- `actor.py` gains `_build_model_with_attn_bridge` and its two call sites. This is the patch's own code, with one dead assignment (`self._use_triton_bridge`, read nowhere) dropped and a `getattr` that guarded nothing tightened to a plain attribute access.
- `sglang_attn_bridge/` moves from `docker/amd_patch/latest/` into `miles/backends/fsdp_utils/`, the package that imports it — pure rename, no content change, so the move stays reviewable.
- `docker/amd_patch/latest/miles.patch` and its Dockerfile steps are deleted, and with them the only use of `--unidiff-zero` in the repo. This was also the only zero-context patch under `docker/`.
- The build now imports both backends after installing miles, mirroring the existing `import sglang; import sgl_kernel` check. That is what closes the class rather than this one instance: on the broken tree `git apply` exits 0 and this check exits 1, so the image would have failed to build the day this landed.

**Nothing changes off ROCm.** The bridge is gated on `torch.version.hip`, following `miles/ray/train_actor.py`, so elsewhere `attn_implementation="triton"` reaches `from_pretrained` and is rejected there exactly as before — this code was previously only ever present inside the ROCm image. Every other value takes the same path on every platform, and the bridge is imported lazily, so nothing new loads unless ROCm asks for it.

A CPU-only test pins all three branches. It would not have caught the original failure — that was a build step — but this code has had no coverage until now.

## Scope

This restores `import miles.backends.fsdp_utils`, and with it the FSDP backend on ROCm for `flash_attention_2`, `sdpa` and `eager`. It does **not** make `--attn-implementation triton` work end to end: the bridge's forward calls `sglang.srt.layers.attention.triton_ops.extend_attention`, which sglang moved to `sglang.kernels.ops.attention.extend_attention` in #30044 and whose signature gained `k_scale`/`v_scale` before `qo_indptr`. That path is already broken on the pinned `sglang-miles` branch, independently of this change, and fixing it needs a real forward/backward to validate. Happy to follow up separately.

## Verified

On a ROCm container: `pre-commit` clean, the new tests pass, and `miles.backends.fsdp_utils` imports.


XinyuJiangCMU:
Looks good to me. The true on-policy path can be handled in a follow-up PR.

Arist12:
> Looks good to me. The true on-policy path can be handled in a follow-up PR.

Thanks! @Zhichenzzz Could you help review this PR when you got a time?

Zhichenzzz (approved):

PR #2404, opened 2026-08-12
Install the kubernetes toolchain the charts and the k8s provider need

fzyzcjy:
Nothing on `main` uses these yet; they land ahead of the kubernetes work so the image is ready when it does.

## What each one costs

Measured on `radixark/miles:dev` (h200 devbox, linux/amd64). "Pulled" is what a `docker pull` actually transfers, "on disk" is what the layer occupies once unpacked.

| Addition | Pulled | On disk | Notes |
| --- | --- | --- | --- |
| `kubectl` v1.36.3 | 16.9 MiB | 56.8 MiB | single static Go binary |
| `helm` v4.2.3 | 18.9 MiB | 61.3 MiB | single static Go binary |
| `kubernetes_asyncio` 36.1.0 | 2.9 MiB (wheel) | 26 MiB, plus ~21 MiB of `__pycache__` generated at install | pure python, mostly generated api clients |
| **total** | **38.7 MiB** | **~145–165 MiB** | |

`radixark/miles:dev` is currently **19.12 GB compressed**, so the pull grows by **0.2%**. The two binaries land in one layer each Dockerfile adds once, and the script deletes its tarballs inside the same `RUN`, so nothing but the binaries is committed.

## Downsides, such as they are

- `kubernetes_asyncio` pulls in **no new transitive dependency**: `aiohttp`, `certifi`, `python-dateutil`, `pyyaml`, `six` and `urllib3` are all already in the image (verified against the installed set).
- Two more build-time downloads (`dl.k8s.io`, `get.helm.sh`). Both are pinned and sha256-verified, so a corrupted or moved artifact fails the build rather than shipping silently.
- Two more versions to keep current. They sit in one script both Dockerfiles call, so a bump is one edit rather than two.
- Nothing here is imported or executed on any training path: the binaries are only run by the launcher and the chart tests, and the wheel is only imported by the k8s worker provider.

## Why it fails quietly today

- Without `kubernetes_asyncio`, the modules importing it fail at collection.
- Without `helm`, the chart tests **skip themselves** — the run reports success while covering nothing.
- Without `kubectl`, the launcher has nothing to drive inside the pod.

### DIFF
--- a/.github/workflows/_run-cpu-ci.yml
+++ b/.github/workflows/_run-cpu-ci.yml
@@ -94,6 +94,11 @@ jobs:
           uv pip install -r requirements.txt -r tests/ci/requirements-ci-cpu.txt
           python tests/ci/verify_source_resolution.py
 
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v4.2.3
+
       - name: Resolve suite plan
         shell: bash
         run: ${{ inputs.execute_command }} --list-only
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -6,6 +6,7 @@ on:
       - main
     paths:
       - 'docker/Dockerfile'
+      - 'docker/install-kube-tools.sh'
       - 'docker/verify_transformer_engine.py'
       - 'requirements.txt'
 
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -180,8 +180,11 @@ jobs:
 
       - name: Install Python + dependencies
         run: |
-          apt-get update && apt-get install -y python3 python3-pip
-          pip3 install --break-system-packages typer
+          if ! command -v python3 >/dev/null || ! command -v pip3 >/dev/null; then
+            apt-get -o DPkg::Lock::Timeout=600 update
+            apt-get -o DPkg::Lock::Timeout=600 install -y python3 python3-pip
+          fi
+          pip3 install --break-system-packages --ignore-installed typer
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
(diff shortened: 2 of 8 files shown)


guapisolo (approved):

fzyzcjy:
Smoke-tested the image this PR's own `docker-build` produced — `radixark/miles:pr-2404` (19.17 GB, amd64 `sha256:1bcfa38c…`) — on a fresh 1×H200 devbox, nothing installed on top of it.

**The three additions are there, at the pinned versions**

```
$ kubectl version --client=true
Client Version: v1.36.3
Kustomize Version: v5.8.1
$ helm version --short
v4.2.3+g43e8b7f
$ python -c "import kubernetes_asyncio; print(kubernetes_asyncio.__version__)"
36.1.0
$ which kubectl helm
/usr/local/bin/kubectl
/usr/local/bin/helm
```

**The stack they sit next to is untouched**

```
torch 2.11.0+cu130   cuda True   NVIDIA H200
sglang 0.5.17.dev36+gcb05a44
ray 2.57.0
megatron-core 0.16.0rc0
```

A bf16 4096² matmul on the gpu returns.

**miles itself runs.** The checkout baked into the image (`6a6843f1d`) passes its own fast tests:

```
$ python -m pytest tests/fast/utils tests/fast/rollout -q
3322 passed, 28 skipped in 364.90s
```

And it serves a real model — `sglang.launch_server` on Qwen2.5-0.5B-Instruct came up in 40s and answered:

```
$ curl -s :30000/generate -d '{"text": "The capital of France is", "sampling_params": {"max_new_tokens": 16, "temperature": 0}}'
{"text":" Paris. It is the largest city in Europe and the second largest city in the", ...}
```

**No dependency was disturbed.** `pip check` on this image names 5 pre-existing conflicts (protobuf, cudnn, pillow), none of them `kubernetes_asyncio` or any of its requirements — the released `radixark/miles:dev` reports more than this image does. `kubernetes_asyncio`'s six requirements (`aiohttp`, `certifi`, `python-dateutil`, `pyyaml`, `six`, `urllib3`) were all already in the image, so pip installed the one wheel and nothing else.

PR #2347, opened 2026-08-10
[AMD] Enable amd pr ci

JessicaJiang-123:
Co-author-with: @XinyuJiangCMU, @yuankaichen-amd, @Xinyu-Kang, @indianspeedster, @sreerohi, @lizamd 

## Summary

Bring up the ROCm CI lane on MI300X/MI355X and enable the cases, without changing shared NVIDIA behavior. ROCm-specific recipes live under `scripts/amd/`. Tests select them through `MILES_PLATFORM=rocm`, while smaller ROCm-only differences are passed through backend-neutral `extra_args` / `extra_env_vars`.

## Test status
| case | status | blocked on |
|---|---|---|
| `test_amd_mtp1_spec_v2_r3` | disabled | — |
| `test_inkling_small_4layer_ci` | disabled | — |
| `test_qwen3_4B_offload_disk_stream` | disabled | — |
| `test_deepseek_v4_flash_4layer_ci` | disabled | sgl-project/sglang#34330 |
| `test_qwen3_4B_fsdp_hybrid_shard_r2s2` | disabled | radixark/miles#2265 |
| `test_amd_moriep_fp8_bridge` | disabled | sgl-project/sglang#34006, sgl-project/sglang#34203, sgl-project/sglang#34220 |
| `test_glm5_2_744b_a40b_5layer_ci` | disabled | sgl-project/sglang#34223 |
| `test_glm47_flash/test_amd_r3_mtp` | disabled | ROCm/aiter#4653, sgl-project/sglang#34223, sgl-project/sglang#34238 |

## How AMD coverage is added

ROCm-specific behavior is selected through `MILES_PLATFORM=rocm`, set by `docker/Dockerfile.rocm`. Shared NVIDIA paths do not contain ROCm-specific branches.

There are three patterns:
### 1. Different recipe → use `scripts/amd/`
If ROCm needs a different recipe, the test dispatches to an AMD-specific script:
```python
if os.getenv("MILES_PLATFORM") == "rocm":
    from scripts.amd.run_inkling import _MODEL_REGISTRY, ScriptArgs, _train
else:
    from scripts.run_inkling import _MODEL_REGISTRY, ScriptArgs, _train
```
### 2. Same recipe, extra flags → use `CaseConfig`
If ROCm only needs extra arguments or environment variables, pass them through backend-neutral fields:
```python
extra_args: str = ""
extra_env_vars: dict[str, str] = field(default_factory=dict)
```
### 3. AMD-only case → keep it in the AMD test
[...]

PR #2407, opened 2026-08-12
[DO NOT MERGE] CI 1

fzyzcjy:
CI-only PR for the refactor chain up to deliver-1, rebased onto upstream main `8d60ec7dbc` (`deliver-1-base` now points at that commit). Fix commits are folded into the owning ops.

The chain no longer touches `tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py` at all: upstream #2873 landed the same Anthropic session configuration, so the four lines that were folded into op position 79 are dropped and that file is byte-identical to base. The rebase onto the new base replayed with zero conflicts, and the chain's whole base-to-tip delta is byte-identical before and after the base change.

ci-sglang-pr: bef37b5c21b1c370ff8a3395475061bba4f0d490

Megatron stays unpinned: `miles-main` tracks the branch the old pin `235952df6` came from, and CI resolves it per run.

SGLang has to stay pinned, and neither existing ref works on its own. `sglang-miles` (`c16b821`) has the `register_lora_adapter` RPC that upstream main's #2756 LoRA weight-update rewrite calls, but its CLI allowlist still rejects `out_proj`, which `scripts/run_qwen3_5_35b_a3b_lora.py:61` passes as a LoRA target, so `test_qwen3_5_35b_a3b_lora_ci.py` dies in argparse and hangs to the 1800s timeout. The older `d4b1974` accepts `out_proj` but predates the register RPC, so every LoRA lane then fails with `404 Not Found` for `/register_lora_adapter`. The pin above is `sglang-miles` plus one commit that adds `out_proj` back to the CLI allowlist, which the LoRA runtime already supports end to end (sgl-project/sglang branch `tom/allow-gdn-out-proj-lora-target`). Historical old-SGLang and old-Megatron compatibility code has been removed from the Miles branch.


fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** I downloaded the completed CPU shard's full log, identified the first root failure, and classified it before taking any rerun action. Please push back if any conclusion is off.

The [`stage-a-cpu (1) / run-cpu` job](https://github.com/radixark/miles/actions/runs/31558060297/job/93996695890) failed with 97 cascading tokenizer-test failures after Hugging Face returned `HTTP 429 Too Many Requests` for the Qwen3-4B Xet token endpoint. The first root fingerprint is:

```text
ConnectionError: Network error: Request error: HTTP status client error (429 Too Many Requests), domain: https://huggingface.co/api/models/Qwen/Qwen3-4B/xet-read-token/...
```

This is an external download-rate-limit failure rather than a Miles assertion failure, as shown by the [same job log](https://github.com/radixark/miles/actions/runs/31558060297/job/93996695890). GPU siblings are already active in the [PR checks](https://github.com/radixark/miles/pull/2407/checks), so I will leave them undisturbed and rerun the failed job after the workflow settles.

fzyzcjy:
> CI triage update from Codex (AI agent), based on the complete logs of every completed GPU failure below.

The completed GPU failures on head `8021c9073f` are one code bug, not independent test regressions:

- [8x H200 shard 0](https://github.com/radixark/miles/actions/runs/31558060297/job/93998603055)
- [8x H200 shard 1](https://github.com/radixark/miles/actions/runs/31558060297/job/93998603115)
- [8x H100 shard 0](https://github.com/radixark/miles/actions/runs/31558060297/job/93998603220)
- [2x H200 shard 1](https://github.com/radixark/miles/actions/runs/31558060297/job/93998603099)

All four logs have the same primary fingerprint:

```
AssertionError: cli argv roundtrip mismatch on device: parsed 'None' != wanted 'cuda'
```

Classification: **our code bug, fixed on the new head**.

Root cause: `_compute_server_args` preserves argparse's raw `device=None`. `ServerArgs` then auto-detects and resolves that value to `cuda`, but the `always_render_fields` path preferred the present raw key even when its value was `None`. It consequently emitted `--device None`, and the next parse produced the literal string `"None"` rather than the resolved accelerator.

Fix: commits [`0778e163a5`](https://github.com/radixark/miles/commit/0778e163a5) and [`a58b3f6c39`](https://github.com/radixark/miles/commit/a58b3f6c39) make raw `None` mean unspecified for always-rendered fields and fall back to the constructor-resolved value. Both `tom/refactor-miles/deliver-1` and `tom/refactor-miles/op13-142` now point to `a58b3f6c39`; `tom/refactor-miles/deliver-1-base` remains at the chain root.

Pre-push validation on a 1x H200 devbox:

- Regression proof: the two new tests failed before the source fix with the exact fingerprint above, then passed after it.
- Relevant argv/ServerArgs/RouterArgs fast tests: 624 collected, successful.
- Full fast suite on the absorbed chain: 7,011 passed, 30 skipped, 4 xfailed.
- `pre-commit run --all-files`: passed.
- Post-fix gate: no test weakening; no documentation contract became stale.

The new final-head CI is now running on `a58b3f6c390e176d17510a39afa2e2ff032c90f5`. I will continue tracking it and triage any new failure independently.

fzyzcjy:
CI audit update (AI-assisted by Codex)

- Final Miles head: `a58b3f6c390e176d17510a39afa2e2ff032c90f5`.
- One duplicate same-SHA CPU check failed in the check suite triggered by stacked PR #2176 (`tom/refactor-miles/op13-142`): [`stage-a-cpu (1) / run-cpu`](https://github.com/radixark/miles/actions/runs/31561390491/job/94007887353).
- Failure fingerprint: `AttributeError: 'ServerArgs' object has no attribute 'gated_launch_port'` in `test_the_command_launches_sglang_with_the_allocated_addressing`.
- Classification: dependency-context mismatch, not a Miles code regression. The #2176 event body had no SGLang override and resolved `sglang=sglang-miles`; the equivalent same-SHA check under this PR used `tom/refactor-miles-repo-sglang/deliver-1` and passed: [`stage-a-cpu (1) / run-cpu`](https://github.com/radixark/miles/actions/runs/31561390573/job/94007975298).
- Corrective action: added the same `ci-sglang-pr` pin to #2176 and dispatched a targeted `run-unit-test` run ([PR Test](https://github.com/radixark/miles/actions/runs/31563479231)); the temporary label was removed after dispatch. No code change was made for this dependency-only failure.
- Already green on the final SHA: both pre-commit runs, both complete CPU suites except the superseded unpinned duplicate above, and all four MI300X jobs. The authoritative #2407 H100/H200 GPU jobs are still queued/running and remain under observation.

fzyzcjy:
Final CI audit (AI-assisted by Codex)

- Final head: `a58b3f6c390e176d17510a39afa2e2ff032c90f5`.
- Authoritative #2407 checks are complete and green: pre-commit, all CPU shards, all ROCm MI300X shards, and all H100/H200 GPU shards in [PR Test run 31561390573](https://github.com/radixark/miles/actions/runs/31561390573).
- The earlier duplicate #2176 CPU failure was a superseded unpinned-SGLang context. After adding the same `ci-sglang-pr` pin, targeted PR Test run [31563479231](https://github.com/radixark/miles/actions/runs/31563479231) and ROCm run [31563479644](https://github.com/radixark/miles/actions/runs/31563479644) completed successfully, including the previously failing CPU shard.
- GitHub now reports `mergeStateStatus=CLEAN`; there are no queued or running checks.
- Final refs verified: `deliver-1` = `op13-142` = `a58b3f6c390e176d17510a39afa2e2ff032c90f5`; `deliver-1-base` = chain root `d2e3ad9f197d86c138347063fdd9750dbd3d296e`.
- All temporary validation devboxes were unbound and released.

fzyzcjy:
https://github.com/radixark/miles/actions/runs/31583550632

fzyzcjy:
Final CI audit for head 0963e0efc9521c78d041a767af9b22250144f42f:

- Ref integrity: tom/refactor-miles/deliver-1 and tom/refactor-miles/op13-142 both point to this head; tom/refactor-miles/deliver-1-base remains at chain root d2e3ad9f197d86c138347063fdd9750dbd3d296e.
- Remote validation: targeted argv suite passed with 620 passed and 4 xfailed; the full fast suite passed with 6890 passed, 30 skipped, and 4 xfailed; all pre-commit hooks passed.
- Authoritative full workflow run: https://github.com/radixark/miles/actions/runs/31583550632
- Every H100/H200 lane was validated from its complete log, not merely from the check conclusion. Each selected a nonzero test count and ended with a matching N/N Test Summary.
- GPU summaries: stage-b 2-GPU H200 2/2; stage-c 2-GPU H200 4/4 and 5/5; stage-c 4-GPU H200 8/8, 9/9, and 9/9; stage-c 8-GPU H200 3/3 and 4/4; stage-c 8-GPU H100 10/10 and 12/12.
- The initial stage-b CPU job failed while downloading from Hugging Face with HTTP 429. Its rerun job 94118753909 passed with 86 passed and 18 skipped.
- Run 31583560405 came from another PR/branch sharing the same SHA. Its seconds-long GPU jobs reported No tests found / No tests to run, so they were treated as no-op checks and were not used as coverage evidence.

Final rollup: 40 successful checks, 2 skipped approval-holder checks, no pending or failing checks. PR is MERGEABLE/CLEAN.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** CI is green on `4802a93e9a`. This comment records the two real failures that were fixed, and the three GPU failures that were diagnosed as infra flakes and cleared by a rerun. Please push back if any conclusion is off.

## Green on [`4802a93e9a`](https://github.com/radixark/miles/pull/2407/commits/4802a93e9a80944a91b3b1d6cfa30a9f238825bd)

All 20 non-skipped checks pass; `mergeStateStatus=CLEAN`. Run: https://github.com/radixark/miles/actions/runs/31619024942 ([checks view](https://github.com/radixark/miles/pull/2407/checks))

`Run pre-commit`, `resolve-ci-policy`, `docker-paths`, `docker-build`, `resolve-ci-image`, `stage-a-cpu (0..3)`, `stage-b-cpu`, `stage-b-2-gpu-h200`, `stage-c-2-gpu-h200 (0,1)`, `stage-c-4-gpu-h200 (0,1,2)`, `stage-c-8-gpu-h100 (0,1)`, `stage-c-8-gpu-h200 (0,1)`.

## Real failures that were fixed

### 1. pre-commit — formatting only

`ruff-check` (5), `isort` (4 files), `black` (7 files), all `files were modified by this hook`, no remaining errors, all under `tests/`. Fixed in `bf85b6be65`.

### 2. Two stale tests

- `tests/fast/ray/rollout/test_cell_state.py` — `StateInitializing` gained a required `start_time` field but the test's `_STATE_FACTORIES` was not updated. Fixed in `6834694247`.
- `tests/fast/utils/api_server/test_server.py` — the test still stubbed `uvicorn.run` and `server.threading`, but `_start_api_server_raw` had moved to `uvicorn.Server(uvicorn.Config(...))` plus `_start_and_wait_thread`. Rewritten to assert against the real startup path in `85dc42f52b`.

### 3. `stage-a-cpu` shard 0 — missing optional dependency at collection

Failing job: https://github.com/radixark/miles/actions/runs/31612662504/job/94184605724

```
ERROR collecting tests/fast/backends/megatron_utils/update_weight/test_connection_contract.py
  .../p2p_transfer_utils.py:11: from mooncake.engine import TransferEngine
E ModuleNotFoundError: No module named 'mooncake'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

`mooncake` is an optional RDMA dependency present in neither `requirements.txt` nor `tests/ci/requirements-ci-cpu.txt`, so it does not exist on the GitHub-hosted CPU runner. The module-scope import predates this branch; what is new is `test_connection_contract.py`, the first CPU-side test to import the p2p updater. Moving the import to its single use site in `create_transfer_engine()` fixes it, matching how `miles/utils/object_store.py` already treats mooncake as optional. Fixed in `4802a93e9a`.

## Three GPU failures, all diagnosed as infra flakes and cleared on rerun

All three were on attempt 1 of run 31619024942 and all three pass on attempt 2. Each rerun was checked against the anti-fake-green criteria (non-zero `Enabled N test(s)`, plausible wall-clock, explicit per-test PASS) rather than accepted on the green badge alone.

| Lane | Test | Attempt-1 fingerprint | Attempt-2 |
|---|---|---|---|
| `stage-c-4-gpu-h200 (1)` | `test_dp_attention.py` | gloo `Connection closed by peer` in `_send_to_colocated_engine` | `PASS elapsed=763s`, 9/9 tests |
| `stage-c-4-gpu-h200 (2)` | `test_r3_baseline.py` | trainer rank0 vanished with no output; TCPStore EOF + raylet `SYSTEM_ERROR` | `PASS elapsed=941s`, 9/9 tests |
| `stage-c-8-gpu-h100 (0)` | `test_r3_mtp.py` | `Cuda failure 999 'unknown error'` via NCCL inside TransformerEngine CP attention | `PASS elapsed=802s`, 10/10 tests, CI Gate `TRUSTED (18 metric(s))` |

Attempt-1 jobs: [4-gpu (1)](https://github.com/radixark/miles/actions/runs/31619024942/job/94190952842), [4-gpu (2)](https://github.com/radixark/miles/actions/runs/31619024942/job/94190952931), [8-gpu-h100 (0)](https://github.com/radixark/miles/actions/runs/31619024942/job/94190952903).

Why these read as infra rather than regressions:

- **None of the three raised from miles Python logic.** Two are a peer training process disappearing with no self-reported error (the gloo FIN and the TCPStore EOF are two transports noticing the same class of event; raylet's own message lists OOM-killer / SIGSEGV as the usual causes). The third is a CUDA driver error surfacing inside third-party TransformerEngine/NCCL code right after a sleep/wake GPU memory reload.
- **`test_dp_attention` failed in code this branch did not change since a green run.** The same lane passed at [`0963e0efc`](https://github.com/radixark/miles/actions/runs/31583550632/job/94118757472), and `git diff 0963e0efc..4802a93e9a -- miles/backends/megatron_utils/update_weight/` is only the two-line mooncake change above.
- **`reloadable_process_group.py`, on the causal path of the CUDA-999 failure, has zero commits** in `deliver-1-base..HEAD`.
- **Sibling tests in the same jobs passed**, including other tests exercising the same wake_up / colocate pattern.
- Where this branch *is* in the blast radius — the new `RayWorkerHandle` death-confirmation path and `TrainerCell` error handling — it behaved correctly: it detected the dead worker, marked the cell errored, and failed fast with a clear `WorkerUnreachableError` rather than hanging or silently corrupting state.

Residual uncertainty worth stating: GitHub Actions logs carry no host `dmesg`, so a kernel-level OOM-kill of the vanished processes can be neither confirmed nor excluded. If the "silent trainer process death" signature recurs on this branch, host memory pressure is the first thing to check.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** I compared the two duplicate `PR Test` runs on this head, cancelled the one that resolved to an empty label scope, verified the green CPU stages actually executed tests rather than no-opping, and ran the full `stage-a-cpu` suite on an H200 devbox against this exact head. Please push back if any conclusion is off.

## Head under test

`1530b0b4398eda9284de1dda88bd84ed4dddd75f` — https://github.com/radixark/miles/pull/2407/commits/1530b0b4398eda9284de1dda88bd84ed4dddd75f

## Two duplicate runs on the same SHA; the no-coverage one is cancelled

Two `pull_request` runs were created 14s apart for this head and resolved to different policies:

| Run | Resolved policy | Verdict |
|---|---|---|
| [31663956556](https://github.com/radixark/miles/actions/runs/31663956556) | `cadence=regular labels=[] bypass_fastfail=false` ([policy job log](https://github.com/radixark/miles/actions/runs/31663956556/job/94334478713)) | empty label scope → near-zero coverage. **Cancelled.** |
| [31663968951](https://github.com/radixark/miles/actions/runs/31663968951) | `cadence=regular labels=[run-ci-image bypass-fastfail] bypass_fastfail=true` ([policy job log](https://github.com/radixark/miles/actions/runs/31663968951/job/94334515674)) | **authoritative** |

The `CANCELLED` entries in the checks list all belong to the first run and are intentional, not test failures. Everything below refers to the authoritative run [31663968951](https://github.com/radixark/miles/actions/runs/31663968951).

## CPU gate is green, and it really executed tests

`pre-commit` SUCCESS: https://github.com/radixark/miles/actions/runs/31663968813/job/94334512898
`resolve-ci-policy`, `docker-paths`, `docker-build`, `resolve-ci-image` all SUCCESS: https://github.com/radixark/miles/actions/runs/31663968951

A green GPU/CPU job can still be a no-op, so each was checked for a non-zero `Enabled N test(s)` and a real result summary:

| Job | Enabled | Result | Duration |
|---|---|---|---|
| [stage-a-cpu (0)](https://github.com/radixark/miles/actions/runs/31663968951/job/94336422962) | 80 | 1136 passed, 12 skipped | 2:59 |
| [stage-a-cpu (1)](https://github.com/radixark/miles/actions/runs/31663968951/job/94336423010) | 79 | 2458 passed, 4 skipped | 2:30 |
| [stage-a-cpu (2)](https://github.com/radixark/miles/actions/runs/31663968951/job/94336422995) | 95 | 1953 passed, 15 skipped | 5:09 |
| [stage-a-cpu (3)](https://github.com/radixark/miles/actions/runs/31663968951/job/94336422965) | 100 | 2233 passed, 8 skipped, 4 xfailed | 4:09 |
| [stage-b-cpu](https://github.com/radixark/miles/actions/runs/31663968951/job/94336422833) | 3 | 86 passed, 18 skipped | 3:55 |
| [stage-b-2-gpu-h200](https://github.com/radixark/miles/actions/runs/31663968951/job/94349324649) | 2 | `Test Summary: 2/2 passed` | 1:42 |

`stage-b-2-gpu-h200` finishing in 1:42 is short enough to look like a no-op, but it declares `Enabled 2 test(s)` and reports `2/2 passed`, so it is a genuine — merely tiny — lane. It should not be counted as meaningful GPU coverage; that comes from the `stage-c` lanes.

The four `stage-a-cpu` shards sum to 354 enabled files, which matches the 354 the full suite enables when run unsharded, so no file silently dropped out of the suite.

The nine `stage-c` lanes (2-gpu-h200 ×2, 4-gpu-h200 ×3, 8-gpu-h100 ×2, 8-gpu-h200 ×2) are still in progress; I will follow up with their results.

## Dependency pin is current

The PR body pins `ci-sglang-pr: tom/refactor-miles-repo-sglang/deliver-1`. Against its target branch:

```
gh api repos/sgl-project/sglang/compare/sglang-miles...tom/refactor-miles-repo-sglang/deliver-1
→ {"ahead_by": 4, "behind_by": 0}
```

`behind_by = 0`, so CI is not running against a stale sglang.

## Devbox pre-verification of the same head: zero real failures

Run on an H200 devbox with `radixark/miles:dev`, miles at `1530b0b439` and `/sgl-workspace/sglang` at `637fce854f` (the pinned branch tip):

| Pass | Command | Result |
|---|---|---|
| 1 — emulating the CPU runner | `CUDA_VISIBLE_DEVICES= python tests/ci/run_suite.py --hw cpu --suite stage-a-cpu --labels run-ci-image --continue-on-error` | `7519 passed, 22 skipped, 4 xfailed, 280 errors` in 649s; **zero `FAILED`** |
| 2 — GPU visible, re-running only the affected files | `pytest` over the 18 files that errored in pass 1 | **`313 passed`** in 1891s |

All 280 errors in pass 1 collapse to one cause — `ValueError: Attempting to start raylet with 8 GPU, but CUDA_VISIBLE_DEVICES contains []` — the known artifact of blanking `CUDA_VISIBLE_DEVICES` on a GPU host, not a test failure. Pass 2 confirms each of those tests passes with the GPU visible.

Caveat: a green devbox is not evidence for a green CPU runner. The hosted CPU runner has no accelerator and roughly 16GB of RAM, so accelerator-probing and resource-ceiling failures cannot appear on a devbox. The `stage-a-cpu` results above are the actual gate, and they are green.

## Note on the earlier queueing

These CPU stages sat `QUEUED` for about 50 minutes. That was repo-wide runner starvation, not a problem with this PR: at the time the repo had 196 queued runs against 15 in progress, 167 of them `PR Test` on `tom/refactor-miles/op<N>-<M>` marker branches re-triggered by a force-push that restacked the refactor chain. The queue drained on its own and no bulk cancellation was needed.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** The chain was rebased onto upstream main and promoted; I fixed the three integration gaps that broke the CPU stages, then root-caused the two remaining GPU failures. Please push back if any conclusion is off.

## Head under test

`e6999a73d612beceaa1c51edb939ca7c9d90a462` — authoritative run [31676416822](https://github.com/radixark/miles/actions/runs/31676416822)

The chain now sits on upstream main `6033ac0eafb2f86733d37c5dd2958573a88c2735` (previously `d2e3ad9f197d86c138347063fdd9750dbd3d296e`), so this PR's base branch moved with it.

## CPU gate is green

| Job | Result |
|---|---|
| `stage-a-cpu (0)` … `(3)` | SUCCESS |
| `stage-b-cpu` | SUCCESS |
| `stage-c-8-gpu-h200 (1)` | SUCCESS |
| `pre-commit`, `docker-build`, `docker-paths`, `resolve-ci-{policy,image}` | SUCCESS |

The previous run on this chain had six CPU failures. All six came from three integration gaps left by the rebase onto new main, each fixed in the commit that owns it (no new commits, chain invariant preserved):

| Gap | Symptom | Fixed in |
|---|---|---|
| `miles/rollout/session/core.py` still called `merge_samples_with_addition_r3(self.args, …)` after `SessionCore` switched to a `SessionServerConfig` | `AttributeError: 'SessionCore' object has no attribute 'args'` | `ff588a6d8e` *Avoid session server input using full args* |
| `SessionServerConfig` gained the upstream fields `session_message_matcher` and `pause_generation_mode`, but the test literal was not updated | `ValidationError: 2 validation errors for SessionServerConfig` | `4b1a3e55da` *Render the sglang server and router argv through one renderer* |
| Upstream #2485 moved `examples/fully_async` to `examples/infra_features/fully_async`; one of the two references was left behind | `ModuleNotFoundError: No module named 'examples.fully_async'` | `54490c3bac` *Move load_function into miles.utils.function_registry* |

Each fix landed in `max(last commit touching the file, commit introducing the problem)`, so replaying the chain produced no conflicts and the tree diff against the pre-fix chain is exactly those three hunks.

## Two GPU failures remain, and both are upstream defects this rebase inherited

### 1. `stage-c-2-gpu-h200 (0)` — [job 94374011017](https://github.com/radixark/miles/actions/runs/31676416822/job/94374011017)

```
FAILED tests/e2e/sglang/test_session_v1_v2_parity.py::test_qwen3_8b_h200_fa3_agentic_v2_drop_retries_matches_v1_training_payload_bitwise
AssertionError: sample.rollout_id is not bitwise equal
left = None, right = 0, path = 'sample.rollout_id'
```

`miles/rollout/generate_hub/agentic_tool_call.py` assigns `rollout_id` **only on the v2 branch**:

```python
samples = result.samples
if use_v2:
    # FIXME: handle sample index issues.
    rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
    assert rollout_id is not None, "v2 agentic samples require input Sample.rollout_id or Sample.index"
    for sample in samples:
        sample.rollout_id = rollout_id
```

so v1 keeps `None` while v2 gets `0`, and the parity test compares every field. Evidence that this is upstream's, not the chain's:

- That block is **byte-identical between `6033ac0eaf` and the chain tip** — the chain never touched it.
- It does **not exist** in the previous base `d2e3ad9f19` (0 occurrences of `rollout_id` in that file there, 3 in the new base), which is why the same lane was green before the rebase.
- `tests/e2e/sglang/test_session_v1_v2_parity.py` is byte-identical across the old base, the new base and both chain tips.
- Neither the upstream nor the chain copy of `tests/session_parity_utils.py` mentions `rollout_id`, so no field-exclusion was dropped by the chain.

### 2. `stage-c-8-gpu-h200 (0)` — [job 94374011197](https://github.com/radixark/miles/actions/runs/31676416822/job/94374011197)

```
FAILED: tests/e2e/fsdp/r3/test_qwen3_5_35b_a3b_r3.py (exit code 1)
Error: 'Qwen3_5MoeTextConfig' object has no attribute 'intermediate_size'. Did you mean: 'moe_intermediate_size'?
  File "miles/backends/fsdp_utils/actor.py", line 120, in init
  File "miles/utils/flops_utils.py", line 180, in flops_args_from_hf_config
```

`flops_utils.py:180` reads the attribute bare, while every neighbouring field goes through a fallback helper:

```python
ffn_hidden_size=config.intermediate_size,
kv_channels=_first(config, "head_dim", default=hidden_size // num_attention_heads),
moe_router_topk=_first(config, "num_experts_per_tok", "moe_topk", default=1),
```

Evidence that this is upstream's:

- `miles/utils/flops_utils.py` is **byte-identical to `6033ac0eaf`** — the chain does not modify it.
- The call site is `flops_args_from_hf_config(self.hf_config)` at **line 120 of `miles/backends/fsdp_utils/actor.py` in both the base and the chain tip**, and the chain's diff on that file contains no change to this path.

Both are reported here rather than patched, since patching upstream semantics inside this chain would pollute its diff. Happy to carry a workaround on `deliver-1` if that is preferred.

## Devbox verification

Run on a 1×H200 devbox with `radixark/miles:dev`, `/sgl-workspace/sglang` at the pinned `tom/refactor-miles-repo-sglang/deliver-1` tip. `tests/fast --collect-only` at the chain tip collects 9331 tests with 4 collection errors; three of them (`test_scenario_ft_random.py`, `test_misc.py`, `test_process_identity.py`) were verified to be pre-existing — the imported symbols are absent in both the old and the new chain and the three test files are byte-identical between them — and the fourth is an artifact of the sync excluding `.git`, so `git ls-files` fails on the devbox but not in CI.

A full old-chain-versus-new-chain comparison run is still in flight; I will follow up with the attribution.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Follow-up to the previous comment: I ran the old chain and the new chain through the same suite on the same devbox and diffed the failing test-id sets to attribute every failure. Please push back if any conclusion is off.

## The rebase onto upstream main introduces zero test regressions

Same command, same devbox, same pinned sglang/Megatron, same `--ignore`/`--deselect` set, run against each chain tip:

| | old chain `db77cc1196` | new chain `30179d1317` |
|---|---|---|
| failed | 212 | 212 |
| passed | 8497 | 8736 |
| errors | 309 | 320 |

Counts alone prove nothing, so the sets were diffed by test id:

- **211 of the 212 failures are identical on both sides.**
- **Introduced: 1** — `tests/fast/doc/test_sync_example_docs.py::test_docs_examples_matches_the_readmes`
- **Fixed: 1** — `tests/fast/utils/test_command_utils.py::TestExecuteTrain::test_lets_the_environment_override_nvls`
- The 11 extra errors are all `test_train_data_conversion.py::TestPostProcessRewards`, tests upstream added in the new base. They hit the same raylet artifact as every other error here, and have no counterpart on the old chain.

### The one "introduced" failure is a test-environment artifact, not a regression

```
error: mirrored but not listed in examples/README.md — add a bullet for:
  examples/fully_async
```

`examples/fully_async` and `docs/examples/fully_async` contain **zero files** in the new base, in this PR's head, and at the chain tip — only the old chain has that directory. The devbox still had leftovers of it because the sync step does not delete removed files, and my earlier cleanup pass only swept `*.py`, so `README.md` and `run-qwen3-4b-fully_async.sh` survived and tripped the docs drift gate. After removing them, that file reports **19 passed**.

So the attribution is **introduced = 0**, which matches what the rebase report claimed on static grounds.

The 320 errors on both sides are the known consequence of blanking `CUDA_VISIBLE_DEVICES` on a GPU host (`ValueError: Attempting to start raylet with 8 GPU, but CUDA_VISIBLE_DEVICES contains []`), and the 211 shared failures come from running the full `pytest tests/fast` surface rather than the 354 files `run_suite.py --hw cpu --suite stage-a-cpu` actually registers — that registered set is green in CI here. Neither set is attributable to this rebase.

## CI status on `e6999a73d612beceaa1c51edb939ca7c9d90a462`

Run [31676416822](https://github.com/radixark/miles/actions/runs/31676416822): 13 green, 2 red, remaining GPU shards still running.

Green: `stage-a-cpu (0..3)`, `stage-b-cpu`, `stage-b-2-gpu-h200`, `stage-c-2-gpu-h200 (1)`, `stage-c-4-gpu-h200 (1)`, `stage-c-8-gpu-h200 (1)`, plus `pre-commit`, `docker-build`, `docker-paths`, `resolve-ci-policy`, `resolve-ci-image`.

The two red shards are the upstream defects detailed in the previous comment — the v2-only `rollout_id` assignment that breaks the v1/v2 parity test, and the bare `config.intermediate_size` read in `flops_utils.py` that `Qwen3_5MoeTextConfig` does not satisfy. Both live in code byte-identical to `6033ac0eaf`, on paths this chain does not modify.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Final CI status for this head, with the two upstream workarounds now verified green on the lanes that previously failed. Please push back if any conclusion is off.

## Green on `e924ab99f02deb90634865098e04934bd62ff391`

Run [31685938763](https://github.com/radixark/miles/actions/runs/31685938763) — **23 success, 1 failure, 2 skipped**. The single failure is an upstream/infra one reproduced on `main`, detailed at the bottom.

| Lane | Result | Duration |
|---|---|---|
| `stage-a-cpu (0)` / `(1)` / `(2)` / `(3)` | success | 5 / 5 / 3 / 8 min |
| `stage-b-cpu` | success | 5 min |
| `stage-b-2-gpu-h200` | success | 1 min |
| `stage-c-2-gpu-h200 (0)` / `(1)` | success | 51 / 44 min |
| `stage-c-4-gpu-h200 (0)` / `(1)` / `(2)` | success | 91 / 84 / 83 min |
| `stage-c-8-gpu-h100 (0)` / `(1)` | success | 91 / 118 min |
| `stage-c-8-gpu-h200 (0)` / `(1)` | success | 48 / 64 min |
| `pre-commit`, `docker-build`, `docker-paths`, `resolve-ci-policy`, `resolve-ci-image` | success | — |

Durations are in line with these lanes' historical runtimes, so none of them is a no-op green. `stage-b-2-gpu-h200` is short by design — it declares `Enabled 2 test(s)` and reports `2/2 passed`.

## The two previously-red GPU lanes

Both failures were defects in upstream `main` that this chain inherited by rebasing onto `6033ac0eaf`, so the workarounds live on `deliver-1` rather than inside any op commit — that keeps the chain's own diff clean.

| Lane | Before | After |
|---|---|---|
| `stage-c-2-gpu-h200 (0)` | `Enabled 5 test(s)` → 4/5, `test_session_v1_v2_parity` failed on `sample.rollout_id` | `Enabled 5 test(s)` → **5/5 passed** |
| `stage-c-8-gpu-h200 (0)` | `Enabled 3 test(s)` → 2/3, `test_qwen3_5_35b_a3b_r3.py` exit 1 on `intermediate_size` | `Enabled 3 test(s)` → **3/3 passed** |

`bdd8cc5e61` gives `ffn_hidden_size` the same fallback every neighbouring field in `flops_args_from_hf_config` already uses, so a config that only carries `moe_intermediate_size` no longer raises. `e924ab99f0` lifts the `rollout_id` derivation out of the `use_v2` branch, leaving v2 behaviour byte-identical — the assertion still guards only that path — while v1 stops reporting `None` where v2 reports the index.

## Chain integrity after the rebase and promotion

```
verify origin/main..origin/tom/refactor-miles/op22-6 against the issue and the PR chain
  chain ops: 340  issue entries: 360 (18 merged, 2 closed, 340 open)
  verify: OK
```

Every chain commit has exactly one op branch, each PR's base is the previous op's branch, the chain bottoms out at `main`, and the tracking issue's ordering matches the live chain.

## Remaining failure is not this PR's

`stage-c-4-gpu-mi300x (0)` fails with:

```
ImportError: cannot import name 'shuffle_scale' from 'aiter.ops.shuffle' (/sgl-workspace/aiter/aiter/ops/shuffle.py)
```

`aiter` is the AMD kernel library shipped inside the ROCm image, not miles code. The scheduled run on `main` — [31612060884](https://github.com/radixark/miles/actions/runs/31612060884), 2026-08-12 — fails the **same shard** with the **same fingerprint**, so this is a pre-existing image/dependency mismatch rather than anything introduced here.

fzyzcjy:
## CI status after rebasing the chain onto latest main

Head is now `97bba45562e6a5bb5daffeb28d125be97a0fa052`, rebased from base `6033ac0eaf` onto `1200fca8f8f3fe011d9a95707d5a562478060843` (14 upstream commits). 356 commits replayed, 0 merge commits, 0 empty commits.

**Result: 24 success / 2 failure / 9 skipped.** Every CPU lane, `Run pre-commit`, and every CUDA GPU lane is green. The only red is the ROCm `stage-c-4-gpu-mi350` pair, and the evidence below shows it is a missing sglang feature in the ROCm CI image, not something this branch introduces.

Runs: [PR Test #31774484373](https://github.com/radixark/miles/actions/runs/31774484373) · [PR Test (ROCm) #31774484412](https://github.com/radixark/miles/actions/runs/31774484412)

### Green lanes, with durations

Durations are included because a fast "success" can mean the lane never actually ran. These are real runs.

| Lane | Result | Duration |
|---|---|---|
| `Run pre-commit` | success | — |
| `stage-a-cpu (0..3)` | success | 4m / 5m / 6m / 7m |
| `stage-b-cpu` | success | 5m |
| `stage-b-2-gpu-h200` | success | 1m |
| `stage-c-2-gpu-h200 (0,1)` | success | 51m / 44m |
| `stage-c-4-gpu-h200 (0,1,2)` | success | 77m / 69m / 93m |
| `stage-c-8-gpu-h100 (0,1)` | success | 115m / 101m |
| `stage-c-8-gpu-h200 (0,1)` | success | 44m / 55m |

### The two red ROCm shards

| Shard | Duration | Failing tests |
|---|---|---|
| [`stage-c-4-gpu-mi350 (0)`](https://github.com/radixark/miles/actions/runs/31774484412/job/94687155502) | 37m | `tests/e2e/megatron/test_qwen3_30B_A3B/test_r3_baseline.py` |
| [`stage-c-4-gpu-mi350 (1)`](https://github.com/radixark/miles/actions/runs/31774484412/job/94687155460) | 99m | `tests/e2e/megatron/test_qwen3_30B_A3B/test_baseline.py`, `tests/e2e/megatron/test_mimo_7B_mtp_only_grad.py`, `tests/e2e/lora/test_lora_qwen2.5_0.5B.py` |

Every failing test in both shards fails the same single way, with no second failure mode: the sglang engine starts and serves normally (`The server is fired up and ready to roll!`, `Uvicorn running on http://172.19.0.2:20000`), but the launch gate port never listens, so `retry_until_deadline` gives up after 1800s.

### Root cause: the ROCm image's sglang has no gated-launch support

The decisive evidence is sglang's own `server_args=ServerArgs(...)` dump in each lane's log:

| | CUDA image ([passing job](https://github.com/radixark/miles/actions/runs/31685938763/job/94404301572)) | ROCm image ([failing job](https://github.com/radixark/miles/actions/runs/31774484412/job/94687155502)) |
|---|---|---|
| `gated_launch_port` in `ServerArgs` | present: `20034`, `20035`, `20036`, `20037`, `20069`, … | **field absent entirely** |
| `engine_info_bootstrap_port` (the sibling port miles passes the same way) | present | present (`20035`) |
| Gate behaviour | cell reaches `init end` carrying `gate_url='http://172.20.0.2:20037'`, then `mark_pending_weights` | no gate server; `retry_until_deadline` gives up after 343 / 3094 attempts |

The sibling port arrives while the gate port does not, so the argument plumbing on the miles side is fine — that sglang build simply does not have the field. Both `docker/Dockerfile` and `docker/Dockerfile.rocm` fetch `${SGLANG_BRANCH}` (`sglang-miles`), but the ROCm image is built on the prebuilt `rocm/sgl-dev` base with a pinned `SGLANG_COMMIT`, and the sglang that actually lands there predates the feature.

### Why this is not introduced by this branch

- The launch gate is code that exists only on this chain: `git grep gate_url 1200fca8f8 -- miles/` returns nothing on main. This failure shape cannot occur on main, so there is no main-side comparison to run.
- The ROCm PR lane itself is new: #2347 enabled AMD PR CI and moved ROCm from mi300x to mi350 on 2026-08-13. This rebase produced the first state of the chain that includes #2347, so the chain had never run against this lane before.
- It reproduces deterministically across two separate CI rounds (`66c26ce840` and `97bba45562`) with identical attempt counts and timeouts, so it is not a flake.

Three possible ways forward, none of them applied here since the ownership is not obvious: rebuild the ROCm CI image on an sglang that has gated launch; teach miles to probe backend capability and take the existing no-gate path (`miles/ray/specs/inference.py` currently declares the gate port unconditionally); or keep this chain out of the ROCm lane for now.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Triaged the CPU failures of run [33253668762](https://github.com/radixark/miles/actions/runs/33253668762). Both are ours and both look like the pinned sglang moving under us, not CI infra. Please push back if any conclusion is off.

`stage-a-cpu` shards 1 and 2 and `stage-b-cpu` are green. Two shards are red; `bypass-fastfail` is on, so the GPU stages are proceeding regardless and this run will still produce GPU evidence.

### `stage-a-cpu (0) / run-cpu` — ours

- Job: https://github.com/radixark/miles/actions/runs/33253668762/job/99103600884

```
FAILED tests/fast/backends/sglang_utils/test_server_args_utils.py::TestEveryServerArgsFieldIsRenderable::test_a_field_renders_to_argv_that_parses_back_to_the_same_value[_speculative_draft_quantization_explicitly_set]
AssertionError: '_speculative_draft_quantization_explicitly_set' cannot be rendered: the parser registers no option for dest '_speculative_draft_quantization_explicitly_set'. Add an entry to field_to_dest, or pass the value through the native passthrough path.
1 failed, 1858 passed, 9 skipped, 4 xfailed in 135.54s
```

### `stage-a-cpu (3) / run-cpu` — ours

- Job: https://github.com/radixark/miles/actions/runs/33253668762/job/99103600929

```
FAILED tests/fast/backends/sglang_utils/test_sglang_engine.py::TestLoraTargetModules::test_gdn_attention_targets_are_named_one_by_one
argparse.ArgumentError: argument --lora-target-modules: invalid choice: 'in_proj_qkvz'
  (choose from 'q_proj', 'k_proj', 'v_proj', 'o_proj', 'q_a_proj', 'kv_a_proj_with_mqa', 'q_b_proj', 'kv_b_proj', 'wq_b', 'wk', 'weights_proj', 'gate_proj', 'up_proj', 'down_proj', 'qkv_proj', 'gate_up_proj', 'embed_tokens', 'lm_head', 'qkvr', 'wo_ud', 'all')
1 failed, 2492 passed, 21 skipped in 217.86s
```

- Common shape: both are places where this branch encodes a fact about sglang's `ServerArgs` surface, and the pinned sglang (`tom/refactor-miles-repo-sglang/ac8770`, v0.5.18) no longer matches — a new private field with no CLI option, and a rewritten `--lora-target-modules` choice list that drops the fused GDN slice names. Neither is CI infra and neither reruns green.
- Both are also in the pre-existing failure set recorded for this chain's tip, so they are being fixed on the unit-test track; the fixes land on `tom/refactor-miles/deliver-1` as well as `deliver-2` because both tests are inside `deec6fce05`.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Followed the `--lora-target-modules` failure across into the pinned sglang and found it is a pin gap, not a miles bug. Recording it here because it needs a change I am not authorized to push. Please push back if any conclusion is off.

### `TestLoraTargetModules` needs GDN names the pinned sglang's CLI does not accept

- Failing job: https://github.com/radixark/miles/actions/runs/33253668762/job/99103600929
- The pin in this PR's body is `ci-sglang-pr: tom/refactor-miles-repo-sglang/ac8770` (`2643d77070`).
- On that branch, `python/sglang/srt/lora/utils.py` still lists `in_proj_qkvz`, `in_proj_ba` and `out_proj` in `_KNOWN_LORA_TARGET_MODULES`, so the LoRA runtime supports them — but `SUPPORTED_LORA_TARGET_MODULES` in `python/sglang/srt/utils/common.py`, which is what feeds `choices=` for `--lora-target-modules`, does not. So the engine can serve those adapters but the CLI refuses to name them, which is exactly what the four `TestLoraTargetModules` cases assert against.
- `sgl-project/sglang@main` has the same gap. The three names were added on this campaign's own sglang branches: both `tom/refactor-miles-repo-sglang/deliver-1` and `.../deliver-2` carry them, `ac8770` does not.
- The missing piece on `ac8770` is the tail of `SUPPORTED_LORA_TARGET_MODULES`:

```python
    # GDN linear attention (Qwen3.5 / Qwen3-Next): the two slices the fused in_proj is
    # served as, and its output projection.
    "in_proj_qkvz",
    "in_proj_ba",
    "out_proj",
```

- There is no miles-side fix that is not a weakening: passing `all` instead is the auto-detecting shorthand these tests exist to rule out, and every other route goes through the same rejecting parser. `ac8770` needs the three names.
- The sibling failure on shard 0 (`_speculative_draft_quantization_explicitly_set`) is the other direction of the same pin choice — that field exists only on `ac8770`, not on the sglang deliver branches — and is fixed on the miles side, since sglang declares it `Arg(no_cli=True)` on purpose.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Re-checked the LoRA failure after this PR was switched to the default dependencies. It still fails, and the reason is now on the default `sglang-miles` branch rather than on a pin. Please push back if any conclusion is off.

### `stage-a-cpu (3) / run-cpu` — still red, and no miles-side fix exists

- Job: https://github.com/radixark/miles/actions/runs/33255544858/job/99108513959 — `1 failed, 2492 passed, 21 skipped in 217.21s`
- Only failure: `tests/fast/backends/sglang_utils/test_sglang_engine.py::TestLoraTargetModules::test_gdn_attention_targets_are_named_one_by_one`

```
argparse.ArgumentError: argument --lora-target-modules: invalid choice: 'in_proj_qkvz'
  (choose from 'q_proj', 'k_proj', 'v_proj', 'o_proj', 'q_a_proj', 'kv_a_proj_with_mqa', 'q_b_proj', 'kv_b_proj', 'wq_b', 'wk', 'weights_proj', 'gate_proj', 'up_proj', 'down_proj', 'qkv_proj', 'gate_up_proj', 'embed_tokens', 'lm_head', 'qkvr', 'wo_ud', 'all')
```

- This PR now carries no dependency pins, so it resolves sglang to `sglang-miles` (`3bbb2812e2`). That branch's `SUPPORTED_LORA_TARGET_MODULES` in `python/sglang/srt/utils/common.py` — the list that feeds `choices=` for `--lora-target-modules` — does not contain `in_proj_qkvz`, `in_proj_ba` or `out_proj`, even though the LoRA runtime's own `_KNOWN_LORA_TARGET_MODULES` in `python/sglang/srt/lora/utils.py` does.
- The rebuilt `tom/refactor-miles-repo-sglang/deliver-2` (`853a75eac2`) does carry the three names, so #2542 should clear this test on its current pin; this PR will not.
- The rest of this PR's CPU stage is green: `stage-a-cpu` shards 0, 1 and 2 and `stage-b-cpu` all passed, including shard 0, which the previous round failed on `_speculative_draft_quantization_explicitly_set` and which `1b9041d37e` fixed.
- Recorded for the owner to decide: either give this PR a pin to an sglang branch rebuilt the same way `deliver-2` was (latest `sglang-miles` plus the two GDN LoRA commits), or land those two commits on `sglang-miles` itself. Pinning the existing `tom/refactor-miles-repo-sglang/deliver-1` (`637fce854f`) as-is would not be safe: it is 1365 commits behind `sglang-miles`.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Correcting an earlier claim I made on this PR: I said `bypass-fastfail` would let the GPU stages run despite a red CPU shard. Run [33255544858](https://github.com/radixark/miles/actions/runs/33255544858) disproves it. Please push back if any conclusion is off.

### `bypass-fastfail` does not carry the GPU stages past a red `stage-a-cpu`

- Observed in that run: `stage-a-cpu (3)` red, `docker-build / docker-decide` and `docker-build / docker-build` both green (53m33s), then `resolve-ci-image` **skipped in 0s** and all five CUDA stages (`stage-b-2-gpu-h200`, `stage-c-2-gpu-h200`, `stage-c-4-gpu-h200`, `stage-c-8-gpu-h100`, `stage-c-8-gpu-h200`) skipped in 0s.
- Mechanism: `docker-build` clears the failed CPU gate through its own `if: always() && ...`, but `resolve-ci-image` deliberately carries no `if` ("a detector or build failure must stop the suites rather than run them against a stale image"), so the CPU failure propagates through it. Every GPU stage's `if` then requires `needs.resolve-ci-image.result == 'success'`, and that clause is `&&`-ed with the `bypass_fastfail` alternative rather than or-ed.
- Practical consequence: **`stage-a-cpu` must be fully green before any GPU stage runs at all.** `bypass-fastfail` still does its real job — `--continue-on-error` inside a shard, so one round reports every CPU failure instead of stopping at the first.
- So this PR gets no GPU coverage until the `--lora-target-modules` gap on the sglang side is closed. The already-built `radixark/miles:pr-2407` image should let `docker-decide` skip the 53-minute rebuild on the next round.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** First round with real GPU coverage on this PR. Classifying every CUDA failure and pushing four fixes. Please push back if any conclusion is off.

Run [33259383440](https://github.com/radixark/miles/actions/runs/33259383440) is the first on this PR where `stage-a-cpu` went fully green, so the GPU stages actually ran. `stage-b-2-gpu-h200`, both `stage-c-2-gpu-h200 (0)`, all three `stage-c-4-gpu-h200` shards and `stage-c-8-gpu-h100 (1)` passed. Three CUDA jobs were red.

### Group A — `--debug-train-only` runs never size a rollout fleet (8 tests)

- [stage-c-8-gpu-h200 (1)](https://github.com/radixark/miles/actions/runs/33259383440/job/99119625104) (5 tests) and [stage-c-8-gpu-h200 (0)](https://github.com/radixark/miles/actions/runs/33259383440/job/99119625113) (3 tests)

```
miles/backends/sglang_utils/sglang_config.py:311, in _compute_raw_sglang_config
    server_groups=[_RawServerGroupConfig(worker_type="regular", num_gpus=args.rollout_num_gpus)]
pydantic_core._pydantic_core.ValidationError: 1 validation error for _RawServerGroupConfig
num_gpus
  Input should be a valid integer [type=int_type, input_value=None, input_type=NoneType]
```

- All eight pass `--debug-train-only`, directly or through `--load-debug-rollout-data`. `rollout_num_gpus` is only derived for a colocated `debug_rollout_only` run, so it stays `None`, and every one of them dies within ninety seconds of launch.
- Already solved on the M35 chain by `d4315b9be7` (op35-9, "Give a run with no engine fleet nothing to serve"), which is in `deliver-2` and not here. Cherry-picked.

### Group B — a run replaying recorded data has no rollout function to checkpoint

- [stage-c-2-gpu-h200 (1)](https://github.com/radixark/miles/actions/runs/33259383440/job/99119625090), `tests/e2e/fsdp/test_qwen3_0.6B_megatron_fsdp_align.py`

```
miles/ray/placement_group.py:157, in create_training_models
    await rollout_executor.load.remote(args.start_rollout_id - 1)
  miles/ray/rollout/rollout_executor.py:256, in load
    self.generate_rollout.load(rollout_id)
AttributeError: 'NoneType' object has no attribute 'load'
```

- `RolloutExecutor.__init__` sets `generate_rollout = None` on purpose under `--load-debug-rollout-data`. No op on the M35 chain fixes this and `deliver-2` carries the same defect — it took out five of the eight tests in that branch's own FT job — so the fix goes to both branches.

### Group C — healing never completes because a restarted cell is not yet addressable

- `test_trainer_ft_with_failure_dp2_cp2_real_rollout_dense` and `test_trainer_ft_deterministic_dp2_cp2_real_rollout`, both failing `assert_reconfigure_events` with only the crash event recorded and no heal event.
- Between the restart and the assertion the reconcile sweep failed four times, five seconds apart, on `TrainerCell.__init__` reading `worker_infos[0].self_addrs['master']` with `KeyError: 'master'`. `alloc_ports` published `self_addrs` as an empty dict up front and filled it after awaiting the actor's node ip, so the cell was reported alive with no endpoints for the seventeen seconds its actor spent importing torch and Megatron. One unaddressable worker fails the whole sweep, and the cell was never observed again.
- Solved on the M35 chain by `5fd8d7c1a0` (op35-1) and `5f3e58aa23` (op35-2). Cherry-picked; both conflicted only on surrounding context that later ops introduce, and the ops' own changes were applied unaltered.

### Still open, not fixed here

- `tests/e2e/long/test_qwen3_0.6B_verifiers.py` — killed at the harness's 1800 s ceiling after ~17000 consecutive `POST /v1/chat/completions -> 502`. The 502s come from the `verifiers` package's own interception server, which relays every upstream failure as a 5xx so the harness retries; the test's generated TOML sets no `[timeout] rollout`, so nothing bounds that retry loop. The failure path is byte-identical between `deliver-1` and `deliver-2`, so this is very likely pre-existing rather than a chain regression. **I could not identify what made the upstream call start failing** — `http_utils._post`'s retry logging never appears in the log — and I am not touching the threshold or the timeout until a rerun says whether it is deterministic.

### Three cherry-picked M35 ops = three ops that want moving earlier

`op35-9`, `op35-1` and `op35-2` all fix defects that exist from well before op13-142. They sit in M35 only because their landing point was not allowed to absorb them. Flagging that as a signal for whoever folds these fix commits back into the chain.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** I read the complete log of the one CUDA failure in the current run, traced it to a missing happens-before in the FT test-action injector, and prepared a fix for both delivery branches. Please push back if any conclusion is off.

## `stage-c-8-gpu-h200 (1)` on `0ab6457db6` — five FT e2e tests, one root cause

[Job](https://github.com/radixark/miles/actions/runs/33267309055/job/99140645251) of run [33267309055](https://github.com/radixark/miles/actions/runs/33267309055), `Test Summary: 3/8 passed`. Failing: `test_trainer_ft_deterministic_dp2_cp2_pp2`, `test_trainer_ft_deterministic_dp4_cp2`, `test_trainer_ft_with_failure_dp2_cp2_pp2`, `test_trainer_ft_with_failure_dp2_cp2_real_rollout_dense`, `test_trainer_ft_with_failure_dp2_cp2_tp2_ep2`. The three that pass (`fsdp/r3/test_qwen3_30b_a3b_r3`, `test_trainer_ft_no_failure_dp2_cp2_real_rollout`, `test_trainer_ft_no_failure_dp2_cp2_tp2_ep2`) are exactly the three that schedule no `stop_cell_at_end` / `start_cell_at_end` action.

**Classification: ours, and it is the test harness, not the runner.** The actor deaths in this job are the scenario's own `crash_before_allreduce` calling `os._exit(1)`; Ray's "consider the OOM killer" sentence is its generic guess for that exit. No CUDA OOM, no segfault, `free_GB: 79` throughout.

Root cause: `FTTestActionControllerExecutor.run_after_step` issues the stop and the start back to back and waits for neither, while `TrainerController.cell_ids` only changes when the watcher's reconcile loop runs. From the log:

```
19:28:37.272  FT test action: stop_cell_at_end  cell trainer-actor-1
19:28:37.276  FT test action: start_cell_at_end cell trainer-actor-1
19:28:40.956  ft op=health phase=stop name=trainer-cell-trainer-actor-1
19:28:59.432  ft op=refresh phase=start rollout=3 alive=0 healing= all_states="(0, 'StateAllocatedAlive')"
```

The start is issued 4 ms after the stop, against a view that has not seen the stop yet, and rollout 3 then trains on whatever cells were left. The three downstream shapes are all consequences of that:

- `deterministic_dp2_cp2_pp2`: one cell left running `pp2` deadlocks in pipeline p2p, `[rank1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=21, OpType=SEND, ...) ran for 600082 ms`, then `NonRetryableError: Cannot recover when all cells are dead`.
- `deterministic_dp4_cp2`: three cells trip `miles/backends/training_utils/data.py:530 assert len(set(sum(micro_batch_indices, []))) == num_local_samples` eight times, cascading into two extra `CellReconfigureEvent`s.
- `with_failure_dp2_cp2_pp2` and `..._tp2_ep2`: the cell is never observed back, so the rollout-3 healing event the assertion waits for is never produced.

Fix pushed on the next round: wait for each transition to be observed before moving on, deadlined so it fails loudly instead of racing on. The same gap exists on `deliver-2` in a different shape — [`da47fa6f66`](https://github.com/radixark/miles/commit/da47fa6f66ccf4d538b306cc4e66a17c1d64b145) added a wait after the resume only, and that wait passes on the same stale pre-suspend view, so it gets the same fix.

`test_trainer_ft_with_failure_dp2_cp2_real_rollout_dense` fails differently, on `Step 7, metric 'train/grad_norm': rel_diff=0.060683 > rtol=0.05`. That is bit-for-bit the same number this scenario produced on `deliver-2`, so it is reproducible rather than noise, and it is plausibly the same race training one step on a configuration the scenario never asked for. **The tolerance is not being touched**; it gets re-judged on the round after the injector fix.

## `tests/e2e/long/test_qwen3_0.6B_verifiers.py` — 1800s timeout on both branches

Second data point in [this job](https://github.com/radixark/miles/actions/runs/33267764531/job/99142152721) on `deliver-2`, so it is not a flake. The rollout never converges at any time budget: the Verifiers interception server answers every `POST /v1/chat/completions` with 502 and the harness SDK retries about 17000 times, because the generated TOML has no `[timeout]` section and `TimeoutConfig.rollout` defaults to `None`, i.e. `asyncio.wait_for(..., None)`.

Why the cause is not in the log: `verifiers/__init__.py` ends with `setup_logging(os.getenv("VF_LOG_LEVEL"))`, and with the variable unset that call clears the `verifiers` logger's handlers and sets `propagate = False`. The job log therefore has 16687 aiohttp access lines reporting 502 and zero `model call failed` warnings — the package is muted, not silent about a missing error.

Two changes go out with the same round: bound the rollout stage, and ask for `VF_LOG_LEVEL=INFO` so the next run names the exception. The failures are instantaneous (several 502s inside one millisecond) and no outbound HTTP retry is logged, so the exception is raised inside `TrainClient.get_response` before any request reaches SGLang.

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Follow-up: I found the cause of the `test_qwen3_0.6B_verifiers.py` timeout, and it is not this branch. Please push back if any conclusion is off.

`verifiers` 0.2.0 declares `renderers>=0.1.8.dev40`. A specifier that names a pre-release makes every later pre-release of that package eligible, so `examples/experimental/verifiers/requirements.txt` asking for `renderers>=0.1.8` installs whichever dev wheel was uploaded most recently. `renderers 0.1.11.dev3` was published 2026-08-29 at 00:28 UTC.

The nightly runs on `main` draw the line by themselves — same job, `stage-c-2-gpu-h200 (1)`:

| date | run | renderers | result |
| --- | --- | --- | --- |
| 2026-08-27 18:40Z | [33104087365](https://github.com/radixark/miles/actions/runs/33104087365) | `0.1.10` | `test_qwen3_0.6B_verifiers.py PASS elapsed=237s`, zero 502 |
| 2026-08-28 19:11Z | [33202173180](https://github.com/radixark/miles/actions/runs/33202173180) | `0.1.11.dev1` | `PASS elapsed=240s`, zero 502 |
| 2026-08-29 15:09Z | [33259484101](https://github.com/radixark/miles/actions/runs/33259484101/job/99144469000) | `0.1.11.dev3` | still answering every model call with 502 when I sampled its log |

So this is a **pre-existing upstream regression, not a delivery-branch failure**, and `main` is red on it right now. Note that `main` reaches it through `RolloutManager` while these branches reach it through the legacy `RolloutExecutor`, which rules the refactor out again.

Going out with the next push: pin the example to `renderers==0.1.10`, the last release CI has actually passed on. The two hardening changes stay as well — a bounded `[timeout] rollout` so a broken model call can no longer eat a whole 1800s slot (a healthy run takes 240s), and `VF_LOG_LEVEL=INFO` so the next upstream break names its exception in the log instead of leaving 16687 anonymous 502 access lines.

fzyzcjy:
> 🤖 **Posted autonomously by Codex acting on the user's behalf.** I diagnosed the first failure on the newly published head and am preparing the minimal dependency compatibility fix. Please push back if any conclusion is off.

- **Head:** `a4cea18e00b4edfb25ac1f6e6a0702efa3b040a5`.
- **Failure:** [`stage-a-cpu (0) / run-cpu`](https://github.com/radixark/miles/actions/runs/33321730211/job/99284805877) fails during collection because the pinned SGLang SHA does not export `resolve_dsv4_reasoning_effort_profile`.
- **Fingerprint:** `ImportError: cannot import name 'resolve_dsv4_reasoning_effort_profile' from 'sglang.srt.entrypoints.openai.chat_encoding'`.
- **Classification:** dependency compatibility gap exposed by the complete CPU suite, not a threshold or test weakening issue.
- **Next step:** build and verify the smallest SGLang compatibility ref on top of the H200-proven SHA, then trigger a fresh `synchronize` run with that immutable ref.

fzyzcjy:
> 🤖 **Posted autonomously by Codex acting on the user's behalf.** I archived and classified the separate ROCm failure while continuing to babysit the authoritative CUDA run. Please push back if any conclusion is off.

- **Failure:** [`stage-c-4-gpu-mi350 (1) / run`](https://github.com/radixark/miles/actions/runs/33322322465/job/99286403137).
- **Test:** `tests/e2e/megatron/test_qwen3_5_35B_A3B_mtp/test_amd_mtp1_spec_v2_r3.py`.
- **Fingerprint:** `timeout after 2400s`; the other two selected tests passed.
- **Classification:** allowed ROCm timeout exception for this Deliver-1 gate; there is no product assertion fingerprint in the complete log.
- **Action:** no threshold or timeout weakening. The authoritative CUDA run remains active and currently has no red checks.

fzyzcjy:
> 🤖 **Posted autonomously by Codex acting on the user's behalf.** I archived and diagnosed the first authoritative CUDA failure while the remaining CUDA jobs continue. Please push back if any conclusion is off.

- **Failure:** [`stage-c-2-gpu-h200 (1) / run`](https://github.com/radixark/miles/actions/runs/33322322332/job/99287424101).
- **Test:** `tests/e2e/sglang/test_session_server_multi_role/test_glm47.py`.
- **Root fingerprint:** `ValueError: Unsupported TP style 'mla_kv_a_proj' for Transformers backend.`
- **Consequence:** the failed SGLang scheduler leaves the harness waiting until `timeout after 1800s`.
- **Classification:** dependency compatibility gap in the old SGLang pin, not an accuracy threshold or timeout-budget failure.
- **Next step:** extend the H200-proven SGLang compatibility ref with the smallest upstream GLM-4.7 support delta, validate that focused model-resolution path, and trigger a fresh `synchronize` run. No timeout weakening.

fzyzcjy:
CI triage update for head `1ec8552275945f64b354f29d51eb19b34a873744`:

- ROCm job [stage-c-4-gpu-mi350 (1)](https://github.com/radixark/miles/actions/runs/33341465589/job/99337776530) timed out in `test_amd_mtp1_spec_v2_r3.py` after 2400 seconds.
- The other two selected tests passed.
- This matches the explicitly allowed ROCm exception for this acceptance run. No product code, test predicate, or timeout was weakened.
- The complete terminal log and SHA-256 completeness manifest were archived locally for audit.

The CUDA run remains under active observation and currently has no failing job.

This comment was generated by an AI agent.

fzyzcjy:
https://github.com/radixark/miles/pull/2407/commits/34a780705689adffedab69bb2f6bb162a78f0941

image

fzyzcjy:
/rerun-test tests/e2e/fsdp/test_qwen3_4B_fsdp_true_on_policy.py

fzyzcjy:
/rerun-test tests/e2e/megatron/test_qwen3_30B_A3B/test_baseline.py

fzyzcjy:
> 🤖 **Posted autonomously by Claude Code acting on the user's behalf.** Every failure below was classified from the complete job log, not from the check badge. Please push back if any conclusion is off.

## Head `4b6d3e115f` — three red checks, none is a code regression

| Lane | Failing test | Fingerprint (from the full log) | Classification | Evidence |
|---|---|---|---|---|
| [`stage-c-4-gpu-mi350 (0)`](https://github.com/radixark/miles/actions/runs/33416839609/job/99569499784) | `tests/e2e/megatron/model_scripts/test_inkling_small_4layer_ci.py` | sglang startup: `ValueError: Cannot find model module. 'InklingForConditionalGeneration' is not a registered model` — the CI retry classifier mislabels it as `OOM` | **Pre-existing on ROCm**: the mi350 image's sglang has no Inkling model. The same test failed identically on the previous deliver-1 head `34a7807056` ([job 99394329710](https://github.com/radixark/miles/actions/runs/33361738858/job/99394329710)); reproduced 5/5 tonight. | not caused by this branch |
| [`stage-c-2-gpu-h200 (1)`](https://github.com/radixark/miles/actions/runs/33416839635/job/99572272534) | `tests/e2e/fsdp/test_qwen3_4B_fsdp_true_on_policy.py` | `train.py:53 check_weights → sglang_api_client._make_request("weights_checker") → httpx.ReadError` right after a weight update, while the engine still answered `GET /health 200` in the same second | **Transport flake**: same shard passed on the previous head `13f9047afe` ([run 33401213041](https://github.com/radixark/miles/actions/runs/33401213041)), and the three commits in between do not touch this path. | `/rerun-test` of the file **passed** — [run 33426572708](https://github.com/radixark/miles/actions/runs/33426572708) |
| [`stage-c-4-gpu-h200 (0)`](https://github.com/radixark/miles/actions/runs/33416839635/job/99572272653) | `tests/e2e/megatron/test_qwen3_30B_A3B/test_baseline.py` | raylet: `A worker died or was killed while executing a task by an unexpected system error`; the `train` call got `ActorUnavailableError: RPC error: Socket closed`; FT confirmed the trainer cell dead after 131 s and gave up (`all alive cells failed`). The engine kept serving `/health 200` throughout. | **Infra flake** (trainer worker killed by the runner; the job was already on a runner-retried attempt). Same shard passed on `13f9047afe` ([run 33401213041](https://github.com/radixark/miles/actions/runs/33401213041)). | `/rerun-test` of the file **passed** in 18m21s — [run 33441158958](https://github.com/radixark/miles/actions/runs/33441158958) |

Everything else on this head is green: pre-commit, `stage-a-cpu (0..3)`, `stage-b-cpu`, `stage-b-2-gpu-h200`, `stage-c-2-gpu-h200 (0)`, `stage-c-4-gpu-h200 (1,2)`, `stage-c-8-gpu-h200 (0,1)`, `stage-c-4-gpu-mi350 (1)`. Still pending: `stage-c-8-gpu-h100 (0,1)` (queued behind H100 capacity most of the night).

Logs archived locally under `~/domains/human/artifacts/miles/2026-08-31-deliver1-m39-ci/`.

fzyzcjy:
/rerun-test tests/e2e/megatron/test_qwen3_5_35B_A3B_mtp/test_mtp1_spec_v2_r3.py

fzyzcjy:
image

fzyzcjy:
## CI status on `06a03ae4ab` (rebase round 4)

| Workflow | Run | Result |
| --- | --- | --- |
| pre-commit | [33730558857](https://github.com/radixark/miles/actions/runs/33730558857) | success |
| PR Test | [33730559131](https://github.com/radixark/miles/actions/runs/33730559131) | 2 GPU shards red, everything else green |
| PR Test (ROCm) | [33730559059](https://github.com/radixark/miles/actions/runs/33730559059) | failure (out of scope for this PR) |

All CPU work is green: `stage-a-cpu (0..3)` and `stage-b-cpu`. The previous round's
`fixture 'train_one_step_env' not found` in `stage-a-cpu (1)` is fixed and confirmed by CI.

Two GPU shards are red. They are unrelated to each other.

### 1. `stage-c-2-gpu-h200 (0)` — pre-existing breakage on `main`, not caused by this PR

`tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py` aborts before doing any work:

```
File ".../test_session_server_multi_role/_common.py", line 81, in run_one
    raise ValueError("Anthropic per-model verification requires an intermediate-system expectation")
```

Evidence that this is broken on `main` itself, independent of this PR:

- The whole `tests/e2e/sglang/test_session_server_multi_role/` directory is **byte-identical** between this PR's tip and its base `6b7ac37283` (`git diff 6b7ac37283 HEAD -- <dir>` is empty). This PR never touches these files.
- `_common.py` has required `anthropic_intermediate_system_expectation` for every Anthropic run since #2711 (merged 2026-08-29).
- `test_qwen38small.py` was added by #2760 (merged 2026-08-31) **without** that field. #2760 branched before #2711, so its own branch had no such check — a semantic merge conflict that neither side could see.
- #2760's own CI ([run 33139416968](https://github.com/radixark/miles/actions/runs/33139416968)) has `stage-c-2-gpu-h200` **skipped**, so the new test never actually ran before merging.

Fix (one line, in the test config): the expected value is derivable, not a guess.
`_verify_intermediate_system()` returns `route_supports and "system" in fixed_template_append_roles(tito_model)`,
and `Qwen38SmallTITOTokenizer.FIXED_TEMPLATE.allowed_append_roles` is `{"tool", "user", "assistant"}` —
no `system`. So the contract is `"forbidden"`, matching the sibling `test_qwen35.py` / `test_qwen36.py`,
whose templates also omit `system`.

```python
anthropic_intermediate_system_expectation="forbidden",
```

Root cause of the class of bug: no test enumerates the real `ModelConfig` objects in that directory,
so a model config missing a required field is only caught by a GPU E2E lane that #2760 skipped.

### 2. `stage-c-4-gpu-h200 (1)` — `test_qwen36.py`, under investigation

The `v2` + OpenAI leg fails on a hard TITO gate:

```
AssertionError: tito_session_mismatch_rate/v2/special_token_count=0.0303 must be 0
```

`0.0303` is 1 sample out of 33. Notes so far:

- `test_qwen36.py` is also untouched by this PR.
- It **passed** on this PR's pre-rebase tip `6b8f242b6a` in the same `stage-c-4-gpu-h200 (1)` shard
  ([run 33586187008](https://github.com/radixark/miles/actions/runs/33586187008), elapsed 902s), so this
  appeared when the chain was rebased onto the newer `main`.
- The lane runs EAGLE speculative decoding (`enable_spec=True`), so a single-sample gate breach may be
  nondeterministic rather than a hard regression.
- `use_addition_r3` semantics changed upstream in #2834 (`in_place` only → everything except `retract`).
  The rebased tip carries the new upstream semantics verbatim
  (`self.use_addition_r3 = config.pause_generation_mode != "retract"`), so the rebase did not drop it.
  That path additionally requires `return_routed_experts`, which this lane does not set, so it is
  probably not the cause.

Next step: re-run the failed job on the same SHA to separate flake from regression before changing anything.

`ServerCell ... was garbage collected without dispose()` lines in that log are teardown noise after the
job already failed, not the cause.

fzyzcjy:
/rerun-test tests/e2e/sglang/test_session_server_multi_role/test_qwen36.py

fzyzcjy:
/rerun-test tests/e2e/megatron/test_qwen3_5_35b_a3b_lora_ci.py

fzyzcjy:
## Follow-up: all three reds resolved, new head `016b31c8c5`

I re-ran the two suspicious files as single-file fresh runs (`/rerun-test`) against the exact
head `06a03ae4ab`. They gave opposite answers, and one of them **corrects my previous comment**.

| File | Rerun | Verdict |
| --- | --- | --- |
| `test_qwen36.py` | [33743331503](https://github.com/radixark/miles/actions/runs/33743331503) success | flaky test |
| `test_qwen3_5_35b_a3b_lora_ci.py` | [33743312591](https://github.com/radixark/miles/actions/runs/33743312591) failure | missing dependency pin, **not** infra |

### `test_qwen36.py` — flake, no code change

The rerun passed on the same SHA with `rollout/tito_session_mismatch_rate/v2/special_token_count: 0.0`,
against `0.0303` (1 sample in 33) on the first attempt. Same lane, same command, different outcome.
The lane runs EAGLE speculative decoding. Treating it as flaky rather than patching anything.

### `test_qwen3_5_35b_a3b_lora_ci.py` — correction to my earlier comment

I previously called this infra because the shard log ends in `timeout after 1800s`. That was wrong.
The single-file rerun shows the actual error, which is also present in the original shard log but
buried tens of thousands of lines up:

```
default_worker.py: error: argument --lora-target-modules: invalid choice: 'out_proj'
  (choose from 'q_proj', ..., 'in_proj_qkvz', 'in_proj_ba', 'all')
```

argparse rejects the argument outright, the process then hangs until the 1800s timeout.

Root cause is the removal of the `ci-sglang-pr:` pin from this PR's description:

| Run | Resolved SGLang | Checked out | Result |
| --- | --- | --- | --- |
| pre-rebase tip `6b8f242b6a`, fully green [33586187008](https://github.com/radixark/miles/actions/runs/33586187008) | `d4b1974323...` | `d4b1974 Allow GDN LoRA targets in the CLI` | PASS, 771s |
| `06a03ae4ab` after unpinning | `sglang-miles` | `c16b821 [sglang-miles] Streamed LoRA weight updates...` | timeout 1800s |

`sglang-miles` does carry #37466, which restored the GDN `in_proj` target choices — visible in the
`choose from` list above. What it still lacks is `out_proj`, and
`scripts/run_qwen3_5_35b_a3b_lora.py:61` passes exactly `f"{_LAYERS}.self_attention.out_proj"` as a
LoRA target. Both that script and the test are upstream `main` files this PR never touches, so
upstream `main` currently needs an SGLang newer than `sglang-miles` for this lane.

Fix: restored `ci-sglang-pr: d4b1974323d8640f6dc2dcebe88660023699a7b6` in the PR description, with the
reason written there. The Megatron pin stays off — `miles-main` already resolves to `235952df6`, which
is the commit that used to be pinned, so unpinning it changed nothing.

### `test_qwen38small.py` — fixed in the chain

Folded `anthropic_intermediate_system_expectation="forbidden"` into the owning op (chain position 79),
not as a new commit on the tip. New tip `016b31c8c5`: 236 commits, no conflict markers,
`git diff --stat 06a03ae4ab HEAD` is exactly that one file `+1`, `range-diff` shows `!` only at
position 79, and `pre-commit run --all-files` is clean.

tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py
```diff
@@ -17,6 +17,7 @@
     mamba_full_memory_ratio=4.59,
     cycles=2,
     tool_call_failure_mode="append_tool",
+    anthropic_intermediate_system_expectation="forbidden",
```
fzyzcjy, line 23: **Why this line is added.**

Without it `stage-c-2-gpu-h200 (0)` fails before running anything:

```
File ".../test_session_server_multi_role/_common.py", line 81, in run_one
    raise ValueError("Anthropic per-model verification requires an intermediate-system expectation")
```

`_common.py` has required this field for every Anthropic run since #2711 (merged 2026-08-29).
`test_qwen38small.py` was added by #2760 (merged 2026-08-31) without it, because #2760 branched
before that check existed, and #2760's own CI had `stage-c-2-gpu-h200` **skipped**, so the new
test never actually ran. Nothing in the repo enumerates the real `ModelConfig` objects in this
directory, so a missing required field is only observable from this GPU lane.

**This is not a bug this PR introduced.** `git diff 6b7ac37283 HEAD -- tests/e2e/sglang/test_session_server_multi_role/`
was empty before this line: the whole directory was byte-identical to the base. The chain never
touched these files; the rebase just moved onto a `main` where the lane finally runs.

**The value is derived, not guessed.** `_verify_intermediate_system()` is

```python
return route_supports and "system" in fixed_template_append_roles(tito_model)
```

and `Qwen38SmallTITOTokenizer.FIXED_TEMPLATE.allowed_append_roles` is `{"tool", "user", "assistant"}` —
no `"system"`. So the live capability is always `False`, and `_assert_intermediate_system_expectation`
only passes when the contract says `"forbidden"`. Picking `"required"` would swap the immediate
`ValueError` for an assertion failure later in the run.

Cross-check: the sibling models whose fixed templates also omit `"system"` — `test_qwen35.py` and
`test_qwen36.py` — both declare `"forbidden"`. The ones that allow appending `system`
(`test_qwen3.py`, `test_glm47.py`, `test_inkling.py`, `test_nemotron3.py`) declare `"required"`.

Landed by amending chain position 79, the op that already absorbed the earlier
upstream-new-consumer fixes, rather than as a new commit on the tip.

fzyzcjy:
image

PR #2414, opened 2026-08-12
Stop printing from the launch command layer

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils.py
+++ b/miles/utils/external_utils/command_utils.py
@@ -5,6 +5,7 @@ This file is not for miles framework itself, but as an optional utility to easil
 import base64
 import datetime
 import json
+import logging
 import os
 import random
 import shlex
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils.py
```diff
@@ -254,7 +257,7 @@ def check_has_nvlink():
 
 def get_default_wandb_args(test_file: str, run_name_prefix: str | None = None, run_id: str | None = None):
     if not os.environ.get("WANDB_API_KEY"):
-        print("Skip wandb configuration since WANDB_API_KEY is not found")
+        logger.info("Skip wandb configuration since WANDB_API_KEY is not found")
```
fzyzcjy, line 277: [P2] Initialize logging before the WandB notice

When `WANDB_API_KEY` is absent on a train-only entrypoint such as `scripts/amd/run_glm5_2_744b_a40b.py train` or `scripts/amd/run_inkling.py train`, this helper runs before any `exec_command_*` call. The only new logger setup is inside `_exec_command()`, so the default WARNING threshold discards this INFO record and the previous user-visible notice vanishes. Deliver-2 still has these call orders, so initialize the launcher logger before assembling `train_args`.

fzyzcjy (commented):
Logging is initialized too late for a previously unconditional user-facing message on train-only launcher paths. The regression remains in the designated safe deliver.

PR #2415, opened 2026-08-12
Turn command_utils into a package

fzyzcjy:
Part of #1837


miles/utils/external_utils/command_utils/__init__.py
```diff
@@ -0,0 +1,57 @@
+"""
+This package is not for miles framework itself, but as an optional utility to easily launch miles jobs and tests.
+"""
+
+from miles.utils.external_utils.command_utils.base_backend import (
```
fzyzcjy, line 5: [P1] Preserve the NPU patch across the package split

When an NPU user follows `docker/npu_patch/readme.md:116`, `git apply ../npu_patch/miles.patch` now fails with `command_utils.py: No such file or directory`, because its `execute_train_npu` hunk still targets the module removed by this split; `scripts/run_qwen3_4b_npu.py:4` therefore never becomes importable. The patch remains unchanged at deliver-2 (`f5a18edf`), so this is still a blocker: retarget the hunk to the package implementation and re-export `execute_train_npu` here.

fzyzcjy (commented):
The package split breaks the documented NPU patch application, and the same stale patch and absent module remain in deliver-2. This leaves an existing launcher unusable on its supported setup.

PR #2416, opened 2026-08-12
Generate every chart's values schema from one set of python types

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The generated production schemas preserve the prior validation semantics, and no in-scope regression introduced by this op remains at deliver-2.

PR #2417, opened 2026-08-12
Let a checkout on shared storage override the code baked into the image

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the deliver-2 safe point.

PR #2418, opened 2026-08-12
Hand the workbench's own infra values to every run launched from it

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-workbench/templates/infra-configmap.yaml
+++ b/charts/miles-workbench/templates/infra-configmap.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "miles-workbench.infraConfigMapName" . | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "miles-workbench.labels" . | nindent 4 }}
+data:
+  {{ include "miles-workbench.infraValuesFileName" . }}: |
+    infra:
+      {{- toYaml .Values.infra | nindent 6 }}
--- a/charts/miles-workbench/templates/statefulset.yaml
+++ b/charts/miles-workbench/templates/statefulset.yaml
@@ -29,17 +29,22 @@ spec:
           image: {{ include "miles-common.image" . }}
           imagePullPolicy: {{ .Values.infra.image.pullPolicy | quote }}
           command: ["sleep", "infinity"]
-          {{- with include "miles-common.env" . }}
+          {{- with include "miles-workbench.env" . }}
           {{- . | nindent 10 }}
           {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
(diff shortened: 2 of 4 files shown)


charts/miles-workbench/templates/statefulset.yaml
```diff
@@ -29,17 +29,22 @@ spec:
...
+              mountPath: {{ include "miles-workbench.infraValuesDir" . | quote }}
+            {{- with $mounts }}
             {{- . | nindent 12 }}
-          {{- end }}
-      {{- with include "miles-common.sharedStorageVolume" . }}
+            {{- end }}
       volumes:
+        - name: infra-values
+          configMap:
+            name: {{ include "miles-workbench.infraConfigMapName" . | quote }}
```
fzyzcjy, line 49: [P2] Wait for upgraded infra before reporting readiness

When an existing workbench is upgraded with a ConfigMap-only change such as `infra.paths.runsRoot`, this volume reference and the pod template remain unchanged. In deliver-2, `install()` runs `helm upgrade --install` and then `kubectl rollout status`; that returns immediately for the unchanged StatefulSet even though projected ConfigMap updates are only eventual, so a launch after a successful upgrade can read the previous `infra.yaml`. Tie the ConfigMap content to the pod revision or explicitly wait until the mounted file matches.

charts/miles-workbench/templates/infra-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
```
fzyzcjy, line 2: [P2] Check ConfigMap permissions before installing

Adding this object requires the account running Helm to manage ConfigMaps. Deliver-2's `CHART_RULES` checks StatefulSets and Helm Secrets but not ConfigMaps; when `--no-rbac` uses pre-created identities, or the caller has escalate/bind without ConfigMap CRUD, preflight reports success and `helm upgrade --install` then fails on this object. Add ConfigMap verbs to the installer-side preflight.

charts/miles-workbench/templates/statefulset.yaml
```diff
@@ -29,17 +29,22 @@ spec:
...
           volumeMounts:
+            - name: infra-values
+              mountPath: {{ include "miles-workbench.infraValuesDir" . | quote }}
+            {{- with $mounts }}
             {{- . | nindent 12 }}
-          {{- end }}
-      {{- with include "miles-common.sharedStorageVolume" . }}
+            {{- end }}
       volumes:
+        - name: infra-values
```
fzyzcjy: [P2] Reserve the internal volume name and mount path

At deliver-2 the shared infra schema accepts either an `infra.volumes` mount at `/etc/miles` or a volume named `infra-values`. These fixed entries then create duplicate `mountPath` or `spec.volumes[].name` values, which the Kubernetes API rejects, so otherwise schema-valid cluster values cannot install the workbench. Reserve or validate both internal identifiers.

fzyzcjy (commented):
The handoff leaves production failures that are still present at deliver-2: stale values after upgrades, a false-positive RBAC preflight, and schema-valid collisions with internal volume identifiers. The op is therefore not safe to deliver as-is.

PR #2419, opened 2026-08-12
Give the namespace one account that may uninstall a run

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-workbench/templates/uninstaller-rbac.yaml
+++ b/charts/miles-workbench/templates/uninstaller-rbac.yaml
@@ -1,0 +1,36 @@
+{{- if .Values.rbac.create }}
+{{- $name := .Values.uninstaller.serviceAccount }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $name | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "miles-workbench.labels" . | nindent 4 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ $name | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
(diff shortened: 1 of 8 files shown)


charts/miles-workbench/values.yaml
```diff
@@ -38,3 +38,6 @@ rbac:
 
 serviceAccount:
   name: ""
+
+uninstaller:
+  serviceAccount: miles-uninstaller
```
fzyzcjy: [P1] Carry the configured account into run releases

When the workbench is installed with `--set uninstaller.serviceAccount=custom`, runs launched from it do not inherit that value at deliver-2: `infra-configmap.yaml` exports only `infra`, while `charts/miles-run/values.yaml` still selects `miles-uninstaller`. On a clean namespace the escape Job therefore references a nonexistent ServiceAccount and the finished release retains its workloads; either make this name fixed or propagate it into every run.

charts/miles-workbench/templates/uninstaller-rbac.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.rbac.create }}
+{{- $name := .Values.uninstaller.serviceAccount }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $name | quote }}
```
fzyzcjy, line 6: [P2] Reject identical workbench and uninstaller account names

When `serviceAccount.name` equals `uninstaller.serviceAccount`—for example `--set serviceAccount.name=miles-uninstaller`—`serviceaccount.yaml` and this template emit the same ServiceAccount twice. The schema accepts this configuration, so a fresh Helm install fails when creating the duplicate object; deliver-2 adds no guard, so reject the collision before rendering.

fzyzcjy (commented):
Deliver-2 still leaves the configurable uninstaller name disconnected from run releases, and accepted account-name collisions can make the chart fail to install. These are persistent functional defects rather than transient chain states.

PR #2420, opened 2026-08-12
Give the workbench the actions that talk to a cluster

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/miles_workbench/actions.py
+++ b/miles/utils/external_utils/miles_workbench/actions.py
@@ -1,0 +1,115 @@
+from __future__ import annotations
+
+import logging
+import shutil
+
+from miles.utils.external_utils.command_utils.common import run_process
+from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Helm, Kubectl
+from miles.utils.external_utils.miles_workbench.naming import CHART_DIR, PACKAGE, object_name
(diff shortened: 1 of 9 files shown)


miles/utils/external_utils/miles_workbench/actions.py
```diff
@@ -0,0 +1,115 @@
...
+
+    logger.error("FAIL  could not read namespace %s (%s)", namespace, output)
+    raise SystemExit(1)
+
+
+def _probe_namespace_from_inside(namespace: str) -> None:
+    result = Kubectl.run_raw("get", "serviceaccounts", "-n", namespace, "-o", "name")
+    output = (result.stdout + result.stderr).strip()
+    if result.returncode == 0:
+        return
```
fzyzcjy, line 72: [P2] Do not infer namespace existence from a collection LIST

When cluster-scoped `get namespaces` is forbidden but the account has namespaced `list serviceaccounts` rights, Kubernetes returns an empty successful collection LIST even when this Namespace does not exist. The unconditional return therefore treats the missing namespace as present; preflight can pass and `helm upgrade` then fails because no namespace was created. Use an operation that validates the Namespace, or treat an empty LIST as inconclusive.

fzyzcjy (commented):
The namespace fallback relies on a namespaced collection LIST that does not prove the Namespace exists. This remains unchanged at the specified safe deliver and can produce a false preflight success followed by a Helm failure.

PR #2421, opened 2026-08-12
Ask the cluster what this account may do before installing anything

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/miles_workbench/preflight/checker.py
+++ b/miles/utils/external_utils/miles_workbench/preflight/checker.py
@@ -1,0 +1,102 @@
+from __future__ import annotations
+
+import logging
+import shutil
+import subprocess
+from typing import NamedTuple
+
+from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Kubectl
+from miles.utils.external_utils.miles_workbench.preflight.reporter import Reporter
+
+logger = logging.getLogger(__name__)
+
+_ROLES_RESOURCE = "roles.rbac.authorization.k8s.io"
+
+
+class _Answer(NamedTuple):
+    ok: bool
+    output: str
+
+
+class Checker:
+    def __init__(self, namespace: str) -> None:
+        self.namespace = namespace
+        self._reporter = Reporter()
+        self._answered: dict[tuple[str, str], bool] = {}
+
+    @property
+    def failed(self) -> bool:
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/miles_workbench/preflight/rules.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+CHART_RULES: dict[str, tuple[str, ...]] = {
+    "statefulsets.apps": ("create", "delete", "get", "patch"),
+    "secrets": ("create", "delete", "get", "list", "update"),
```
fzyzcjy, line 6: [P2] Include ConfigMap verbs in the install permission set

When `rbac.create=false`, the integrated runner uses only `CHART_RULES` for install rights, but the chart still always renders `charts/miles-workbench/templates/infra-configmap.yaml`. An account with the listed StatefulSet and Secret rights but no ConfigMap lifecycle rights therefore passes preflight and fails only when `helm upgrade --install` reaches that object, defeating the promise to catch missing permissions before installation; add the required ConfigMap verbs here.

miles/utils/external_utils/miles_workbench/preflight/checker.py
```diff
@@ -0,0 +1,102 @@
...
+        self._reporter.report(ok, message, counted)
+
+    def warn(self, message: str) -> None:
+        self._reporter.warn(message)
+
+    def check_binary(self, binary: str) -> None:
+        self.report(shutil.which(binary) is not None, f"{binary} is installed", counted=False)
+
+    def check_cluster_reachable(self) -> None:
+        answer = self._query("get", "--raw", "/version")
```
fzyzcjy: [P2] Authenticate the cluster probe

On clusters that grant anonymous access to `/version`, as the standard Kubernetes public-info binding does, a context with no credentials returns success here, so the checker skips its credential-error path and misreports subsequent permission-query failures as missing RBAC. Conversely, a hardened cluster can deny this URL to an otherwise valid namespaced account; use a request or identity check that actually authenticates the caller rather than treating `/version` as proof that credentials were accepted.

fzyzcjy (commented):
The permission inventory omits an object the chart always installs, and the reachability check cannot establish the authentication property it claims. Both behaviors remain in deliver-2, so they are blockers rather than transient issues.

PR #2422, opened 2026-08-12
Check the objects, the API and the namespace, and let the verdict gate the install

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/miles_workbench/preflight/checker.py
+++ b/miles/utils/external_utils/miles_workbench/preflight/checker.py
@@ -6,7 +6,22 @@ import subprocess
 from typing import NamedTuple
 
 from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Kubectl
+from miles.utils.external_utils.miles_workbench.options import InstallArgs
 from miles.utils.external_utils.miles_workbench.preflight.reporter import Reporter
+from miles.utils.external_utils.miles_workbench.preflight.rules import (
+    AVAILABLE_CONDITION,
+    CHART_FAMILY,
+    CLUSTER_PROVIDED_RESOURCES,
+    DEFAULT_TOKEN_PREFIX,
+    LWS_API_GROUP,
+    LWS_CONTROLLER_DEPLOYMENT,
+    LWS_CONTROLLER_NAMESPACE,
+    LWS_RESOURCE,
+    MANAGED_BY,
+    NAMESPACE_KINDS,
+    UNSERVED_RESOURCE_MARKERS,
+)
+from miles.utils.external_utils.miles_workbench.render import RbacPlan, rbac_plan_of, render_chart
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


miles/utils/external_utils/miles_workbench/preflight/runner.py
```diff
@@ -0,0 +1,83 @@
...
+
+    plan = checker.check_rbac_plan(args)
+    if plan.grants_leader_worker_sets:
+        checker.check_leader_worker_set_api()
+        checker.check_leader_worker_set_controller()
+
+    if plan.creates_role:
+        _check_it_may_grant_the_role(checker, args, plan=plan)
+    else:
+        checker.check_rules("install the chart", CHART_RULES)
```
fzyzcjy: [P1] Check ConfigMap rights before starting Helm

When `rbac.create=false`, this is the only install-permission set consulted, but `charts/miles-workbench/templates/infra-configmap.yaml` always renders a ConfigMap while `CHART_RULES` contains no ConfigMap verbs. An installer lacking ConfigMap access can therefore pass preflight and then be rejected when Helm applies that object; include the required ConfigMap operations in the unconditional chart checks.

fzyzcjy (commented):
The new gate can falsely pass missing install or delegation rights and emits false namespace-sharing warnings during normal upgrades. These defects remain present at the target safe deliver.

miles/utils/external_utils/miles_workbench/preflight/runner.py
```diff
@@ -0,0 +1,83 @@
...
+def _check_it_may_grant_the_role(checker: Checker, args: InstallArgs, *, plan: RbacPlan) -> None:
+    checker.check_rules("install the chart", CHART_RULES, CHART_SERVICE_ACCOUNT_RULES, CHART_RBAC_RULES)
+
+    message = f"may grant the workbench its Role in namespace {args.namespace}"
+
+    if not (denied := checker.denied_rules(plan.granted_rules)):
+        checker.report(True, message)
+        return
+
+    if checker.may_delegate_rules_it_does_not_hold(object_name(args.release)):
```
fzyzcjy: [P2] Validate delegation against every rendered Role

The chart renders both the release Role and a separately named uninstaller Role, but `rbac_plan_of` merges their grants and this fallback checks name-scoped `escalate`/`bind` only on `object_name(args.release)`. If a caller lacks, for example, `delete deployments.apps` granted only by the uninstaller Role but has delegation for the release Role, preflight returns success even though Kubernetes rejects the uninstaller Role or RoleBinding; retain and validate each rendered Role separately.

miles/utils/external_utils/miles_workbench/preflight/checker.py
```diff
@@ -52,6 +67,90 @@ def check_cluster_reachable(self) -> None:
...
+        where = f" in namespace {namespace}" if namespace else ""
+        message = f"{kind} {name} exists{where}"
+        if not answer.ok:
+            self._report_failed_query(message, answer, unverifiable="(Forbidden)" in answer.output, counted=False)
+
+    def check_namespace_holds_only(self) -> None:
+        message = f"namespace {self.namespace} holds nothing but Miles releases"
+        family = ",".join(CHART_FAMILY)
+        selectors = [
+            f"app.kubernetes.io/managed-by!={MANAGED_BY}",
```
fzyzcjy: [P2] Exempt Miles Helm records from foreign-object warnings

This selector also matches Secrets lacking the standard app label. Helm's `sh.helm.release.v1.*` storage Secrets use `owner=helm` rather than `app.kubernetes.io/managed-by=Helm`, and `_is_cluster_provided` does not exempt them, so every reinstall or namespace containing a live Miles Helm release is falsely reported as foreign; account for Miles release records while retaining warnings for genuinely foreign releases.

PR #2423, opened 2026-08-12
Give the workbench a command line

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the target safe deliver. Intermediate issues were repaired before that point and excluded as requested.

PR #2425, opened 2026-08-12
Skip the values a CLI action cannot render

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Skipping an unrenderable zero-arity action remains guarded by the parsed-object round-trip, preventing silent configuration drift. The target safe deliver retains these semantics without an unresolved correctness regression.

PR #2426, opened 2026-08-12
Make launching a training run a method on a backend object

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The extracted Ray launch behavior remains equivalent, and no correctness regression introduced by this op survives in the target safe deliver.

PR #2427, opened 2026-08-12
Read and rewrite train argv through one small helper

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/base_backend.py
+++ b/miles/utils/external_utils/command_utils/base_backend.py
@@ -10,6 +10,7 @@ from functools import partial
 from pathlib import Path
 
 from miles.utils.external_utils.command_utils.common import (
+    ArgvManipulator,
     MOONCAKE_MASTER_LOG_PATH,
     MOONCAKE_MASTER_METRICS_PORT,
     MOONCAKE_MASTER_PORT,
@@ -75,7 +76,8 @@ class BaseCommandBackend(ABC):
         if not os.path.isabs(train_script):
             train_script = f"{repo_base_dir}/{train_script}"
 
-        train_backend_fsdp = "--train-backend fsdp" in train_args
+        train_argv = shlex.split(train_args)
+        train_backend_fsdp = "fsdp" in ArgvManipulator.values_of(train_argv, "--train-backend")
         assert train_backend_fsdp == (megatron_model_type is None)
 
         self._execute_train_inner(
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/command_utils/base_backend.py
```diff
@@ -78,7 +79,8 @@ def execute_train(
         if not os.path.isabs(train_script):
             train_script = f"{repo_base_dir}/{train_script}"
 
-        train_backend_fsdp = "--train-backend fsdp" in train_args
+        train_argv = shlex.split(train_args)
+        train_backend_fsdp = "fsdp" in ArgvManipulator.values_of(train_argv, "--train-backend")
```
fzyzcjy, line 86: [P2] Derive the backend from the effective declaration

When `train_args` contains `--train-backend=fsdp --train-backend megatron` with a non-`None` Megatron model type, argparse uses the final `megatron`, but this membership check sets `train_backend_fsdp` to true and the following assertion rejects an otherwise valid launch. The parent accepted this form, and deliver-2 retains the membership check; use the final value or explicitly reject conflicting repeats to preserve the argparse contract required by `.claude/rules/general-code-style.md:12-15`.

miles/utils/external_utils/command_utils/common.py
```diff
@@ -86,6 +86,36 @@ def get_env_enable_infinite_run():
...
+    def with_flag(argv: list[str], flag: str, value: str) -> list[str]:
+        if ArgvManipulator.declares(argv, flag):
+            return list(argv)
+        return [*argv, flag, value]
+
+    @staticmethod
+    def replacing_value(argv: list[str], flag: str, value: str) -> list[str]:
+        assert flag in argv, f"{flag} is not among the arguments, so there is no value of it to replace"
+        rewritten = list(argv)
+        rewritten[rewritten.index(flag) + 1] = value
```
fzyzcjy, line 118: [P2] Rewrite the effective declaration in both argv forms

When a Kubernetes Mooncake launch supplies its launcher-local address through argparse's valid `--mooncake-store-init-kwargs=<json>` form, deliver-2's `MooncakeInfo.with_cluster_master` uses this rewriter, but `flag in argv` is false and the launch aborts. With repeated space-form declarations, `.index(flag)` rewrites the first while argparse and the pods consume the last, leaving the effective launcher-local address unchanged; rewrite the final occurrence in both accepted forms to preserve the framework contract required by `.claude/rules/general-code-style.md:12-15`.

fzyzcjy (commented):
The new argv helper mishandles argparse's last-wins and equals-form semantics in two production paths. Both defects remain in deliver-2, with no later safe deliver available.

PR #2428, opened 2026-08-12
Let a launch carry the commands that prepare the machines it runs on

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/ray_backend/backend.py
+++ b/miles/utils/external_utils/command_utils/ray_backend/backend.py
@@ -30,6 +30,9 @@ class RayCommandBackend(BaseCommandBackend):
                 f"ray start --head --node-ip-address {master_addr} --num-gpus {request.num_gpus_per_node} --disable-usage-stats"
             )
 
+        for cmd in request.prepare_cmd.values():
+            self.exec_command_multi_node(cmd)
+
         if (f := request.before_ray_job_submit) is not None:
             f()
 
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/ray_backend/backend.py
```diff
@@ -30,6 +30,9 @@ def _execute_train_inner(self, request: ExecuteTrainRequest) -> None:
                 f"ray start --head --node-ip-address {master_addr} --num-gpus {request.num_gpus_per_node} --disable-usage-stats"
             )
 
+        for cmd in request.prepare_cmd.values():
+            self.exec_command_multi_node(cmd)
```
fzyzcjy, line 34: [P1] Join Ray workers before preparing them

When a launcher combines `prepare_cmd` with the supported `before_ray_job_submit=ssh_start_ray_workers(...)` path, only the freshly started head is in Ray when `exec_command_multi_node` snapshots `ray.nodes()`. The callback then joins the workers too late, so trainers scheduled there can see missing node-local checkpoints or other unstaged state. Run the cluster-completion hook before this fan-out, or otherwise wait for every target node; deliver-2 retains this ordering, so the failure is not transient.

fzyzcjy (commented):
The new preparation path can omit worker nodes in the existing SSH fan-out launch flow because preparation runs before those workers join Ray. The target deliver retains the faulty ordering.

PR #2429, opened 2026-08-12
Split the typer wrapper into pieces a second caller can reuse

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The extracted helpers preserve the existing CLI parameter construction, dataclass instantiation, and argument-printing behavior, with no correctness or regression issue remaining at the target deliver.

PR #2430, opened 2026-08-12
Let a launch script read its own config out of the environment

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this commit remains at the target safe deliver.

PR #2431, opened 2026-08-12
Give every launch script the same three subcommands

fzyzcjy:
Part of #1837


scripts/run_deepseek.py
```diff
@@ -93,15 +92,15 @@ def _prepare_megatron_ckpt(args: ScriptArgs):
...
-    )
+def _prepare_cmd(args: ScriptArgs) -> dict[str, str]:
+    copies = [
+        U.rsync_cmd(
+            f"{args.model_dir}/{args.model_name}_torch_dist",
+            f"{args.model_local_dir}/{args.model_name}_torch_dist",
+        ),
+        U.rsync_cmd(f"{args.model_dir}/{args.model_name}", f"{args.model_local_dir}/{args.model_name}"),
+    ]
+    return {"trainer": " && ".join(copies)}
```
fzyzcjy, line 103: [P1] Stage checkpoints on inference workers too

When `--cluster-backend kubernetes` is used with `model_local_dir != model_dir` (the default here), this map attaches both rsyncs only to the trainer role. In deliver-2, `helm_backend/launcher/values/pool_entry.py::_with_prepare_cmd` skips inference-engine pods, while `compute_engine_launch_cmd` passes the local HF path to SGLang as `model_path`; disaggregated rollout nodes therefore never receive the model, and colocated inference pods can start before the trainer's rsync completes. Add inference-side staging; the same pattern affects the V4/GLM launchers.

scripts/run_inkling.py
```diff
@@ -412,41 +405,26 @@ def _train(args: ScriptArgs):
...
         megatron_path=args.megatron_path,
+        prepare_cmd=_prepare_cmd(args),
     )
 
 
 @app.command()
 @U.dataclass_cli
 def train(args: ScriptArgs):
-    """Run training. Assumes HF checkpoint / torch_dist are already staged."""
+    """Run training; each trainer pod first copies torch_dist to node-local NVMe when the two differ."""
```
fzyzcjy, line 420: [P1] Keep Inkling's CLI as a command group

After `prepare_cp` and `full_train` are removed, this app has exactly one registered command and no callback. Typer collapses such an app into a root command, so the newly documented `python scripts/run_inkling.py train ...` treats `train` as an unexpected argument; deliver-2 retains this shape. Restore the common commands or add a callback so `train` remains a subcommand.

scripts/run_deepseek_v4.py
```diff
@@ -357,23 +356,19 @@ def prepare_spmd(args: ScriptArgs):
     _prepare_spmd(args)
 
 
-@app.command()
-@U.dataclass_cli
-def prepare_cp(args: ScriptArgs):
-    _prepare_cp(args)
+def _prepare_cmd(args: ScriptArgs) -> dict[str, str]:
```
fzyzcjy, line 368: [P2] Add the common prepare command to both V4 launchers

At deliver-2, neither this launcher nor `scripts/amd/run_deepseek_v4.py` registers a `prepare` subcommand; both expose only granular `prepare-*` steps plus `train` and `full-train`. Callers relying on this op's advertised three-command contract still get `No such command 'prepare'`, so add a wrapper composing the existing preparation steps (`.claude/rules/launch-and-model-scripts.md:63-70`).

scripts/run_deepseek.py
```diff
@@ -299,27 +298,37 @@ def _execute_train(args: ScriptArgs, before_ray_job_submit=None):
...
+def prepare(args: ScriptArgs) -> None:
+    _prepare_download(args)
+    _prepare_bf16_ckpt(args)
     _prepare_megatron_ckpt(args)
-    _prepare_cp(args)
+
+
+@app.command()
+@U.dataclass_cli
+def train(args: ScriptArgs) -> None:
```
fzyzcjy, line 330: [P2] Update every caller of the renamed launcher roles

This split makes `train` submit-only and removes `prepare-cp`, but deliver-2 still documents the old interfaces in `docs/models/deepseek/deepseek.md:27-38`, `docs/models/glm/glm4-5.md:60-73`, `docs/models/deepseek/deepseek-v3-2.md:96-111`, and `docs/models/glm/glm5-2.md:39-69`. The first two commands no longer perform the documented full pipeline, while the latter two fail with an unknown subcommand; update all affected pages as required by `.claude/rules/launch-and-model-scripts.md:135-140`.

fzyzcjy (commented):
The patch leaves Kubernetes inference workers without staged node-local models and breaks Inkling's documented CLI. It also leaves the advertised command migration and its documentation incomplete, and these defects persist at deliver-2.

PR #2432, opened 2026-08-12
Have every launch script talk to a backend object

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/__init__.py
+++ b/miles/utils/external_utils/command_utils/__init__.py
@@ -1,18 +1,4 @@
-"""
-This package is not for miles framework itself, but as an optional utility to easily launch miles jobs and tests.
-"""
-
-from miles.utils.external_utils.command_utils.base_backend import (
-    ExecuteTrainConfig,
-    check_has_nvlink,
-    convert_checkpoint,
-    execute_train,
-    fp8_cast_bf16,
-    hf_download_dataset,
-    rsync_simple,
-    ssh_start_ray_workers,
-    start_mooncake_master,
-)
+from miles.utils.external_utils.command_utils.base_backend import ExecuteTrainConfig, default_config
 from miles.utils.external_utils.command_utils.common import (
     GENERATION_HARDWARE,
     NUM_GPUS_OF_HARDWARE,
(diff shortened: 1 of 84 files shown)


miles/utils/external_utils/command_utils/__init__.py
```diff
@@ -1,19 +1,8 @@
...
-"""
-
 from miles.utils.external_utils.command_utils.base_backend import (
     ExecuteTrainConfig,
-    check_has_nvlink,
-    convert_checkpoint,
-    execute_train,
-    fp8_cast_bf16,
-    hf_download_dataset,
+    default_config,
```
fzyzcjy, line 3: [P1] Migrate remaining launchers before deleting exports

After these exports are removed, `scripts/amd/run_glm5_2_744b_a40b.py` and `scripts/amd/run_inkling.py` still import this package as `U` and call `U.execute_train` plus checkpoint/download/rsync helpers, so their advertised prepare/train/full-train commands raise `AttributeError` before launching. Target deliver `f5a18edf2aa9` retains those calls; its opt-in `.legacy` module does not repair the unchanged imports. Migrate these launchers to `args.create_backend()` or retain the old surface (`.claude/rules/general-code-style.md:41-43`).

fzyzcjy (commented):
The patch leaves supported launchers calling symbols removed from `command_utils`, and it shadows a module alias with a backend while retaining module-only constant lookups. Both runtime failures remain in the specified safe deliver.

scripts/run_deepseek_v4.py
```diff
@@ -283,6 +287,7 @@ def _prepare_mxfp8(args: ScriptArgs):
     head/wo_a/ffn.gate/compressor/norms/embed and the DSA indexer weights_proj
     are all kept BF16 by SKIP_WEIGHT_SUBSTRINGS in tools/convert_hf_to_mxfp8.py.
     """
+    U = args.create_backend()
     if not args.rollout_mxfp8:
         return
     assert U.GENERATION_HARDWARE[args.hardware] == "Blackwell", "rollout_mxfp8 requires Blackwell"
```
fzyzcjy: [P1] Read hardware constants from command_utils

When `--rollout-mxfp8` is enabled, `U` is now a backend object, which has no `GENERATION_HARDWARE` attribute, so this path raises `AttributeError` before conversion; the same invalid lookup occurs in `_train` whenever either MXFP8 flag is enabled. Target deliver `f5a18edf2aa9` leaves both lookups unchanged, so use `command_utils.GENERATION_HARDWARE` for them.

PR #2433, opened 2026-08-12
Move the ray-only command helpers under the ray backend

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The moved helpers preserve their previous execution semantics, all in-scope production imports are updated, and no correctness regression from this op remains at deliver-2.

PR #2434, opened 2026-08-12
Start the mooncake master when a run asks for that object store

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/ray_backend/backend.py
+++ b/miles/utils/external_utils/command_utils/ray_backend/backend.py
@@ -4,12 +4,18 @@ import shlex
 
 from miles.utils.external_utils.command_utils.base_backend import BaseCommandBackend, ExecuteTrainRequest
 from miles.utils.external_utils.command_utils.common import (
+    MOONCAKE_BACKEND_NAME,
+    OBJECT_STORE_BACKEND_FLAG,
+    ArgvManipulator,
     _parse_extra_env_vars,
     _pythonpath_with_sources,
     get_bool_env_var,
     run_shell_command,
 )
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/ray_backend/backend.py
```diff
@@ -30,6 +36,11 @@ def _execute_train_inner(self, request: ExecuteTrainRequest) -> None:
                 f"ray start --head --node-ip-address {master_addr} --num-gpus {request.num_gpus_per_node} --disable-usage-stats"
             )
 
+        if MOONCAKE_BACKEND_NAME in ArgvManipulator.values_of(
+            shlex.split(request.train_args), OBJECT_STORE_BACKEND_FLAG
+        ):
+            start_mooncake_master()
```
fzyzcjy, line 43: [P1] Preserve a caller-configured Mooncake master

When a caller has already started Mooncake on a non-default port and supplies that endpoint through `--mooncake-store-init-kwargs` or `get_mooncake_object_store_args(master_port=...)`, this no-argument call probes only `127.0.0.1:50051`. It then reaches `command.py:94-97`, kills the configured `mooncake_master`, starts a replacement on port 50051, and submits unchanged arguments that still target the old port, so initialization fails and other users of that master may be interrupted. This violates the caller-managed endpoint contract in `docs/advanced/mooncake-rollout-transfer.md:21-23` and remains present in target deliver `f5a18edf`; derive the configured endpoint and ownership before starting a process.

fzyzcjy (commented):
The unconditional default-port startup can terminate a valid caller-managed Mooncake endpoint while leaving the job configured for the now-dead port. This regression remains unresolved at the target safe deliver.

PR #2435, opened 2026-08-12
Call a trainer pool a trainer engine

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the target safe deliver. Production consumers consistently derive trainer pool and cell IDs through the updated helper.

PR #2437, opened 2026-08-12
Install a run's workers, one workload per kind of pool

fzyzcjy:
Part of #1837


miles/utils/external_utils/command_utils/helm_backend/launcher/values/helm_values_types.py
```diff
@@ -36,6 +39,24 @@ def as_values(self) -> dict[str, Any]:
         return self.model_dump(by_alias=True, exclude_none=True)
 
 
+class PortEntry(ValuesModel):
+    name: Annotated[str, Field(min_length=1, max_length=_PORT_NAME_MAX)]
```
fzyzcjy, line 43: [P1] Validate Kubernetes port names before rendering

When PD disaggregation is enabled, the target deliver's `_port_name("disaggregation_bootstrap")` truncates the name to `disaggregation-`; this field accepts it and the new pool template copies it into `container.ports[].name`. Kubernetes requires an IANA service name to end in an alphanumeric character, so the worker Pod is rejected and the run never starts; sanitize before truncation and validate the result here.

fzyzcjy (commented):
The target deliver generates an invalid Kubernetes port name for PD-prefill engines, preventing their worker Pods from being created. This blocker remains present at the target deliver.

PR #2438, opened 2026-08-12
Compute a pool's chart entry from the worker spec that declares it

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/values/pool_entry.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/values/pool_entry.py
@@ -1,0 +1,170 @@
+from __future__ import annotations
+
+import shlex
+import sys
+
+from miles.utils.external_utils.command_utils.base_backend import TRAINER_ROLE
+from miles.utils.external_utils.command_utils.helm_backend import naming
+from miles.utils.external_utils.command_utils.helm_backend.launcher.values.helm_values_types import (
+    PoolEntry,
+    PortEntry,
+)
+from miles.utils.external_utils.command_utils.helm_backend.launcher.values.misc import (
+    SECTION_OF_CATEGORY,
+    TRAINER_ENGINES_SECTION,
+    LaunchPlan,
+)
+from miles.utils.workers.worker_provider.kubernetes.helm import env
+from miles.utils.workers.worker_spec import (
+    BaseWorkerSpec,
+    CommandWorkerSpec,
+    HostAndPort,
+    LaunchCommandContext,
+    NamedHostAndPorts,
+    ServeWorkerSpec,
+)
+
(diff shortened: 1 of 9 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/values/pool_entry.py
```diff
@@ -0,0 +1,170 @@
...
+    embedded = [argument for argument in argv if sentinel in argument and argument != sentinel]
+    assert not embedded, (
+        f"Spec '{spec.name}' builds {embedded} out of its pod index; kubelet substitutes a whole "
+        f"argument, so the index has to reach the command unchanged"
+    )
+    return [_WORKER_INDEX_PLACEHOLDER if argument == sentinel else argument for argument in argv]
+
+
+def _port_name(name: str) -> str:
+    return name.replace("_", "-")[:15]
```
fzyzcjy, line 180: [P1] Keep shortened port names valid for Kubernetes

When a PD-disaggregated prefill pool includes the existing `disaggregation_bootstrap` port, this returns `disaggregation-`. Kubernetes requires a named container port to end in an alphanumeric character, so the API server rejects the LeaderWorkerSet and the run cannot install; sanitize and validate the final shortened IANA service name.

miles/utils/external_utils/command_utils/helm_backend/launcher/values/pool_entry.py
```diff
@@ -0,0 +1,170 @@
...
+def build_entry(
+    spec: BaseWorkerSpec, plan: LaunchPlan, addresses: dict[str, dict[str, NamedHostAndPorts]]
+) -> PoolEntry:
+    assert spec.scheduling.num_cells > 0, (
+        f"Spec '{spec.name}' asks for {spec.scheduling.num_cells} cells; a spec a run has turned off is dropped "
+        f"before conversion, because a values entry always renders at least one pod"
+    )
+    context = _launch_context(
+        spec, addresses=addresses, cell_index=_RENDERED_CELL_INDEX, worker_in_cell_index=_WORKER_INDEX_SENTINEL
+    )
```
fzyzcjy, line 49: [P2] Use the StatefulSet replica's cell index

When `--num-session-servers > 1` on Kubernetes, this renders one StatefulSet command with cell index 0 for every replica. `spec_session_server` bakes it into `--instance-id`, so replicas 1..N expose `<run_uuid>-0` from `/health` while `router_manager` records `<run_uuid>-1`, etc. for those addresses; substitute `MILES_CELL_INDEX` or defer command construction into the pod.

fzyzcjy (commented):
The patch can emit an API-rejected LeaderWorkerSet for PD prefill engines and assigns the same cell identity to every multi-session-server replica. Both defects remain in the target safe deliver.

PR #2439, opened 2026-08-12
Assemble a run's values from its specs and its launch plan

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/values/builder.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/values/builder.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+from miles.utils.external_utils.command_utils.helm_backend import naming
+from miles.utils.external_utils.command_utils.helm_backend.launcher.values.helm_values_types import (
+    MilesRunChartValues,
+    ObjectNames,
+    OrchestratorSection,
+    PoolEntry,
+    RunValues,
+)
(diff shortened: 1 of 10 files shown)


fzyzcjy (commented):
Deliver-2 still emits an invalid Kubernetes port name for PD-prefill engines, preventing those workloads from starting.

miles/utils/external_utils/command_utils/helm_backend/launcher/values/builder.py
```diff
@@ -0,0 +1,90 @@
...
+            _assert_worker_ports_fit(spec)
+    addresses = _compute_addresses(specs, plan.release)
+
+    entries: dict[str, list[PoolEntry]] = {
+        STATIC_WORKERS_SECTION: [],
+        INFERENCE_ENGINES_SECTION: [],
+        TRAINER_ENGINES_SECTION: [],
+    }
+    for spec in specs:
+        entries[SECTION_OF_CATEGORY[spec.category]].append(build_entry(spec, plan=plan, addresses=addresses))
```
fzyzcjy, line 43: [P1] Keep truncated Kubernetes port names valid

When this path builds a PD-prefill engine, `build_entry()` shortens `disaggregation_bootstrap` to `disaggregation-`. Kubernetes requires a named container port to end with an alphanumeric character, so the API server rejects the generated LeaderWorkerSet and the run never starts. Strip trailing separators after truncation and validate the emitted name; the defect remains in deliver-2.

PR #2440, opened 2026-08-12
Write a run's verdict where the launcher can read it

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/naming.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/naming.py
@@ -1,11 +1,48 @@
 from __future__ import annotations
 
+import random
+import time
+from pathlib import Path
+
 from miles.utils.workers.worker_provider.kubernetes.helm.naming import CHART_NAME
 
 ORCHESTRATOR_COMPONENT = "orchestrator"
+
+_RUNS_DIR_NAME = "miles-runs"
+_STATE_DIR_NAME = "state"
+_VALUES_DIR_NAME = "values"
+_STATE_FILE_GLOB = "orchestrator-*.state"
 
 
 class RunNames:
     @staticmethod
     def release(*, run_id: str) -> str:
         return f"{CHART_NAME}-{run_id}"
+
+
+class RunFiles:
+    @staticmethod
+    def run_dir(*, shared_root: str | Path, run_id: str) -> Path:
+        return Path(shared_root) / _RUNS_DIR_NAME / run_id
+
+    @staticmethod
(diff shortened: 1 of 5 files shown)


miles/utils/external_utils/command_utils/helm_backend/naming.py
```diff
@@ -1,11 +1,48 @@
...
+
+    @staticmethod
+    def new_state_file(*, run_directory: str | Path) -> Path:
+        return _orchestrator_state_path(run_directory, _new_launch_token())
+
+    @staticmethod
+    def latest_state_file(*, run_directory: str | Path) -> Path | None:
+        """The newest launch's file, by the launch token in its name; see _new_launch_token."""
+        written = sorted((Path(run_directory) / _STATE_DIR_NAME).glob(_STATE_FILE_GLOB))
+        return written[-1] if written else None
```
fzyzcjy: [P2] Derive the latest generation before globbing state files

When a relaunch has allocated a fresh state path but its pod has not written it yet, such as during Pending or ImagePullBackOff, this glob sees only the prior generation and returns its verdict as the newest. In the target deliver, `collect_diagnosis_command` then copies that old verdict and does not report the current diagnosis as incomplete; determine the latest generation from a launcher-written marker or launch record before checking for its state file.

miles/utils/external_utils/command_utils/helm_backend/naming.py
```diff
@@ -1,11 +1,48 @@
...
+
+
+def _orchestrator_state_path(run_directory: str | Path, launch_token: str) -> Path:
+    return Path(run_directory) / _STATE_DIR_NAME / f"orchestrator-{launch_token}.state"
+
+
+def _new_launch_token() -> str:
+    return f"{time.strftime('%y%m%d-%H%M%S')}-{random.Random().randint(0, 999999):06d}"
```
fzyzcjy: [P2] Make launch tokens preserve creation order

When two launches of one run allocate paths in the same wall-clock second, they share the timestamp prefix and their independent random suffixes do not encode call order. Because `latest_state_file()` sorts these names lexicographically, it selects the older state about half the time; this remains in the target deliver and can attach manual diagnosis to the wrong verdict. Use a lexicographically monotonic per-launch identifier while retaining uniqueness.

fzyzcjy (commented):
The state-writing protocol is largely coherent, but the new latest-state discovery cannot reliably identify the newest launch. Both failure modes remain in deliver-2 and can cause diagnosis to report a previous generation's verdict.

PR #2441, opened 2026-08-12
Read a rendered manifest as objects rather than as a string

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_types.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_types.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from miles.utils.external_utils.command_utils.helm_backend.orchestrator.state import STATE_FILE_FLAG
+from miles.utils.pydantic_utils import FrozenOpenBaseModel, FrozenStrictBaseModel
+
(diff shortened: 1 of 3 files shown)


fzyzcjy (commented):
The parser applies the chart's workload schema to every rendered object. Deliver-2 still exposes this behavior to pass-through manifests, so valid custom resources can prevent an install.

miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_types.py
```diff
@@ -0,0 +1,74 @@
...
+    containers: list[Container] = []
+
+
+class PodTemplate(FrozenOpenBaseModel):
+    spec: PodSpec | None = None
+
+
+class ObjectSpec(FrozenOpenBaseModel):
+    replicas: int | None = None
+    template: PodTemplate | None = None
```
fzyzcjy, line 27: [P1] Keep unknown object specs out of the Pod schema

In deliver-2, an `extra_manifests` custom resource whose schema legitimately defines `spec.template` as something other than a Pod template—for example, a Knative Service with an unnamed container—is still validated through `ObjectSpec` and `PodSpec`; `Manifest.parse()` therefore raises before `Helm.upgrade()` can install it. Because the pass-through API accepts objects Miles does not model, preserve their raw bodies and decode pod fields only for known workload kinds.

PR #2442, opened 2026-08-12
Run the orchestration script under a wrapper that publishes its verdict

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the specified safe-deliver commit.

PR #2443, opened 2026-08-12
Wait for a run's verdict from outside the cluster

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/orchestrator/observer.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/orchestrator/observer.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import logging
+import time
+from collections.abc import Callable
+from pathlib import Path
+
+from miles.utils.external_utils.command_utils.helm_backend.orchestrator.state import OrchestratorState
+from miles.utils.pydantic_utils import FrozenStrictBaseModel
+
+logger = logging.getLogger(__name__)
+
+_POLL_INTERVAL_SECONDS = 10.0
+
+_DEAD_POD_PHASES = frozenset({"Failed", "Succeeded"})
+_UNREADABLE_PHASE = "Unknown"
+_MISSING_POD_POLLS = 3
+_NO_VERDICT_EXIT_CODE = 1
+
+
+class _RunOutcome(FrozenStrictBaseModel):
+    exit_code: int
+    reason: str
+
+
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/helm_backend/orchestrator/observer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    if phase is None:
+        if state is None or missing_polls < _MISSING_POD_POLLS:
+            return None
+        return _RunOutcome(
+            exit_code=_NO_VERDICT_EXIT_CODE,
+            reason=f"the orchestrator pod has been gone for {missing_polls} polls and left no exit code",
+        )
+
+    if phase in _DEAD_POD_PHASES:
```
fzyzcjy, line 64: [P1] Detect crash-looping orchestrator containers

When the wrapper fails before its first state-file write, `charts/miles-run/templates/orchestrator.yaml` runs it with `restartPolicy: Always`; Kubernetes therefore leaves the Pod in `Running` or `Pending` during `CrashLoopBackOff` or `ImagePullBackOff`, rather than entering either phase checked here. With no terminal state, `wait_for_run` sleeps forever instead of returning the no-verdict exit code, so the observer must inspect container/workload failure states rather than only Pod phase.

miles/utils/external_utils/command_utils/helm_backend/orchestrator/observer.py
```diff
@@ -0,0 +1,70 @@
...
+    if state is not None and state.is_terminal:
+        if state.exit_code is None:
+            return _RunOutcome(
+                exit_code=_NO_VERDICT_EXIT_CODE, reason="the orchestrator reported a terminal state with no exit code"
+            )
+        return _RunOutcome(exit_code=state.exit_code, reason="the orchestrator reported its exit code")
+
+    if phase is None:
+        if state is None or missing_polls < _MISSING_POD_POLLS:
+            return None
```
fzyzcjy, line 58: [P2] Stop after a pre-start release disappears

If the Helm release is removed after observation starts but before the wrapper writes `STARTED`, both the state and Pod phase remain `None`; this condition returns `None` forever regardless of `missing_polls`. The launcher consequently hangs instead of returning `_NO_VERDICT_EXIT_CODE`; after an initial creation grace, persistent absence of the active workload must terminate even when no state file was created.

fzyzcjy (commented):
The observer has no terminating path for crash-looping or pull-failing containers, nor for a workload that disappears before publishing `STARTED`. Both behaviors remain in deliver-2 and can leave the launcher blocked indefinitely.

PR #2444, opened 2026-08-12
Report what the pods of a run are actually doing

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/pod_facts.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/pod_facts.py
@@ -1,0 +1,106 @@
+from __future__ import annotations
+
+from typing import NamedTuple
+
+from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Kubectl
+from miles.utils.workers.k8s_types import Event, EventList, Pod, PodList
+
+_SCHEDULING_GATED_REASON = "SchedulingGated"
+_POD_WARNINGS_SELECTOR = "involvedObject.kind=Pod,type=Warning"
+
+
+class ContainerKey(NamedTuple):
+    pod: str
+    container: str
+    previous: bool
+
+
+class ContainerRun(NamedTuple):
+    key: ContainerKey
+    container_id: str
+    running: bool
+
+
+def selected_pods(namespace: str, selector: str) -> list[Pod]:
+    listed = Kubectl.get_json("pods", return_type=PodList, namespace=namespace, selector=selector)
+    return listed.items if listed is not None else []
+
+
+def pod_events(namespace: str, pods: list[Pod]) -> list[Event]:
+    wanted = {pod.metadata.name for pod in pods}
+    if not wanted:
+        return []
+
+    listed = Kubectl.get_json(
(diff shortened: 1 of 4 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/observability/pod_facts.py
```diff
@@ -0,0 +1,106 @@
...
+    return listed.items if listed is not None else []
+
+
+def pod_events(namespace: str, pods: list[Pod]) -> list[Event]:
+    wanted = {pod.metadata.name for pod in pods}
+    if not wanted:
+        return []
+
+    listed = Kubectl.get_json(
+        "events", return_type=EventList, namespace=namespace, field_selector=_POD_WARNINGS_SELECTOR
```
fzyzcjy, line 47: [P1] Report controller failures before a pod exists

When the StatefulSet controller cannot create any Pod, such as after a Pod quota or admission-webhook rejection, Kubernetes records `FailedCreate` against the StatefulSet rather than a Pod. This early return and the Pod-only selector discard that warning; because the target waiter treats an absent state file plus absent pod as startup, the launcher can remain forever at `No pods yet`. Include warnings for the release's controller objects even when the pod list is empty.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/pod_facts.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    events = listed.items if listed is not None else []
+    return [event for event in events if event.involved_object.name in wanted]
```
fzyzcjy: [P2] Match warning events to the current pod UID

When a StatefulSet pod is recreated under the same ordinal name, warning Events for the old incarnation remain until the event TTL expires and are distinguished by `involvedObject.uid`. Filtering only by name returns those old warnings for the current pod, so a healthy replacement can be reported as having the previous pod's scheduling or image failure. Retain the involved-object UID and match it against `pod.metadata.uid`.

fzyzcjy (commented):
The pod observation helpers miss controller-side startup failures and mishandle pod lifecycle details. These defects remain in the target safe deliver, including a silent indefinite wait when no Pod can be created.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/pod_facts.py
```diff
@@ -0,0 +1,106 @@
...
+        for condition in pod.status.conditions
+    )
+
+
+def _described(pod: Pod) -> str:
+    if is_pod_scheduling_gated(pod):
+        return "Pending (scheduling gated)"
+    phase = phase_of_pod(pod)
+    if phase == "Running" and not is_pod_ready(pod):
+        return "Running (not ready yet)"
```
fzyzcjy, line 132: [P2] Include restart counts for unready running pods

When a crashing container leaves its Pod in `Running` with `Ready=False`, which is the normal CrashLoopBackOff shape, this early return omits the restart count. Consecutive snapshots therefore remain `Running (not ready yet)` while `restartCount` rises, so `status_changes` emits no restart updates. Build the readiness qualifier and restart suffix together.

PR #2445, opened 2026-08-12
Build a run's environment in one place before a second backend needs it

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The extraction preserves the Ray runtime environment's values, precedence, and serialization order. No correctness regression remains at the specified deliver-2 safe point.

PR #2446, opened 2026-08-12
Install a training run into Kubernetes and wait for its verdict

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/base_backend.py
+++ b/miles/utils/external_utils/command_utils/base_backend.py
@@ -11,15 +11,14 @@ from pathlib import Path
 
 from miles.utils.external_utils.command_utils.common import (
     ArgvManipulator,
-    create_run_id,
-    run_shell_command,
     _pythonpath_with_sources,
+    create_run_id,
     repo_base_dir,
+    run_shell_command,
 )
 from miles.utils.external_utils.model_args_utils import shell_safe_model_args
-from miles.utils.logging_utils import configure_logger_raw
-from miles.utils.typer_utils import dataclass_from_env
 from miles.utils.pydantic_utils import FrozenStrictBaseModel
+from miles.utils.typer_utils import dataclass_from_env
 from miles.utils.workers.types import ClusterBackend
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -0,0 +1,107 @@
...
+    run_id = config.run_id
+    assert _RUN_ID_PATTERN.fullmatch(
+        run_id
+    ), f"run_id {run_id!r} names every object this run installs, so it has to match {_RUN_ID_PATTERN.pattern}"
+
+    namespace = config.namespace
+    release = RunNames.release(run_id=run_id)
+    pod_argv, args = _compute_train_argv(request, run_id=run_id)
+
+    specs = compute_specs(args)
```
fzyzcjy: [P1] Compute worker specs under the effective run environment

When `extra_env_vars` overrides a variable also emitted by an inference `CommandWorkerSpec`—for example `SGLANG_ENABLE_TP_MEMORY_INBALANCE_CHECK=true`—this builds the spec before the merged run environment exists. `compute_inference_engine_env_vars()` therefore captures the launcher's ambient/default value, and the chart gives that per-pool value precedence over `run.env`, so Kubernetes engines silently receive the wrong setting while Ray honors the override. The target deliver retains this ordering; compute specs under the effective run environment or make run-level overrides win.

miles/utils/external_utils/command_utils/base_backend.py
```diff
@@ -100,6 +105,7 @@ def execute_train(
         train_argv = shlex.split(train_args)
         train_backend_fsdp = "fsdp" in ArgvManipulator.values_of(train_argv, "--train-backend")
         assert train_backend_fsdp == (megatron_model_type is None)
+        _assert_train_args_name_no_other_backend(train_argv, cluster_backend=self.config.cluster_backend.value)
```
fzyzcjy, line 110: [P2] Validate the backend after argument overlays

When a Kubernetes launch's `--custom-config-path` YAML contains `cluster_backend: ray`, this raw-argv check sees no conflicting flag; `_compute_train_argv()` injects Kubernetes, but `parse_args()` subsequently applies the YAML and resolves the backend back to Ray. The installed pods then select Ray providers despite being rendered as Kubernetes workers, causing the run to fail or hang. The target deliver still lacks a post-parse equality check, so validate the resolved backend or reserve this overlay key.

fzyzcjy (commented):
The Kubernetes launch path can silently render worker environments differently from the requested runtime environment and can allow post-parse configuration to switch the backend after validation. Both defects remain in the target safe deliver.

PR #2447, opened 2026-08-12
Point a kubernetes run's api server at the orchestrator's own service

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/naming.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/naming.py
@@ -4,7 +4,7 @@ import random
 import time
 from pathlib import Path
 
-from miles.utils.workers.worker_provider.kubernetes.helm.naming import CHART_NAME
+from miles.utils.workers.worker_provider.kubernetes.helm.naming import CHART_NAME, component_name
 
 ORCHESTRATOR_COMPONENT = "orchestrator"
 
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/helm_backend/naming.py
```diff
@@ -19,6 +19,14 @@ class RunNames:
     def release(*, run_id: str) -> str:
         return f"{CHART_NAME}-{run_id}"
 
+    @staticmethod
+    def service_fqdn(*, name: str, namespace: str) -> str:
+        return f"{name}.{namespace}.svc.cluster.local"
```
fzyzcjy, line 26: [P2] Avoid hard-coding the Kubernetes DNS domain

On clusters configured with a DNS domain other than `cluster.local`, this host never resolves, so callers cannot reach the orchestrator API. The existing Kubernetes worker path uses the domain-neutral `<pod>.<service>.<namespace>.svc` form; use that form here or source the configured cluster domain. The literal remains in deliver-2 (`f5a18edf`), so this is a blocker.

fzyzcjy (commented):
The generated orchestrator address assumes Kubernetes's default cluster DNS domain. This remains unfixed at the target safe deliver and breaks API connectivity on clusters using a custom domain.

PR #2448, opened 2026-08-12
Run the mooncake master as part of the run that needs it

fzyzcjy:
Part of #1837


fzyzcjy (commented):
At the specified safe-deliver boundary, the op's transient correctness gaps have been repaired. No regression introduced by this op remains as a blocker.

PR #2449, opened 2026-08-12
Stream every container's logs while the run is going

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/log_follower.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/log_follower.py
@@ -1,0 +1,158 @@
+from __future__ import annotations
+
+import logging
+import subprocess
+import tempfile
+import threading
+from collections.abc import Iterator
+from contextlib import contextmanager
+from datetime import datetime
+from typing import IO
+
+from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Kubectl
+from miles.utils.external_utils.command_utils.helm_backend.launcher.observability import polling
+from miles.utils.external_utils.command_utils.helm_backend.launcher.observability.pod_facts import (
+    ContainerKey,
+    ContainerRun,
+    container_runs,
+    selected_pods,
+)
+from miles.utils.external_utils.command_utils.helm_backend.launcher.observability.polling import polling_in_background
+
+logger = logging.getLogger(__name__)
+
+_STOP_GRACE_SECONDS = 5.0
+_ERROR_TAIL_CHARACTERS = 500
+
+
+@contextmanager
+def with_log_following(*, namespace: str, selector: str) -> Iterator[None]:
+    follower = _LogFollower(namespace=namespace, selector=selector)
+    try:
+        with polling_in_background(
+            follower.reconcile,
+            description="look for pods to follow",
+            join_timeout=polling.POLL_INTERVAL_SECONDS + _STOP_GRACE_SECONDS,
+        ):
+            yield
+    finally:
+        follower.stop_all_streams()
+
+
+class _LogFollower:
+    def __init__(self, *, namespace: str, selector: str) -> None:
+        self._namespace = namespace
+        self._selector = selector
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/observability/log_follower.py
```diff
@@ -0,0 +1,158 @@
...
+        timestamp, _, rest = line.partition(" ")
+        if _is_timestamp(timestamp):
+            self._last_timestamp = timestamp
+            line = rest
+        logger.info(f"{self._prefix} {line}")
+
+
+def _is_timestamp(token: str) -> bool:
+    try:
+        datetime.fromisoformat(token.replace("Z", "+00:00"))
```
fzyzcjy: [P1] Parse every RFC3339Nano fractional width

`setup.py:63` declares Python 3.10 support, but on 3.10 `datetime.fromisoformat` accepts fractional seconds only with three or six digits, while kubectl commonly emits nine-digit RFC3339Nano prefixes. `_emit()` consequently retains the prefix and never advances `_last_timestamp`, so reconnecting replays the complete log without `--since-time`. Parse or normalize RFC3339Nano explicitly. This file is byte-identical at safe deliver `f5a18edf2aa9fade59d0cff05d865c310fd52e7a`, so the issue remains a blocker.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/log_follower.py
```diff
@@ -0,0 +1,158 @@
...
+        self._thread = threading.Thread(target=self._read_until_done, args=(self._process, errors), daemon=True)
+        self._thread.start()
+
+    def _read_until_done(self, process: subprocess.Popen, errors: IO[str]) -> None:
+        for line in process.stdout:
+            self._emit(line.rstrip("\n"))
+        process.wait()
+
+        if process.returncode and not self._stopped:
+            self._stopped = True
```
fzyzcjy: [P1] Retry streams after transient kubectl exits

When a live `kubectl logs --follow` exits nonzero because of a transient API-server or network reset, this marks the stream `_stopped`; every later `reconcile()` calls `resume_if_dropped()`, which immediately returns on that flag. The container is therefore never followed again, potentially hiding all subsequent and crash logs. Reserve the stop flag for explicit teardown and keep operational failures retryable. This code is unchanged at safe deliver `f5a18edf2aa9fade59d0cff05d865c310fd52e7a`, so the issue remains a blocker.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/polling.py
```diff
@@ -0,0 +1,32 @@
...
+def polling_in_background(step: Callable[[], None], *, description: str, join_timeout: float) -> Iterator[None]:
+    stop = threading.Event()
+    thread = threading.Thread(target=_poll_until_stopped, args=(step, description, stop), daemon=True)
+    thread.start()
+
+    try:
+        yield
+    finally:
+        stop.set()
+        thread.join(timeout=join_timeout)
```
fzyzcjy: [P2] Wait for the poller before running cleanup

When context exit overlaps a `step()` that exceeds `join_timeout`, such as an unbounded `kubectl get` during an API outage, this timed join returns while the daemon still executes. `with_log_following()` then cleans up existing streams, but the late `reconcile()` can subsequently add new `_LogStream`s and leave `kubectl logs` children running after Ctrl+C or launcher exit. Make the step bounded or cancellable and ensure it can no longer mutate state before cleanup. The helper is unchanged at safe deliver `f5a18edf2aa9fade59d0cff05d865c310fd52e7a`, so this remains a blocker.

fzyzcjy (commented):
The log follower mishandles common timestamps on a supported Python version, permanently disables streams after transient kubectl failures, and can race cleanup with a still-running poller. These production files remain unchanged at the target safe deliver, so the defects are blockers rather than transient chain issues.

PR #2450, opened 2026-08-12
Announce the cluster events that explain a stuck run

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/cluster_info.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/cluster_info.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+
+import logging
+from collections.abc import Iterator
+from contextlib import contextmanager
+
+from miles.utils.external_utils.command_utils.helm_backend.launcher.observability import polling
+from miles.utils.external_utils.command_utils.helm_backend.launcher.observability.pod_facts import (
+    is_pod_ready,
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/observability/cluster_info.py
```diff
@@ -0,0 +1,109 @@
...
+        f"{event.involved_object.name}: {event.reason} x{event.count}: {(event.message or '').strip()}"
+        for event in ranked[:_REPORTED_EVENTS]
+    ]
+    if len(ranked) > _REPORTED_EVENTS:
+        lines.append(f"... and {len(ranked) - _REPORTED_EVENTS} more warning events")
+    return lines
+
+
+def _event_key(event: Event) -> tuple[str | None, str | None]:
+    return event.involved_object.name, event.reason
```
fzyzcjy: [P2] Key warnings by pod generation

When a StatefulSet or LeaderWorkerSet pod is deleted and recreated under the same name, this key retains the warning reasons reported for the old pod. Any warning with the same reason from the replacement is then filtered as already reported, so the current pod can remain stuck without its diagnostic being shown. Include the involved-object UID or clear entries when the pod generation changes; this remains unchanged in the target deliver.

fzyzcjy (commented):
The watcher conflates warning events across pod generations, suppressing diagnostics from replacement pods. The target safe deliver retains the same implementation, so the issue remains a blocker.

PR #2451, opened 2026-08-12
Watch a run while it runs, and say how to look at it afterwards

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
@@ -17,7 +17,8 @@ from miles.utils.external_utils.command_utils.base_backend import (
 )
 from miles.utils.external_utils.command_utils.common import ArgvManipulator, chart_dir, repo_base_dir, train_env_vars
 from miles.utils.external_utils.command_utils.helm_backend import naming
-from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Helm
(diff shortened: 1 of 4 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -83,14 +84,16 @@ def execute_train(*, request: ExecuteTrainRequest, config: ExecuteTrainConfig) -
...
-        state_file=state_file,
-        read_pod_phase=lambda: pod_phase(namespace, orchestrator_workload),
-    )
+    with with_observability(namespace=namespace, selector=Kubectl.release_selector(release)):
+        outcome = wait_for_run(
+            state_file=state_file,
+            read_pod_phase=lambda: pod_phase(namespace, orchestrator_workload),
+        )
 
+    logger.info(farewell(namespace=namespace, release=release, workload=orchestrator_workload))
```
fzyzcjy, line 100: [P2] Emit the farewell when Ctrl+C stops watching

When the user follows the explicitly advertised `ctrl+c` path, `KeyboardInterrupt` escapes `wait_for_run()` and unwinds the context, so this line is skipped. The launcher therefore exits without the reattach and teardown commands precisely while the remote run is still active; handle the interrupt and emit the farewell before propagating it. This control flow remains unchanged in safe deliver `f5a18edf2aa9fade59d0cff05d865c310fd52e7a`.

fzyzcjy (commented):
The advertised Ctrl+C workflow omits the newly added recovery instructions because the farewell is only emitted after a terminal outcome. This regression persists through the target safe deliver.

PR #2452, opened 2026-08-12
Collect the scene when a run fails, and let the workbench collect it too

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/diagnosis.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/observability/diagnosis.py
@@ -1,0 +1,82 @@
+from __future__ import annotations
+
+import logging
+import time
+from pathlib import Path
+
+from miles.utils.external_utils.command_utils.common import run_process
+from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Kubectl
+from miles.utils.pydantic_utils import FrozenStrictBaseModel
+from miles.utils.workers.k8s_types import PodList
+
+logger = logging.getLogger(__name__)
+
+
+class Diagnosis(FrozenStrictBaseModel):
+    directory: Path
+    missing: tuple[str, ...] = ()
+
+    @property
+    def is_complete(self) -> bool:
+        return not self.missing
+
+
+def collect_diagnosis(
+    *, namespace: str, output_dir: Path, selector: str | None = None, state_file: Path | None = None
+) -> Diagnosis:
+    directory = output_dir / f"miles-diagnosis-{namespace}-{time.strftime('%Y%m%d-%H%M%S')}"
+    directory.mkdir(parents=True, exist_ok=True)
+
+    missing: list[str] = []
+    if not _capture(
+        path=directory / "events.txt",
+        command=["kubectl", "get", "events", "-n", namespace, "--sort-by=.lastTimestamp"],
+    ):
+        missing.append("events")
+
+    pods = _pod_names(namespace=namespace, selector=selector)
+    if pods is None:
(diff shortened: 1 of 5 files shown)


fzyzcjy (commented):
Deliver-2 still contains diagnosis-integrity gaps: collection can falsely report completeness after losing previous logs or finding no run pods, and concurrent snapshots can overwrite one another. These defects undermine the patch's core failure-preservation behavior.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/diagnosis.py
```diff
@@ -0,0 +1,82 @@
...
+def _capture_pod(pod: str, *, namespace: str, directory: Path) -> list[str]:
+    missing = []
+    if not _capture(
+        path=directory / f"{pod}.log", command=["kubectl", "logs", pod, "-n", namespace, "--all-containers"]
+    ):
+        missing.append(f"logs of {pod}")
+    _capture(
+        path=directory / f"{pod}.previous.log",
+        command=["kubectl", "logs", pod, "-n", namespace, "--all-containers", "--previous"],
+        skip_when_it_fails=True,
```
fzyzcjy: [P2] Report unexpected failures of previous-log collection

When `kubectl logs --previous` fails for anything other than the expected absence of a previous container, this flag makes `_capture` return success before preserving stderr or adding the failure to `missing`. A transient API error can therefore make the workbench exit zero and the launcher report a complete diagnosis despite omitting the crash log; suppress only the specific no-previous-container response.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/diagnosis.py
```diff
@@ -0,0 +1,82 @@
...
+        path=directory / f"{pod}.describe.txt", command=["kubectl", "describe", "pod", pod, "-n", namespace]
+    ):
+        missing.append(f"describe of {pod}")
+    return missing
+
+
+def _pod_names(*, namespace: str, selector: str | None) -> list[str] | None:
+    try:
+        listed = Kubectl.get_json("pods", return_type=PodList, namespace=namespace, selector=selector)
+        return [pod.metadata.name for pod in listed.items] if listed is not None else []
```
fzyzcjy, line 76: [P2] Report when a run has no pods left to capture

When a failed run's pods are deleted before this listing, `kubectl get pods` succeeds with an empty `PodList`, this returns `[]`, and nothing is added to `missing`. The selector-based launcher path then reports an events-only directory as a complete diagnosis even though every pod log and description was lost; an empty selected pod set should be marked incomplete.

miles/utils/external_utils/command_utils/helm_backend/launcher/observability/diagnosis.py
```diff
@@ -0,0 +1,82 @@
...
+    @property
+    def is_complete(self) -> bool:
+        return not self.missing
+
+
+def collect_diagnosis(
+    *, namespace: str, output_dir: Path, selector: str | None = None, state_file: Path | None = None
+) -> Diagnosis:
+    directory = output_dir / f"miles-diagnosis-{namespace}-{time.strftime('%Y%m%d-%H%M%S')}"
+    directory.mkdir(parents=True, exist_ok=True)
```
fzyzcjy: [P2] Avoid overwriting diagnoses created in the same second

When two collectors target the same namespace and output directory within one second, they derive the same path, and `exist_ok=True` lets both truncate and rewrite the same event and pod files. Concurrent launcher or workbench collections can therefore silently overwrite or interleave snapshots; create each directory with a collision-resistant token or an exclusive-create retry.

PR #2453, opened 2026-08-12
Say which fields of an installed run a relaunch may change

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_diff.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_diff.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+from typing import Any
+
+from miles.utils.external_utils.command_utils.helm_backend.launcher.manifest_types import Manifest, ManifestObject
+from miles.utils.pydantic_utils import FrozenStrictBaseModel
+
+_SCALABLE_KINDS = frozenset({"LeaderWorkerSet"})
+
+
+class ManifestDiff(FrozenStrictBaseModel):
+    changed: list[str] = []
+    added: list[str] = []
+    removed: list[str] = []
+    scaled: list[str] = []
+
+    @property
+    def is_allowed(self) -> bool:
+        return not (self.changed or self.added or self.removed)
+
+    def summarize_scaling(self) -> str:
+        return "\n".join(f"  {entry}" for entry in self.scaled) or "  (nothing to change)"
+
+    def describe(self) -> str:
+        lines = []
+        for label, entries in (("changed", self.changed), ("added", self.added), ("removed", self.removed)):
+            lines += [f"  {label}: {entry}" for entry in entries]
+        return "\n".join(lines) or "  (no difference)"
+
+
+def diff_manifests(*, before: Manifest, after: Manifest) -> ManifestDiff:
+    old = before.by_key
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_diff.py
```diff
@@ -0,0 +1,75 @@
...
+        ],
+    )
+
+
+def _disallowed_differences(old: ManifestObject, new: ManifestObject) -> list[str]:
+    return _differing_paths(old.body, new.body, (), allowed=_allowed_of_kind(old.kind))
+
+
+def _allowed_of_kind(kind: str) -> tuple[tuple[str, ...], ...]:
+    return (("spec", "replicas"),) if kind in _SCALABLE_KINDS else ()
```
fzyzcjy, line 52: [P2] Key the scaling exception by API group and kind

When a release includes a CR from another API group that also uses `kind: LeaderWorkerSet`, changing that object's `spec.replicas` is removed from `changed` because this decision checks only the Kind. Kubernetes Kind names are unique only within an API group, and deliver-2 accepts arbitrary extra manifests, so the relaunch guard can silently permit an unrelated CRD update; restrict the exception to the supported LeaderWorkerSet group/version.

miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_diff.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _differing_paths(old: Any, new: Any, path: tuple[str, ...], *, allowed: tuple[tuple[str, ...], ...]) -> list[str]:
+    if path in allowed or old == new:
```
fzyzcjy, line 56: [P2] Distinguish scalar types before declaring manifests equal

When an unstructured field changes from YAML `true` to `1` or `false` to `0`, Python considers the scalars and their enclosing dictionaries equal, so `_differing_paths` reports no change and `is_allowed` lets Helm apply it. Deliver-2's arbitrary extra manifests can contain CRDs with preserved unknown fields where both JSON types are valid, so equality must include the scalar type or use a type-sensitive canonical representation.

miles/utils/external_utils/command_utils/helm_backend/launcher/manifest_diff.py
```diff
@@ -0,0 +1,75 @@
...
+        return []
+
+    if isinstance(old, dict) and isinstance(new, dict):
+        differences = []
+        for key in sorted(set(old) | set(new)):
```
fzyzcjy, line 61: [P2] Normalize YAML mapping keys before sorting them

When an extra ConfigMap has a valid unquoted numeric key alongside a string key, such as `data: {1: "old", x: "keep"}`, PyYAML leaves the first key as an `int` while Kubernetes normalizes it to the string `"1"`; changing its value makes this `sorted(...)` raise `TypeError`. The relaunch gate therefore crashes instead of reporting or refusing the change, and the same code remains in deliver-2; normalize mapping keys to Kubernetes JSON strings before sorting and building paths.

fzyzcjy (commented):
The manifest gate can still classify two non-scaling changes as allowed and can raise on a Kubernetes-valid manifest. All three defects remain at the specified deliver-2 commit.

PR #2454, opened 2026-08-12
Let a relaunch change a run's size and nothing else

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
@@ -17,7 +17,9 @@ from miles.utils.external_utils.command_utils.base_backend import (
 )
 from miles.utils.external_utils.command_utils.common import ArgvManipulator, chart_dir, repo_base_dir, train_env_vars
 from miles.utils.external_utils.command_utils.helm_backend import naming
+from miles.utils.external_utils.command_utils.helm_backend.launcher import manifest_diff
 from miles.utils.external_utils.command_utils.helm_backend.launcher.command_wrapper import Helm, Kubectl
+from miles.utils.external_utils.command_utils.helm_backend.launcher.manifest_types import Manifest
 from miles.utils.external_utils.command_utils.helm_backend.launcher.observability import farewell, with_observability
 from miles.utils.external_utils.command_utils.helm_backend.launcher.observability.diagnosis import collect_diagnosis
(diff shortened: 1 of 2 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -119,6 +132,34 @@ def _compute_train_argv(
...
+    release: str,
+    namespace: str,
+    chart: Path,
+    values_files: list[str | Path],
+    skip_upgrade_check: bool,
+) -> None:
+    proposed_manifest = Helm.render_upgrade(
+        release=release, namespace=namespace, chart=chart, values_files=values_files
+    )
+    diff = manifest_diff.diff_manifests(before=installed_manifest, after=proposed_manifest)
```
fzyzcjy, line 162: [P1] Decouple resize inputs from pod commands

When an existing run is relaunched with a normal size input such as `--rollout-num-gpus`, that value is part of `pod_argv`, which is embedded in both `LaunchPlan.orchestrator_command` and served-worker commands. This comparison therefore sees command changes in addition to `LeaderWorkerSet.spec.replicas`, so `diff.is_allowed` is false and the default path exits unless `--skip-upgrade-check` is used; the resize this change is meant to permit cannot be performed safely. Build the proposal so size-only inputs update replicas without rewriting installed commands.

miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -57,6 +59,7 @@ def execute_train(*, request: ExecuteTrainRequest, config: ExecuteTrainConfig) -
 
     Helm.build_dependencies(chart)
 
+    installed_manifest = Helm.get_manifest(release, namespace)
```
fzyzcjy, line 66: [P2] Serialize validation and upgrade per release

When two launchers target the same release, `installed_manifest` is a snapshot with no lock or revision token. Launcher A can validate against revision N, launcher B can complete an upgrade using `--skip-upgrade-check`, and then A's `Helm.upgrade` applies its stale values against N+1, reverting B's image or command and restarting pods despite A having passed the gate. Serialize the check and upgrade per release, or enforce a revision precondition.

fzyzcjy (commented):
The gate blocks ordinary flag-driven resizes and has a stale-manifest race that can permit non-size changes. Both issues remain in the target safe deliver.

PR #2455, opened 2026-08-12
Attach a relaunch to the verdict the installed run is already writing

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
@@ -60,7 +60,9 @@ def execute_train(*, request: ExecuteTrainRequest, config: ExecuteTrainConfig) -
     Helm.build_dependencies(chart)
 
     installed_manifest = Helm.get_manifest(release, namespace)
-    state_file = RunFiles.new_state_file(run_directory=run_directory)
+    state_file = _compute_state_file(
+        installed_manifest=installed_manifest, run_directory=run_directory, release=release
+    )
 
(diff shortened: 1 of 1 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -132,6 +134,18 @@ def _compute_train_argv(
     return pod_argv, args
 
 
+def _compute_state_file(*, installed_manifest: Manifest | None, run_directory: Path, release: str) -> Path:
+    if installed_manifest is None:
+        return RunFiles.new_state_file(run_directory=run_directory)
+
+    attached_state_file = installed_manifest.state_file(container=naming.ORCHESTRATOR_COMPONENT)
```
fzyzcjy, line 156: [P1] Revalidate the release before reusing its state file

At deliver-2, if the run's auto-uninstall completes after `Helm.get_manifest()` but before `Helm.upgrade --install`, this path belongs to a release that no longer exists. The command then performs a fresh install with the completed launch's terminal state file, so the observer accepts the old verdict instead of starting training; the completed uninstall Job can also prevent a replacement Job from being armed, leaving the recreated release running. Revalidate the release revision/existence immediately before upgrading and recompute a fresh state file when it changed.

fzyzcjy (commented):
The attachment decision uses a non-atomic Helm snapshot. Deliver-2 still permits auto-uninstall to remove the release before upgrade, turning an intended attachment into a fresh install bound to a stale terminal verdict.

PR #2456, opened 2026-08-12
Remove the releases a previous ci run left behind

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
@@ -14,6 +14,8 @@ from miles.utils.workers.worker_provider.kubernetes.helm.env import INSTANCE_LAB
 
 _ModelT = TypeVar("_ModelT", bound=BaseModel)
 
+CI_LABEL = "miles.radixark.io/ci-run"
+
 
 class Helm:
     @staticmethod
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
```diff
@@ -63,8 +66,24 @@ def build_dependencies(chart: str | Path) -> None:
...
+    def uninstall(*, release: str, namespace: str) -> None:
+        _run(["helm", "uninstall", release, "--namespace", namespace], capture_output=False)
+
+    @staticmethod
+    def upgrade_command(
+        release: str, namespace: str, chart: str | Path, values_files: list[str | Path], *, ci_run: bool
+    ) -> list[str]:
         command = ["helm", "upgrade", "--install", release, str(chart), "--namespace", namespace]
+        if ci_run:
+            command += ["--labels", f"{CI_LABEL}=true"]
```
fzyzcjy, line 86: [P2] Unset CI ownership on non-CI upgrades

When a release first installed with `ci_run=True` is later relaunched under the same run ID with `ci_run=False`, Helm preserves existing release labels unless a label is explicitly set to `null`; omitting `--labels` here does not make it human-owned. It therefore remains selectable by `miles.radixark.io/ci-run=true`, so the next unrelated CI launch in that namespace uninstalls the developer's live release. Explicitly unset `CI_LABEL` on the false branch; deliver-2 retains this behavior.

fzyzcjy (commented):
The current-run relaunch regression is repaired before deliver-2 and is therefore transient. However, deliver-2 still leaves CI ownership sticky after a non-CI upgrade, allowing cleanup to delete a live developer release.

PR #2457, opened 2026-08-12
Ship a run with the job that will uninstall it

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-run/templates/uninstall-manifest.yaml
+++ b/charts/miles-run/templates/uninstall-manifest.yaml
@@ -1,0 +1,38 @@
+{{- if include "miles-run.autoUninstallEnabled" . }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.run.objectNames.uninstallManifest | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "miles-run.labels" (dict "context" . "component" "uninstall") | nindent 4 }}
+data:
+  {{ include "miles-run.uninstallManifestFileName" . }}: |
+    apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: {{ .Values.run.objectNames.uninstall | quote }}
+      namespace: {{ .Release.Namespace | quote }}
+      labels:
+        miles.radixark.io/uninstall-of: {{ .Release.Name | quote }}
+    spec:
+      backoffLimit: 2
+      ttlSecondsAfterFinished: 3600
+      activeDeadlineSeconds: 1020
+      template:
+        metadata:
+          labels:
+            miles.radixark.io/uninstall-of: {{ .Release.Name | quote }}
+        spec:
+          {{- include "miles-run.podDefaults" . | nindent 10 }}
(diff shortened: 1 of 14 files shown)


charts/miles-run/templates/uninstall-manifest.yaml
```diff
@@ -0,0 +1,38 @@
...
+    apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: {{ .Values.run.objectNames.uninstall | quote }}
+      namespace: {{ .Release.Namespace | quote }}
+      labels:
+        miles.radixark.io/uninstall-of: {{ .Release.Name | quote }}
+    spec:
+      backoffLimit: 2
+      ttlSecondsAfterFinished: 3600
```
fzyzcjy, line 20: [P2] Replace stale uninstall Jobs on direct reinstall

When the chart is installed directly under a release name reused within this one-hour TTL, the completed out-of-band Job from the prior release still owns this fixed name. If the new run finishes before expiry, deliver-2's `Kubectl.create_if_absent` treats `AlreadyExists` as success, but a completed Job cannot execute again, so the new release retains its GPUs. Use a launch-specific name or replace stale Jobs on every install path.

charts/miles-run/templates/uninstall-manifest.yaml
```diff
@@ -0,0 +1,38 @@
...
+      activeDeadlineSeconds: 1020
+      template:
+        metadata:
+          labels:
+            miles.radixark.io/uninstall-of: {{ .Release.Name | quote }}
+        spec:
+          {{- include "miles-run.podDefaults" . | nindent 10 }}
```
fzyzcjy: [P2] Avoid inheriting run-pod anti-affinity

When `infra.scheduling.affinity` contains required pod anti-affinity that spreads run pods across every eligible node, all matching nodes can already host a release pod when this cleanup Job is created. Copying that affinity wholesale makes the uninstall pod unschedulable, leaving the release and its GPUs installed; deliver-2 retains this behavior. Preserve the necessary node placement and tolerations without inheriting run-pod anti-affinity.

fzyzcjy (commented):
Deliver-2 repairs the transient wrapper-argument incompatibility, but these production paths still prevent auto-uninstall from executing. They can leave completed releases holding GPUs, so the patch is not correct at the requested safe-delivery point.

charts/miles-run/values.yaml
```diff
@@ -41,6 +43,9 @@ run:
   inferenceEngines: []
   trainerEngines: []
   env: {}
+  autoUninstall:
+    enabled: true
+    serviceAccount: miles-uninstaller
```
fzyzcjy: [P2] Propagate the configured uninstaller account

When `miles-workbench` is installed with its supported `uninstaller.serviceAccount` override, it creates only that custom account, while its mounted `infra.yaml` forwards no corresponding run value. Every launched run therefore falls back to this independent `miles-uninstaller` default; the uninstall pod cannot start because that ServiceAccount is absent, so completed releases retain their GPUs. Deliver-2 keeps this disconnect, so propagate the workbench setting into `run.autoUninstall.serviceAccount`.

PR #2458, opened 2026-08-12
Arm the escape job once the run has reported its verdict

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
@@ -15,6 +15,7 @@ from miles.utils.workers.worker_provider.kubernetes.helm.env import INSTANCE_LAB
 _ModelT = TypeVar("_ModelT", bound=BaseModel)
 
 CI_LABEL = "miles.radixark.io/ci-run"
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/command_wrapper.py
```diff
@@ -94,6 +95,24 @@ class Kubectl:
     def run_raw(*arguments: str) -> subprocess.CompletedProcess[str]:
         return Kubectl._run(list(arguments))
 
+    @staticmethod
+    def create_if_absent(manifest_path: str) -> bool:
+        result = Kubectl._run(["create", "-f", manifest_path])
```
fzyzcjy, line 101: [P2] Bound each kubectl create attempt

When the API server accepts a connection but stops answering, this call reaches `subprocess.run` without a timeout while kubectl's `--request-timeout` defaults to zero. The first attempt can therefore block forever, so the retry schedule never advances and the finished release retains its pods indefinitely; this path remains unchanged in target deliver `f5a18edf2`, so each create attempt needs a finite request or process timeout.

fzyzcjy (commented):
The nominally bounded cleanup retry can hang permanently in its first kubectl invocation, leaving the completed release installed. The target safe deliver retains this behavior.

PR #2459, opened 2026-08-12
Defuse the previous launch's escape job before installing over it

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
@@ -81,7 +81,9 @@ def execute_train(*, request: ExecuteTrainRequest, config: ExecuteTrainConfig) -
     _write_helm_values(values_path, build_values(specs, plan).as_values())
     values_files: list[str | Path] = [*config.helm_values, values_path]
 
-    if installed_manifest is not None:
+    if installed_manifest is None:
+        _remove_pending_uninstall(release, namespace=namespace)
+    else:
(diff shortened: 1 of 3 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -185,6 +187,12 @@ def _uninstall_leftover_ci_releases(namespace: str) -> list[str]:
     return releases
 
 
+def _remove_pending_uninstall(release: str, *, namespace: str) -> None:
+    job = RunNames.uninstall_job(release=release)
+    logger.info(f"Deleting {job} if it is pending, so it cannot uninstall the release this launch installs")
+    Kubectl.delete_job(job, namespace=namespace, check=True)
```
fzyzcjy, line 208: [P1] Wait for the old uninstall pod before upgrading

When a previous release is manually removed while its auto-uninstall Job is still sleeping, this call is not a barrier: `Kubectl.delete_job` uses plain `kubectl delete job`, whose default background cascade can return while the Job's pod is still running. `Helm.upgrade` can then install the replacement before that pod terminates, allowing the old pod to execute `helm uninstall` against the replacement. Use foreground propagation or explicitly wait for the Job pod to disappear; `f5a18edf2a` retains the same helper, so this remains a deliver-2 blocker.

fzyzcjy (commented):
The new defusing step does not ensure that the auto-uninstall pod has terminated, so a stale pod can still uninstall the replacement release. This race remains in the specified deliver-2 commit.

PR #2460, opened 2026-08-12
Pin what a typical run renders, and the commands one launch runs

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Within the requested scope and exclusions, no correctness or regression blocker introduced by this op remains at the target safe deliver.

PR #2461, opened 2026-08-12
Render a one-off command as a Job of the run chart

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-run/templates/_helpers.tpl
+++ b/charts/miles-run/templates/_helpers.tpl
@@ -33,6 +33,25 @@ affinity:
 {{- end }}
 {{- end }}
 
+{{- define "miles-run.podDefaultsWithAntiAffinity" -}}
+{{- $context := .context -}}
+{{- $scheduling := $context.Values.infra.scheduling | default dict -}}
+enableServiceLinks: false
+{{- with include "miles-common.imagePullSecrets" $context }}
+{{ . }}
+{{- end }}
+{{- with $scheduling.nodeSelector }}
+nodeSelector:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+{{- with $scheduling.tolerations }}
+tolerations:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+affinity:
+  {{- toYaml (merge (dict "podAntiAffinity" .podAntiAffinity) (deepCopy ($scheduling.affinity | default dict))) | nindent 2 }}
+{{- end }}
+
 {{- /* Every container of the release: the run's own identity, from which a worker that has to
        reach another worker recomputes its address and builds a backend capability of its own. */ -}}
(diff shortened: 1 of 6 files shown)


charts/miles-run/templates/command-job.yaml
```diff
@@ -0,0 +1,72 @@
...
+    {{- include "miles-run.labels" (dict "context" . "component" .Values.commandJob.name) | nindent 4 }}
+spec:
+  backoffLimit: 0
+  activeDeadlineSeconds: {{ .Values.commandJob.activeDeadlineSeconds }}
+  ttlSecondsAfterFinished: {{ .Values.commandJob.ttlSecondsAfterFinished }}
+  completions: {{ .Values.commandJob.completions }}
+  parallelism: {{ .Values.commandJob.completions }}
+  {{- if gt (int .Values.commandJob.completions) 1 }}
+  completionMode: Indexed
+  {{- end }}
```
fzyzcjy: [P1] Index one-pod jobs used with index-based placeholders

At deliver-2, `command_job.run_on_nodes` substitutes `{{node_rank}}` with `${JOB_COMPLETION_INDEX}` and `{{master_addr}}` with `<job>-0.<service>` for every completion count; `convert_checkpoint(multinode=True)` can reach this path with one completion. Because this conditional leaves that Job NonIndexed, Kubernetes provides neither the index environment variable nor the deterministic `-0` hostname, so the command receives an empty rank and an unresolvable rendezvous address. Render single-completion Jobs as Indexed too, or special-case both substitutions.

charts/miles-run/templates/_helpers.tpl
```diff
@@ -33,6 +33,25 @@ affinity:
...
+{{- with $scheduling.nodeSelector }}
+nodeSelector:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+{{- with $scheduling.tolerations }}
+tolerations:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+affinity:
+  {{- toYaml (merge (dict "podAntiAffinity" .podAntiAffinity) (deepCopy ($scheduling.affinity | default dict))) | nindent 2 }}
```
fzyzcjy: [P1] Preserve configured required anti-affinity rules

When a multi-pod command runs with `infra.scheduling.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution` configured, Sprig's `merge` keeps the first non-empty list, so the generated hostname rule replaces every required rule supplied by the cluster values. The Job can consequently run in placements those values explicitly forbid, and deliver-2 retains the same merge; concatenate the configured and generated required terms before rendering.

fzyzcjy (commented):
Deliver-2 still contains two blockers introduced here: single-pod jobs lack the index identity expected by command placeholders, and multi-pod jobs discard configured required anti-affinity rules.

PR #2462, opened 2026-08-12
Extract the placeholder substitution the ray path does inline

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The extracted helper preserves placeholder substitution behavior for the Ray path's production values. No correctness regression remains at the target safe deliver.

PR #2463, opened 2026-08-12
Run a command in the cluster the same way a run is launched

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/backend.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/backend.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 
 from miles.utils.external_utils.command_utils.base_backend import BaseCommandBackend, ExecuteTrainRequest
+from miles.utils.external_utils.command_utils.common import chart_dir, repo_base_dir
+from miles.utils.external_utils.command_utils.helm_backend import command_job
 from miles.utils.external_utils.command_utils.helm_backend.launcher import entrypoint
 from miles.utils.external_utils.command_utils.helm_backend.naming import RunNames
 
@@ -13,7 +15,9 @@ class KubernetesCommandBackend(BaseCommandBackend):
     def exec_command_gpu(
         self, cmd: str, capture_output: bool = False, num_gpus_per_node: int | None = None
     ) -> str | None:
-        raise NotImplementedError("A later milestone runs a command as a job in the cluster")
+        return self.exec_command_multi_node(
+            cmd, capture_output=capture_output, num_nodes=1, num_gpus_per_node=num_gpus_per_node
+        )[0]
 
     def exec_command_multi_node(
         self,
@@ -22,7 +26,19 @@ class KubernetesCommandBackend(BaseCommandBackend):
         num_nodes: int | None = None,
         num_gpus_per_node: int | None = None,
     ) -> list[str | None]:
-        raise NotImplementedError("A later milestone runs a command as a job in the cluster")
+        assert self.config.namespace, "Set ExecuteTrainConfig.namespace to run a command somewhere"
+        return command_job.run_on_nodes(
+            command_job.CommandJobContext(
+                namespace=self.config.namespace,
+                chart_dir=chart_dir(repo_base_dir=repo_base_dir),
+                helm_values_files=tuple(self.config.helm_values),
+                gpus_per_node=num_gpus_per_node if num_gpus_per_node is not None else 1,
+            ),
+            cmd,
+            capture_output=capture_output,
+            completions=num_nodes or 1,
+            step="command",
+        )
 
     def api_server_host(self) -> str:
         assert self.config.run_id and self.config.namespace, (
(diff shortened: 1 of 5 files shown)


miles/utils/external_utils/command_utils/helm_backend/backend.py
```diff
@@ -22,7 +26,19 @@ def exec_command_multi_node(
...
+        return command_job.run_on_nodes(
+            command_job.CommandJobContext(
+                namespace=self.config.namespace,
+                chart_dir=chart_dir(repo_base_dir=repo_base_dir),
+                helm_values_files=tuple(self.config.helm_values),
+                gpus_per_node=num_gpus_per_node if num_gpus_per_node is not None else 1,
+            ),
+            cmd,
+            capture_output=capture_output,
+            completions=num_nodes or 1,
```
fzyzcjy, line 39: [P1] Preserve all-node semantics when count is omitted

`exec_command_multi_node(..., num_nodes=None)` means all live nodes in the Ray backend, and `scripts/run_glm45_355b_a32b.py::_prepare_megatron_ckpt` calls `convert_checkpoint(multinode=True)` without supplying a count. This line silently creates one completion and substitutes `{{nnodes}}` with `1`, so a command selected for multi-node execution can fail when it requires aggregate GPU memory. Default from the configured cluster size or require and propagate an explicit count, preserving the established contract (`.claude/rules/general-code-style.md:12-15`).

miles/utils/external_utils/command_utils/helm_backend/command_job.py
```diff
@@ -0,0 +1,174 @@
...
+    capture_output: bool,
+    completions: int,
+    step: str,
+) -> list[str | None]:
+    job = _CommandJob(context=context, step=step, completions=completions, gpus_per_pod=context.gpus_per_node)
+    prepared = substitute_placeholders(
+        cmd,
+        node_rank="${JOB_COMPLETION_INDEX}",
+        nnodes=str(completions),
+        master_addr=job.master_address,
```
fzyzcjy, line 74: [P1] Give one-pod jobs an indexed identity

For a one-completion call containing distributed placeholders, the chart omits `completionMode: Indexed`. Kubernetes therefore supplies neither `JOB_COMPLETION_INDEX` nor the `<job>-0` hostname assumed here, so `{{node_rank}}` expands empty and `{{master_addr}}` names a nonexistent pod DNS record. Render one-completion jobs as Indexed too, or substitute literal single-node values.

miles/utils/external_utils/command_utils/helm_backend/command_job.py
```diff
@@ -0,0 +1,174 @@
...
+        node_ip="$(hostname -i)",
+    )
+    return _run_job(job, command=["bash", "-c", prepared], capture_output=capture_output)
+
+
+def _run_job(job: _CommandJob, *, command: list[str], capture_output: bool) -> list[str | None]:
+    manifest = _render_job(job, command=command)
+
+    Kubectl.delete_job(job.object_name, namespace=job.context.namespace)
```
fzyzcjy, line 87: [P1] Give concurrent commands unique object names

Every backend call uses step `command`, so every process in one namespace targets `miles-run-command-command`. If two launchers prepare different runs concurrently, the second deletes the first Job here; the first command is killed, and both waiters can then observe or delete the replacement Job and report the wrong outcome. Use per-invocation names for both the Service and Job, or enforce serialization with explicit ownership.

miles/utils/external_utils/command_utils/helm_backend/command_job.py
```diff
@@ -0,0 +1,174 @@
...
+    return sorted((_completion_index_of_pod(pod), pod.metadata.name) for pod in listed.items)
+
+
+def _completion_index_of_pod(pod: Pod) -> int:
+    raw = pod.metadata.labels.get(_COMPLETION_INDEX_KEY, pod.metadata.annotations.get(_COMPLETION_INDEX_KEY))
+    return int(raw) if raw is not None else 0
+
+
+def _logs_of_target(target: str, *, namespace: str) -> str:
+    return Kubectl.logs(target, namespace=namespace, tail=_TERMINAL_LOG_LINES)
```
fzyzcjy: [P2] Preserve full raw stdout in capture mode

`capture_output=True` uses this diagnostic-tail helper, so successful output is capped at 200 lines; `Kubectl.logs_command` also adds timestamps, and `Kubectl.logs` turns a failed log read's stderr into a normal result. Parsers and commands with larger output therefore receive altered or truncated data instead of the stdout returned by the Ray backend. Use a checked, unbounded, non-timestamped read for successful captures and reserve this tail for errors (`.claude/rules/general-code-style.md:41-43`).

fzyzcjy (commented):
The Kubernetes command path has correctness regressions in node-count handling, single-pod identity, concurrent ownership, and captured-output semantics. These paths remain present in target deliver f5a18edf2aa9fade59d0cff05d865c310fd52e7a, so they are blockers.

PR #2464, opened 2026-08-12
Compute which engine pod pairs with which trainer pod

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/external_utils/command_utils/helm_backend/launcher/values/colocate.py
+++ b/miles/utils/external_utils/command_utils/helm_backend/launcher/values/colocate.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+from miles.utils.external_utils.colocate_pairing.config import InferencePool, PairingConfig, PairingLayout
+from miles.utils.external_utils.command_utils.helm_backend.launcher.values.misc import (
+    INFERENCE_ENGINES_SECTION,
+    SECTION_OF_CATEGORY,
+    TRAINER_ENGINES_SECTION,
+    LaunchPlan,
+)
+from miles.utils.workers.worker_spec import BaseWorkerSpec
+
+
+def pairing_config(specs: list[BaseWorkerSpec], plan: LaunchPlan) -> PairingConfig:
+    inference_specs = [spec for spec in specs if SECTION_OF_CATEGORY[spec.category] == INFERENCE_ENGINES_SECTION]
+    trainer_specs = [spec for spec in specs if SECTION_OF_CATEGORY[spec.category] == TRAINER_ENGINES_SECTION]
+    assert len(trainer_specs) == 1, (
+        f"colocate pins inference pools onto one trainer pool's nodes, but this run has {len(trainer_specs)} "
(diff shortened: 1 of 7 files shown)


miles/utils/external_utils/command_utils/helm_backend/launcher/values/colocate.py
```diff
@@ -0,0 +1,90 @@
...
+    TRAINER_ENGINES_SECTION,
+    LaunchPlan,
+)
+from miles.utils.workers.worker_spec import BaseWorkerSpec
+
+
+def pairing_config(specs: list[BaseWorkerSpec], plan: LaunchPlan) -> PairingConfig:
+    inference_specs = [spec for spec in specs if SECTION_OF_CATEGORY[spec.category] == INFERENCE_ENGINES_SECTION]
+    trainer_specs = [spec for spec in specs if SECTION_OF_CATEGORY[spec.category] == TRAINER_ENGINES_SECTION]
+    assert len(trainer_specs) == 1, (
```
fzyzcjy, line 16: [P1] Handle the critic pool in colocated PPO

With `--advantage-estimator ppo --colocate`, as emitted by `examples/ppo/run_qwen3_4b_ppo.py`, `compute_specs()` creates both actor and critic trainer pools, so this exact-one assertion aborts the Kubernetes launch before Helm rendering. Deliver-2 still has this guard (`f5a18edf2a`); select the actor pool for inference pairing rather than rejecting the valid critic pool, preserving the existing launch contract (`.claude/rules/general-code-style.md:14-15`).

miles/utils/external_utils/command_utils/helm_backend/launcher/values/colocate.py
```diff
@@ -0,0 +1,90 @@
...
+        f"colocate pins inference pools onto one trainer pool's nodes, but this run has {len(trainer_specs)} "
+        f"trainer pools; which one an inference pool belongs beside would be undefined"
+    )
+
+    trainer = trainer_specs[0]
+    trainer_total_gpus = trainer.scheduling.num_cells * trainer.scheduling.gpus_per_cell()
+    colocated_inference_specs = [
+        spec for spec in inference_specs if spec.scheduling.pg_slot_offset < trainer_total_gpus
+    ]
+    assert colocated_inference_specs, (
```
fzyzcjy, line 26: [P2] Omit pairing when no local inference pool exists

When `scripts/run_deepseek_v4.py` replays saved rollout data, it emits both `--colocate` and `--debug-train-only`; `specs_inference_engine()` then returns no engine specs while `args.colocate` remains true, so this assertion aborts Helm rendering even though no pairing controller is needed. Deliver-2 leaves the condition unchanged (`f5a18edf2a`); omit the colocate section when the local inference list is empty, preserving the script contract (`.claude/rules/general-code-style.md:14-15`).

fzyzcjy (commented):
The pairing builder rejects established colocated launch shapes for shared actor/critic PPO and train-only replay. Both launch-blocking guards remain in deliver-2.

PR #2465, opened 2026-08-12
Hold a colocated engine pod until something places it

fzyzcjy:
Part of #1837

### DIFF
--- a/charts/miles-run/templates/_pool.tpl
+++ b/charts/miles-run/templates/_pool.tpl
@@ -1,6 +1,16 @@
+{{- define "miles-run.colocateRole" -}}
+{{- $colocate := .context.Values.run.colocate | default dict -}}
+{{- $poolId := .poolId -}}
+{{- $isInference := false -}}
+{{- range $colocate.inference_pools }}{{- if eq .pool_id $poolId }}{{- $isInference = true }}{{- end }}{{- end }}
+{{- if $isInference }}inference{{ else if eq $poolId ($colocate.trainer_pool_id | default "") }}trainer{{ end }}
+{{- end }}
+
 {{- define "miles-run.pool" -}}
 {{- $context := .context }}
 {{- $pool := .pool }}
+{{- $role := include "miles-run.colocateRole" (dict "context" $context "poolId" (default $pool.name $pool.poolId)) }}
+{{- $gated := eq $role "inference" }}
 {{- $name := $pool.objectName }}
 {{- $labels := dict "context" $context "component" $pool.name }}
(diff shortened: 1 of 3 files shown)


charts/miles-run/templates/_pool.tpl
```diff
@@ -28,7 +38,14 @@ spec:
...
         {{- end }}
       spec:
-        {{- include "miles-run.podDefaults" $context | nindent 8 }}
+        {{- include "miles-run.podDefaultsFor" (dict "context" $context "gated" $gated) | nindent 8 }}
+        {{- if $role }}
+        hostIPC: true
+        {{- end }}
+        {{- if $gated }}
+        schedulingGates:
+          - name: "miles.radixark.io/colocate-pairing"
```
fzyzcjy, line 47: [P1] Re-pair engines after trainer rescheduling

When a trainer LeaderWorkerSet group is recreated and its indexed pods land on different nodes while the inference LeaderWorkerSet stays alive, this gate was removed during the first placement and is never restored. The deliver-2 implementation of `miles/utils/external_utils/colocate_pairing/controller.py::PairingController.reconcile` considers only `is_gated` inference pods, so it cannot detect the mismatch; the engine remains pinned to the old hostname and no longer shares the intended trainer's GPUs or IPC namespace. Recreate or re-gate paired inference pods whenever their trainer node changes.

fzyzcjy (commented):
Two correctness defects remain at deliver-2: placement is enforced only once and can be lost after trainer rescheduling, and explicit GPU requests are not cleared. These can respectively break colocation after recovery and prevent valid resource configurations from producing pods.

charts/miles-run/templates/_pool.tpl
```diff
@@ -51,6 +71,12 @@ spec:
               {{- end }}
             {{- end }}
             {{- $resources := default dict $pool.resources }}
+            {{- if $gated }}
+            {{- $limits := deepCopy ($resources.limits | default dict) }}
+            {{- $resources = deepCopy $resources }}
+            {{- $_ := set $limits "nvidia.com/gpu" 0 }}
+            {{- $ignored := set $resources "limits" $limits }}
```
fzyzcjy, line 78: [P2] Clear GPU requests for gated engines

When a pool explicitly supplies `resources.requests.nvidia.com/gpu`, which `PoolEntry.resources` accepts, this block changes only the limit and leaves the positive request intact. The resulting Pod has a GPU request above its zero limit, so Kubernetes rejects it instead of allowing the engine to share the trainer; the same template remains in deliver-2. Remove or zero the GPU request together with the limit.

PR #2466, opened 2026-08-12
Place each gated engine pod on the node its trainer landed on

fzyzcjy:
Part of #1837


charts/miles-run/templates/colocate-pairing.yaml
```diff
@@ -0,0 +1,81 @@
...
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ $name | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "miles-run.labels" $labels | nindent 4 }}
+spec:
+  replicas: 1
```
fzyzcjy, line 48: [P2] Prevent pairing-controller overlap during upgrades

When an upgrade changes `run.colocate` while gated engine Pods are recreated, Kubernetes' default RollingUpdate may run both controller versions even though the desired replica count is one. The old replica can remove a new Pod's gate using stale placement data before the replacement handles it, and the released Pod cannot then be re-gated or repinned. Deliver-2 retains this blocker; use a non-overlapping strategy such as `Recreate` or fence controller generations.

miles/utils/external_utils/colocate_pairing/pods.py
```diff
@@ -0,0 +1,55 @@
...
+        pool_id=parsed.pool_id, cell_index=parsed.cell_index, pod_in_cell_index=parsed.pod_in_cell_index
+    )
+
+
+def release_patch(*, node_name: str, gates: list[str], has_node_selector: bool) -> list[dict[str, object]]:
+    index = gates.index(_GATE_NAME)
+    pin = (
+        {"op": "add", "path": f"/spec/nodeSelector/{_escape_pointer(_HOSTNAME_LABEL)}", "value": node_name}
+        if has_node_selector
+        else {"op": "add", "path": "/spec/nodeSelector", "value": {_HOSTNAME_LABEL: node_name}}
```
fzyzcjy, line 37: [P2] Pin engines by the trainer node's actual identity

On clusters where a Node's `metadata.name` differs from its `kubernetes.io/hostname` label, `trainer_pod.spec.node_name` supplies the former but this patch uses it as the latter. Removing the gate then leaves the engine with a selector that matches no node, so colocation never starts. Deliver-2 retains this blocker; use name-based node affinity such as `matchFields` on `metadata.name`, or resolve the Node's actual hostname label first.

fzyzcjy (commented):
The controller can irreversibly apply stale placement during overlapping Deployment rollouts, and it conflates Kubernetes node names with hostname-label values. Both correctness issues remain in deliver-2.

PR #2467, opened 2026-08-12
Skip a backend's tests when that backend is not there

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The standalone backend-capture and Kubernetes-probe defects are repaired by the specified safe deliver, so they are transient. No in-scope correctness or regression blocker remains at f5a18edf2aa9fade59d0cff05d865c310fd52e7a.

PR #2468, opened 2026-08-12
Let every e2e script pick its backend from the environment

fzyzcjy:
Part of #1837

### DIFF
--- a/tests/e2e/fsdp/test_qwen3_0.6B_megatron_fsdp_align.py
+++ b/tests/e2e/fsdp/test_qwen3_0.6B_megatron_fsdp_align.py
@@ -2,7 +2,7 @@ import os
 
 from tests.ci.ci_register import register_cuda_ci
 
-import miles.utils.external_utils.command_utils as U
+from miles.utils.external_utils import command_utils
 
 register_cuda_ci(
     est_time=900,
@@ -20,6 +20,7 @@ MEGATRON_PP_SIZE = 1
 
 
 def prepare():
+    U = command_utils.default_config().create_backend()
     U.exec_command_cpu("mkdir -p /root/models /root/datasets")
     U.exec_command_cpu(f"hf download Qwen/{MODEL_NAME} --local-dir /root/models/{MODEL_NAME}")
     U.hf_download_dataset("zhuzilin/dapo-math-17k")
@@ -33,6 +34,7 @@ def prepare():
 
 
 def execute():
+    U = command_utils.default_config().create_backend()
     ckpt_args = f"--hf-checkpoint /root/models/{MODEL_NAME}/"
 
     rollout_args = (
@@ -83,7 +85,7 @@ def execute():
         f"{rollout_args} "
         f"{optimizer_args} "
         f"{ppo_args} "
-        f"{U.get_default_wandb_args(__file__)} "
+        f"{command_utils.get_default_wandb_args(__file__)} "
         f"{sglang_args} "
         f"{ci_args} "
         f"{misc_args} "
--- a/tests/e2e/megatron/model_scripts/test_glm5_2_744b_a40b_5layer_ci.py
+++ b/tests/e2e/megatron/model_scripts/test_glm5_2_744b_a40b_5layer_ci.py
@@ -22,7 +22,6 @@ else:
 from tests.ci.ci_register import register_cuda_ci, register_rocm_ci
 from tests.ci.metric_history import register_ci_gate
 
-import miles.utils.external_utils.command_utils as U
 
 # Smoke test for the GLM-5.2 (glm_moe_dsa) training script. Exercises the DSA
 # cross-layer index-sharing path (5 layers = 3 dense + 2 MoE, computing layers
@@ -56,6 +55,7 @@ def _args() -> ScriptArgs:
 
 
 def prepare(args: ScriptArgs):
+    U = args.create_backend()
     U.exec_command_cpu(f"mkdir -p {args.output_dir}")
     _prepare_download(args)
     _validate_glm_checkpoint(args)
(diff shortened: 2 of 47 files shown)


tests/e2e/megatron/model_scripts/test_glm5_2_744b_a40b_5layer_ci.py
```diff
@@ -57,6 +56,7 @@ def _args() -> ScriptArgs:
 
 
 def prepare(args: ScriptArgs):
+    U = args.create_backend()
```
fzyzcjy, line 59: [P1] Build model-script configs from the environment

When `MILES_SCRIPT_CLUSTER_BACKEND=kubernetes`, `_args()` directly instantiates `ScriptArgs`, so inherited launcher fields never pass through `default_config`: `cluster_backend` remains `RAY` and `namespace` remains empty. Consequently this new `args.create_backend()` and `_execute_train(args)` still launch Ray; the other changed GLM, Inkling, and Kimi model-script drivers have the same pattern. Seed the launcher fields from the environment before applying recipe overrides. This remains in `f5a18edf2a`, so it blocks the target deliver.

tests/e2e/fsdp/test_qwen3_0.6B_megatron_fsdp_align.py
```diff
@@ -38,6 +39,7 @@ def prepare():
 
 
 def execute():
+    U = command_utils.default_config().create_backend()
```
fzyzcjy, line 42: [P1] Give each Kubernetes training run a fresh run id

When Kubernetes is selected, this single backend retains one `run_id` while `execute()` invokes it three times with different FSDP/Megatron commands. The first release still exists during the immediate second launch, so the launcher treats it as an in-place Helm upgrade and rejects the command changes as non-scaling; `precision/test_qwen3_0.6B_parallel_check.py` similarly reuses one backend throughout its sweep. Use a distinct run config/id for every independent launch. This remains in `f5a18edf2a`.

tests/e2e/sglang_config/test_sglang_config.py
```diff
@@ -31,12 +31,14 @@
...
 
 def prepare():
+    U = command_utils.default_config().create_backend()
     U.exec_command_cpu("mkdir -p /root/models /root/datasets")
     U.exec_command_cpu(f"hf download Qwen/{MODEL_NAME} --local-dir /root/models/{MODEL_NAME}")
     U.hf_download_dataset("zhuzilin/gsm8k")
 
 
 def execute():
+    U = command_utils.default_config().create_backend()
```
fzyzcjy, line 40: [P1] Put generated SGLang configs on shared storage

With the Kubernetes backend and default mounts, `NamedTemporaryFile` creates the YAML only in the launcher pod, but `--sglang-config` is subsequently consumed inside the orchestration/training pods, where that path does not exist. The two mixed-offload SGLang-config scripts repeat this pattern. Materialize the YAML on configured shared storage or pass it as portable inline content. All three cases remain unchanged in `f5a18edf2a`.

tests/e2e/short/test_dumper.py
```diff
@@ -100,6 +101,7 @@ def prepare(dump_dir: str, mode: str) -> None:
 
 
 def _execute(perf_args: str, dump_subdir: str, dump_dir: str) -> None:
+    U = command_utils.default_config().create_backend()
```
fzyzcjy, line 104: [P1] Put dumper artifacts on shared storage

With Kubernetes selected, `_execute()` runs in separate pods while this test's source-patcher files and dump directory remain under launcher-local `tempfile.mkdtemp()` storage; the pods cannot read the configs, and the launcher cannot see their dumps for verification. The FT harness's source-patcher config and `sglang/test_r3_router_equivalence.py`'s `/tmp` dumps have the same cross-pod problem. Use portable config payloads and a configured shared output root. These paths remain in `f5a18edf2a`.

tests/e2e/short/test_qwen2.5_0.5B_opd_sglang.py
```diff
@@ -85,6 +86,7 @@ def _launch_teacher_server(teacher_gpu: str):
 
 
 def execute():
+    U = command_utils.default_config().create_backend()
```
fzyzcjy, line 89: [P2] Launch a reachable OPD teacher on Kubernetes

When the environment selects Kubernetes, `before_ray_job_submit` is never invoked because only `RayCommandBackend` runs that callback, so `launch_teacher` does not start; moreover, the workers' `--rm-url` uses `127.0.0.1`, which addresses each worker pod rather than the launcher. The first reward request therefore cannot reach a teacher. Deploy the teacher with a cluster-reachable address or keep this script explicitly Ray-only. This remains in `f5a18edf2a`.

tests/e2e/megatron/test_qwen3_4B_offload_disk.py
```diff
@@ -69,6 +70,7 @@ def _assert_offloaded_to_disk():
 
 
 def execute():
+    U = command_utils.default_config().create_backend()
```
fzyzcjy, line 73: [P2] Verify disk offload without Ray-only logs

When Kubernetes is selected, training runs in worker pods rather than local Ray actors, but after completion `_assert_offloaded_to_disk()` still searches the launcher's `/tmp/ray/session_latest/logs/worker-*`. That glob is empty even after a successful Kubernetes run, so the test always fails; the streaming sibling has the same dependency. Emit backend-neutral evidence to shared storage or retrieve the pod logs. This remains in `f5a18edf2a`.

fzyzcjy (commented):
Several e2e drivers either still pin Ray or select Kubernetes while retaining launcher-local and Ray-specific assumptions. These failures remain at the target safe deliver, so the op is not safely deliverable.

PR #2469, opened 2026-08-12
Document the cluster backend for the people who will use it

fzyzcjy:
Part of #1837

### DIFF
--- a/docs/advanced/cluster-backend.md
+++ b/docs/advanced/cluster-backend.md
@@ -1,0 +1,101 @@
+---
+title: Ray and Kubernetes Backend
+description: Run the same training script on Ray, or let Kubernetes schedule every worker of the run.
+---
+Miles runs a job on one of two cluster backends. Ray is the default and needs no configuration.
+Kubernetes installs the run as a helm release, so the cluster schedules every worker. The training
+script is the same either way.
+
+<Warning>
+
+**Status.** Under active development: flags, chart values and failure semantics still change. Ray
+also runs *inside* Kubernetes, and that is the well-trodden path — take this one only if you want
+Miles to create the cluster's objects itself.
+
+</Warning>
+
+## Launch
+
+Run everything from the repository root, with `kubectl` and `helm` on your PATH.
+
+Install a workbench — the long-lived pod you launch from — once per namespace, from your
+cluster's [`infra.yaml`](#for-cluster-administrator):
+
+```bash
+export MILES_NS="miles-$USER"
+python -m miles.utils.external_utils.miles_workbench install -n "$MILES_NS" -f infra.yaml
+```
+
+It creates the namespace if missing, checks your rights, installs, and waits for Ready. Runs
+launched from the pod inherit those values.
+
+Launch from inside it:
+
+```bash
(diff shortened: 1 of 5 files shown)


docs/advanced/cluster-backend.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+It creates the namespace if missing, checks your rights, installs, and waits for Ready. Runs
+launched from the pod inherit those values.
+
+Launch from inside it:
+
+```bash
+python -m miles.utils.external_utils.miles_workbench exec -n "$MILES_NS" -- \
+  bash -lc "cd /root/miles && python scripts/run_qwen3_4b.py train"
```
fzyzcjy, line 36: [P2] Put the launcher's default paths on shared storage

When a reader uses the supplied `infra.yaml`, only `/cluster-storage` is mounted, but this command invokes `train` with `/root/models`, `/root/datasets`, and `/root/shared_data` defaults; `train` also does not prepare those inputs. Worker pods therefore use private root filesystems, causing missing data or unshared checkpoints. Deliver-2 (`f5a18edf2a`) retains this mismatch; point these paths at shared storage and document preparation, or mount the defaults.

fzyzcjy (commented):
The documented launch does not work with its own storage example, and the administrator guidance omits essential multi-node storage and isolation requirements. All three issues remain in the target deliver.

docs/advanced/cluster-backend.md
```diff
@@ -0,0 +1,101 @@
...
+Everything above assumes this was done once.
+
+**Install LWS.** Miles deploys its worker pools as
+[LeaderWorkerSets](https://github.com/kubernetes-sigs/lws). Install the CRDs and controller, and
+grant users rights over them explicitly: LWS ships no aggregation labels, so a namespace `admin`
+role does not include them.
+
+**Give each user a namespace.** The namespace is the real boundary, not the Role: anything that
+may create workloads can name another ServiceAccount and read its token. Keep privileged accounts
+out of it.
```
fzyzcjy, line 85: [P1] Require admission controls before calling namespace a boundary

In a multi-tenant cluster without an admission policy, namespaced RBAC is not the boundary claimed here: the root-running workbench may create Deployments and Jobs, so a user can submit a privileged workload or arbitrary `hostPath` and access the node or other tenants without any privileged ServiceAccount in their namespace. Deliver-2 retains this guidance and capability; require appropriate Pod Security or admission allowlists, or qualify the guarantee.

docs/advanced/cluster-backend.md
```diff
@@ -0,0 +1,101 @@
...
+
+```yaml
+infra:
+  image:
+    repository: radixark/miles
+    tag: dev
+  sharedStorage:
+    type: hostPath
+    hostPath: /cluster-storage
+    mountPath: /cluster-storage
```
fzyzcjy, line 97: [P2] State that hostPath must already be shared

When `/cluster-storage` is an ordinary directory on each node, Kubernetes `hostPath` gives pods scheduled on different machines different filesystems; the workbench then cannot see the orchestrator's state file and workers cannot share models or checkpoints. The example must state that this path is backed by a cluster-wide filesystem or use an RWX PVC. Deliver-2 still presents an unqualified `hostPath` example.

PR #2470, opened 2026-08-12
Assert that the rollout executor keeps receiving weight versions

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/weight_version.py
+++ b/miles/utils/weight_version.py
@@ -11,6 +11,20 @@ SGLANG_DEFAULT_WEIGHT_VERSION = "default"
 
 _NUMERIC_VERSION_PATTERN = re.compile(r"[0-9]+")
 
+MAX_ROLLOUTS_WITHOUT_PUBLISHED_WEIGHT_VERSION = 3
+
+
+def assert_weight_version_is_published(args: Namespace, *, rollouts_since_publish: int) -> None:
+    if args.debug_rollout_only or args.debug_train_only or args.debug_skip_weight_update or is_lora_enabled(args):
+        return
+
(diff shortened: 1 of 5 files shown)


miles/utils/weight_version.py
```diff
@@ -11,6 +11,20 @@
...
 _NUMERIC_VERSION_PATTERN = re.compile(r"[0-9]+")
 
+MAX_ROLLOUTS_WITHOUT_PUBLISHED_WEIGHT_VERSION = 3
+
+
+def assert_weight_version_is_published(args: Namespace, *, rollouts_since_publish: int) -> None:
+    if args.debug_rollout_only or args.debug_train_only or args.debug_skip_weight_update or is_lora_enabled(args):
+        return
+
+    assert rollouts_since_publish <= MAX_ROLLOUTS_WITHOUT_PUBLISHED_WEIGHT_VERSION, (
```
fzyzcjy: [P1] Derive the publish gap from the configured cadence

With `train_async.py --update-weights-interval >= 3` and at least four rollouts, the supported one-ahead prefetch enters `RolloutExecutor.get()` four times before the scheduled publish, so this fixed limit aborts a correctly wired run on the fourth call. Account for `args.update_weights_interval` plus the prefetched rollout, or reject incompatible intervals; the same cap remains in deliver-1 `deec6fce` and deliver-2 `f5a18edf`, so this is still a blocker.

fzyzcjy (commented):
The hard-coded rollout limit crashes valid asynchronous runs whose configured weight-update cadence exceeds two rollouts, and the defect persists through both safe delivers.

PR #2471, opened 2026-08-12
Fail the run when the FT api server cannot start

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No production correctness or regression blocker remains at deliver-1. The transient trainer type/import mismatch is repaired there, while API-server startup failure propagation remains intact through both safe delivers.

PR #2472, opened 2026-08-12
Run the weight equality check before leaving StateInitializing

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/server_cell.py
+++ b/miles/ray/rollout/server_cell.py
@@ -166,6 +166,14 @@ class ServerCell:
             await api_client.resume_memory_occupation(tags=[GPU_MEMORY_TYPE_WEIGHTS])
 
         serve_without_weight_update: bool = not self.meta.update_weights or self.args.debug_rollout_only
+        if not serve_without_weight_update and self.args.check_weight_update_equal:
+            await self.check_weights(
+                action="reset_tensors",
+                allow_quant_error=False,
+                selector="all",
+                skip_list=self.args.check_weight_update_skip_list,
+            )
+
         if serve_without_weight_update:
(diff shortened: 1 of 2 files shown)


miles/ray/rollout/server_cell.py
```diff
@@ -166,20 +166,21 @@ async def _tick_when_initializing(self) -> None:
             await api_client.resume_memory_occupation(tags=[GPU_MEMORY_TYPE_WEIGHTS])
 
         serve_without_weight_update: bool = not self.meta.update_weights or self.args.debug_rollout_only
+        if not serve_without_weight_update and self.args.check_weight_update_equal:
+            await self.check_weights(
+                action="reset_tensors",
+                allow_quant_error=False,
+                selector="all",
+                skip_list=self.args.check_weight_update_skip_list,
+            )
```
fzyzcjy, line 177: [P2] Preserve the baseline when retrying a destructive reset

When `reset_tensors` mutates any tensors/ranks and then raises, or completes but its response is lost or cancelled, this await leaves the cell initializing. The next tick re-executes the earlier `snapshot`, replacing the pristine checkpoint baseline with already-randomized weights; after a successful retry and correct trainer sync, `compare` reports a false mismatch. Preserve the first snapshot/initialization progress or restart the engine rather than replaying the whole sequence; this ordering still exists in deliver-1 `deec6fce05` and deliver-2 `f5a18edf2a`.

fzyzcjy (commented):
The state move makes reset failures retryable, but the retry re-snapshots weights after the destructive reset may already have executed. This can corrupt the comparison baseline and remains unfixed at both stated safe deliveries.

PR #2473, opened 2026-08-12
Bound and wrap the free-port scan

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The bounded scan covers each valid candidate once, wraps safely, and rejects exhaustion instead of probing invalid ports indefinitely. No correctness regression was found, and the behavior remains unchanged at deliver-1 and deliver-2.

PR #2474, opened 2026-08-12
Report a rollout cell stuck initializing as unhealthy

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No production correctness or regression blocker persists at deliver-1. The startup deadline is scoped to actively initializing cells and feeds the existing unhealthy-cell recovery path as intended.

PR #2475, opened 2026-08-12
Let workers deregister from the miles router

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/router/router.py
+++ b/miles/router/router.py
@@ -60,6 +60,7 @@ class MilesRouter:
         """Setup all the HTTP routes except catch-all proxy"""
         # sglang-router api
         self.app.post("/add_worker")(self.add_worker)
+        self.app.post("/remove_worker")(self.remove_worker)
         self.app.get("/list_workers")(self.list_workers)
         # Catch-all route for proxying to SGLang - must be registered LAST
         self.app.api_route("/{path:path}", methods=["GET", "POST", "PUT", "DELETE"])(self.proxy)
@@ -94,6 +95,8 @@ class MilesRouter:
                 results = await asyncio.gather(*(self._check_worker_health(url) for url in urls))
 
                 for url, is_healthy in results:
+                    if url not in self.worker_request_counts:
+                        continue
                     if not is_healthy:
                         failures = self.worker_failure_counts.get(url, 0) + 1
                         self.worker_failure_counts[url] = failures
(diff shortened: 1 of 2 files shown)


miles/router/router.py
```diff
@@ -202,6 +195,29 @@ async def add_worker(self, request: Request):
...
 
+    async def remove_worker(self, request: Request):
+        """Remove a worker from the router, using the same URL conventions as add_worker."""
+        worker_url = await self._parse_worker_url(request)
+        if worker_url is None:
+            return JSONResponse(
+                status_code=400, content={"error": "worker_url is required (use query ?url=... or JSON body)"}
+            )
+
+        self.worker_request_counts.pop(worker_url, None)
```
fzyzcjy, line 209: [P1] Preserve accounting for in-flight proxy requests

When `--use-miles-router` deregisters a URL while `do_proxy()` is awaiting an in-flight request, this pop deletes the counter that `_finish_url()` unconditionally accesses in `finally`, so a completed response becomes an `AssertionError`/500. If the same URL is re-registered first, the old completion decrements the replacement's counter below zero and subsequent requests can keep failing. Preserve a draining or generation-scoped record until outstanding requests complete. The router blob is unchanged at deliver-1 and deliver-2, so this is a blocker rather than transient.

miles/router/router.py
```diff
@@ -202,6 +195,29 @@ async def add_worker(self, request: Request):
...
+        self.worker_failure_counts.pop(worker_url, None)
+        self.dead_workers.discard(worker_url)
```
fzyzcjy, line 211: [P2] Ignore stale health results after deregistration

When deregistration overlaps `_health_check_loop()`'s awaited probe, the loop still holds this URL in its snapshot and can recreate `worker_failure_counts[url]` and `dead_workers` after these lines. With `health_check_failure_threshold=1`, a stale failure leaves the URL quarantined; a later `add_worker` initializes its counters but never clears the stale dead marker, so a replacement at the same address receives no traffic. Apply results only to the registration generation that was probed. Identical code remains at both safe delivers, so this is also a blocker.

fzyzcjy (commented):
The endpoint deletes registration state while existing asynchronous router operations can still own that URL. Both races survive deliver-1 and remain in deliver-2, so they are blockers rather than transient issues.

Issue #2406, opened 2026-08-12
FSDP VLM: text-only microbatch in a multimodal batch desyncs vision-tower collectives (30-min stall / NCCL watchdog kill)


nblintao:
## Symptom

Training a VLM (Qwen3-VL-4B, FSDP2, 8xH200) on a batch that mixes multimodal samples with a few **text-only** samples (in our case: multi-turn CUA episodes whose usable prefix ended before the first screenshot):

- With the default `--distributed-timeout-minutes 10`: the job dies at the end of the first rollout with `Watchdog caught collective operation timeout: _ALLGATHER_BASE`.
- With a raised timeout: every train step stalls ~30 minutes (the gloo default 1800s), then continues. GPUs show 100% util at ~118W (NCCL spin-wait, no compute). `gather_log_data` also times out (`kind=log_gather success=false degraded=true`) because ranks arrive up to 30 min apart.

Observed live 2026-08-12 on `tao/hud-v6-cua` (HUD v6 recipe); 5 of 32 samples were text-only, ranks 3/5/6 held them.

## Mechanism

`_get_model_inputs_args` (miles/backends/fsdp_utils/actor.py:668) passes `multimodal_train_inputs` only when present. A microbatch without images never calls `model.visual`, so that rank skips the vision tower's FSDP weight all-gathers while other ranks issue them. Collectives on the communicator pair up across ranks in issue order, so the ranks that did call the ViT block until some later phase happens to post a matching collective — the smoking gun in our logs: ranks stuck 30:00 in `log_gather` were released within 200ms of the other ranks' gloo timeout firing.

Standalone ViT forward on the same data is 0.4s — this is pure collective-schedule desync, not compute.

Note `remove_sample` write-offs don't escape this: train_data_conversion.py:96 zeroes their loss mask but keeps them in the batch, so a pad-token write-off is itself a text-only microbatch.

## Repro shape

Any FSDP VLM run where per-rank microbatches differ in whether they contain images. `mbs=1` + a batch with ≥1 image-free sample is the minimal case.

## Fix directions

The standard guard (verl et al.): when a microbatch has no images, run the vision tower on a dummy image and add its output to the loss with weight 0, so every rank issues the same collectives. Alternatively document + assert that multimodal batches must not contain text-only samples.

Our recipe works around it by making every sample — including write-offs — carry one minimal image (`examples/experimental/hud/rollout.py::_failed`), but the trap is generic to any multimodal + FSDP user.

[...]

Issue #2393, opened 2026-08-11
Promote the credential-supply contract out of the openenv example into miles/


nblintao:
Ray records a job's `runtime_env` verbatim, `exec_command_cpu` echoes the whole
submit command into the driver log, and Ray persists it in job metadata — all in
plaintext. So any launcher that forwards a provider API key **by value** leaks it
into logs that outlive the run.

The openenv example solved this properly, in `openenv_launch_common.py`:

- workers get the credential from their own environment or from a file they can
  read (dotfile, K8s Secret mount, shared FS); the launcher forwards only the
  **path**
- a launch-time preflight resolves the supply and prints which one is in effect,
  and fails loudly when neither is available
- the SDK is preflighted too, because both failures otherwise first appear
  *inside an episode*, where the sample is dropped and the rollout loop refills
  forever — silent GPU-burning churn instead of an error
- addresses that could smuggle a credential (a URL with `user:token@host`) are
  refused rather than forwarded

That logic is currently private to `examples/experimental/openenv/`. Writing the
HUD example (#TBD) needed the same contract, and I reimplemented a compact
version rather than import across examples — so the pattern now exists twice, and
the next sandbox-backed connector will make it three.

### Proposal

Promote the contract to `miles/` (e.g. `miles/utils/credential_supply.py`) with
the provider table staying in each example:

```python
forward_credential_path(
    provider="Daytona",
    key_env_vars=("DAYTONA_API_KEY",),
    file_env_var="DAYTONA_API_KEY_FILE",
    default_path="~/.config/daytona/api_key",
    provision_hint="mkdir -p ~/.config/daytona && echo dtn_... > ~/.config/daytona/api_key",
) -> dict[str, str]   # env vars safe to forward
```

Then migrate `examples/experimental/openenv/` and `examples/experimental/hud/`
onto it, and keep `_forward_address`'s `@` check with it.

### Worth deciding alongside

Whether `exec_command_cpu` should redact `*_API_KEY` / `*_TOKEN` / `*_SECRET`
values when it echoes the submit command. That is defence in depth for launchers
that forward a secret anyway, and it is cheap.

PR #2479, opened 2026-08-12
examples: computer-use RL on HUD v6 environments

nblintao:
Trains a VLM to operate a real GUI through [HUD](https://hud.ai) v6
environments. HUD owns the episode — it boots the environment (one Daytona
sandbox per episode), runs the agent loop, and grades — while Miles owns
training. Nothing under `miles/` changes.

Verified on HUD's public browser starter (`hud init --preset browser`, no API
key), training Qwen3-VL-4B-Instruct with GRPO on 8×H200.

## What had to be built, and why

HUD's harness already collects per-turn token-level samples, so the connector's
job is not recording but three narrower things:

- **A screen for self-hosted models** — HUD ships computer-use tools for
  Claude/Gemini/OpenAI's native protocols; its OpenAI-*compatible* harness, the
  one an sglang policy speaks through, has none. `computer_tool.py` puts HUD's
  own VNC primitives behind a plain function-calling schema.
- **A token-id bridge** — the harness asks for `return_token_ids`, a vLLM
  extension; stock sglang has the same data under `return_prompt_token_ids` +
  `return_meta_info`. `sglang_compat.py` is a client-side transport that
  rewrites the request and synthesizes `choice.token_ids` from the meta_info
  logprob triples. No change on either side; plausibly upstreamable to sglang.
- **The stitch** — the server re-renders the chat template every turn, so
  `rollout.py` verifies per episode that turn k+1's prompt ids start with turn
  k's prompt+output, then concatenates into one sequence with loss mask 1 on
  policy tokens and 0 on scaffolding. One divergence is tolerated because it
  is canonicalization rather than rewriting: the tool parser collapses the
  whitespace the model emitted before its tool call (`.\n\n<tool_call>` →
  `.\n<tool_call>`), so a divergence whose two sides are text-equal once
  whitespace is stripped — and whose remainder realigns token-for-token —
  adopts the re-rendered tokens with mask 0 and continues. Anything else ends
  the episode at the last verified turn. Vision inputs are rebuilt from the
  screenshots the policy saw and checked against the image-token count in the
  ids; a mismatch drops the episode loudly, because a mis-stitched sequence
  corrupts training silently.

## Layering

Dependencies run one way, and each layer is ignorant of the one above it:

| layer | files | knows nothing about |
|---|---|---|
| transport | `sglang_compat.py` | HUD |
| computer use | `computer_tool.py`, `agent.py` | which task is running |
| HUD | `rollout.py`, `make_hud_data.py` | which taskset it runs, or what its reward means |
| the 2048 recipe | `hud2048_config.yaml`, `run_hud2048.py` | — |

A change lands in one of three places by kind: a **mechanism** in the general
layers, a **number** in the recipe config (each annotated with what it trades
against), and the **reward shape** in the data — which task template to train
on, retuned via `make_hud_data.py --args-json`. No code decides reward shape.

## What makes it learnable

Four things have to be true before RL has any signal:

- **The policy has to see the task.** At the starter's default 1280x800 the
  browser's own chrome crops the 4×4 board out of the viewport *and* gives the
  policy a sidebar to click, which takes keyboard focus away from the game.
  The fix is one edit to the env package (kiosk, taller desktop; the README
  has it).
- **Turns are the scarce resource, keystrokes are not.** Every turn re-prefills
  the whole screenshot history through the vision encoder. The system prompt
  says so, and the tool takes a key *sequence* per call — one call with several
  keys would be a chord, not several moves.
- **The reward has to be shaped to the reachable range.** The starter ships
  three graders for 2048; `reach_tile` is a step function that goes flat the
  moment a population converges on one tile, zeroing GRPO's within-group
  variance. This trains on `reach_score` (linear) with the target retuned.
- **Failures must be write-offs, not exceptions.** A long run starts hundreds
  of cloud sandboxes; a provider hiccup is routine. Episodes that error or fail
  to stitch become `remove_sample` write-offs with every field shaped as the
  trainer expects — including one dummy image, because under FSDP a text-only
  microbatch skips the vision tower's weight all-gathers and desynchronizes
  the ranks' collective schedules (#2406). One bad episode costs one sample
  rather than the rollout.

## Results

20 GRPO steps, 32 episodes/step, ~15 min/step on 8×H200:

- Mean episode reward 0.037 → ~0.07 (`reach_score` against the retuned
  target 1024); by the last steps all 32 episodes in a batch score, and
  `zero_std/all_zero_percentage` stays 0 throughout — every group keeps a
  gradient.
- Behaviourally: arrow keys per episode up ~40%, malformed tool calls → 0.
  The model first learns to reliably reach the board and spend its move
  budget — what a linear score reward buys first.
- 0 rollout failures over 640 episodes; the whitespace resync recovers every
  would-be stitch break (148/148), which roughly doubles the trained tokens
  per episode in the early steps.

## Tests

28 offline cases against fakes — no GPU, no network, no Daytona, ~3s:

```
python -m pytest examples/experimental/hud/tests/ -q
```

They cover the three translations this example owns: the stitch (including
whitespace resync, the prefix break, and the trailing screenshot the policy
never saw), the tool's dispatch and coordinate scaling, and the sglang
rewrite. `MODE=smoke` runs two real episodes for what a fake cannot cover.

## Notes

- Supersedes #2371, which reached the same task through HUD's pre-v6 taskset
  format. That leg's Daytona lifecycle code and multi-turn token accounting are
  both gone here — v6's `DaytonaRuntime` and harness provide them — which also
  removes that PR's dependency on private helpers in another example.
- The Daytona credential travels as a *path*, never a value: Ray records
  `runtime_env` verbatim into driver logs and job metadata. Promoting that
  contract out of the openenv example is #2393.
- The FSDP text-only-microbatch guard this recipe carries locally belongs in
  the trainer; that's #2406.

### DIFF
--- a/docs/user-guide/environments.md
+++ b/docs/user-guide/environments.md
@@ -19,6 +19,7 @@ where the environment itself comes from:
 | Integration | Plugs in at | Guide |
 |---|---|---|
 | [Harbor](https://github.com/harbor-framework/harbor) | agent function | [guide](/user-guide/harbor) |
+| [HUD](https://hud.ai) | generate function³ | [example](https://github.com/radixark/miles/tree/main/examples/experimental/hud) |
 | [NeMo Gym](https://github.com/NVIDIA-NeMo/Gym) | agent function | [guide](/user-guide/nemo-gym) |
 | [OpenEnv](https://github.com/huggingface/openenv) | agent function | [guide](/user-guide/openenv) |
 | [Strands Agents](https://strandsagents.com/) | generate function | [example](https://github.com/radixark/miles/tree/main/examples/experimental/strands_sglang) |
@@ -31,7 +32,7 @@ Sandbox providers are a different axis: they provision the task containers
 | Sandbox provider | Used within | Guide |
 |---|---|---|
 | [AgentENV](https://github.com/kvcache-ai/AgentENV) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/agentenv) |
-| [Daytona](https://www.daytona.io/) | OpenEnv, Harbor, NeMo Gym | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
+| [Daytona](https://www.daytona.io/) | Harbor, HUD, NeMo Gym, OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 | [E2B](https://e2b.dev/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 | [Modal](https://modal.com/) | OpenEnv | [example](https://github.com/radixark/miles/tree/main/examples/experimental/openenv) |
 
@@ -62,3 +63,9 @@ rather than the session-server chat endpoint Miles' own recording uses.
 ² The environment may grade an episode itself (Harbor and τ-bench do); the
 score still enters training through Miles' `Sample.reward` / RM hooks, and
 group-level reward handling stays in Miles.
+
+³ HUD's harness records per-turn token ids and sampling logprobs itself when
+the inference server returns them, so the connector's job is stitching those
+into one training sequence rather than recording. Computer-use observations
+are screenshots, which Miles' session-server recording does not carry yet —
+once it does, this connector can also sit in the agent function.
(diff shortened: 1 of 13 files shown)


Shi-Dong (approved):
LGTM!

PR #2481, opened 2026-08-12
docs: mirror examples/ READMEs into the Examples tab

nblintao:
> [!NOTE]
> **Stacked on #2482** (examples README content fixes) — merge that first, then retarget this to `main`. The generator depends on those fixes: its dead-link check, required-H1 rule and description-length gate all fail against the unfixed READMEs.

## Review guide — what to actually read

**Hand-written (the real diff, ~500 lines):**

| File | What it is |
|---|---|
| `scripts/tools/sync_example_docs.py` | The generator — the one file that deserves real review |
| `.pre-commit-config.yaml` | `sync-example-docs` hook (regenerate + fail on drift) |
| `.gitattributes` | marks `docs/examples/**` linguist-generated |
| `docs/docs.json` | **redirects block is hand-written**; the Examples tab under `navigation` is generated |
| `examples/README.md` | +14 lines: the "These READMEs are the documentation site" section (docs:exclude-wrapped, GitHub-only) |
| `docs/README.md` | contributor note: don't edit `docs/examples/` by hand |
| `docs/developer/debug.md`, `docs/developer/index.md`, `docs/user-guide/rollout-endpoints.md` | repoint links that targeted the deleted pages |

**Generated — skip in review** (GitHub collapses them via linguist-generated; verify instead by checking out the branch and running `python scripts/tools/sync_example_docs.py --check`):

- `docs/examples/**` — 15 pages mirrored from `examples/**/README.md`, plus deletion of the 4 stale hand-written pages (`search-r1`, `multi-agent`, `reproducibility`, `openhermes-sft`; agreed in Slack, each 301s to its nearest replacement)

## Problem

The Examples tab on the docs site was hand-written and had drifted badly from the repo:

- 3 of its 5 pages documented `experimental/` recipes (search-r1, multi-agent, reproducibility) — the ones we agreed not to surface yet.
- 1 page (openhermes-sft) documented a script under `scripts/`, not an example at all.
- The only in-scope page (fully-async) was out of date, and search-r1 carried three broken file references.
- 9 non-experimental examples (geo3k_vlm, multi_lora, retool_v2, true_on_policy, ...) were never mentioned anywhere on the site.

Per the Slack discussion: `examples/` on main is the single source of truth, and the old pages are thrown away.

## What this does

**`scripts/tools/sync_example_docs.py`** mirrors every `README.md` outside `examples/experimental/` onto the docs site — 15 pages plus the Examples tab navigation in `docs.json` (Recipes and Infra Features groups, 1:1 with the directory layout). The converter handles what Mintlify's MDX pipeline rejects or silently drops, each case verified against a local `mint dev` render:

- unescaped `{`/`}` and non-self-closed void tags (``, `<br>`) fail the build → escaped/self-closed
- `<details>` passes validation but renders empty → converted to `<Accordion>`
- relative links/images are dead on the site → resolved to site pages, GitHub blob/tree URLs, or raw.githubusercontent for images
- `<!-- docs:exclude:start/end -->` markers keep contributor-facing content GitHub-only
- a page's meta description comes from its bullet in `examples/README.md` (or its own first sentence)

**Drift is now impossible**: the pre-commit hook regenerates the mirror (`--check` fails CI on divergence; pre-commit.ci autofixes PRs from authors who don't run hooks). Day to day you edit only the README — the mirror commit is made for you.

## Content deliberately dropped with the deleted pages

Called out explicitly since `linguist-generated` collapses exactly these diffs. Per the Slack decision ("throw them away"), the old hand-written pages are removed even where the replacement README is shorter:

| Deleted page | Old lines | Replacement | Net |
|---|---|---|---|
| `/examples/search-r1` | 275 | `experimental/search-r1/README.md` (305 lines, GitHub) | nothing lost |
| `/examples/multi-agent` | 197 | `experimental/multi_agent/README.md` (52 lines, GitHub) | ~145 lines dropped |
| `/examples/reproducibility` | 144 | `experimental/reproducibility/README.md` (52 lines, GitHub) | ~92 lines dropped |
| `/examples/openhermes-sft` | 163 | SFT section of `/models/qwen/qwen3` | walkthrough dropped |

The dropped walkthroughs live in git history; the natural moment to bring content back is when an example is promoted out of `experimental/`, at which point its README (the SOT) is the place to enrich.

## Verification

- `mint validate` (strict) and `mint broken-links` pass; the one remaining broken-links warning (`qwen3.md` → `/advanced/architecture-support`) pre-exists on main and is covered by a redirect.
- All 8 raw.githubusercontent image URLs return 200.
- Every page rendered via local `mint dev` and screenshot-checked; a paragraph-level audit compared source READMEs against served HTML (158 paragraphs, no real loss).
- `python scripts/tools/sync_example_docs.py --check` is idempotent; full pre-commit run passes.


Zhichenzzz:
mostly LGTM! need to rebase to latest main first

nblintao:
> mostly LGTM! need to rebase to latest main first

rebased

Zhichenzzz:
## Claude review

Requested by @Zhichenzzz. I read the generator in full, then checked its output against the sources on `tao/examples-readme-fixes`.

The design is sound and the execution is careful: `docs:exclude` markers for GitHub-only sections, the MDX hazards handled (unescaped braces, non-self-closing void tags, `<details>` rewritten to `<Accordion>`), unknown tags like `<tool_call>` escaped rather than parsed as JSX, relative links resolved to the site page when mirrored and to GitHub otherwise, missing link targets turned into a hard failure, and both `docs/README.md` and `examples/README.md` updated to say where the source of truth is. CI runs `pre-commit run --all-files`, so drift is gated even when only the script changes.

Three things below: one I think should land before merge, one that needs an explicit decision, and a list of smaller ones.

---

### 1. Inline math is corrupted in the generated output

`convert()` ends with `body.replace("{", "\\{")`. `mask_code()` protects fenced blocks, inline backticks and `$$…$$` display math, but **not** inline `$…$`, so every brace inside inline math gets escaped.

`docs/examples/infra-features/train-infer-mismatch-helper.md` line 71, as generated in this PR:

```
**proximal policy** $\pi_\{\textcolor\{blue\}\{\text\{old\}\}\}$
```

The source is `$\pi_{\textcolor{blue}{\text{old}}}$`. KaTeX reads `\{` as a literal brace, so the subscript grouping and the `\textcolor` / `\text` arguments fall apart. 18 occurrences on that page (line 71 and the table on 103-105). Display math on the same page is fine, because it is masked. The only other escaped brace in the whole output is in `geo3k-vlm/multi-turn.md`, where it is correct prose.

One more mask in `mask_code`, next to the `$$` one, fixes it:

```python
text = re.sub(r"(?<!\$)\$(?!\$)(?:\\.|[^$\n])+\$", lambda m: keep(m.group(0)), text)
```

Bounded to a single line, so a stray `$5` in prose cannot swallow the rest of the document.

### 2. Three pages of content disappear, and the diff hides it

The mirror can only publish what lives in `examples/`, so pages whose sources sit under `experimental/` (or nowhere) are deleted. Redirects were added, which is the right instinct, but the content behind three of them does not exist anywhere else:

| Deleted page | Lines | Where the content goes |
|---|---|---|
| `examples/search-r1` | 275 | fine, `examples/experimental/search-r1/README.md` is 305 lines |
| `examples/multi-agent` | 197 | README is 52 lines, ~145 lines lost |
| `examples/reproducibility` | 144 | README is 52 lines, ~92 lines lost |
| `examples/openhermes-sft` | 163 | no example exists at all; redirected to `/models/qwen/qwen3` |

Internal links were repointed at GitHub (`docs/developer/debug.md`, `docs/developer/index.md`), so nothing dangles. But the redirect sends a reader looking for the multi-agent walkthrough to an index page that does not contain it. Either port the surplus into the example READMEs first, so the mirror brings it back, or keep those three as hand-written pages.

If you take the second option, note that `main()` unlinks every `*.md` under `docs/examples/` that the generator did not produce, so a hand-written page cannot coexist there today; it would need an allowlist or a different directory.

Worth calling this out in the PR description either way: `linguist-generated=true` collapses exactly the diffs where a reviewer would otherwise notice it.

### 3. Smaller things

- **`escape_html` whitelist is narrow.** `HTML_TAGS` omits `h1`-`h6`, `center`, `figure`, `figcaption`, `picture`, `video`, `source`, `s`, `u`, `mark`, `dl`/`dt`/`dd`. None appear in today's READMEs (only `img`, `p`, `details`, `summary` plus placeholder tokens), but the failure is silent: a perfectly ordinary `<h3 align="center">` in a future README renders as literal `&lt;h3 align="center">` text with no error anywhere.
- **Slug collisions overwrite silently.** `slug_for` lowercases and maps `_` to `-`, and `rendered` is keyed by output path, so `foo_bar` and `foo-bar` would collapse into one page, last one wins, with `--check` still passing. No collisions among today's 15 pages. A guard when inserting into `rendered` closes it.
- **pre-commit `files:` does not include the script.** Editing the generator alone will not re-run it locally. CI's `--all-files` catches it, so this only costs a round trip; adding `|scripts/tools/sync_example_docs\.py` to the pattern makes the two agree.
- **`MD_LINK` mangles nested image-in-link.** `[badge](target)` matches as text `![badge`, href `img.png`; the outer href is never rewritten and never reported as broken. No occurrences today, so this is latent.
- **`docs.json` is now partly generated.** The script round-trips the whole file through `json.dumps(indent=1)` on every run. Formatting matches the current file so the diff is clean, but future hand-formatting elsewhere in that file will be normalized silently.
- **Nav order is now alphabetical.** `build_navigation` sorts by directory name, replacing the curated order (fully-async, search-r1, multi-agent, …). Fine as a rule; it just means importance ordering is no longer expressible without renaming directories. The `infra_features` grouping is also hardcoded by prefix.

### Nits

- `first_sentence` strips the duplicated opening sentence via `body.startswith(description)`, but the description has had its markdown removed by then, so the dedup silently no-ops when the first sentence contains a link or emphasis, and the page shows it twice.
- Braces inside an `<Accordion title="…">` produced by `convert_details_blocks` are escaped afterwards and would render as `\{`. No current occurrence.
- Links that escape the repo (`../../..`) return `None`, are kept verbatim, and are not reported as broken.

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+                out_lines.append(keep("\n".join(buffer)))
+                fence = None
+        else:
+            out_lines.append(line)
+    if fence is not None:
+        out_lines.append(keep("\n".join(buffer)))
+    text = "\n".join(out_lines)
+
+    text = re.sub(r"\$\$.*?\$\$", lambda m: keep(m.group(0)), text, flags=re.DOTALL)
+    text = re.sub(r"(`+)(.+?)\1", lambda m: keep(m.group(0)), text, flags=re.DOTALL)
```
Zhichenzzz: `re.DOTALL` on the inline-code pattern makes a single unpaired backtick swallow everything up to the next backtick, across paragraphs. Everything in between lands in the stash, so it is exempt from link rewriting, brace escaping and tag escaping, and a broken relative link inside it is never reported.

Reproduction against this file:

```python
text = "Use the `--flag option here.\n\nSee [the guide](./guide.md) for details.\n\nThen run `python x.py` to finish.\n"
masked, stash = mask_code(text)
# masked == 'Use the \x000\x00python x.py` to finish.\n'   <- the link is gone
```

None of the 15 mirrored READMEs has an odd backtick count today, so this is latent, but the failure is silent. Dropping `flags=re.DOTALL` (inline code does not span paragraphs) or bounding the match to a single line would close it.

Zhichenzzz (commented):
Follow-up from the review above, anchored to the lines. Two new latent bugs in the generator (unpaired backtick, title scan inside code fences), plus the inline-math fix and three smaller ones.

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+
+    return HTML_TAG.sub(replace, text)
+
+
+def convert(readme_text, rel_dir, mirrored, broken):
+    readme_text = EXCLUDE_BLOCK.sub("", readme_text)
+    lines = readme_text.split("\n")
+    title = None
+    for i, line in enumerate(lines):
+        if line.startswith("# "):
```
Zhichenzzz, line 330: The title scan runs before `mask_code`, so it sees inside fenced blocks: a shell comment picked up here becomes the page title *and* is deleted from the code block.

```python
readme = "```bash\n# Download the model\nhf download foo\n```\n\n# Real Title\n\nBody.\n"
title, body = convert(readme, "demo", {}, [])
# title == 'Download the model'
```

No current README opens with a fence, so this is latent too. Skipping fenced regions in this loop (or running it after `mask_code`) fixes it.

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def rewrite(match):
+        bang, text, href, hint = match.groups()
+        resolved = resolve_link(href, rel_dir, mirrored, broken)
+        return f"{bang}[{text}]({resolved or href}{hint or ''})"
+
+    body = MD_LINK.sub(rewrite, body)
+    body = convert_details_blocks(body)
+    body = escape_html(body)
+    body = body.replace("{", "\\{").replace("}", "\\}")
```
Zhichenzzz, line 354: This is the line that corrupts inline math, as described in the summary comment: `mask_code` protects `$$…$$` but not `$…$`, so `$\pi_{\textcolor{blue}{\text{old}}}$` ships as `$\pi_\{\textcolor\{blue\}\{\text\{old\}\}\}$` (18 occurrences in `docs/examples/infra-features/train-infer-mismatch-helper.md`).

Suggested fix in `mask_code`, next to the `$$` rule:

```python
text = re.sub(r"(?<!\$)\$(?!\$)(?:\\.|[^$\n])+\$", lambda m: keep(m.group(0)), text)
```

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+GITHUB_TREE = f"https://github.com/radixark/miles/tree/{BRANCH}"
+GITHUB_BLOB = f"https://github.com/radixark/miles/blob/{BRANCH}"
+GITHUB_RAW = f"https://raw.githubusercontent.com/radixark/miles/{BRANCH}"
+
+IMAGE_SUFFIXES = {".png", ".jpg", ".jpeg", ".gif", ".svg", ".webp"}
+MAX_DESCRIPTION = 160
+
+# Tags Mintlify renders as JSX. Anything else in prose (<tool_call>, <search>, ...) is a
+# model token, not markup, and is escaped so acorn never sees it.
+HTML_TAGS = {
```
Zhichenzzz, line 46: The whitelist is missing tags that are ordinary in a README: `h1`-`h6`, `center`, `figure`, `figcaption`, `picture`, `video`, `source`, `s`, `u`, `mark`, `dl`/`dt`/`dd`, `caption`. Anything not listed is escaped, so a future `<h3 align="center">Results</h3>` renders as literal `&lt;h3 align="center">` text with no error from the sync or the build.

Today's READMEs only use `img`, `p`, `details`, `summary`, so nothing is broken yet; it is the silence that makes it worth widening now.

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+            if body.startswith(description):
+                body = body[len(description) :].lstrip()
+        if not description:
+            raise SyncError(f"{repo_dir_of(rel_dir)}/README.md has no description; add a bullet in examples/README.md")
+        if len(description) > MAX_DESCRIPTION:
+            raise SyncError(
+                f"description for {repo_dir_of(rel_dir)} is {len(description)} characters, "
+                f"over the {MAX_DESCRIPTION} the docs site allows; shorten it at the source"
+            )
+        rendered[OUT_DIR / f"{slug_for(rel_dir)}.md"] = render_page(title, description, rel_dir, body)
```
Zhichenzzz: `slug_for` lowercases and maps `_` to `-`, and `rendered` is keyed by the output path, so two directories that differ only in case or separator (`foo_bar` vs `foo-bar`) collapse into one page with the last one silently winning, and `--check` still passes. No collisions among the current 15 pages.

A guard here would make it a hard error instead:

```python
out = OUT_DIR / f"{slug_for(rel_dir)}.md"
if out in rendered:
    raise SyncError(f"{repo_dir_of(rel_dir)} and an earlier directory both map to {out.name}")
```

scripts/tools/sync_example_docs.py
```diff
@@ -0,0 +1,427 @@
...
+    "tr",
+    "ul",
+    # Mintlify's own component, which <details> blocks are rewritten into.
+    "accordion",
+}
+VOID_TAGS = {"br", "hr", "img", "input"}
+
+# - **[fully_async](./fully_async)**: Demonstrates fully asynchronous rollout generation.
+INDEX_BULLET = re.compile(r"^\s*[-*]\s+\*\*\[([^\]]+)\]\(([^)]+)\)\*\*:\s*(.+?)\s*$")
+MD_LINK = re.compile(r"(!?)\[([^\]]*)\]\(\s*([^)\s]+)(\s+\"[^\"]*\")?\s*\)")
```
Zhichenzzz, line 106: `[^\]]*` for the link text means a nested image link is mis-parsed: `[badge](target)` matches as text `![badge`, href `img.png`, so the inner image URL is rewritten while the outer href is left relative, and it is never reported as broken. No occurrences in the current READMEs.

.pre-commit-config.yaml
```diff
@@ -51,6 +51,13 @@ repos:
 
   - repo: local
     hooks:
+      - id: sync-example-docs
+        name: Mirror examples/ READMEs into the docs site
+        language: system
+        entry: python3 scripts/tools/sync_example_docs.py
+        files: '^(examples/.*README\.md|docs/examples/.*|docs/docs\.json)$'
```
Zhichenzzz: The pattern does not include the generator itself, so editing `scripts/tools/sync_example_docs.py` alone does not re-run it locally and the pages silently go stale until something under `examples/` or `docs/examples/` is touched. CI's `pre-commit run --all-files` still catches it, so this only costs a round trip:

```yaml
files: '^(examples/.*README\.md|docs/examples/.*|docs/docs\.json|scripts/tools/sync_example_docs\.py)$'
```

Zhichenzzz:
maybe move this scripts/tools/sync_example_docs.py‎ into tests/fast/doc (if what i understand is true, you wanted to implement the pre-commit testing for doc <-> examples)and the current implementation is a bit fragile; we should make it more stable and robust.

nblintao:
@Zhichenzzz Thanks! Everything is fixed, other than:

1. Deleted pages: intended — we want to throw away the old examples, and not include `experimental/` for now.
2. docs.json round-trip: intended — one formatting authority prevents drift; documented in the docstring.

nblintao:
> maybe move this scripts/tools/sync_example_docs.py‎ into tests/fast/doc (if what i understand is true, you wanted to implement the pre-commit testing for doc <-> examples)and the current implementation is a bit fragile; we should make it more stable and robust.

I've added some tests for the script in test/. The script itself should better not be in tests/ since it changes code.

Zhichenzzz (approved):

PR #2485, opened 2026-08-12
clean up fully async example and mv to examples/infra_features

yueming-yuan:
The fully-async example is an infra feature and belongs under `examples/infra_features/`, next to `random_async`, `true_on_policy` and the rest — `random_async/README.md` already calls itself its "minimal sibling".

## The shell launcher goes with the move

`run-qwen3-4b-fully_async.sh` is deleted rather than ported. It was the only shell launcher of the three in that directory, and it demonstrated nothing the two python ones do not — the three changes the README calls out (`train_async.py`, `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=1`, `--fully-async`) are all present in `run_qwen3_5_4b_fully_async_eval.py`, which is the same 4B size class and is now the documented quick start. The only thing unique to it was a commented-out `--max-weight-staleness`, which the README explains anyway.

## What the move touches beyond the paths

`external_eval_fn` is loaded by **dotted module path**, not by file path, so the move is not purely mechanical. Three call sites follow it:

- `tests/fast/rollout/test_checkpoint_eval.py` — `importlib.import_module(...)`, and this one runs in CI
- `tests/e2e/megatron/test_qwen3_4b_fully_async_eval.py` — `--eval-function-path ...`
- the launcher itself, `run_qwen3_5_4b_fully_async_eval.py`

Plus the docstring reference in `miles/rollout/checkpoint_eval.py`, the `random_async` README, and the two doc pages. Incidentally the example README's file list never mentioned `run_qwen3_30b_a3b_fully_async.py`; it does now.

## Testing

`pytest tests/fast/rollout/test_checkpoint_eval.py tests/manual/launch_scripts tests/fast/launch_scripts tests/fast/test_ray_launcher_unbuffering.py` — 586 passed. The first of those is what proves the dotted path still resolves; the last is the guard that counts ray launchers, which drops by one with the deleted script. `pre-commit run --all-files` clean, and no doc page links at the old location.


nblintao:
probably need to rebase and fix conflicts. i added auto doc webpage generation. now docs/examples fully mirrors /examples

nblintao (approved):
LGTM. Need to rebase - we no longer need manual changes to docs/examples/fully-async.md

Zhichenzzz (approved):

Shi-Dong (approved):

PR #2486, opened 2026-08-12
docs: point AMD installs at the maintained rocm/sgl-dev images

Arist12:
## Problem

The AMD installation example still uses legacy `rlsys/miles` image names instead of the maintained `rocm/sgl-dev` ROCm images.

## Change

Update the pull and run commands, list ROCm 7.0/7.2 tags for MI300X and MI350X/MI355X, and document daily rebuilds plus dated tags.


Arist12:
@Zhichenzzz @yushengsu-thu  should be a quick doc fix; could you help review when you got time? Thanks!

PR #2497, opened 2026-08-13
Temporarily remove vanilla fault tolerance for the eval fleet

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/eval_fleet.py
+++ b/miles/ray/rollout/eval_fleet.py
@@ -9,6 +9,7 @@ import asyncio
 import logging
 from argparse import Namespace
 
+from miles.backends.sglang_utils.sglang_api_client import SGLangApiClient
 from miles.rollout.checkpoint_eval import EvalSkip, retarget_args
 from miles.rollout.inference_rollout.inference_rollout_common import GenerateState
 from miles.utils.http_utils import wait_http_ok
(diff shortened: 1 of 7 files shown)


miles/ray/rollout/eval_fleet.py
```diff
@@ -24,24 +24,13 @@ class EvalFleet:
...
-            if not self._rollout_ft_enabled:
-                # Otherwise RolloutHealthMonitor owns the probing for these engines.
-                await self._srv.probe_and_mark_dead()
-            await self._srv.recover()
-            await self._srv.wait_all_engines_alive()
-        except Exception as e:
-            logger.warning(f"Eval fleet unhealthy: {e}")
-            raise EvalSkip("unhealthy") from e
-
         if not await self._pin_fleet(checkpoint_dir, weight_version):
```
fzyzcjy, line 44: [P1] Switch eval pinning to RolloutServer API clients

With `--eval-num-gpus > 0`, removing the health block makes `pin()` reach `_pin_fleet()`, but its real `_srv` is a `RolloutServer`, which has no `engines` attribute—only `api_clients`. The resulting `AttributeError` occurs before the retry `try` and is not converted to `EvalSkip`, so every dedicated checkpoint eval fails; deliver-1 retains this code. Deliver-2 fixes this by snapshotting `srv.api_clients` under `srv.context_lock` and invoking their async HTTP methods; the minimal forward-port is that access/call-shape change and helper, which depends only on the existing `api_clients` and `context_lock`, not the later controller/executor split.

fzyzcjy (commented):
The patch activates a pinning path that dereferences an interface already removed from `RolloutServer`, so dedicated eval fleets cannot evaluate checkpoints. The defect remains at deliver-1 and therefore blocks that safe delivery.

PR #2498, opened 2026-08-13
fix: require shared rewards within rollouts

guapisolo:
## Summary
Require rollout siblings to share one reward before rollout-level normalization.

## Symptom & Reproduction
- **Symptom:** Siblings with different rewards were silently reduced to the reward of the leaf with the most trainable tokens, so masking or leaf length could change the rollout advantage.
- **Reproduction:** `pytest -q tests/fast/ray/rollout/test_train_data_conversion.py -k test_grpo_rejects_different_sibling_rewards` expected a contract error for rewards `2.0` and `6.0`; the previous implementation instead selected one sibling as `mainstream`.

## Root Cause
1. `_normalize_rewards_by_rollout` grouped siblings without requiring a shared selected reward.
2. `_trainable_token_count` ranked siblings using the final training mask and response length.
3. `mainstream_segments` silently used the ranked leaf as the rollout normalization input.

## Fix
Validate that sibling rewards are equal after `reward_key` selection, normalize one scalar per rollout, and broadcast the normalized value to every sibling. A mismatch now fails with rollout, row, and reward details instead of applying a heuristic. Custom reward post-processing remains unchanged.

## Verification
- `tests/fast/ray/rollout/test_train_data_conversion.py`: 53 passed, including mismatch rejection, mask independence, and selected `reward_key` coverage.
- Related rollout/session suites: 175 passed across train-data conversion and session v2 reward-broadcast paths.
- Commit hooks: Ruff, autoflake, isort, formatting, and repository policy checks passed.

## Review Focus
- `_normalize_rewards_by_rollout`: equality is checked after selecting the configured scalar reward.
- `remove_sample` and token length no longer influence which rollout reward is normalized.
- Existing session v2 trajectory-reward broadcasting remains the upstream source of shared rewards.


Shi-Dong (approved):
Thanks!

yueming-yuan (approved):

PR #2500, opened 2026-08-13
Synchronize disk-delta baseline reloads across trainer ranks

fzyzcjy:


guapisolo (approved):
Approved for the #2176 safe-merge checkpoint: #2154 makes rank-0 baseline reload failures trigger bounded whole-cell teardown (strengthened by #2173–#2175), and the fix remains present through the checked chain tip, so this no longer blocks the stack.

PR #2501, opened 2026-08-13
Start dynamic worker ports above Ray's worker range

fzyzcjy:


miles/utils/workers/addr_allocator.py
```diff
@@ -5,6 +5,8 @@
 
 logger = logging.getLogger(__name__)
 
+_DYNAMIC_PORT_START = 20000
```
guapisolo, line 8: [P2] Avoid assuming the default Ray worker-port ceiling

`MILES_SCRIPT_EXTERNAL_RAY=1` accepts existing clusters whose worker range can overlap `20000+`; the [Ray Slurm example](https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html) uses `20002-29999`. Both Ray and this allocator probe before the eventual bind, so concurrent startup can select the same port and fail a rollout cell. Please make the Miles range configurable and validate it is disjoint from the external Ray `min/max/list`, or otherwise reserve non-overlapping ranges. #2473 only bounds/wraps the scan, so this remains unresolved at #2176 and checked tip #2649.

PR #2502, opened 2026-08-13
Publish the new weight version to the rollout executor

fzyzcjy:


fzyzcjy (commented):
The weight version is correctly propagated from both backend actors through the trainer group and published by all three in-repository training drivers. No blocking correctness or regression issue remains at deliver-1.

PR #2503, opened 2026-08-13
Let every miles process name itself as the component it is

fzyzcjy:


fzyzcjy (commented):
The new multi-LoRA process identity does not take effect for Kubernetes-served workers because argument parsing has already fixed the logger identity as `main`. This remains unresolved at the target deliver.

miles/ray/multi_lora/controller.py
```diff
@@ -62,6 +64,8 @@ def _load_subclass(path: str | None, base_cls):
 
 class MultiLoRAController:
     def __init__(self, *, args, router_providers: Sequence[BaseWorkerProvider], host: str = "0.0.0.0") -> None:
+        configure_logger(args, source=SimpleProcessIdentity(component="multi_lora_controller"))
```
fzyzcjy, line 67: [P2] Set the worker identity before parsing Kubernetes specs

When `MultiLoRAController` is launched through the Kubernetes `ServeWorker` path, `serve_inner` first calls `compute_specs_from_argv()`, whose `parse_args()` call sets `_LOGGER_CONFIGURED` via `configure_logger_raw("main")`; the call here therefore cannot change the logging format. Controller and backend records remain tagged `main` rather than `multi_lora_controller`, and deliver-2 retains this path. Avoid claiming the main identity during spec parsing or set the worker identity before it.

PR #2504, opened 2026-08-13
Split env_report.py into a package

fzyzcjy:


fzyzcjy (commented):
Within the requested production correctness and regression scope, the package split preserves existing behavior and all production imports are updated. No unresolved blocker remains at the target deliver.

PR #2505, opened 2026-08-13
Record the environment as a redacted audit event

fzyzcjy:


miles/utils/env_report/redaction.py
```diff
@@ -0,0 +1,93 @@
...
+)
+_SGLANG_SECRET_ARG_BASE_NAMES = frozenset({"api_key", "admin_api_key", "ssl_keyfile_password"})
+_SECRET_ARG_NAMES = frozenset(
+    {
+        "wandb_key",
+        "router_api_key",
+        "router_oracle_password",
+        "router_control_plane_api_keys",
+        *(f"{prefix}{name}" for prefix in ("sglang_", "eval_sglang_") for name in _SGLANG_SECRET_ARG_BASE_NAMES),
+    }
```
fzyzcjy, line 18: [P1] Redact credentials in URL-valued audit arguments

When a run that stores events passes an authenticated `--http-proxy` such as `http://user:password@host`, `http_proxy` is absent from this set and `_ARGV_VALUE_REDACTORS` is derived from the same set. `process.args` therefore exposes the credentials in every process report, while the main process's `process.argv` exposes them again. Redact URL user-info before persistence; the target deliver retains this allowlist.

miles/utils/logging_utils.py
```diff
@@ -5,22 +5,28 @@
...
     name = source.to_name()
     configure_logger_raw(name)
 
     if (event_dir := getattr(args, "save_debug_event_data", None)) is not None:
         if not is_event_logger_initialized():
             set_event_logger(EventLogger(log_dir=event_dir, file_name=f"{name}.jsonl", source=source))
 
+    global _ENV_REPORTER
+    if report_env and _ENV_REPORTER is None:
+        _ENV_REPORTER = start_env_reporting(args)
```
fzyzcjy, line 28: [P2] Restore event history before startup reports are written

When resuming from a checkpoint containing `debug_events`, the main and worker-manager processes execute this synchronous report before the rollout-executor actor calls `event_logger_checkpoint.restore()`. That restore renames the live event directory and replaces it with the snapshot, leaving both new-incarnation startup reports under `.trash_*`; in this op they are one-shot, so the active resumed log never records those processes' environments. Restore before launching any event writer; the target still preserves this ordering.

miles/ray/train_actor.py
```diff
@@ -62,6 +58,10 @@ def __init__(
         # os.environ["LOCAL_RANK"] = str(ray.get_gpu_ids()[0])
         os.environ["LOCAL_RANK"] = str(get_local_gpu_id())
 
+        configure_logger(
+            args, source=TrainProcessIdentity(component=role, cell_index=cell_index, rank_within_cell=rank)
+        )
```
fzyzcjy, line 63: [P2] Bind reports after train actor arguments are finalized

When `--num-epoch` leaves `num_rollout` unresolved, train actors report here before `create_rollout_components()` fills it and before `configure_master_addr_and_port()` installs `MASTER_ADDR`/`MASTER_PORT`. `init()` later installs a separately serialized finalized Namespace, but the reporter remains bound to the constructor copy, so this op records incomplete state and target-deliver periodic reports keep serializing stale arguments. Start or rebind reporting after `init()` receives the final args; this is the changing-input case covered by [`.claude/rules/general-code-style.md:22-28`](.claude/rules/general-code-style.md#L22-L28).

miles/utils/arguments.py
```diff
@@ -3225,6 +3233,9 @@ def miles_validate_args(args):
         args.save_debug_trajectory_data = f"{args.dump_details}/trajectory/{{rollout_id}}.jsonl"
         args.save_debug_event_data = f"{args.dump_details}/events"
 
+    if args.save_debug_event_data is None and args.save is not None:
+        args.save_debug_event_data = f"{args.save}/events"
```
fzyzcjy, line 3299: [P2] Keep reference-checkpoint events out of fresh runs

When a fresh fine-tune has no resumable `--load`, validation rewrites `args.load` to `args.ref_load`; this new default means a prior Miles reference checkpoint can contain `iter_N/debug_events`, so `RolloutExecutor` restores that unrelated history into the new run. Reused rollout IDs then combine old and new checksum/witness events, and FT's analyzer can reject the new run as a false mismatch. Gate event restore on a genuinely resumable requested load rather than the resolved weight source; the target tracks `requested_load` but still restores from `args.load`.

fzyzcjy (commented):
The patch can persist live credentials and can lose, stale, or cross-contaminate audit events during initialization and checkpoint loading. These defects remain in the stated target deliver, so they are blockers rather than transient chain issues.

PR #2506, opened 2026-08-13
Re-record the environment periodically, not only at startup

fzyzcjy:


miles/utils/env_report/reporter.py
```diff
@@ -20,11 +31,33 @@ def start_env_reporting(args: Any) -> "EnvReporter | None":
...
+        self._thread.start()
+
+    def stop(self) -> None:
+        self._stopped.set()
+        self._thread.join(timeout=_STOP_TIMEOUT_SECONDS)
+
+    def _run(self) -> None:
+        delay = min(SETTLED_DELAY_SECONDS, self._interval_seconds) if self._interval_seconds > 0 else 0.0
+        while True:
+            self._report_once()
```
fzyzcjy, line 63: [P2] Refresh the active trainer namespace for periodic reports

Trainer workers later receive a separately deserialized namespace in `TrainRayActor.init()`, replace `self.args`, and rewrite `rank`/`world_size` (and sometimes `distributed_backend`), while `EnvReporter` retains the constructor namespace. This loop therefore re-snapshots the stale object: nonzero ranks remain recorded as rank 0 and independent-DP workers can retain the pre-cell world size. `f5a18edf` still has the same ownership split; update the reporter when the active namespace changes, consistent with [`.claude/rules/general-code-style.md` lines 26–28](.claude/rules/general-code-style.md#L26-L28).

miles/utils/env_report/reporter.py
```diff
@@ -20,11 +31,33 @@ def start_env_reporting(args: Any) -> "EnvReporter | None":
...
+            f"--env-report-interval-seconds is {interval_seconds}, which is neither a delay nor a way to say "
+            f"'only at startup'; pass a finite number"
+        )
+
         self._args = args
+        self._interval_seconds = interval_seconds
+        self._jitter = random.Random()
+        self._stopped = threading.Event()
+        self._thread = threading.Thread(target=self._run, name=_REPORTER_THREAD_NAME, daemon=True)
+        _log_env_report(report=collect_unprobed_env_report(snapshot=collect_env_report_snapshot(args)))
```
fzyzcjy: [P2] Keep periodic reporting alive after startup write failures

For normal processes that configure logging once, if the synchronous startup snapshot or event-log write raises—for example while the shared event directory is temporarily unavailable—the exception escapes `__init__`, `start_env_reporting()` returns `None`, and the thread is never started. Consequently no later report retries the transient failure, despite `_report_once()` explicitly tolerating such failures; this remains unchanged in `f5a18edf`. Catch the initial reporting failure without aborting thread startup.

miles/utils/env_report/collector.py
```diff
@@ -66,6 +66,18 @@ def collect_env_report(*, snapshot: EnvReportSnapshot) -> EnvReport:
         editable_packages=editable_packages,
         git_repos=git_repos,
         full_pip_list=full_pip_list,
+        packages_probed=True,
```
fzyzcjy: [P2] Mark failed package probes as unprobed

When `pip inspect` exits nonzero, times out, emits invalid JSON, or cannot start, `_collect_pip_info()` returns `([], [])`, but the report is still marked `packages_probed=True`. The persisted event is therefore indistinguishable from a successful probe of an empty environment—the ambiguity this field was introduced to remove. `f5a18edf` still hard-codes this value; propagate probe success or mark failures as unprobed.

fzyzcjy (commented):
The periodic reporter can retain stale trainer arguments, permanently stop after a transient startup failure, and misclassify failed package probes. All three defects remain present in the target safe deliver.

PR #2507, opened 2026-08-13
Hash the uncommitted changes a run actually ran with

fzyzcjy:


miles/utils/env_report/git_state.py
```diff
@@ -1,44 +1,140 @@
...
-            timeout=10,
-            cwd=location,
-        )
-        diff_stat = diff_result.stdout.strip()
-        dirty = bool(diff_stat)
+
+        diff_stat_result = _run_git(args=["diff", "--stat", "HEAD"], location=location)
+        diff_stat = _decode(diff_stat_result.stdout).strip()
+
+        patch_result = _run_git(args=["diff", "HEAD"], location=location)
```
fzyzcjy: [P2] Capture the stat and patch from one snapshot

When a tracked file is written or restored between the separate `--stat` and patch commands, `dirty`/`diff_stat` and `uncommitted_hash` describe different trees. The final guard only compares HEAD, so it accepts this contradiction; a clean-to-dirty transition even makes `_summarise_repo` omit the captured dirty hash. Generate the stat and patch from one diff result or also validate worktree stability.

fzyzcjy (commented):
The patch can omit untracked files for subdirectory editable installs and can accept contradictory state during concurrent source updates. Both issues remain in the specified deliver commit.

miles/utils/env_report/git_state.py
```diff
@@ -1,44 +1,140 @@
...
+    for entry in untracked.hash_entries:
+        digest.update(b"untracked\0" + entry + b"\0")
+    if untracked.paths_truncated:
+        digest.update(b"untracked-truncated\0%d\0" % untracked.total_count)
+
+    return digest.hexdigest()
+
+
+def _collect_untracked(*, location: str) -> _UntrackedFiles:
+    result = _run_git(args=["ls-files", "--others", "--exclude-standard", "-z"], location=location)
```
fzyzcjy: [P2] Enumerate untracked files from the repository root

For editable packages whose location is a repository subdirectory, this command only returns untracked paths beneath that subdirectory. SGLang is installed with `pip install -e "python[all]"` (`docker/Dockerfile:203-215`), so its location is `/sgl-workspace/sglang/python`, while `rev-parse` and `git diff HEAD` describe the entire repository. An untracked runtime file in a sibling directory is therefore absent from `dirty` and the hash; resolve and use `git rev-parse --show-toplevel` first.

PR #2508, opened 2026-08-13
Record what the kubernetes launcher launched

fzyzcjy:


miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -140,6 +159,12 @@ def _compute_train_argv(
     return pod_argv, args
 
 
+def _compute_pod_record_file(*, installed_manifest: Manifest | None, record_path: Path) -> str | None:
+    if installed_manifest is None:
+        return str(record_path)
+    return installed_launch_record_file(manifest=installed_manifest, container=naming.ORCHESTRATOR_COMPONENT)
```
fzyzcjy, line 182: [P2] Attach the current record when pods are rebuilt

When an installed release is relaunched with `--skip-upgrade-check` after changing a global pod field such as `infra.image.tag`, every pod is rebuilt anyway, but this branch still reuses the old `MILES_SCRIPT_ENV_REPORT` path (or no path for a pre-feature release). The replacement processes therefore load stale or missing launch metadata while the newly written record is orphaned; use `record_path` for templates that are already being rebuilt. This remains in deliver-2 `f5a18ed`, so it is a blocker.

miles/utils/external_utils/command_utils/helm_backend/launcher/entrypoint.py
```diff
@@ -127,6 +142,10 @@ def _compute_train_argv(
     request: ExecuteTrainRequest, *, run_id: str, release: str, namespace: str
 ) -> tuple[list[str], Any]:
     argv = [*shlex.split(shell_safe_model_args(request.megatron_model_type)), *shlex.split(request.train_args)]
+    assert not ArgvManipulator.declares(argv, _ENV_REPORT_FLAG), (
+        f"{_ENV_REPORT_FLAG} is what this launcher tells the pods about the launch that installed them, and an "
+        f"argument of that name outranks it, so the pods would report a launch that never happened; drop it"
+    )
```
fzyzcjy, line 153: [P2] Reject the report variable inside trainer env overrides

With `--train-env-vars '{"MILES_SCRIPT_ENV_REPORT":"/other.json"}'`, this assertion passes because it only checks for a top-level `--env-report`. `compute_trainer_env_vars` then includes the mapping and `miles/utils/workers/serving/serve.py::main` overlays it in `os.execve`, causing trainer processes to read the supplied path instead of the chart-injected record. Reject this reserved key in dynamic worker environments as well; deliver-2 `f5a18ed` retains the same data flow, so this is a blocker.

fzyzcjy (commented):
The patch can attach stale launch metadata to replacement pods and leaves a supported trainer-environment channel able to override the supposedly reserved report path. Both defects remain at the target safe deliver.

PR #2509, opened 2026-08-13
Record the environment every sglang engine serves with

fzyzcjy:


fzyzcjy (commented):
No correctness or regression blocker was found in the reviewed op, and the relevant behavior remains coherent at the target deliver.

PR #2510, opened 2026-08-13
Hand the weight-update group the gpus each engine really has

fzyzcjy:


fzyzcjy (commented):
The change correctly forwards the per-engine GPU counts to the existing group builder, keeping NCCL world size and rank offsets aligned with heterogeneous engine topologies. The production change remains unchanged at the target deliver and introduces no identified regression.

PR #2511, opened 2026-08-13
Give a worker provider a lifecycle hook and let it size its own fleet

fzyzcjy:


fzyzcjy (commented):
The lifecycle hook runs before engine-provider reads, and the fleet-size override correctly preserves both fallback and explicit-zero semantics. No correctness or regression issue remains at the target safe deliver.

PR #2512, opened 2026-08-13
Support cells without launch gates

fzyzcjy:


fzyzcjy (commented):
No in-scope correctness or regression blocker remains at deliver-2. Later endpoint-publication hardening removes the transient risk of mistaking an incomplete address map for a gateless cell.

PR #2513, opened 2026-08-13
Describe an external engine fleet on the command line

fzyzcjy:


miles/utils/arguments.py
```diff
@@ -81,6 +81,9 @@ def _resolve_rollout_functions(args) -> None:
         "--eval-num-gpus and a CheckpointEvalFn --eval-function-path each select an eval "
         "backend; the fleet would boot and then hand the work to the other one."
     )
+    assert not (
+        args.eval_num_gpus > 0 and _compute_rollout_external(args)
+    ), "eval_num_gpus cannot be set with external rollout engines."
```
fzyzcjy, line 86: [P2] Recheck external eval after loading custom config

When `--custom-config-path` supplies `rollout_external_engine_addrs` while `--eval-num-gpus > 0` is on the CLI, this check runs before the YAML is applied in `miles_validate_args` at lines 3586-3594, so `_compute_rollout_external()` still returns false. Validation later marks the run external, while `specs_inference_engine()` drops every local engine spec; at the safe deliver the dedicated eval fleet consequently has no cells and evaluations cannot pin engines. Move or repeat this gate after applying the custom config; the same ordering remains in `f5a18edf2a`.

fzyzcjy (commented):
The external-engine and dedicated-eval mutual exclusion remains bypassable through the supported YAML override path at the target safe deliver. This can leave a configured dedicated eval fleet without any engines.

PR #2514, opened 2026-08-13
Discover a static fleet of external sglang engines

fzyzcjy:


fzyzcjy (commented):
Authenticated external engines cannot be discovered, configured credentials may be exposed in logs, and several accepted external topologies cannot be represented by the current router or weight-update planners. These defects persist at the specified target deliver.

miles/ray/rollout/external_engine_provider.py
```diff
@@ -0,0 +1,259 @@
...
+            int(x) if (x := server_info.get("disaggregation_bootstrap_port")) is not None else None
+        ),
+    )
+
+
+async def _fetch_server_info_with_retry(
+    url: str, *, timeout_seconds: float = DISCOVERY_TIMEOUT_SECONDS
+) -> dict[str, Any]:
+    async def _attempt(_remaining_seconds: float) -> dict[str, Any]:
+        return await SGLangApiClient(server_url=url).get_server_info()
```
fzyzcjy: [P1] Authenticate external-engine discovery

When an external SGLang server is launched with `--api-key` and Miles receives the same value via `--sglang-api-key`, `/server_info` requires a bearer token. This client is constructed without `api_key`, so every attempt returns 401, is retried as an `HTTPError` for 600 seconds, and provider initialization fails; pass the configured key through discovery.

miles/ray/rollout/external_engine_provider.py
```diff
@@ -0,0 +1,259 @@
...
+        _assert_engines_match_args(self._args, engines=engines)
+
+        self._cells = _compute_cells(args=self._args, engines=engines)
+        self._cell_id_by_worker_name = {
+            worker_name: cell_id
+            for cell_id, cell in self._cells.items()
+            for worker_name in cell.cell_info.worker_names
+        }
+
+        logger.info(f"Discovered external inference engines: {self.cell_infos}")
```
fzyzcjy: [P1] Redact API keys before logging discovered cells

When `args.sglang_api_key` is non-empty and discovery succeeds, every `CellInfo.meta` contains that raw value, so formatting `self.cell_infos` writes the credential verbatim to controller logs, which may be persisted or centralized. Log only non-secret topology fields or redact the metadata.

miles/ray/rollout/external_engine_provider.py
```diff
@@ -0,0 +1,259 @@
...
+        return mode
+    return "regular"
+
+
+# ============================== args cross-check ==============================
+
+
+def _assert_engines_match_args(args: Any, *, engines: list[_ExternalEngineInfo]) -> None:
+    discovered_total = sum(engine.num_gpus for engine in engines)
+    assert discovered_total == args.rollout_num_gpus, (
```
fzyzcjy, line 260: [P1] Validate discovered sizes against the selected updater

When engines have different GPU counts and training uses FSDP or Megatron's `p2p` transfer mode, this aggregate-only check accepts the topology even though those planners still derive a uniform layout from `args.rollout_num_gpus_per_engine` and ignore `engine_gpu_counts`. For counts `[4, 2]` with the per-engine argument set to 2, FSDP assigns overlapping NCCL ranks while P2P plans three engines instead of two, causing a hang or invalid endpoint access; reject these combinations or make both planners consume the discovered layout.

miles/ray/rollout/external_engine_provider.py
```diff
@@ -0,0 +1,259 @@
...
+        f"the external engines report {discovered_total} gpus in total "
+        f"({[(engine.url, engine.num_gpus) for engine in engines]}), but --rollout-num-gpus is "
+        f"{args.rollout_num_gpus}. That argument sizes the placement group and the router, so let it "
+        f"describe the fleet that is actually running"
+    )
+
+    pd_engine_urls = [engine.url for engine in engines if engine.worker_type != "regular"]
+    assert bool(pd_engine_urls) == args.rollout_external_router_pd, (
```
fzyzcjy, line 274: [P2] Reject incomplete or mixed PD fleets

When `--rollout-external-router-pd` is set, a list containing only prefill engines, only decode engines, or regular engines plus any PD engine passes because this checks only whether at least one PD engine exists. The PD router requires a prefill/decode pair and ignores regular workers, so requests either cannot select a pair or silently strand declared cells; require both roles and no regular engines.

PR #2516, opened 2026-08-13
fix(lora): persist distributed optimizer parameter state

Arist12:
Part of #2705.

## Problem

LoRA checkpoints only save `optimizer.state_dict()`. Megatron `DistributedOptimizer` keeps its sharded FP32 master weights and Adam moments outside that state dict, so resume restores neither. The stale masters can overwrite the loaded adapter on the first optimizer step, and the incomplete optimizer restore can produce NaNs.

Expert-only LoRA adds a second failure: Megatron keeps an empty non-expert `DistributedOptimizer` stub inside `ChainedOptimizer`. Calling the chained `state_dict()` dereferences that stub's `optimizer=None`, so the checkpoint fails before parameter-state serialization.

## Fix

Save each non-stub distributed optimizer child with Megatron’s `save_parameter_state` API and restore it with `load_parameter_state_from_dp_zero`. Refresh master weights after loading adapter weights, coordinate rank-local I/O failures before collectives, and validate the complete parameter-state payload before scatter.

For chained metadata state, preserve Megatron’s existing format when no stub is present. When stubs exist, save only active children keyed by their original chain indices; restore requires the exact same active-child layout and synchronizes child steps afterward.

Checkpoints created before this change fall back to a weight-only warm start. A partially present or malformed parameter-state checkpoint is rejected.

## Tests

- 12 focused parameter-state tests, including malformed payloads, stub-child save/restore, active-child layout mismatch, and legacy-format preservation
- Full `tests/fast/backends/megatron_utils`: 301 passed, 2 skipped (before the stub-child follow-up)
- The 4-GPU LoRA E2E is registered for H200 and MI350. It checks the real distributed-optimizer files and reloads one for a resumed step.
- 4× MI355X: restored 192 adapter tensors per rank, optimizer and scheduler at iteration 2, then completed two steps with finite grad norms (0.4721, 0.4671).
- 8× MI355X GPT-OSS-20B expert-only LoRA: reproduced the non-expert stub crash at checkpoint save; the active-child fix then saved optimizer/scheduler state successfully at iteration 13.
- Changed-file pre-commit hooks pass.

## Merge with `main` (multi-LoRA v2, `--no-save-optim`, `--no-load-optim`)

`[multi-lora v2 0/7]` moved `lora_utils.py` to `lora/utils.py`, and #3083 / #3245 added the
`--no-save-optim` / `--no-load-optim` gates to the same three places this PR rewrites. None of
that covers what this PR fixes: `save_parameter_state` and `load_parameter_state_from_dp_zero`
still appear nowhere under `miles/` on `main`, and the per-slot path added by
`[multi-lora v2 3/7]` is a separate `lora/checkpoint.py` + `SlotOptimizer` flow that does not
touch `save_lora_checkpoint` / `load_lora_adapter`. `reload_model_params()` on `main` covers
the HF-load path and the inkling adapter path (#2122), not the native adapter path.

How the two gates compose:

- `--no-save-optim` now also skips the parameter-state shards, not just the optimizer entry.
  The flag is uniform across ranks, so the `_save_optimizer_param_state` collective is either
  entered by every rank or by none.
- `--no-load-optim` skips both the training-state restore and the parameter-state restore, and
  the restore itself moved into `_restore_optimizer_param_state` so the gate stays one line and
  mirrors `_save_optimizer_param_state`.
- Reading `training_state["optimizer"]` is now `None`-guarded, because a `--no-save-optim`
  checkpoint writes `None` there.

One test-side consequence: `load_lora_adapter` calls `optimizer.reload_model_params()`, so the
fake optimizer in #3245's `_recorder()` needed that method. `tests/fast/backends/megatron_utils/`
is 325 passed, 2 skipped on the merge.


jhinpan (commented):
I reviewed the current head (`55dc933d6c9070940f854a88f9111f018d98853c`) and exercised the implementation with a real two-rank Megatron distributed Adam optimizer on AMD Instinct MI355X GPUs.

The happy path restored model-side adapter tensors, fp32 master weights, first/second moments, optimizer metadata, and iteration exactly. The following failure paths still need work.

**Blocker — parameter-state structure is not validated before DP scatter.**

`_load_optimizer_param_state` coordinates only `torch.load`. With a valid pickle containing an invalid state structure, DP rank 0 raised `KeyError` while checking `numel_unpadded`, while rank 1 had already entered `scatter` and failed after the peer disconnected. This was reproduced with the real Megatron `DistributedOptimizer`, not a mock.

DP roots need a non-mutating structural/numel preflight, followed by global consensus, before any rank calls `load_parameter_state_from_dp_zero`.

**The final rebase must preserve #2733 before any mutation.**

#2516 conflicts exactly at the native-load site and its side contains the permissive `if name in state_dict: copy_` loop. The combined flow must be:

1. every rank reads its selected native shard;
2. strict key/shape/dtype/finiteness validation runs locally;
3. ranks reach global error consensus;
4. only then copy model parameters and reload optimizer masters;
5. preflight and restore optimizer/scheduler state.

Taking #2516's conflict side would silently revert #2733. Its new positional `args` parameter also requires updating #2733/#2669 tests and callers.

**Distinguish legacy warm start from a torn current checkpoint.**

`_optimizer_param_state_is_available` maps both “no new-format files exist” and “only a subset exists” to the same weight-only warm start. The former is valid legacy compatibility; the latter is a damaged current checkpoint and should fail rather than silently discard optimizer, scheduler, and step state.

**Enforce topology compatibility.**

Adapter files become topology-keyed after #2669, while training/optimizer files remain keyed by global rank. Persist world/TP/PP/EP/DP topology metadata and refuse exact restore when it differs; otherwise the same world size with a different layout can pair correct adapter shards with another coordinate's optimizer state.

**Coordinate each remaining mutation boundary.**

Scheduler loading is currently unguarded after optimizer mutation. Likewise, optimizer child save failures should be checked after each child so all ranks stop in lockstep, rather than waiting until every child has run. The parameter-state payload also round-trips with `weights_only=True, map_location="cpu"`; use that safer load mode where possible.

jhinpan (commented):
Follow-up review on the unchanged head (`55dc933d6c9070940f854a88f9111f018d98853c`). These findings are new relative to the first review.

**Blocker — rank-local checkpoint generations and iterations are never compared.**

Each rank independently loads `training_state_rank{rank}.pt`, and the only consensus covers whether `torch.load` raised. There is no equality check for checkpoint generation, iteration, or scheduler position. I reproduced this with two valid files: rank 0 contained iteration 10 and rank 1 iteration 11; `_load_training_state` returned `[10, 11]` without an error.

This can arise from an interrupted overwrite because native, training-state, and optimizer-state files are published independently with no generation manifest. Persist one generation ID plus the expected shard/file set, write a completion marker only after all required state succeeds, and require all ranks to agree on generation and iteration before any model/optimizer mutation.

**`--no-load-optim` can leave the LR/WD scheduler at step zero.**

The no-load branch returns the saved nonzero iteration without restoring `opt_param_scheduler`. Later, `model.py` skips `opt_param_scheduler.step(iteration * global_batch_size)` whenever `use_checkpoint_opt_param_scheduler` is true and iteration is nonzero, assuming the checkpoint restored scheduler state. In this path that assumption is false, so training resumes at the old iteration with a fresh scheduler.

Return whether scheduler state was actually restored, and reposition it from iteration whenever it was not.

**Security — no-load paths still perform unrestricted pickle deserialization.**

`training_state_rank*.pt` is loaded with `weights_only=False` before checking `optimizer is None` or `args.no_load_optim`. A user-supplied adapter directory can therefore execute pickle payloads even when optimizer loading was explicitly disabled. The real distributed-Adam training-state payload produced by this PR successfully round-tripped with `weights_only=True, map_location="cpu"` in local validation.

Use safe deserialization for the supported state schema, or split iteration/scheduler metadata from any payload that genuinely requires trusted pickle loading.

**Validate supported process-group and optimizer-child layouts up front.**

The legacy distributed-optimizer parameter-state API requires Megatron DP Gloo subgroups. Miles passes `args.enable_gloo_process_groups` into optimizer construction, so `--disable-gloo-process-groups` currently fails only at the first LoRA save interval. Reject this incompatible combination during argument validation rather than failing after training has begun.

The active-child compatibility guard is also one-directional: checkpoints written with `optimizer_children` are checked, but a checkpoint written while all children were active uses the `optimizer` key and bypasses the guard when the current run has a stub child. Record active child indices unconditionally and compare them in both directions before calling `load_state_dict`.

Arist12:
Thanks — this review led to a substantial rewrite. The PR is now narrowed to one problem: `DistributedOptimizer` keeps master weights and Adam moments out of `state_dict()`, so a LoRA checkpoint that saves only `state_dict()` silently resumes from base-model masters.

**Adopted — the DP-root preflight.** Confirmed against Megatron: `load_parameter_state_from_dp_zero` asserts `numel_unpadded` *between* two scatters (`distrib_optimizer.py:2369-2375`), so a mismatch raising on the DP root hangs the rest of the group. The shape check now runs before any collective, and its failure is agreed on across ranks.

**Adopted — `weights_only=True` for the parameter state.** Verified it round-trips the `(dtype, dtype)`-keyed payload; the `weights_only=False` load is now confined to the training state, which genuinely holds non-tensor objects, with a comment saying so.

**Adopted — distinguish "legacy checkpoint" from "torn checkpoint".** All shards present → restore. None present → warn and `reload_model_params()` (warm-start from the adapter). Anything in between → raise, rather than silently warm-starting from a partially written checkpoint.

**Adopted — collectives must not sit below a per-rank early return.** Re-reviewing, `_load_training_state` returned early on a per-rank `state_path.exists()` above the newly-added collectives, and the `torch.load` / `load_state_dict` were uncoordinated. Both now go through the consensus helper.

**Not adopted — guarding the scheduler load.** Nothing collective runs after it, so a rank that skips it cannot strand anyone. Adding a consensus there is cost with no failure mode behind it.

**Not adopted — per-child consensus around the parameter-state save.** Stub-ness is rank-dependent, so a global all-reduce *per child* would itself deadlock whenever the ranks disagree about how many non-stub children they have. The save loop records the first error and reaches one consensus at the end instead.

**Not adopted — `data_parallel_group_gloo`.** Megatron's own `load_parameter_state` uses `data_parallel_group` (`distrib_optimizer.py:2624`); matching it is the conservative choice.

**Deferred (listed in the PR body).** `--no-save-optim` / `--no-load-optim` / `--finetune` handling, topology metadata in the checkpoint, and cross-rank agreement on iteration/generation. All are separate behaviours; folding them in was what made the earlier revision hard to review. Also stripped: the `checkpoint.py` change and the `args` plumbing — `isinstance(child, DistributedOptimizer)` answers the same question locally, so the load path no longer needs `args` at all.

Note on stacking: the failure-consensus helpers here duplicate `lora_checkpoint.py` from #2669. Whichever lands second should delete its copy.

Validation: 4 focused tests exercising round-trip through the DP root, the legacy warm-start, the torn-shard rejection, and the pre-scatter shape rejection; `tests/fast/backends/megatron_utils` — 263 passed, plus the two `TestUpdateWeightsZeroChunks` failures that reproduce on `main`. pre-commit clean.

PR #2517, opened 2026-08-13
fix(example): remove invalid Kimi LoRA flag

Arist12:
## Problem

The Kimi K2.5 LoRA launcher and documentation include `--filter-zero-reward-samples`, an unsupported argument.

## Change

Remove the argument from the launcher and documentation, and update the generated launch-script snapshot to match.


Arist12:
@yushengsu-thu should be quick fix to remove non-existent flag, could you help review? Thanks!

PR #2518, opened 2026-08-13
fix(kimi): stop forcing --attention-backend flash on an MLA model

Arist12:
## Problem

`--attention-backend flash` does not only enable FlashAttention. Megatron's `LanguageModule._set_attention_backend` turns it into

```python
check_and_set_env_variable("NVTE_FLASH_ATTN",   1, AttnBackend.flash)
check_and_set_env_variable("NVTE_FUSED_ATTN",   0, AttnBackend.flash)
check_and_set_env_variable("NVTE_UNFUSED_ATTN", 0, AttnBackend.flash)
```

so it is the one setting that leaves TransformerEngine a single candidate. Kimi-K2.5 is MLA with asymmetric head dims (qk_nope 128 + qk_rope 64 = 192 against v 128), which FlashAttention 2 refuses. Where FlashAttention 3 is also unavailable, nothing is left and the first training step exits with

```
ValueError: No dot product attention backend is available for the provided inputs.
```

naming neither the restriction nor the flag that caused it. Reproduced by driving TE's `DotProductAttention` at K2.5's head dims on ROCm 7.2 (MI350X, TE 2.14, FA 2.8.3):

```
flash: Disabling FusedAttention due to NVTE_FUSED_ATTN=0
       Disabling UnfusedDotProductAttention due to NVTE_UNFUSED_ATTN=0
       Disabling FlashAttention 2 as it does not support MLA.
       Selected backend = NoBackend                                -> ValueError
auto:  Disabling FlashAttention 2 as it does not support MLA.
       Selected backend = UnfusedDotProductAttention               -> OK
```

## Fix

Use `auto`, Megatron's default, in the two Kimi-K2.5 launchers, and update the sentence in `docs/models/kimi/kimi-k2.5.md` that documents the old value. `scripts/run_deepseek.py` already keeps the flag commented out on its MLA model with "need to comment this when using model with MLA"; this applies the same rule to K2.5.

Scope is the two recipes I ran. `scripts/run_kimi_k2.py`, `scripts/run_deepseek_v32.py` and the GLM launchers pass the same flag on models with the same head dims and are left alone.

### DIFF
--- a/docs/models/kimi/kimi-k2.5.md
+++ b/docs/models/kimi/kimi-k2.5.md
@@ -167,7 +167,7 @@ SGLANG_ARGS=(
 )
 ```
 
-The `--use-rollout-routing-replay` flag replays the rollout-time MoE routing decisions during training so the two stages stay consistent. On the Megatron side, attention uses the Flash backend (`--attention-backend flash`).
+The `--use-rollout-routing-replay` flag replays the rollout-time MoE routing decisions during training so the two stages stay consistent. On the Megatron side, attention is left to TransformerEngine (`--attention-backend auto`): forcing `flash` also disables the fused and unfused backends, and FlashAttention 2 rejects K2.5's asymmetric MLA head dims.
 
 The launcher sets the required env vars for you, including the INT4 QAT pair (`OPEN_TRAINING_INT4_FAKE_QAT_FLAG=1`, `OPEN_TRAINING_INT4_GROUP_SIZE=32`), a long NCCL timeout (`NCCL_TIMEOUT=3600`), `CUDA_DEVICE_MAX_CONNECTIONS=1`, and NVLink-gated NVLS (`NCCL_NVLS_ENABLE` follows the script's NVLink autodetection).
 
--- a/examples/lora/run-kimi-k25-megatron-lora.sh
+++ b/examples/lora/run-kimi-k25-megatron-lora.sh
@@ -145,7 +145,9 @@ MISC_ARGS=(
    # should be good for model performance
    --accumulate-allreduce-grads-in-fp32
    --attention-softmax-in-fp32
-   --attention-backend flash
+   # not flash: that also disables the fused and unfused backends, and FA2 rejects this
+   # model's asymmetric MLA head dims, so it leaves TE nothing to select without FA3
+   --attention-backend auto
    --no-check-for-nan-in-loss-and-grad
 )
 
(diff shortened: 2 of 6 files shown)


Arist12:
Closing this: the change does not fix the recipes it edits, so the premise is wrong.

`auto` only helps because TE falls through to `UnfusedDotProductAttention`, and TE drops that backend under context parallelism ([`dot_product_attention/utils.py:763-767`](https://github.com/NVIDIA/TransformerEngine), "Disabling UnfusedDotProductAttention as it does not support context parallelism"). Both recipes here run with CP > 1, so nothing changes for them. Driving `get_attention_backend` at K2.5's MLA dims (qk 192 / v 128) on MI350X, ROCm 7.2, TE 2.14.0.dev0, FA 2.8.3:

| `--attention-backend` | CP off | CP on |
|---|---|---|
| `flash` | NoBackend | NoBackend |
| `auto` | UnfusedDotProductAttention | **NoBackend** |

`examples/lora/run-kimi-k25-megatron-lora.sh` is `--context-parallel-size 8` and `scripts/run_kimi_k25.py` is `--context-parallel-size 4` on its multi-node path, so the only configuration this patch fixes is the single-node 2-layer debug path (CP 1), where the fallback is the O(s²) reference kernel rather than anything usable at scale.

Two more reasons not to land it as written:

- On H200 `auto` is not neutral. TE prefers FusedAttention over FlashAttention on Hopper+ for performance, so the recipe would move from FlashAttention 3 to cuDNN fused attention — and `docs/developer/debug.md` points at `--attention-backend flash` for deterministic attention under context parallelism, which both recipes use.
- The scope note was wrong: the GLM launchers do not share K2.5's head dims. `scripts/models/glm5.2-744B-A40B.py` is `--qk-head-dim 192 --qk-pos-emb-head-dim 64 --v-head-dim 256`, i.e. 256/256 symmetric, which FA2 accepts. `scripts/run_kimi_k2.py` and `scripts/run_deepseek_v32.py` are the launchers with the same asymmetric dims, and `run_kimi_k2.py` carries the "need to comment this when using model with MLA" note and keeps `flash` anyway.

The real gap is that ROCm has no MLA-capable fused attention for asymmetric head dims and no FA3, so no value of `--attention-backend` makes K2.5 train under CP there. That is a platform item, not something a launcher flag fixes, and it wants its own issue rather than this patch.

PR #2520, opened 2026-08-13
fix(lora): restore checker-scrambled base weights once

Arist12:
Part of #2705.

## Problem

`--check-weight-update-equal` randomizes every rollout-engine weight exactly once at startup (`miles/ray/placement_group.py`, `check_weights(action="reset_tensors")`). LoRA runs never push the frozen base afterwards, so on the disaggregated path the check compares its adapter against a base that is still random. On the colocated path the checker instead forced a full base sync on *every* update, which repairs the scramble but re-ships the base for the whole run.

## Change

Restore the frozen base on the first checked update only, on both transports, then return to adapter-only transfer. Retire `_lora_base_synced`, a write-only flag that `weight_version` already expresses.

## Validation

- 31 fast tests in `tests/fast/backends/megatron_utils/test_lora_weight_sync_validation.py`, covering two consecutive updates across the `(is_lora, check_equal)` matrix on both transports
- 4x MI350X disaggregated checker E2E, including a rollout after the first optimizer update

## Reviewer notes

- Land after #2683. This PR opens the base path for LoRA runs, which are always `--megatron-to-hf-mode bridge`; without #2683 that path still converts through the `--spec`-name converters, which fail on bridge-only architectures. The two merge cleanly and the tests pass stacked.
- The gating predicate is rank-uniform: `weight_version` is incremented before any rank branching and `check_weight_update_equal` is parsed once, so no rank can take a different side of a collective.


Arist12:
Deferring until the weight-sync refactor settles. The underlying issue is unchanged — with LoRA on the broadcast path the frozen base is never re-pushed, so `--check-weight-update-equal`'s startup scramble is never undone — but #2752/#2753/#2756 replaced the seam this patch attaches to, so the fix belongs on the new `WeightTransferProtocol` (`needs_base_resync_for_lora`) rather than the old mixin. Will reopen against the new structure.

PR #2522, opened 2026-08-13
Make the class-based rollout the default and convert legacy path to env var gated

yueming-yuan:
`MILES_EXPERIMENTAL_ROLLOUT_REFACTOR` has guarded the class-based rollout path since January (#484/#491), and every feature added since — fully-async, snapshot eval, agentic tool calls, nemo-gym, openenv, verifiers — requires it. The experiment is the product. This PR flips the polarity: **`InferenceRolloutFn` is the default**, and `MILES_USE_LEGACY_ROLLOUT_V1=1` opts back into the v1 `sglang_rollout` path.

## What the old flag actually gated

Six branches: the default rollout-fn path (`resolve_rollout_function_paths`), the loader and the train/eval call sites in `RolloutManager`, the `--fully-async` and snapshot-eval assertions, and the registration of user-provided-function CLI args (`fn.add_arguments`). All six now key on `use_legacy_rollout_v1()`, inverted. The v1 module and its importers (multi-LoRA, benchmarkers, delegated eval) are untouched — removing v1 itself is a later PR.

## Why legacy customize functions keep working

`load_rollout_function` wraps plain functions in `LegacyRolloutFnAdapter`, which reproduces the `(args, rollout_id, data_source, evaluation=...)` calling convention and boxes bare return values; `load_generate_function` sniffs legacy generate signatures into `LegacyGenerateFnAdapter`. The v1 path has loaded custom generate fns through this same adapter all along, so the flip does not change how a legacy `--rollout-function-path` / `--custom-generate-function-path` is invoked. In-repo legacy fns riding the adapter after the flip: `sft_rollout`, `random_async`, multi-LoRA's injected rollout fn.

## Known v1-only features now fail fast

Two features only work on the v1 generate, verified by a full semantic diff of both paths. They now raise at argument validation with a message pointing at the env var, instead of deep in the run:

- **student-side top-k OPD** — `opd_student_top_logprobs` is only produced by the v1 generate
- **`--mask-offpolicy-in-partial-rollout`** — the class-based path does not re-extend the loss mask after resume

Their consumers are pinned to the legacy var: `test_qwen2.5_0.5B_opd_sglang`, the two top-k OPD shell examples, and `test_qwen3_0.6B_verifiers` — previously the repo's only explicit `"0"` — becomes the explicit legacy pin, keeping v1 coverage in CI until the path is removed. Remaining v1-only gaps that have **no in-repo consumers** (audio_data payloads, per-eval-dataset `custom_generate_function_path`, kimi-style `call_processor` multimodal prompts) are not guarded; they surface when someone hits them.

## Everything else deletes its opt-in

14 launchers/examples, 30 e2e tests, the session-verify runner, and the autouse conftest fixture drop the now-redundant `=1`. Six LoRA snapshots lose the var from their recorded runtime env; the two OPD shell snapshots gain the pin. The harness `CLEARED_ENV` gains `MILES_USE_LEGACY_ROLLOUT_V1` so an ambient developer value cannot skew recordings. Ten doc sites are rewritten — notably `generate-endpoint.md` and `customization.md`, which described the v1 path as the shipped default.

## Behavioral notes for the ~30 e2e tests that silently switch paths

They are all plain GRPO/SFT recipes on the default rollout fn (FT scenarios, r3-FSDP, MTP, model-script drivers, offload/p2p/dumper). The class-based path is a near-mirror of v1 (same over-sampling loop, dynamic filter, partial-rollout abort, sample filters, prefill logprob recompute), with intentional differences to watch on their first CI runs: per-group task exceptions are logged and resubmitted instead of crashing the rollout, abort fan-out failures are fatal, and eval prompt caches are per-instance instead of process-global.

## Testing

Launch-script snapshot suite regenerated and green (443 passed). Full `tests/fast` running in a container now; will report. `pre-commit run --all-files` clean.


maocheng23 (approved):

nblintao (approved):

miles/utils/arguments.py
```diff
@@ -29,10 +29,10 @@
 
 def resolve_rollout_function_paths(args) -> tuple[str, str]:
     """The (rollout, eval) function paths the arguments select."""
-    if enable_experimental_rollout_refactor():
-        standard_path = "miles.rollout.inference_rollout.inference_rollout_common.InferenceRolloutFn"
-    else:
+    if use_legacy_rollout_v1():
         standard_path = "miles.rollout.sglang_rollout.generate_rollout"
+    else:
+        standard_path = "miles.rollout.inference_rollout.inference_rollout_common.InferenceRolloutFn"
```
guapisolo, line 35: [P2] Preserve trajectory lifecycle events on the new default path

Making `InferenceRolloutFn` the default sends ordinary single-turn jobs through `inference_rollout_common.generate_and_rm`, which never calls the `TrajectoryLifecycle` sink. The legacy path emits `attempt_start`, `gen_start`, and `attempt_end` around the semaphore/generate call, and the dashboard consumes those events for its per-sample queue/generation timeline. Consequently, `--use-miles-dashboard` jobs on the new default path have an empty trajectory stream even though generation succeeds. Please carry the same outer lifecycle instrumentation into the class-based path, while avoiding duplicate per-turn spans from specialized generators.

guapisolo, line 35: [P2] Preserve bounded rollout logging on the new default path

Making `InferenceRolloutFn` the default also makes its first/finish sample logs the common path. `inference_rollout_train.py` interpolates the complete prompt, response, and raw reward, while the legacy path limits the text preview to 512 characters and summarizes nested reward fields. This matters for long-context and OPD jobs, where a reward can contain per-token logprob payloads: every rollout can now send very large messages through Ray log forwarding and storage, obscuring useful diagnostics and adding avoidable I/O. Please reuse the legacy preview/reward-summary formatting in the class-based path before making it the default.

examples/experimental/verifiers/run.py
```diff
@@ -88,9 +88,9 @@ def execute(args: ScriptArgs):
     # function plug-point selects the adapter, which resolves as a bare module
     # because PYTHONPATH carries this directory into the rollout actor.
     rollout_fn = (
-        "verifiers_rollout.VerifiersRolloutFn"
-        if os.environ.get("MILES_EXPERIMENTAL_ROLLOUT_REFACTOR") == "1"
-        else "verifiers_rollout.generate_rollout"
+        "verifiers_rollout.generate_rollout"
+        if os.environ.get("MILES_USE_LEGACY_ROLLOUT_V1") == "1"
```
guapisolo: [P2] --extra-env-vars MILES_USE_LEGACY_ROLLOUT_V1=1 is applied only when execute_train builds the Ray runtime env, while this lookup reads the launcher os.environ. That can select VerifiersRolloutFn while RolloutManager uses the v1 ABI, raising TypeError on the first rollout. I put the root-cause fix in #2528: adapter selection and the propagated flag now use the same effective environment, with CLI overrides taking precedence over ambient values.

miles/rollout/inference_rollout/inference_rollout_common.py
```diff
@@ -73,15 +74,24 @@ async def generate_and_rm(
...
     async with state.generate_fn_semaphore:
         if state.aborted:
             sample.status = Sample.Status.ABORTED
+            if sink is not None:
+                sink.attempt_end(sample)
             return sample
 
         logger.debug(f"{log_prefix} Acquired semaphore, calling generate_function")
+        if sink is not None:
+            sink.gen_start(sample)
```
guapisolo: [P2] Close lifecycle attempts on generation failure and cancellation

After `sink.gen_start(sample)`, `state.generate_function()` can raise, or its task can be cancelled when a sibling in the group fails. Both paths skip the normal-path `attempt_end()` below the semaphore. The synchronous rollout path records the failure and keeps submitting work, so the dashboard can leave an already-finished trajectory in the running state indefinitely. Please emit `attempt_end()` from the exception/cancellation cleanup path while preserving the original exception.

guapisolo (approved):

PR #2526, opened 2026-08-13
fix(swe-agent example): stop logging unmeasured agent metrics as zero

Shi-Dong:
## What this does

`aggregate_agent_metrics` in the SWE-agent example now logs only the agent metrics that an agent actually reported. Previously every key an agent did not report was substituted with `0` and logged as if it had been measured.

## Why

Agent harnesses report different key sets. One that instruments every turn fills in `turns`, `tool_calls`, `model_query_time_*`, `env_execution_time_*`, `time_per_turn` and the ratios; another may report only wall-clock totals such as `agent_run_time`, `eval_time` and `total_time`.

`_collect_values` used `m.get(key, 0)`, so the second kind of agent produced a wandb panel like this:

| metric | logged | truth |
|---|---|---|
| `agent/agent_run_time_mean` | 404.0 | measured |
| `agent/eval_time_mean` | 50.8 | measured |
| `agent/total_time_mean` | 490.4 | measured |
| `agent/tool_calls_mean` | **0.0** | never measured |
| `agent/turns_mean` | **0.0** | never measured |
| `agent/model_query_time_avg` | **0.0** | never measured |
| `agent/env_execution_time_avg` | **0.0** | never measured |
| `agent/time_per_turn` | **0.0** | never measured |

A zero sitting next to real numbers reads as a measurement. On a run whose trajectories each contained 23–47 `bash_command` calls, the dashboard said the agent made none, and the flat-zero series looked like a broken agent rather than missing instrumentation.

There is a second, quieter failure. Because the default applied per sample, a mixed batch silently corrupted values that *were* measured: one sample omitting `tool_calls` halved the mean for the sample that reported 47, giving `agent/tool_calls_mean = 23.5`. That is wrong for any batch where a single trial fails to report.

After the change, an unreported key is absent from the log (wandb simply shows no series, which is honest), a genuine `0` measurement is still logged as `0`, and means are taken over the agents that measured the key.

## Verification

Executed `aggregate_agent_metrics` directly against both agent shapes before and after the change:

- fully instrumented agent: all 17 metrics identical before and after (no regression)
- totals-only agent: 12 false zeros before, 0 after; the 5 measured metrics unchanged
- mixed batch: `agent/tool_calls_mean` 23.5 before, 47.0 after

Added `tests/fast/examples/swe_agent_harbor_docker/test_agent_metrics.py` (5 tests, loaded by path like the other example tests since the directory name is not an identifier). Confirmed the tests fail on the pre-fix code — 2 failed, 3 passed — and all 5 pass with the fix.

Not covered here: the dashboard's `tool_calls` column comes from a different code path (`_tool_call_count` in `miles/dashboard/dump_reader.py`, which counts `role == "tool"` messages in `sample.prompt`) and is reported separately.

### DIFF
--- a/examples/swe-agent-harbor-docker/generate.py
+++ b/examples/swe-agent-harbor-docker/generate.py
@@ -44,15 +44,10 @@ async def reward_func(args, samples: Sample | list[Sample], **kwargs) -> float |
 
 
 def _collect_values(all_metrics: list[dict], key: str) -> list[float]:
-    """Values the agents actually reported for ``key``, skipping the ones that did not.
+    """Values agents reported for ``key``; agents that omit it are skipped.
 
-    Agent harnesses report different key sets: one that measures per-turn timing
-    and tool-call counts fills every key, while another may only report the
-    wall-clock totals. Substituting 0 for a key an agent never reports turns
-    "nobody measured this" into "this agent made zero tool calls", which is
-    indistinguishable on a dashboard from a real measurement. It also drags the
-    mean of a mixed batch toward zero. Callers already skip empty lists, so
-    unreported keys are left out of the log entirely.
+    Defaulting to 0 would log a fake measurement and drag the batch mean down.
+    Callers skip empty lists, so an unreported key stays out of the log.
     """
     return [value for metrics in all_metrics if (value := metrics.get(key)) is not None]
 
--- a/examples/swe-agent-harbor-docker/generate.py
+++ b/examples/swe-agent-harbor-docker/generate.py
@@ -44,7 +44,17 @@ async def reward_func(args, samples: Sample | list[Sample], **kwargs) -> float |
 
 
 def _collect_values(all_metrics: list[dict], key: str) -> list[float]:
-    return [m.get(key, 0) for m in all_metrics]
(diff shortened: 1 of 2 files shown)


examples/swe-agent-harbor-docker/generate.py
```diff
@@ -44,7 +44,17 @@ async def reward_func(args, samples: Sample | list[Sample], **kwargs) -> float |
...
+    """Values the agents actually reported for ``key``, skipping the ones that did not.
+
+    Agent harnesses report different key sets: one that measures per-turn timing
+    and tool-call counts fills every key, while another may only report the
+    wall-clock totals. Substituting 0 for a key an agent never reports turns
+    "nobody measured this" into "this agent made zero tool calls", which is
+    indistinguishable on a dashboard from a real measurement. It also drags the
+    mean of a mixed batch toward zero. Callers already skip empty lists, so
+    unreported keys are left out of the log entirely.
+    """
```
Shi-Dong: Shorten the explanation here. This is too long. Make it clear and concise.

nblintao (approved):

PR #2527, opened 2026-08-13
ci: skip GPU stages on docs-only pull requests

nblintao:
## Summary

Docs-only pull requests currently occupy the whole self-hosted GPU fleet (observed on #2489, a pure docs PR that queued every NVIDIA stage). This adds a `docs_only` verdict to `resolve-ci-policy` and skips only the GPU stages when it holds.

`ci_policy.py` owns the decision (workflows only feed it facts, per the existing design); the diff step reuses the merge-commit-vs-`HEAD^1` convention from `docker-paths`.

## Conservative by construction — every rule fails toward the full fleet

`docs_only=true` requires **all** of:

- the trigger is a `pull_request` (scheduled nightly/weekly and manual dispatch never skip),
- the PR carries **no** CI label (`run-ci-*`, `nightly`, `bypass-fastfail` are explicit requests for tests and win),
- the changed-file list is **non-empty** (an empty or failed diff means "unknown", never "docs-only"),
- **every** changed path sits under `docs/` — an allowlist of prefixes, never a denylist. Paths git quotes for non-ASCII names fail closed; docs-adjacent *code* (`scripts/tools/sync_example_docs.py`, `examples/**/README.md`) is deliberately not allowlisted.

Blast radius is limited to the GPU stages (5 NVIDIA + MI300X). The CPU stages ignore `docs_only` entirely, so the docs checks themselves (docs.json navigation, examples-mirror drift gate) still run on every docs PR. In the workflow expressions the gate is `docs_only != 'true'`, so a missing/empty output also runs the fleet.

## Verification

- 11-case unit matrix for `is_docs_only_change` + 4-case `resolve_workflow_inputs` integration + 4-case CLI end-to-end (including a missing change-list file) — all in `tests/ci/test/test_ci_policy.py`, exercised standalone against the real module and CLI.
- Existing exact-`GITHUB_OUTPUT` contract tests updated for the new `docs_only=` line and re-verified byte-for-byte.
- Both workflow YAMLs parse; `WorkflowPolicy` has no constructors outside `ci_policy.py`; the new `resolve_workflow_inputs` parameter defaults to `()` so existing callers are unchanged.
- `docs/ci/00-stage.md` updated in the same change per the workflows' `doc-dev` contract.

This PR itself touches workflows/tests/docs (not docs-only), so its own run exercises the `docs_only=false` path end-to-end; the resolver log prints the verdict (`Resolved CI policy: ... docs_only=...`) for auditability.


nblintao:
@guapisolo will support this in a refactor after v0.1

PR #2129, opened 2026-08-03
test(session): validate H200 v1/v2 agentic parity

guapisolo:
> **Depends on:** #2128
> Stacked on the parent PR's branch. Review / merge the parent first.

## Summary

Add deterministic H200 proof that v2 drop-retries matches v1 training payloads bitwise.

## Motivation

V2 with `drop_retries` must produce exactly the same collected samples as v1 on the linear agentic path, including a real weather-tool interaction rather than a non-agentic shortcut.

## Usage

Run `pytest -q -s tests/e2e/sglang/test_session_v1_v2_parity.py` on an H200 host with the Qwen3-8B checkpoint available.

## Design Notes

- **Key choices:** The test starts SGLang with FA3.
- Generation runs in deterministic mode.
- One fixed seed drives 16 weather-tool samples through both versions.
- The assertion compares canonical training payload fields recursively and bitwise.

## Verification

- **Tests added:** `test_qwen3_8b_h200_fa3_agentic_v2_drop_retries_matches_v1_training_payload_bitwise` passed on NVIDIA H200 in 62.19 seconds.

## Review Focus

- Scrutinize `_build_input_samples` for its fixed test-input contract.
- Scrutinize the parity projection for complete training-payload coverage.
- Scrutinize the weather-tool path for identical v1/v2 interactions.

PR #2536, opened 2026-08-14
Carry the rollout id on both agentic paths

fzyzcjy:
## What breaks

- `tests/e2e/sglang/test_session_v1_v2_parity.py::test_qwen3_8b_h200_fa3_agentic_v2_drop_retries_matches_v1_training_payload_bitwise` fails on `main`:

```
AssertionError: sample.rollout_id is not bitwise equal
left = None, right = 0, path = 'sample.rollout_id'
```

- It is not flaky and not environment-dependent. The scheduled run on `main` reproduces it: [31715141995](https://github.com/radixark/miles/actions/runs/31715141995) → [`stage-c-2-gpu-h200 (0)`](https://github.com/radixark/miles/actions/runs/31715141995/job/94499981993), `Test Summary: 4/5 passed`.

## Root cause

- `miles/rollout/generate_hub/agentic_tool_call.py` stamps the rollout id **only** on the v2 branch:

```python
    samples = result.samples
    if use_v2:
        rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
        assert rollout_id is not None, "v2 agentic samples require input Sample.rollout_id or Sample.index"
        for sample in samples:
            sample.rollout_id = rollout_id
```

- So v1 leaves `sample.rollout_id` at `None` while v2 sets it to the index. The parity test compares every field of the training payload, so the two paths are structurally unable to agree on this one.
- The two halves landed independently and were never reconciled:

| Change | Date | Effect |
| --- | --- | --- |
| #2129 `test(session): validate H200 v1/v2 agentic parity` | 2026-08-05 | Adds the bitwise parity test over all payload fields |
| #2368 `fix(rollout): group session v2 leaf samples` | 2026-08-11 | Adds the v2-only `sample.rollout_id` stamping |

- The rollout id is sample identity, not a v2 concept — nothing about it is specific to the session-server path.

## The fix

- Hoist the derivation out of the branch, keep the assertion where it was, and stamp whenever a value is derivable:

```python
    samples = result.samples
    rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
    if use_v2:
        assert rollout_id is not None, "v2 agentic samples require input Sample.rollout_id or Sample.index"
    if rollout_id is not None:
        for sample in samples:
            sample.rollout_id = rollout_id
```

- **v2 is byte-identical in behaviour**: the same assertion still guards only that path, and `rollout_id` is non-`None` there by construction, so the new `if` is always taken.
- **v1 gains the stamp only when derivable**: if neither `rollout_id` nor `index` is set, v1 keeps its current behaviour rather than acquiring a new failure mode.

## Evidence

- The same change, applied on a chain based on `6033ac0eaf`, turns that lane green: [`stage-c-2-gpu-h200 (0)`](https://github.com/radixark/miles/actions/runs/31685938763) went from `Enabled 5 test(s)` → `4/5 passed` to `Enabled 5 test(s)` → **`5/5 passed`**, running 51 minutes, so it is a real pass rather than a skipped or no-op lane.
- `ruff check miles/rollout/generate_hub/agentic_tool_call.py` passes.

## Note for reviewers

- #2347 edits the same guard to `if use_v2 and len(samples) > 1:` for an unrelated reason. Whichever lands second will need a trivial rebase; the two intents are compatible, since the condition being narrowed there is the stamping loop rather than the derivation.

### DIFF
--- a/miles/rollout/generate_hub/agentic_tool_call.py
+++ b/miles/rollout/generate_hub/agentic_tool_call.py
@@ -114,11 +114,10 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
         return GenerateFnOutput(samples=[sample] if use_v2 else sample)
 
     samples = result.samples
-    # FIXME: handle sample index issues.
-    rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
-    if use_v2:
+    if use_v2 and len(samples) > 1:
+        # FIXME: handle sample index issues.
+        rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
         assert rollout_id is not None, "v2 agentic samples require input Sample.rollout_id or Sample.index"
-    if rollout_id is not None:
         for sample in samples:
             sample.rollout_id = rollout_id
     if not use_v2:
--- a/miles/rollout/generate_hub/agentic_tool_call.py
+++ b/miles/rollout/generate_hub/agentic_tool_call.py
@@ -114,10 +114,11 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
         return GenerateFnOutput(samples=[sample] if use_v2 else sample)
 
     samples = result.samples
+    # FIXME: handle sample index issues.
+    rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
     if use_v2:
-        # FIXME: handle sample index issues.
-        rollout_id = input.sample.rollout_id if input.sample.rollout_id is not None else input.sample.index
         assert rollout_id is not None, "v2 agentic samples require input Sample.rollout_id or Sample.index"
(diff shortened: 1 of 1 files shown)


guapisolo:
Rebased fix direction after #2347: this branch now keeps `rollout_id=None` for v1 and v2 singleton results, and assigns a shared rollout ID only to v2 multi-leaf results via `if use_v2 and len(samples) > 1`.

This preserves ordinary postprocessing, including sample-count trimming and dynamic-global-batch metadata, while retaining v2 leaf grouping. Focused verification passed: `tests/fast/rollout/generate_hub/test_agentic_v2.py` (`7 passed`) and Ruff on `agentic_tool_call.py`.

The PR description still describes the superseded v1-stamping approach; this comment records the final behavior instead.

guapisolo (approved):
Do changes to align with session server v2 design

PR #2538, opened 2026-08-14
release: miles version release workflow

yueming-yuan:
## Summary

Add a CI-gated Miles release workflow with pinned dependencies and official Docker images.

## Motivation

An official Miles release must preserve the exact Miles, SGLang, and Megatron-LM sources that passed CI. Before this change, CI could not run the full release scope against an arbitrary release ref or carry that frozen triplet through branch cut, hotfix, tag, and image publication, leaving maintainers without one reproducible supported path.

## Usage

Follow `docs/ci/04-release.md` from the base-version bump through image verification. The core dispatches use the frozen release branch and its verified SHA:

```bash
gh workflow run release-branch-cut.yml -f branch_name="release/v${BASE_VERSION}"
gh workflow run release-tag.yml -f version="${EXACT_VERSION}" -f ref="${RELEASE_SHA}"
```

## Design Notes

- `release-lock.json` stores the pinned dependency SHAs used by a release branch.
- `cadence=release` runs the full CI scope on the requested ref without writing the rolling performance baseline.
- `release-tag.yml` requires `release-ci=success` before dispatching official image builds from the verified tag.

## Verification

- Full 26-file PR surface: all applicable `pre-commit` hooks passed on commit `027222b8ffc4a6208873214ec9620809c61fae4c`.
- `pytest -q tests/ci/test/test_run_suite.py`: 83 passed.
- Workflow YAML and `docs/docs.json` parsing passed; `python3 scripts/tools/sync_example_docs.py --check` exited 0.

## Review Focus

- `release-tag.yml`: confirm the frozen branch-tip operator check closes the gap between `release-ci` validation and tag publication.
- `release-docker.yml`: confirm manual retries keep `version` and `ref` paired because the workflow does not cross-check them.


.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,140 @@
...
+            echo "::error::The newest branch-cut run for '$REF' concluded '$CONCLUSION', not success: $RUN_URL"
+            exit 1
+          fi
+          # A run older than the branch tip tested a stale tip (e.g. a
+          # cherry-pick landed after it) — require a fresh full run.
+          if ! python3 -c "
+          import sys, datetime as dt
+          run = dt.datetime.fromisoformat(sys.argv[1].replace('Z', '+00:00'))
+          tip = dt.datetime.fromisoformat(sys.argv[2])
+          sys.exit(0 if run > tip else 1)" "$CREATED" "$TIP_TIME"; then
```
guapisolo: [P1] Bind the gate to the commit this run tested

The first release-branch-cut run is created before its cut-release-branch job commits release-lock.json. Consequently CREATED is earlier than TIP_TIME, so this check rejects the very first successful cut run and contradicts the summary telling maintainers to tag after it. Please record and compare the exact post-cut SHA (or otherwise account for the lockfile commit) rather than using run creation time.

.github/workflows/pr-test.yml
```diff
@@ -163,7 +198,9 @@ jobs:
         shell: bash
         env:
           PR_BODY: ${{ github.event.pull_request.body || '' }}
-          INPUT_CI_IMAGE_TAG: ${{ github.event.inputs.ci_image_tag || '' }}
+          # workflow_call image_tag (a frozen release-CI tag) outranks the
+          # dispatch input, which only exists on manual runs anyway.
+          INPUT_CI_IMAGE_TAG: ${{ inputs.image_tag || github.event.inputs.ci_image_tag || '' }}
```
guapisolo, line 203: [P1] Rebuild release images after Docker hotfixes

A workflow_call inherits the caller workflow_dispatch event, so docker-paths has no pull_request.base.sha and writes changed=false; inputs.image_tag then forces the original frozen image. If a Dockerfile/build patch is cherry-picked onto the release branch and release-branch-cut is re-dispatched, CI tests the old image while release-docker later publishes a newly built, untested image. Please build and test an image from the release ref, or explicitly reject Docker-related cherry-picks.

guapisolo (approved):

guapisolo (approved):
need handle permission issue

PR #2543, opened 2026-08-14
Allow --stream-optimizer-state-to-disk without trainer offload

yueming-yuan:
Found while running `examples/experimental/openenv/glm52_tbench2` against `main` on 16 GB300 nodes. It is the first of three blockers that stop that example; the run could not start at all.

## The failure

```
AssertionError: --stream-optimizer-state-to-disk requires --offload-train-target=disk.
Both answer the same pressure and are only ever deployed as a pair, so that is the
only combination validated; streaming alone runs but is untested.
```

## Why the coupling does not hold

The two flags answer different pressures, and `--stream-optimizer-state-to-disk`'s own help text says so:

> `--offload-train-target=disk` **cannot help there**, because pause/resume happen at phase boundaries and everything is resident again by the time Adam launches.

`--offload-train-target` is where the trainer is parked *while the rollout engines generate*, and it is read only under `--offload-train`, which defaults to true under `--colocate` and false otherwise. A disaggregated run never offloads the trainer — the engines are on their own nodes — so satisfying the assertion means switching `--offload-train` on purely to satisfy it, paying a full trainer round-trip to NVMe at every phase boundary for nothing.

That leaves streaming unusable in exactly the topology that needs it most. In the run above, DP1 leaves the optimizer state unsharded at **270 GB/rank on 276 GB GPUs**; streaming is what makes it fit:

```
nvme_stream.py:266 - NVMe optimizer state store: 120 buckets, 270.0 GB at /scratch/opt_state/...
nvme_stream.py:361 - NVMe streaming step: 120 buckets, read 90.0 GB, wrote 270.0 GB in 60.9s
model.py:635      - ft op=train_step rollout=0 step=0 outcome=NORMAL valid_step=true
```

The check has never been exercised by a run that uses streaming this way: job 1835 (2026-08-03) logged `offload_train_target ... cpu` *with* streaming on, from a checkout predating the assertion.

## Change

Scope the assertion to runs that actually offload the trainer:

```python
assert args.offload_train_target == "disk" or not args.offload_train, ...
```

and hoist the `offload_train_disk_dir` default so streaming-without-offload gets one. The NVMe store joins that path unconditionally (`miles_plugins/optimizers/nvme_stream.py:453`), so leaving it `None` was a `TypeError` inside the step.

## Testing

- `tests/e2e/megatron/test_qwen3_4B_offload_disk_stream.py` is colocate and sets both flags, so it is unaffected.
- Verified on 16x GB300: with this change the reference config parses, loads, and completes training steps with the store engaged (log lines above).

### DIFF
--- a/docs/advanced/disk-offload.md
+++ b/docs/advanced/disk-offload.md
@@ -104,10 +104,11 @@ training GPUs, which is the point when you are sizing engines against a fixed cl
   cannot help with: add `--stream-optimizer-state-to-disk`, and consider
   `--stream-optimizer-state-moment-dtype bf16` to claw back some of the I/O cost.
 
-Streaming requires `--offload-train-target=disk`; miles asserts it. The two answer the same
-pressure and are only deployed as a pair, so that is the only combination covered.
-Streaming on its own does run — it is the one case where offload has nothing to do, because
-nothing else wants the training GPUs during rollout — but nothing validates it.
+Streaming on its own is the disaggregated case: nothing else wants the training GPUs during
+rollout, so there is no actor to park and `--offload-train-target` is never read. Pass
+`--stream-optimizer-state-to-disk` alone there. Under `--offload-train`, though, the two go
+together — a run that cannot hold the optimizer state on the GPU for the step will not hold
+a pinned host copy of the whole actor either — and miles asserts that pairing.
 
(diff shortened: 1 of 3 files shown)


Zhichenzzz (approved):
Great, could you also update the related sections in doc?

PR #2544, opened 2026-08-14
Do not kill the run when one sample's collect_samples loses its connection

yueming-yuan:
Found while running `examples/experimental/openenv/glm52_tbench2` against `main` on 16 GB300 nodes. A 64-GPU run reached rollout 0, completed its training step, and died seven seconds later on one sample's HTTP connection.

## The failure

```
train_async.py:109 -> RayTaskError(ReadError)
  fully_async_rollout.py:136 _worker_loop -> await self._output.put(task.result())
  fully_async_rollout.py:119 _generate_group
  inference_rollout_common.py:168 generate_and_rm_group -> asyncio.gather(*tasks)
  agentic_tool_call.py:92  result = await tracer.collect_samples(...)
  openai_endpoint_utils.py:80 collect_samples -> post_bytes_no_retry
  http_utils.py:266 response = await _http_client.post(url, json=payload)
httpx.ReadError
```

`collect_samples` posts to the session server, and `generate()` caught exactly one thing from it:

```python
except asyncio.TimeoutError:
    collect_timed_out = True
```

Anything else escapes `generate` -> `_generate_group` -> the fully-async worker loop -> `train_async`, and the job dies over a single sample.

This is not a rare path. Under `--pause-generation-mode in_place` the engines pause **with requests in flight**, so connections to the session server break as a matter of course at every weight update. The crash landed inside the first `update_weights` window, immediately after `fn=train phase=end ok=true elapsed_s=387.3`.

## Change

The asymmetry is the giveaway: the sibling call in the same function — the agent loop itself — is already wrapped in `except Exception` precisely so one episode cannot take the run down. And a failed collect already has a correct outcome three lines below: mark the sample `ABORTED` and let `check_no_aborted` drop its group.

So the fix is to widen the `except` to the transport errors, named next to the client that raises them rather than at the call site:

```python
COLLECT_SAMPLES_TRANSIENT_ERRORS = (TimeoutError, httpx.TransportError)
```

`RuntimeError` from a non-2xx reply still propagates — a session server answering 500 is not a transient condition, and should fail loudly.

## Testing

- `tests/fast/rollout/generate_hub/test_multi_turn.py::TestAgentCollectionFailure` pinned this contract already; it is now parametrized over the timeout and the transport case and keeps asserting that `RuntimeError` escapes. 2 passed.
- Verified in production, on the same exception that caused the crash:

```
16:31:11 agentic_tool_call.py:97 - [session=59939399e30147a9b78252182fee5868]
                                   Failed collecting samples: ReadError('')
```

The run marked that sample `ABORTED`, dropped its group, and went on to complete rollouts 10 and 11. Three such errors were absorbed over the following hours; before the change, any one of them ended the run.

### DIFF
--- a/miles/rollout/generate_hub/agentic_tool_call.py
+++ b/miles/rollout/generate_hub/agentic_tool_call.py
@@ -24,13 +24,13 @@ Agent function contract:
 """
 
 import argparse
-import asyncio
 import logging
 import time
 from collections.abc import Callable
 from copy import deepcopy
 from typing import Any
 
+import httpx
 from sglang.srt.entrypoints.openai.protocol import ChatCompletionRequest
 
 from miles.rollout.base_types import GenerateFnInput, GenerateFnOutput
@@ -68,7 +68,7 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
     metadata = {**metadata, "session_server_id": tracer.session_server_id}
 
     agent_metadata = None
-    collect_timed_out = False
+    collect_failed = False
     t_start = time.monotonic()
     try:
         logger.debug(f"{log_prefix} Starting agent function call")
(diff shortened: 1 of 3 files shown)


yushengsu-thu (approved):

guapisolo (approved):

guapisolo:
good fix!

PR #2550, opened 2026-08-14
fix(lora): support attention-output-gated fused QKV in CanonicalLoRA

### LINKED ISSUE #2008: [Bug] CanonicalLoRA omits attention-output gate rows in fused QKV projections
## Summary

`CanonicalLoRA` assumes that every fused `linear_qkv` projection has the standard grouped Q/K/V output layout. For model providers with `TransformerConfig.attention_output_gate=True`, including the Qwen3.5/3.6 family, the fused projection instead contains Q/Gate/K/V rows.

The canonical adapter currently allocates rows only for Q/K/V and interleaves only those three components. Its output therefore has a different width from the wrapped `linear_qkv` output, and the first forward or trainer-to-SGLang LoRA publication fails with a tensor shape mismatch.

The standard merged LoRA path is not affected. The issue is triggered by the supported `--lora-type canonical_lora` path on an attention-output-gated model.

## Affected configuration

- Megatron training through Megatron-Bridge
- `--lora-type canonical_lora`
- Attention projections included in `--target-modules`
- A provider whose finalized transformer config has `attention_output_gate=True`

Known affected providers include Qwen3.5, Qwen3.6 through the Qwen3.5 spec, and the gated `linear_qkv` portion of Qwen3-Next.

## Reproduction

A native Miles run reproduced the failure with:

- 8 x NVIDIA H200
- `Qwen/Qwen3.6-27B` at revision `6a9e13bd6fc8f0983b9b99948120bc37f49c13e9`
- `--lora-type canonical_lora`
- attention LoRA targets
- Miles code based on `dfc66ff38752bfa2c5d325e0037ebc4b537c06de`
- image `radixark/miles@sha256:95b3afa9ee4313f5633e6ed3779c8276353cc8e24a2462e4f54ec0d5978fbae7`

The same missing handling is still present on Miles `main` at `2cb59235176a170480765e762e85a952e6ba72b7`, whose Dockerfile pins Megatron-Bridge `7f0fb3456f8ffe47599b5fd167b454605d85f932`.

For Qwen3.6-27B full-attention layers:

- Q: `24 * 256 = 6144` rows
- Gate: another `6144` rows
- K: `4 * 256 = 1024` rows
- V: `4 * 256 = 1024` rows
- Wrapped fused projection: `6144 + 6144 + 1024 + 1024 = 14336` rows
- Current canonical adapter: `6144 + 1024 + 1024 = 8192` rows

The observed failure was consequently a `14336` versus `8192` tensor dimension mismatch.

## Root cause

In the pinned Megatron-Bridge implementation:

- [`CanonicalLoRA.transform()`](https://github.com/radixark/Megatron-Bridge/blob/7f0fb3456f8ffe47599b5fd167b454605d85f932/src/megatron/bridge/peft/canonical_lora.py#L382-L393) calculates `q_out_features` as `kv_channels * num_attention_heads`, without including output-gate rows.
- [`LoRALinearSplitQKV._interleave_qkv()`](https://github.com/radixark/Megatron-Bridge/blob/7f0fb3456f8ffe47599b5fd167b454605d85f932/src/megatron/bridge/peft/canonical_lora.py#L90-L143) packs only Q/K/V groups.
- Miles providers correctly declare `attention_output_gate=True`, so the base `linear_qkv` module has the wider Q/Gate/K/V contract.

## Expected behavior

CanonicalLoRA should preserve the complete output contract of the wrapped fused projection. For gated attention, it should create Q and gate rows and pack them in Megatron's grouped Q/Gate/K/V order. Non-gated attention behavior should remain unchanged.

## Proposed fix

At the Megatron-Bridge compatibility boundary:

1. When wrapping `linear_qkv` and `config.attention_output_gate` is true, allocate the query-side canonical adapter with `2 * kv_channels * num_attention_heads` output rows so it represents both Q and gate.
2. Reshape that adapter output as `[num_attention_heads, 2, head_size]` and split it into Q and gate tensors.
3. For every query group, interleave `[Q heads, Gate heads, K, V]` to match the wrapped Megatron projection.
4. Delegate to the existing Q/K/V implementation when `attention_output_gate` is false.

Suggested regression coverage:

- gated row allocation and restoration of the original model config
- exact grouped Q/Gate/K/V interleaving
- unchanged ordinary Q/K/V behavior
- one Qwen3.5 or Qwen3.6 CanonicalLoRA forward and SGLang weight-publication test

This can be implemented directly in Megatron-Bridge and followed by a Miles dependency-pin update. A small Miles-side compatibility shim can provide the same behavior until that update lands.

layahaasini:
## Motivation

Fixes #2008.

`CanonicalLoRA` in Megatron-Bridge assumes every fused `linear_qkv` has the grouped Q/K/V output layout. Providers with `TransformerConfig.attention_output_gate=True` (Qwen3.5, Qwen3.6, and the gated `linear_qkv` of Qwen3-Next) pack Q/Gate/K/V rows instead, so the canonical adapter output is narrower than the wrapped projection and the first forward — or the trainer-to-SGLang LoRA publication — fails with a tensor shape mismatch (14336 vs 8192 on Qwen3.6-27B full-attention layers).

## Modifications

Following the fix proposed in the issue, this adds a Miles-side Megatron-Bridge shim (`miles_plugins/megatron_bridge/canonical_lora_gated_qkv.py`, installed from the plugin package `__init__` like the existing bridge shims) that patches `CanonicalLoRA.transform` for attention-output-gated fused QKV:

- The query-side canonical adapter is allocated with `2 * kv_channels * num_attention_heads` output rows so it represents both Q and gate, using the HF `q_proj` row contract for gated attention (per head: `head_size` Q rows, then `head_size` gate rows). This matches the base-weight mapping in `miles_plugins/mbridge/qwen3_5.py`.
- The wrapper's forward packs adapter output in Megatron's grouped order — per query group: all Q heads, all gate heads, K, V — matching `get_query_key_value_tensors` in Megatron-LM.
- Non-gated `linear_qkv` and every other module delegate to the original `transform`, so ordinary Q/K/V behavior is unchanged.

The shim is idempotent and best-effort (a missing `megatron` import degrades to a warning), and can be dropped once the Megatron-Bridge pin includes the fix upstream.

## Checklist

- New CPU regression tests in `tests/fast/test_canonical_lora_gated_qkv.py` (auto-registered in `stage-a-cpu`, verified with `run_suite.py --list-only`): gated row allocation, exact grouped Q/Gate/K/V interleaving against Megatron's split contract, forward output width matching the wrapped projection, and delegation/unchanged behavior for non-gated modules.

### DIFF
--- a/miles_plugins/megatron_bridge/__init__.py
+++ b/miles_plugins/megatron_bridge/__init__.py
@@ -68,6 +68,14 @@ except Exception as _e:  # best-effort; Qwen3-VL may be unavailable in some envs
     logger.warning("miles Qwen3-VL THD packed mRoPE patch not applied: %s", _e)
 
 
+try:
+    from miles_plugins.megatron_bridge.canonical_lora_gated_qkv import install_canonical_lora_gated_qkv_patch
+
+    install_canonical_lora_gated_qkv_patch()
+except Exception as _e:  # best-effort
+    logger.warning("miles bridge shim install_canonical_lora_gated_qkv_patch not applied: %s", _e)
+
+
 # Model-specific bridge subclasses. Each submodule self-installs on import.
 # Keep imports here so merely importing ``miles_plugins.megatron_bridge`` is
 # enough to pick up every miles bridge (mirrors ``miles_plugins.mbridge``).
--- a/miles_plugins/megatron_bridge/canonical_lora_gated_qkv.py
+++ b/miles_plugins/megatron_bridge/canonical_lora_gated_qkv.py
@@ -7,7 +7,9 @@ the wrapped projection and the first forward fails with a shape mismatch. The pa
 gives ``adapter_q`` per-head [Q; gate] rows (the HF ``q_proj`` layout, as in
 ``miles_plugins/mbridge/qwen3_5.py``) and packs adapter output per query group as
 [Q heads, gate heads, K, V] (Megatron's fused layout). Non-gated modules keep the
-original behavior. Drop once the Megatron-Bridge pin includes the fix upstream.
+original behavior. Like upstream's non-gated ``_interleave_qkv``, the interleave uses
+global head counts, so fused-QKV canonical LoRA remains TP=1-only for now. Drop once
+the Megatron-Bridge pin includes the fix upstream.
 """
 
 from __future__ import annotations
@@ -58,6 +60,7 @@ def install_canonical_lora_gated_qkv_patch() -> None:
         ParallelLinearAdapter,
         get_adapter_attributes_from_linear,
         get_effective_lora_dim,
+        is_modelopt_linear,
     )
     from torch import nn
 
@@ -100,7 +103,8 @@ def install_canonical_lora_gated_qkv_patch() -> None:
             ),
         )
         gated = bool(getattr(getattr(m, "config", None), "attention_output_gate", False))
-        if name != "linear_qkv" or already_wrapped or isinstance(m, nn.Linear) or not gated:
+        plain_linear = isinstance(m, nn.Linear) and not is_modelopt_linear(m)
+        if name != "linear_qkv" or already_wrapped or plain_linear or not gated:
             return _original_transform(self, m, name, prefix)
 
         if (ans := self.match(m, name, prefix)) is None:
--- a/miles_plugins/megatron_bridge/canonical_lora_gated_qkv.py
+++ b/miles_plugins/megatron_bridge/canonical_lora_gated_qkv.py
@@ -1,0 +1,146 @@
+"""CanonicalLoRA support for attention-output-gated fused QKV (radixark/miles#2008).
+
+Megatron-Bridge's ``CanonicalLoRA`` sizes the ``linear_qkv`` adapter for grouped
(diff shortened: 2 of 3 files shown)


nblintao:
@claude review

layahaasini:
@nblintao thanks for kicking off the review! Claude bot suggestions are addressed in the latest commit (transform guard now mirrors upstream's ModelOpt qualifier, and pre-existing TP=1 interleave limitation is noted in module docstring, with ModelOpt tests added).

nblintao:
@yushengsu-thu could you take a look?

behroozazarkhalili:
Verified on top of current main (f211937b5); the branch merges with no conflicts.

- tests/fast/test_canonical_lora_gated_qkv.py on the merge: 13 passed.
- ruff/black/isort at the pre-commit pins are clean on all three files.
- The new module is the only implementation, so there is no meaningful revert arm; the tests fail to import without it.

The semantic question (whether the gate rows land in the right slice of the fused QKV for every attention-output-gated model) needs an actual checkpoint, which I cannot run this week. From the CPU side this looks ready for a maintainer pass.

cc @yushengsu-thu @Zhichenzzz

PR #2556, opened 2026-08-14
[example] GLM-5.2 launcher: log both entropy signals, stable wandb names

Shi-Dong:
## Summary

Two small fixes to `examples/swe-agent-harbor-docker/run_glm52_lora_tb2_daytona.py`:

- Pass `--use-rollout-entropy` alongside `--observe-training-entropy`. The two flags are independent: the training-side flag alone leaves `rollout/entropy` unpopulated and the miles dashboard's per-token entropy view empty (`dashboard/args.py` warns about exactly this). With both set, entropy is observed on both sides at zero optimisation impact (`--entropy-coef` stays 0).
- Pass `--disable-wandb-random-suffix` so the wandb display name is exactly the `--wandb-group` value instead of `<group>_<id>-RANK_<n>`.

## Test plan

- [x] `uvx black==24.3.0 --line-length 119` passes unchanged.
- [x] Validated live: a 4-node x 8 H200 run launched from current main with these flags completed its first GRPO step logging a real `train/entropy_loss` (0.169) instead of the hardcoded 0.0, with the wandb run named exactly by its group.

### DIFF
--- a/examples/swe-agent-harbor-docker/run_glm52_lora_tb2_daytona.py
+++ b/examples/swe-agent-harbor-docker/run_glm52_lora_tb2_daytona.py
@@ -346,7 +346,7 @@ def execute(args: ScriptArgs):
 
     # train/entropy_loss is a hardcoded 0.0 unless this is set, and a falling entropy is the
     # earliest warning of policy collapse on a long agentic run.
-    misc_args += "--observe-training-entropy "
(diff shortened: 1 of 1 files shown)


Shi-Dong:
Closing per Shi's call — launcher-only tweaks aren't worth a standalone PR; the entropy flags can ride along the next substantive change to this example.

Issue #2560, opened 2026-08-14
Verifiers e2e covers only the legacy rollout path; the default VerifiersRolloutFn has no end-to-end coverage


nblintao:
`tests/e2e/long/test_qwen3_0.6B_verifiers.py` is the only end-to-end proof that the chain miles ↔ verifiers SDK ↔ prime taskset ↔ training loop actually runs. It pins the **legacy** rollout path:

```python
    train_args = " ".join([
        ...
        "--rollout-function-path verifiers_rollout.generate_rollout",
        ...
    ])
    U.execute_train(
        ...
        extra_env_vars={
            "MILES_USE_LEGACY_ROLLOUT_V1": "1",
            ...
        },
    )
```

That was correct when it was written (#1739, will brown, 2026-08-06): the class-based rollout was opt-in, and the env var then read `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=0`. #2522 made the class-based rollout the default and renamed the knob — its diff to this file is exactly one line, `MILES_EXPERIMENTAL_ROLLOUT_REFACTOR=0` → `MILES_USE_LEGACY_ROLLOUT_V1=1`, semantics unchanged.

So the default path — `verifiers_rollout.VerifiersRolloutFn` — has no end-to-end coverage at all. What exists instead:

- `tests/fast/examples/experimental/test_verifiers_run.py` (#2522) checks only **which** function name the launcher writes into `--rollout-function-path` and the ray runtime env. It does not run it.
- `tests/fast/rollout/test_verifiers_rollout.py` covers the adapter's pure functions against fakes.

Neither would catch a `VerifiersRolloutFn` that wires up but does not produce trainable samples.

## Why this matters more than it looks

The e2e's assertions are the ones fakes cannot reach — 12 samples in groups of 4, `len(rollout_log_probs) == len(loss_mask) == response_length`, per-sample `metadata["verifiers"]["rewards"] == {correct, fastest, most_concise}`, at least one `passed`, `sum(fastest) == 0.5` per group. Those hold today only for the path nobody defaults to.

## What it should do

Test the default path: drop `MILES_USE_LEGACY_ROLLOUT_V1` from `extra_env_vars` and point `--rollout-function-path` at `verifiers_rollout.VerifiersRolloutFn`.

The legacy path then has no e2e, and that is the intended outcome — it is an env-var escape hatch on its way out, not a configuration worth spending a 2xH200 job on. Coverage belongs on what every user actually runs.

[...]

PR #2561, opened 2026-08-14
test(ci): file the verifiers tests under the example they test

nblintao:
Follows #2558 (merged), which moved these suites' neighbours under `tests/fast/`. Rebased onto main; the diff is three renames and an empty `__init__.py`.

The verifiers integration's tests were filed in three different places, none of them where the example lives:

| Was | Now |
|---|---|
| `tests/fast/rollout/test_verifiers_rollout.py` | `tests/fast/examples/experimental/verifiers/test_verifiers_rollout.py` |
| `tests/fast/rollout/test_verifiers_runtime.py` | `tests/fast/examples/experimental/verifiers/test_verifiers_runtime.py` |
| `tests/fast/examples/experimental/test_verifiers_run.py` | `tests/fast/examples/experimental/verifiers/test_run.py` |

`tests/fast/rollout/` mirrors `miles/rollout/` — `test_fully_async_rollout.py`, `test_checkpoint_eval.py`, `generate_hub/`, `rm_hub/`, `session/` all cover miles' own rollout code. The two verifiers files are the only ones there covering an example: both import `examples.experimental.verifiers.verifiers_rollout`. The launcher test, meanwhile, sat loose at the `experimental/` root while every other example has a directory.

The destination follows what the repo already does for examples — `tests/fast/examples/swe_agent_harbor_docker/`, `tests/fast/examples/infra_features/p2p_weight_transfer/` — and the launcher test drops its redundant prefix so `test_run.py` names `run.py`, matching the p2p precedent.

## Why this is safe

Pure move, no content change. All three reach their subject through the package path (`examples.experimental.verifiers...`), so unlike the suites in #2558 they need no conftest or `sys.path` bootstrap; none of them touches `__file__`; and nothing else in the repo — docs, CI config, other tests — references the old paths.

`run_suite.py --hw cpu --suite stage-a-cpu --list-only` before and after differs by exactly the three names, each still an implicit CPU entry.

## Not in scope

The substantive gap found while reading these files is #2560: the verifiers e2e pins the legacy rollout path, so the default `VerifiersRolloutFn` has no end-to-end coverage. That is a behavior change to a `run-ci-long` GPU job and belongs in its own PR.


nblintao:
@claude review

Shi-Dong (approved):
LGTM

PR #2573, opened 2026-08-15
rollout: make router/session-server ready timeout env-tunable

williammtan:
## Motivation

The router / session-server ready wait is hardcoded to 30 s. The spawned child re-imports the full torch/sglang module graph, which on a CPU-starved or cold-page-cache host can exceed that: observed on Modal gVisor containers with 8 visible cores shared with 8 initializing Megatron ranks, where the router process was alive but not yet listening at 30 s, so the run died on a healthy startup.

## What this does

Make the ceiling env-tunable: `MILES_SERVER_READY_TIMEOUT_SECS`, default 300. The child-death fast path in `wait_for_server_ready` is unchanged, so a generous ceiling costs nothing on the failure path — it only affects runs that would previously have timed out spuriously.

## Testing

`tests/fast/utils/test_http_utils.py` and `tests/fast/router/test_router.py` pass locally (23 passed).

### DIFF
--- a/miles/ray/rollout/router_manager.py
+++ b/miles/ray/rollout/router_manager.py
@@ -1,6 +1,7 @@
 import copy
 import logging
 import multiprocessing
+import os
 import random
 import uuid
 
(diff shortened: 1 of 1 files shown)


williammtan:
> ## Motivation
> The router / session-server ready wait is hardcoded to 30 s. The spawned child re-imports the full torch/sglang module graph, which on a CPU-starved or cold-page-cache host can exceed that: observed on Modal gVisor containers with 8 visible cores shared with 8 initializing Megatron ranks, where the router process was alive but not yet listening at 30 s, so the run died on a healthy startup.
> 
> ## What this does
> Make the ceiling env-tunable: `MILES_SERVER_READY_TIMEOUT_SECS`, default 300. The child-death fast path in `wait_for_server_ready` is unchanged, so a generous ceiling costs nothing on the failure path — it only affects runs that would previously have timed out spuriously.
> 
> ## Testing
> `tests/fast/utils/test_http_utils.py` and `tests/fast/router/test_router.py` pass locally (23 passed).

PR #2574, opened 2026-08-15
feat(session): return 410 for a session the trainer already reclaimed

Shi-Dong:
## What this changes

A session the trainer has deleted currently answers `404`, exactly like a session ID that
never existed. Through an OpenAI-compatible client both arrive as an indistinguishable
`NotFoundError`.

This adds a bounded, short-lived record of reclaimed session IDs and answers **`410 Gone`**
for them, with a message saying the trainer released the session. Unknown IDs keep their
`404`, so the two cases stay distinguishable.

## Why

The two cases really do both happen, and telling them apart matters.

When the trainer stops waiting for an agent — for example the agent-server call exceeds
`AGENT_TRIAL_TIMEOUT` — `agentic_tool_call.generate` runs `collect_samples` in its `finally`,
and that deletes the session. But `asyncio.wait_for` only cancels the *local* request; the
agent server is never told, so the agent can still be running and still issuing requests.
Every one of those then `404`s.

On a long agentic RL run this produced a sustained ~18% of trials failing with a generic
`NotFoundError` that was reported as a plain agent error — indistinguishable from a bad ID
or a model-side fault. Working out that these were abandoned sessions rather than genuine
agent failures took reconstructing the HTTP access log by hand. A distinct status code makes
that a one-line grep, and lets the agent side classify the outcome as infrastructure
abandonment instead of a model failure.

This is a diagnosability change only. It does **not** fix the underlying race — the trainer
still reclaims sessions out from under live agents, and there is still no per-trial
cancellation. It makes that failure legible so it can be measured and attributed.

## Scope note

`SessionRegistryV2` subclasses `SessionRegistry` and inherits `get_session` / `remove_session`,
so v1 and v2 both pick this up from a single change point.

The tombstone is capped at 4096 IDs and one hour. It only has to outlive an abandoned agent's
next request, so it is a diagnostic breadcrumb rather than a record of the run, and it cannot
grow without bound. Expired entries are dropped on lookup and evicted oldest-first on insert.

## Wire-contract change — please weigh this

Two responses change status code:

- a chat/samples/get request against a **deleted** session: `404` -> `410`
- a second `DELETE` of an already-deleted session: `404` -> `410`

A request against a never-existent ID is unchanged at `404`. `SessionReclaimedError` is a
sibling of `SessionNotFoundError`, not a subclass, so catching one will not swallow the other;
nothing in the tree catches `SessionNotFoundError` today (the only consumer is the
`SessionError` base handler, which reads `status_code`), so this is contained.

Three existing tests encoded the old behaviour and are updated accordingly.

## Verification

Executed, not just read — on a machine with the full runtime:

- `tests/fast/router/test_linear_trajectory.py` + `test_session_state.py` — **101 passed**
- `tests/fast/router/test_sessions.py` — **29 passed** (this is the HTTP-level suite, so the
  `410` and its body are asserted over the real wire, including that an unknown ID still
  returns `404`)
- `black` (line-length 119) and `isort` (repo settings) clean on all seven touched files

New tests cover: reclaimed vs unknown discrimination, both status codes, the sibling-not-subclass
relationship, TTL expiry falling back to `404` without leaking the entry, and bounded eviction
where the oldest tombstones degrade to `404`.

### DIFF
--- a/miles/rollout/session/core.py
+++ b/miles/rollout/session/core.py
@@ -18,9 +18,9 @@ from starlette.responses import Response
 from miles.rollout.generate_utils.sample_utils import merge_samples
 from miles.rollout.session.errors import (
     MessageValidationError,
-    SessionNotFoundError,
     TokenizationError,
     UpstreamResponseError,
+    reclaimed_error,
 )
 from miles.rollout.session.linear_trajectory import SessionRegistry
 from miles.rollout.session.samples.codec import encode_samples
@@ -342,7 +342,7 @@ class SessionCore:
     async def delete_session(self, session_id: str) -> Response:
         session = self.registry.get_session(session_id)
         if session.closing:
-            raise SessionNotFoundError(f"session not found: session_id={session_id}")
+            raise reclaimed_error(session_id)
         session.closing = True
         # Acquire the lock so an in-flight chat finishes before we drop the session.
         await session.lock.acquire()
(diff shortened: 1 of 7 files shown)


Shi-Dong:
@claude review

PR #2580, opened 2026-08-17
fix(lora): export PEFT adapters through Megatron-Bridge

Arist12:
Part of #2705.

## Problem

Miles manually assembles PEFT weights and metadata, duplicating Megatron-Bridge conversion logic and producing names or packed-expert layouts that external PEFT loaders may not consume. The two files are also written in place, so a crash between them leaves a config without its weights.

## Reproduction on `main`

The artifact `main` writes at 70b89e117, from the `test_lora_qwen2.5_0.5B.py` recipe:

```
adapter_model.bin   336 tensors
                    0 of them carry PEFT's required `base_model.model.` prefix
                    sample key: model.layers.0.mlp.gate_proj.lora_A.weight
adapter_model.safetensors   not written
```

`PeftModel.from_pretrained` on that directory raises nothing, matches no key, and leaves all 168 `lora_B` tensors at their zero init while the 168 `lora_B` tensors on disk are non-zero. `B @ A = 0`, so the exported adapter is the identity — every consumer silently gets a no-op. The export is also wrapped in a bare `except Exception` that downgrades any failure to a warning.

## Change

Delegate the export to `AutoBridge.save_hf_adapter`. Stage Bridge output in a temporary directory, then promote safetensors before config, so a reader that keys on `adapter_config.json` never sees it without its weights.

Notes for review:

- Staging is created inside the guarded block, so a rank that cannot create it still reaches the setup consensus instead of hanging its peers. That consensus runs on the gloo group: the failure being reported may be a poisoned GPU communicator.
- A failure *inside* Bridge's collective export propagates rather than being swallowed. Bridge runs all-gathers between its own entry and exit barriers, so a rank that drops out cannot be rejoined — an earlier revision of this PR issued a compensating `dist.barrier()`, which reads as recovery but deadlocks against peers still inside the all-gather.
- Promotion runs after Bridge's exit barrier, where the group is quiesced, so promoting on one rank and agreeing afterwards is safe. A promotion failure stays a warning, as before: the native shards and training state are what resume needs.

Known boundary: dense and per-expert MoE export work; `--experts-shared-outer-loras` remains native-checkpoint-only because Bridge currently rejects that PEFT layout.

## Validation

- 65 fast tests in `test_lora_utils.py`, including injected setup and collective-save failures, cross-rank consensus symmetry and temporary-directory cleanup. Full `tests/fast/backends/megatron_utils` green apart from the two `TestUpdateWeightsZeroChunks` failures that reproduce on `main`.
- 8x MI350X, Qwen3-4B colocated LoRA: the 30-rollout run resumed at `iter_14`, exported complete PEFT pairs at 19/24/29, and the final artifact contains 504 finite tensors with no staging residue
- 4x MI350X dense export loaded through the public PEFT API with nonzero LoRA parameters

## Scope

The multi-LoRA half of this PR has been dropped — that code is being replaced by the in-flight refactor. Removing it also drops this PR's duplicate of the collective-failure helper #2669 adds in `lora_checkpoint.py`, and the CPU-CI failure where a fast test reached `megatron.bridge`, which the `stage-a-cpu` environment does not install.

### DIFF
--- a/docs/advanced/lora.md
+++ b/docs/advanced/lora.md
@@ -182,7 +182,7 @@ alternative aligned-expert path.
   Bridge mode.
 - **Checkpoints.** miles saves native per-rank adapter shards and
   optimizer/scheduler state. Exact resume expects the same TP/PP topology. It
-  also attempts a best-effort HF PEFT `adapter_model.bin` plus
+  also attempts a best-effort HF PEFT `adapter_model.safetensors` plus
   `adapter_config.json` export for external serving and warns if that export
   fails. Direct HF PEFT-to-Bridge resume is not implemented yet; native Inkling
   supplies a model-specific HF adapter importer.
--- a/miles/backends/megatron_utils/lora_utils.py
+++ b/miles/backends/megatron_utils/lora_utils.py
@@ -1,9 +1,9 @@
 """LoRA utilities for Megatron backend using Megatron-Bridge PEFT integration."""
 
 import logging
-import os
 from argparse import Namespace
 from collections.abc import Sequence
+from contextlib import ExitStack
 from pathlib import Path
 from typing import Any
 
@@ -356,7 +356,7 @@ def parse_exclude_modules(args: Namespace, lora_type=None) -> list[str]:
     return exclude_modules
 
 
-def create_lora_instance(args: Namespace):
+def create_lora_instance(args: Namespace, *, log_config: bool = True):
     """Create a LoRA or CanonicalLoRA instance based on args.
 
     Returns:
@@ -391,17 +391,46 @@ def create_lora_instance(args: Namespace):
 
     lora = lora_cls(**lora_kwargs)
 
-    logger.info(
-        f"Created {lora_cls.__name__}: rank={args.lora_rank}, alpha={args.lora_alpha}, "
-        f"dropout={args.lora_dropout}, target_modules={target_modules}, "
-        f"exclude_modules={exclude_modules}"
-    )
+    if log_config:
+        logger.info(
+            f"Created {lora_cls.__name__}: rank={args.lora_rank}, alpha={args.lora_alpha}, "
+            f"dropout={args.lora_dropout}, target_modules={target_modules}, "
+            f"exclude_modules={exclude_modules}"
+        )
     return lora
 
 
 # ---------------------------------------------------------------------------
 # Checkpoint save/load
 # ---------------------------------------------------------------------------
+
+
+def _build_peft_export(
+    adapter_weights,
+    *,
+    rank: int,
+    alpha: int,
+    dropout: float,
+    base_model_name_or_path: str,
+) -> tuple[dict[str, torch.Tensor], dict[str, object]]:
+    """Convert Bridge adapter weights to a loadable PEFT state dict and config."""
+    from megatron.bridge.models.conversion.peft_bridge import (
+        build_adapter_config_dict,
+        convert_adapter_weights_to_peft_state,
+        infer_rank_pattern_from_adapter_weights,
+        infer_target_modules_from_adapter_weights,
+    )
+
+    adapter_weights = list(adapter_weights)
+    state_dict, module_weight_names, target_parameters = convert_adapter_weights_to_peft_state(adapter_weights)
+    config = build_adapter_config_dict(
+        Namespace(dim=rank, alpha=alpha, dropout=dropout),
+        target_modules=infer_target_modules_from_adapter_weights(module_weight_names),
+        target_parameters=target_parameters,
+        base_model_name_or_path=base_model_name_or_path,
+        rank_pattern=infer_rank_pattern_from_adapter_weights(adapter_weights, default_rank=rank),
+    )
+    return state_dict, config
 
 
 def save_lora_checkpoint(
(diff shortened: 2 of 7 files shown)


jhinpan (commented):
I reviewed the current head (`99a218d5028f50a327de0f9923c58fab12242bf1`) and tested both successful PEFT loading and rank-local failure behavior on AMD Instinct MI355X GPUs.

The staged safetensors produced by the happy path were loadable through `PeftModel.from_pretrained`, but the distributed failure protocol still has a merge blocker.

**Blocker — staging-directory creation occurs outside setup consensus.**

`TemporaryDirectory(...)` is entered before the inner `try` that assigns `hf_export_err`. If it fails on one rank, that rank skips `all_gather_object` and proceeds toward the trailing barrier while peers wait in `all_gather_object`. Injecting this failure on one of two MI355 ranks reproducibly timed out after 30 seconds.

Move staging-directory creation into the coordinated setup phase so every rank reaches the same consensus before any rank enters Bridge.

**Publication occurs before all-rank save success is known.**

Rank 0 moves weights/config into the public directory before `save_errors` is gathered. A peer failure can therefore be discovered only after files are visible. Overwriting an existing adapter also exposes new weights alongside the old config between the two replacements, and a config-rename failure can make that mismatch permanent.

Use an explicit two-phase protocol:
1. create staging and gather setup errors;
2. perform the collective Bridge export;
3. gather all save results;
4. publish on rank 0 only after global success;
5. gather promotion results.

A versioned directory or explicit commit marker is safer for overwrite/retry semantics.

**The recovery protocol depends on undocumented Bridge internals.**

The compensating `dist.barrier()` assumes `save_hf_adapter` has exactly one entry and one exit barrier. Its public contract only says the method is collective. The object consensuses also use the default RCCL group, unlike the Gloo helper used elsewhere on this path. Please centralize failure consensus on `get_gloo_group()` and avoid encoding Bridge's private barrier count.

Finally, keep Bridge's empty-adapter guard on the multi-LoRA helper path; otherwise an empty safetensors file plus `target_modules: []` can be promoted as a complete checkpoint.

jhinpan (commented):
Follow-up review on the unchanged head (`99a218d5028f50a327de0f9923c58fab12242bf1`): one new retry/topology issue remains after excluding the first review's publication findings.

**A failed multi-LoRA attempt directory is reused and later promoted wholesale.**

`save_multi_lora_checkpoints` uses the deterministic `_tmp_step_{iteration}` directory. Several new coordinated failure points can raise after native shards have already been written, but none removes that directory. On restart, `save_due_adapter_checkpoints` retries because only `final_dir` suppresses a save, `mkdir(exist_ok=True)` reuses the old temporary directory, and `os.replace(tmp_dir, final_dir)` publishes every stale file that remains in it.

Concrete failure sequence:
1. an EP4 attempt writes `_ep0` through `_ep3` shards and then fails during PEFT conversion;
2. the run restarts at EP1, writes the no-EP-suffix shard into the same temporary directory, and promotes it;
3. a later EP4 load sees the stale `_ep0` through `_ep3` set as complete and resumes those pre-failure weights while reporting the retried step as valid.

Use a per-attempt unique staging directory whose name is chosen once and broadcast to all ranks. Alternatively, have one coordinator remove the deterministic temporary directory, reach a barrier, and only then let shard writers recreate/populate it. Concurrent shard writers should not each race to remove the shared directory.

Arist12:
Thanks. Three points adopted; on the publication-ordering redesign I read Bridge's source and came to a different conclusion — details below so it can be checked.

**Adopted.** The export now stages into a temp directory *inside* the failure consensus rather than after it, so a rank that fails mid-export cannot leave a half-written directory that a later run mistakes for complete. The consensus all-gather runs on the gloo group rather than the default group, so it does not serialise behind NCCL work. And the empty-adapter case now raises instead of promoting an empty `adapter_model.safetensors` with `target_modules: []` as a valid checkpoint.

**Adopted (comment only).** The `.clone().float()` cast now says why: it matches `AutoBridge.save_hf_adapter`, so both export paths write the same dtype.

**Not adopted — the 5-step publication protocol / dropping the compensating barrier.** The concern is that the extra `dist.barrier()` after the Bridge call papers over an ordering bug. Looking at `megatron/bridge/models/conversion/auto_bridge.py`:

- `save_hf_adapter` opens with a barrier (`:582-583`) and closes with one (`:633-634`), and there is **no code after that final barrier**. So nothing this PR could add can run after Bridge's own exit synchronisation from inside the call.
- The empty-adapter case raises on *every* rank (`:592-598`), not just rank 0, so it is already collective and cannot desynchronise the group.
- Rank 0 is Bridge's own choice of writer (`is_rank0 = not dist.is_initialized() or dist.get_rank() == 0`, `:606`); the PR does not assume it independently.

Removing the compensating barrier means the non-writer ranks return from Bridge while rank 0 is still writing, then reach the gloo consensus and observe an incomplete directory — which is the failure the barrier exists to prevent. Restructuring the whole publication into a 5-step protocol would mean reimplementing what Bridge already does rather than calling it, which is the opposite of what this PR is for (title: *export PEFT adapters through Megatron-Bridge*). Also relevant: completion is already signalled by `HF_EXPORT_COMPLETE_MARKER` (`hf_export.py:192-193`), so a partially written directory is not consumed as complete downstream.

Happy to revisit if I've misread the barrier placement — that is the load-bearing claim.

**Not adopted — "this PR bundles four changes."** The four hunks are the single change of routing PEFT export through Bridge: key prefixing, rank pattern, target modules and dtype are all what `save_hf_adapter` does and what the multi-LoRA path has to match to produce a loadable adapter. Splitting them yields intermediate states that write adapters PEFT cannot load.

**Deferred — `_tmp_step_{iteration}` reuse across reruns.** Pre-existing and orthogonal; same follow-up as noted on #2669.

Validation: `tests/fast/backends/megatron_utils` — 265 passed, plus the two `TestUpdateWeightsZeroChunks` failures that reproduce on `main`. pre-commit clean.

PR #2581, opened 2026-08-17
dashboard: resolve CP/PP-sharded train dumps offline and fix a partition-reader race

yueming-yuan:
## What was broken

Four user-visible dashboard bugs, all found while debugging a GB300 16-node GLM-5.2 run (TP2/CP4/PP4):

1. **Token view unusable under CP > 1.** Per-token train columns (`log_probs`, `lp_diff`, `imp_ratio`, ...) are dumped as this rank's CP slice, but nothing recorded which slice a rank held. The reader concatenated slices in file order, which is wrong for the zigzag CP split, so the view either crashed on length mismatch or silently showed misaligned values.
2. **Batch view per-token columns empty under PP.** Only the last pipeline stage dumps per-token outputs; the reader picked a rank file arbitrarily and found nothing.
3. **`avg staleness` stat off by one.** The frontend recomputed `rollout_id + 1 - weight_version_min` instead of using the backend value; measured 2.75 vs the native 1.75.
4. **`--follow` dashboards double-count records.** `PartitionedReader._block` is a read-modify-write spanning two GIL-releasing steps (file read, parse). The follow thread and the threadpool request handlers could both read from the same stale offset and append the same byte range twice. Duplicates carry identical timestamps, so every timestamp-based check passes while every count-based aggregate is inflated (observed: a lane rendering 147 segments whose ground truth is 49, shredding the batch-anatomy axis).

## Design

- `save_debug_train_data_for_rank` now records `cp_rank` / `cp_size` / `qkv_format` alongside `rank`, passed explicitly like `rank` is (dumpers stay parallel-state-free for tests).
- New `assemble_log_prob_from_cp` in `cp_utils` is the exact inverse of `slice_log_prob_with_cp`: both sides call the same `get_logits_and_tokens_offset_with_cp`, which now accepts explicit `cp_rank`/`cp_size` (defaulting to parallel state as before). Round-trip pinned by `test_cp_log_prob_assembly` for thd and bshd.
- For dumps that predate the metadata, the reader recovers the layout: group rank files by content (`rollout_log_probs` probe), then place slices by the deterministic per-`(total_length, cp_size)` width pattern. Verified widths must match exactly.
- A sample whose slices cannot be placed is reported (`alignment_failed`, surfaced as an advisory) instead of shown misaligned.
- Samples table gains a `staleness` column (`rollout_id - oldest weight version`); the stat box reads the backend value.
- `_block` takes a lock making offset-read -> parse -> append one transaction. Regression test drives the race deterministically with a delayed parser.

## Verification

- 241 dashboard tests pass; new: CP assembly round-trips (thd x3 sizes, bshd x2, partial-group rejection) and the concurrent-fill regression.
- Real TP2/CP4/PP4 dumps: 64/64 samples aligned (`alignment_failed` 0), token view and batch view fully populated.
- Race fix end-to-end: 24 concurrent `/api/rollout/15/trajectories` requests all return the ground-truth 49 segments (previously 147 = 3x after a follow-tick collision).

### DIFF
--- a/miles/backends/training_utils/cp_utils.py
+++ b/miles/backends/training_utils/cp_utils.py
@@ -21,13 +21,20 @@ def get_logits_and_tokens_offset_with_cp(
     response_length: int,
     qkv_format: str = "thd",
     max_seq_len: int | None = None,
+    cp_rank: int | None = None,
+    cp_size: int | None = None,
 ):
     """
     All offsets start from the begining of the prompt.
+
+    ``cp_rank`` / ``cp_size`` default to this process's parallel state. Pass them
+    explicitly to compute another rank's offsets outside the process group, which
+    is how an offline reader reassembles a sample from per-rank dumps.
     """
-    parallel_state = get_parallel_state()
-    cp_rank = parallel_state.cp.rank
-    cp_size = parallel_state.cp.size
+    if cp_rank is None or cp_size is None:
+        parallel_state = get_parallel_state()
+        cp_rank = parallel_state.cp.rank
+        cp_size = parallel_state.cp.size
     assert cp_size > 1
 
     prompt_length = total_length - response_length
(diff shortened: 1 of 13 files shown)


Zhichenzzz (approved):
LGTM

PR #2587, opened 2026-08-17
openenv: stop the episode at a length-truncated turn

yueming-yuan:
## Problem

In the tbench2 agent loop, a turn that hits `--rollout-max-response-len` comes back with `finish_reason="length"` and closes the trainable sample at that turn — sample collection keeps nothing past a truncated turn. The agent loop did not know that: it executed the cut-off command and kept generating until `OPENENV_MAX_TURNS` (or the episode wall-clock), entirely on tokens training will discard.

Measured on a GB300 16-node GLM-5.2 run (6469 episodes): the post-truncation ghost tail consumed **22.7% of all episode wall-clock** (144.7h of 638h), concentrated at the longest contexts of the run — so its share of engine *compute* is higher still. One concrete sample: truncated at turn 22, then 8 more turns over 16.4 minutes (73% of its episode), all discarded. Samples that rode the tail to the wall-clock cap came back ABORTED and took their whole group with them via `check_no_aborted`.

The tail is invisible in the dashboard's batch anatomy (tool/gen spans after the final policy call cannot be derived from request gaps), which is how it went unnoticed: it renders as a long "queue/waiting" band after the last visible activity.

Secondary effect: ghost turns lengthen episodes, which fattens the fully-async whole-group staleness tail — the carrier population of the train/rollout divergence documented in the smoke-test report.

## Fix

Break the turn loop when `finish_reason == "length"`; skip the truncated turn's cut-off command (mid-stream garbage); scoring (`evaluate`) still runs so the episode is graded as-is.

## Tests

- `_FakePolicy` now carries `finish_reason` like the real SDK response.
- New `test_truncated_turn_ends_the_episode`: one policy call, no exec of the cut-off command, evaluate still runs, `turns == 1`.
- `tests/fast/examples/experimental/openenv/`: 103 passed, 2 skipped (arm64 container).

### DIFF
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -300,6 +300,9 @@ async def multi_turn(
             # (extras like reasoning_content included).
             convo.append(message.model_dump(exclude_none=True))
 
+            if completion.choices[0].finish_reason == "length":
+                break
+
             command = _strip_fence(reply) if "```" in reply else reply.strip()
             if not command or command.upper().startswith("TASK_COMPLETE"):
                 break
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -288,7 +288,8 @@ async def multi_turn(
                 model=model_name, messages=convo, extra_body=request_kwargs
             )
             gen_times.append(time.monotonic() - t0)
-            message = completion.choices[0].message
+            choice = completion.choices[0]
+            message = choice.message
             reply = message.content or ""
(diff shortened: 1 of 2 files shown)


examples/experimental/openenv/openenv_agent_function.py
```diff
@@ -300,6 +300,9 @@ async def body(env: Any) -> tuple[float | None, int, list[float], list[float], f
             # (extras like reasoning_content included).
             convo.append(message.model_dump(exclude_none=True))
 
+            if completion.choices[0].finish_reason == "length":
```
nblintao: TODO: https://github.com/radixark/miles/issues/2592

nblintao (approved):

PR #2588, opened 2026-08-17
glm52_tbench2: collapse levers and launch fixes from the GB300 16-node smoke runs

yueming-yuan:
The exact example-side configuration of the GB300 16-node smoke runs (jobs 2403/2415), extracted so the run is reproducible from main:

```
examples/experimental/openenv/glm52_tbench2/launch_16node_slurm.sh \
  --eval-interval 0 --max-seq-len 65536 --sglang-config low-latency
```

## Collapse levers

Two 16-node runs on unmodified main collapsed the same way: `train_rollout_kl` flat at ~0.02, then an exponential ratchet (57x over 30 steps) concentrated entirely on near-deterministic tokens, followed by response-length blowup, truncation ~0.9 and reward -> 0.03. Root cause chain, per-token forensics on the train dumps:

- Failed samples loop (polling/retry cycles), hit the per-turn cap, truncate, and carry negative advantage over 16k+ near-deterministic scaffolding tokens shared with successful samples.
- `--tis-clip-low 0.5` **clamps** a collapsed importance ratio (exp(-30)) up to 0.5 — manufacturing gradient exactly where importance sampling says there should be none. The recipe now uses the miles default `0.0`, i.e. standard truncated IS (`w = min(ratio, 2)`).
- `rollout_max_response_len` 16384 -> 8192: completed samples never touched the cap in any run; only degenerate loops did. 16384 doubles a capped loop's token mass for no benefit.
- The policy client used SDK defaults (600 s timeout, `max_retries=2`). Slow 16k-token turns time out, the SDK resubmits while the engine keeps decoding the original, and zombies accumulate — measured ~1000 engine-side running requests against 128 live samples (~85-90% of decode capacity wasted), at ~500 duplicate requests/hour. `timeout=3600` matches the episode wall-clock cap; `max_retries=0` because resending a generation is not idempotent.
- A turn whose `finish_reason` is `length` now ends the episode instead of feeding a truncated turn forward.

With these three levers the run holds `kl ~= 0.03`, `tis_clipfrac ~= 0.007`, healthy climbing reward, and zero retry lines for 70+ steps (2.7x the unfixed lifetime; a residual slower ratchet through in-band stale groups is being addressed separately via `--max-weight-staleness`).

## Operational fixes hit while bringing the run up

- `--eval-interval 0` disables eval from the CLI (`int | None` + typer could not express "off").
- `FABRIC_PREFIX` is required: the previous default silently selected an unroutable address and the launch hung 15 minutes with no error.
- `--max-seq-len` is a recipe flag instead of a hard-coded 131072.
- Daytona org-capacity errors ("Total CPU limit exceeded", memory/disk variants) retry with the same jittered backoff as rate limits: failing immediately aborts the episode before its first model call, and `check_no_aborted` then discards the whole group. Retry count configurable via `OPENENV_DAYTONA_CREATE_MAX_RETRIES`.
- README updated to match.

### DIFF
--- a/examples/experimental/openenv/glm52_tbench2/run_glm5_2_744b_a40b_daytona.py
+++ b/examples/experimental/openenv/glm52_tbench2/run_glm5_2_744b_a40b_daytona.py
@@ -55,7 +55,9 @@ class ScriptArgs(U.ExecuteTrainConfig):
     rollout_batch_size: int = 8
     n_samples_per_prompt: int = 8
     global_batch_size: int = 64
-    rollout_max_response_len: int = 16384
+    rollout_max_response_len: int = 8192
+    # whole-session budget; CP splits it, so it also sets per-rank activations
+    max_seq_len: int = 131072
     # Max concurrently generating trajectories, decoupled from the train batch;
     # also sizes the Daytona pool so every in-flight trajectory has a sandbox.
     async_max_concurrent_samples: int = 128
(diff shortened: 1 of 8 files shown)


Shi-Dong (approved):

Zhichenzzz:
fix the failed ci, wait for a second

examples/experimental/openenv/glm52_tbench2/run_glm5_2_744b_a40b_daytona.py
```diff
@@ -55,7 +55,9 @@ class ScriptArgs(U.ExecuteTrainConfig):
     rollout_batch_size: int = 8
     n_samples_per_prompt: int = 8
     global_batch_size: int = 64
-    rollout_max_response_len: int = 16384
+    rollout_max_response_len: int = 8192
+    # whole-session budget; CP splits it, so it also sets per-rank activations
```
nblintao: TODO: https://github.com/radixark/miles/issues/2591

nblintao (approved):
You probably need to rebase this on PR 2587?

PR #2595, opened 2026-08-18
[Sampling Replay 2/8] Represent and transport rollout sampling support

nanjiangwill:
## Summary

- Add a CSR representation for per-token rollout sampling support to Sample, including validation, truncation, reset, and multi-turn merge behavior.
- Preserve the compact representation through the session codec and trainer transport.
- Add passive helpers for forced environment tokens and multi-turn support merging.

This PR defines and transports the data contract only. It does not request support from SGLang or apply support during actor scoring, so existing rollout and training behavior is unchanged.

## Why

Top-p support is ragged and changes at every generated token. Flattened token IDs plus response-aligned offsets preserve the exact realized support without padding it to response_length by top_k during rollout, session storage, or object-store transport.

Keeping representation separate from activation makes the transport contract independently reviewable and leaves no partially enabled sampling behavior.

## Sampling-support replay series

| Part | PR | Responsibility | State |
|---:|---|---|---|
| 1/8 | #2200 | TP/CP-aware masked log-probability primitive | Merged |
| 2/8 | #2595 | Compact support representation and transport | Merged |
| 3/8 | #3327 | Actor-only Megatron and FSDP integration | Open |
| 4/8 | #3328 | Native SGLang support capture | Open |
| 5/8 | #3329 | Session and agentic preservation | Open |
| 6/8 | #3330 | User-facing validation | Open |
| 7/8 | #3331 | GPU E2E | Open |
| 8/8 | #3332 | Documentation | Open |

External prerequisite: #3321 provides the SGLang v0.5.20 base. It is owned by another contributor and is not part of this numbered series.

#2596 is the [DO NOT MERGE] review umbrella for the complete stack.

## Validation

- Repository pre-commit hooks passed on every changed file.
- 95 focused representation, transport, codec, and log-probability tests passed on one H200.

## Acknowledgements

Thank you to [Zilin Zhu](https://github.com/zhuzilin) for adding top-p masking support to slime in [THUDM/slime#2102](https://github.com/THUDM/slime/pull/2102).

PR #2200, opened 2026-08-04
[Sampling Replay 1/8] Add sampling-support log-prob primitives

nanjiangwill:
## Summary

- Add an optional per-token sampling-support mask primitive for actor log-probability computation.
- Preserve the public get_responses() two-tuple contract while deriving context-parallel response-row indices only when support replay needs them.
- Apply an optional support mask through the existing true-on-policy and fused tensor-parallel cross-entropy paths.

No production caller supplies a sampling mask in this PR, so rollout and training behavior remain unchanged.

## Why

The trainer needs a correct way to normalize actor log probabilities over the same sampling support used during rollout. This primitive intentionally landed before capture was activated: rollout log probabilities must not switch to a restricted distribution until actor scoring can use the same support.

The logical input follows rollout data directly: one list of supported token IDs per response token. The enabled path selects response rows owned by the current context-parallel rank, flattens those supports once, and builds one tensor-parallel-local boolean mask. The disabled path does not build response indices or allocate a mask.

## Sampling-support replay series

| Part | PR | Responsibility | State |
|---:|---|---|---|
| 1/8 | #2200 | TP/CP-aware masked log-probability primitive | Merged |
| 2/8 | #2595 | Compact support representation and transport | Merged |
| 3/8 | #3327 | Actor-only Megatron and FSDP integration | Open |
| 4/8 | #3328 | Native SGLang support capture | Open |
| 5/8 | #3329 | Session and agentic preservation | Open |
| 6/8 | #3330 | User-facing validation | Open |
| 7/8 | #3331 | GPU E2E | Open |
| 8/8 | #3332 | Documentation | Open |

External prerequisite: #3321 provides the SGLang v0.5.20 base. It is owned by another contributor and is not part of this numbered series.

#2596 is the [DO NOT MERGE] review umbrella for the complete stack.

## Validation

- Repository lint and formatting checks passed on every changed file.
- Focused CPU checks cover nested-support selection, tensor-parallel vocabulary filtering, validation, and empty local response shards.

## Acknowledgements

Thank you to [Zilin Zhu](https://github.com/zhuzilin) for adding top-p masking support to slime in [THUDM/slime#2102](https://github.com/THUDM/slime/pull/2102).

PR #2596, opened 2026-08-18
[DO NOT MERGE] [Sampling Replay] Bounded sampling-support replay

nanjiangwill:

> [!WARNING]
> Review umbrella only. Do not merge this PR. Parts 3/8 through 8/8 form a linear branch chain. GitHub native stack merge becomes available after those head branches are adopted into `radixark/miles`; GitHub does not support cross-fork stacks.

## Goal

SGLang samples each token after temperature and support filters. A full-vocabulary actor softmax is not the same policy as the bounded rollout distribution. Sampling-support replay records the realized support S_t and defines both behavior and actor scores on that fixed support:

- SGLang returns log mu(a_t | s_t, S_t).
- Miles computes log pi_theta(a_t | s_t, S_t) by masking and renormalizing current actor logits.
- S_t remains fixed during optimization; Miles does not recompute the current actor top-p boundary.

## Merge series

| Part | PR | Commit | Responsibility | State |
|---:|---|---|---|---|
| 1/8 | #2200 | [`f0499c46e7`](https://github.com/radixark/miles/commit/f0499c46e78a2f19a694728020dfcb3e225542be) | TP/CP-aware masked log-probability primitive | Merged |
| 2/8 | #2595 | [`29c2c3aee4`](https://github.com/radixark/miles/commit/29c2c3aee4be0c6d426ca0695880819899b1a2e6) | Compact sampling-support representation and transport | Merged |
| 3/8 | #3327 | [`d5a9a374ae`](https://github.com/nanjiangwill/miles/commit/d5a9a374ae304f0d88a466beb15eb844dc639a07) | Actor-only replay in Megatron and FSDP | Open |
| 4/8 | #3328 | [`431fc267da`](https://github.com/nanjiangwill/miles/commit/431fc267da26540c5367c647b4b1e5a4466392da) | Native SGLang support capture for direct rollouts | Open |
| 5/8 | #3329 | [`a7d99c7d48`](https://github.com/nanjiangwill/miles/commit/a7d99c7d487e44203bf16f361bad47c3a0cd8288) | Session v1/v2 and agentic preservation | Open |
| 6/8 | #3330 | [`de5cb50b3e`](https://github.com/nanjiangwill/miles/commit/de5cb50b3e9e84893cac6cb89338c2cf5f01c385) | User-facing bounds and compatibility validation | Open |
| 7/8 | #3331 | [`a57ca059bd`](https://github.com/nanjiangwill/miles/commit/a57ca059bdff66a4a71a8c1b79a4ce7bab8613f2) | Focused 2xH200 E2E and metric gate | Open |
| 8/8 | #3332 | [`0477a2c4ea`](https://github.com/nanjiangwill/miles/commit/0477a2c4ea5c26c57971269b24f52bb9d4f1a89d) | Advanced guide, navigation, and CLI documentation | Open |

External prerequisite: #3321 provides the SGLang v0.5.20 base. It is owned by another contributor and is deliberately not counted as part of this series.

GitHub native stack merge requires every head branch to live in `radixark/miles`. The current cross-fork PRs preserve the exact linear history and link each part-only diff, but GitHub cannot link them as a native stack. After a Miles maintainer publishes these six refs in `radixark/miles`, recreate or link parts 3/8 through 8/8 so that each PR targets the branch below it; merging 8/8 will then merge the complete open stack.

The umbrella head contains the merged foundations through its base ancestry: #2200 landed as `f0499c46e7`, and #2595 landed as `29c2c3aee4`. GitHub lists only parts 3/8 through 8/8 in this PR's Commits tab because parts 1/8 and 2/8 are already reachable from the base. The head is byte-identical to part 8/8, #3332.

## End-to-end behavior

- Replay activates when rollout top-p is below 1 or rollout top-k is positive.
- SGLang v0.5.20 supplies the realized support and its support-normalized selected-token log probability.
- Standard, legacy, multi-turn, tool-call, and session v1/v2 paths preserve the same contract.
- Environment and tool tokens receive singleton support.
- Megatron and FSDP apply support only to actor scores. Reference, teacher, value, evaluation, and entropy semantics remain full-vocabulary or unchanged.
- Evaluation disables training-only support capture.

## Ownership and fail-closed boundaries

SGLang owns sampling-mask capacity, overflow, custom logit processors, and speculative-decoding compatibility. Miles does not duplicate those policies and does not add a speculative-decoding gate.

Miles rejects cases its trainer cannot replay faithfully:

- finite top-p without a positive top-k;
- request temperature different from rollout temperature;
- frequency, presence, or repetition penalties and logit bias;
- prefill log-probability recomputation;
- a router that drops the native request extension;
- incomplete or inconsistent response metadata.

Request-specific positive top-k may differ from the launch default. SGLang enforces the physical support limit.

## Current KL and distillation limitation

The actor still produces full-vocabulary logits, so sampling replay is not mathematically incompatible with KL or distillation. The current loss interface exposes one actor score per token. Replay uses that slot for the support-normalized policy score, while KL and distillation require a separate full-policy actor score. Until both are carried independently, validation rejects nonzero KL coefficient, KL loss, and on-policy distillation with replay.

## Validation

- 651 relevant CPU tests passed across trainer math, direct rollout, sessions, transport, FSDP, and Megatron integration; 10 tests were skipped.
- Repository pre-commit hooks and diff checks passed on the complete tree.
- #3331 registers a 2xH200 Qwen3-0.6B E2E covering live SGLang TP=2 capture, Megatron TP=2 replay, one GRPO update, weight synchronization, and the existing log-probability consistency assertion.

## Acknowledgements

Thank you to [Zilin Zhu](https://github.com/zhuzilin) for the original top-p replay implementation in [THUDM/slime#2102](https://github.com/THUDM/slime/pull/2102).

### DIFF
--- a/miles/backends/megatron_utils/model.py
+++ b/miles/backends/megatron_utils/model.py
@@ -28,11 +28,13 @@ from megatron.training.training import get_model
 from miles.backends.megatron_utils.ft.indep_dp import allreduce_grads_and_losses_across_replicas
 from miles.backends.megatron_utils.ft.types import TrainStepOutcome
 from miles.backends.megatron_utils.local_weight_checksum import dump_local_weight_checksums
+from miles.backends.training_utils.sampling_mask import get_rollout_sampling_mask
 from miles.utils.audit_utils.witness.allocator import WitnessInfo
 from miles.utils.audit_utils.witness.module import witness_dump_and_clear_stale
 from miles.utils.dumper_utils import DumperMegatronUtil, DumperPhase
 from miles.utils.memory_utils import clear_memory
 from miles.utils.multi_lora import is_multi_lora_enabled
+from miles.utils.sampling import sampling_mask_replay_enabled
 from miles.utils.test_utils.ft_test_actions import FTTestActionActorExecutor
 from miles.utils.tracking_utils.structured_log import log_structured
 
@@ -257,6 +259,7 @@ def forward_only(
     num_microbatches: Sequence[int],
     rollout_id: int,
     store_prefix: str = "",
+    use_rollout_sampling_mask: bool = False,
 ) -> dict[str, list[torch.Tensor]]:
     """Run forward passes only and collect non-loss outputs (e.g., logprobs).
 
@@ -271,6 +274,8 @@ def forward_only(
         num_microbatches: Number of microbatches per rollout step.
         rollout_id: Rollout identifier (selects the per-rollout dump subdirectory).
         store_prefix: Prefix to prepend to stored output keys.
+        use_rollout_sampling_mask: Whether this is actor scoring over captured
+            rollout support.
 
     Returns:
         Aggregated outputs keyed by ``store_prefix + key``.
@@ -305,21 +310,27 @@ def forward_only(
         assert not return_schedule_plan, "forward_only step should never return schedule plan"
 
         # Get the batch.
+        forward_only_keys = [
+            "tokens",
+            "loss_masks",
+            "multimodal_train_inputs",
+            "total_lengths",
+            "response_lengths",
+            "max_seq_lens",
+            "witness_ids",
+        ]
+        if use_rollout_sampling_mask:
+            forward_only_keys.extend(["rollout_sampling_mask_ids", "rollout_sampling_mask_offsets"])
         batch = get_batch(
             data_iterator,
-            [
-                "tokens",
-                "loss_masks",
-                "multimodal_train_inputs",
-                "total_lengths",
-                "response_lengths",
-                "max_seq_lens",
-                "witness_ids",
-            ],
+            forward_only_keys,
             args.data_pad_size_multiplier,
             args.qkv_format,
             allgather_cp=args.allgather_cp,
         )
+        sampling_mask_ids = sampling_mask_offsets = None
+        if use_rollout_sampling_mask:
+            sampling_mask_ids, sampling_mask_offsets = get_rollout_sampling_mask(batch)
         unconcat_tokens = batch["unconcat_tokens"]
         tokens = batch["tokens"]
         packed_seq_params = get_packed_seq_params(batch, args)
@@ -350,6 +361,8 @@ def forward_only(
             response_lengths=response_lengths,
             with_entropy=args.use_rollout_entropy,
             max_seq_lens=batch.get("max_seq_lens", None),
+            rollout_sampling_mask_ids=sampling_mask_ids,
+            rollout_sampling_mask_offsets=sampling_mask_offsets,
         )
 
     # Turn on evaluation mode which disables dropout.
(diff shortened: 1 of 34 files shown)


miles/backends/megatron_utils/model.py
```diff
@@ -361,6 +370,7 @@ def forward_step(
             response_lengths=response_lengths,
             with_entropy=args.use_rollout_entropy,
             max_seq_lens=batch.get("max_seq_lens", None),
+            rollout_sampling_mask=rollout_sampling_mask,
```
guapisolo: codex cmt

[P1] Only pass sampling-mask kwargs to actor replay callbacks

`forward_only()` now unconditionally adds `rollout_sampling_mask` to `partial(f, ...)`. The critic path calls `forward_only(get_values, ...)`, but `get_values()` does not accept this keyword. Setting `use_rollout_sampling_mask=False` only makes the value `None`; it does not remove the keyword, so even default `--rollout-top-p 1` Megatron critic training fails with `TypeError: get_values() got an unexpected keyword argument 'rollout_sampling_mask'` when the callback runs.

Please add this callback kwarg only when actor sampling-support replay is enabled, preserving the existing critic callback signature.

miles/utils/arguments.py
```diff
@@ -2997,6 +3011,27 @@ def miles_validate_args(args):
             "have known issues in SGLang and need to be fixed."
         )
 
+    if not 0.0 < args.rollout_top_p <= 1.0:
+        raise ValueError(f"--rollout-top-p must be in (0, 1], got {args.rollout_top_p}")
+    if args.rollout_top_k != -1 and args.rollout_top_k < 1:
+        raise ValueError(f"--rollout-top-k must be -1 or at least 1, got {args.rollout_top_k}")
+    if top_p_sampling_replay_enabled(args):
```
guapisolo, line 3018: codex cmt

[P2] Reject sampling replay with speculative decoding before worker startup

`--rollout-top-p 0.8 --rollout-top-k 32 --use-miles-router --sglang-speculative-algorithm EAGLE` passes these checks and enables `return_sampling_mask=True` for training requests. SGLang `98bb1455b9bfc922eaf199ea31945144e2f90fef` explicitly aborts every such request in `scheduler.py` with `return_sampling_mask is not supported with speculative decoding.` This combination therefore fails only after rollout/model startup. The workaround is to disable speculative decoding for the training rollout or use `--rollout-top-p 1`.

Please check the effective `speculative_algorithm` for each actual server group of the primary rollout model in `SglangConfig.resolve`, before placement-group creation and worker startup. Respect CLI/YAML override precedence: a group can enable EAGLE when the global option is unset, or explicitly disable it with `null` when the global option enables it. Skip placeholders and preserve independent eval/reward models' speculative configuration.

miles/backends/megatron_utils/model.py
nanjiangwill: Good catch. Fixed in d17f5716d: `rollout_sampling_mask` is now added to the callback only when actor sampling-support replay is enabled, so the critic keeps its existing callback signature. Added a regression test for the unmasked callback path.

miles/utils/arguments.py
nanjiangwill: Agreed for the currently pinned SGLang. Fixed in d17f5716d at resolved-config time: each non-placeholder group of the primary rollout model is checked using the effective `speculative_algorithm` (group override over CLI, including explicit `null`). Secondary reward/eval models remain independent. Tests cover both precedence directions and scope.

nanjiangwill:
Replaced by #3354, rebased onto current main. GitHub cannot reopen or retarget this PR because its deleted base branch (#3321) causes the API to reject both operations.

PR #2597, opened 2026-08-18
fix(weight-sync): load transfer backends lazily

nanjiangwill:
## Summary

- load each mutually exclusive weight-transfer implementation only in its existing selection branch
- keep colocated, broadcast, disk-delta, and P2P backend imports structurally consistent
- prevent non-P2P runs from loading Mooncake RDMA libraries

## Root cause

The Megatron actor imported several transfer implementations at module load time. In particular, the P2P updater transitively imported Mooncake and loaded native RDMA libraries even when another transfer mode was configured.

## Validation

- python -m py_compile miles/backends/megatron_utils/actor.py
- ruff check miles/backends/megatron_utils/actor.py
- git diff --check


nanjiangwill:
Closing as superseded by the weight-update refactor merged in #2752 and #2754. Current main selects the transfer implementation through `get_weight_transfer_protocol()`, where each backend is imported only inside its selected branch. The P2P path also imports Mooncake’s `TransferEngine` only when creating that engine. Non-P2P runs therefore no longer load the P2P/Mooncake dependency chain, and rebasing this actor-level patch would restore an ownership boundary that no longer exists.

PR #2598, opened 2026-08-18
fix(metrics): propagate and correctly aggregate speculative decoding counters

nanjiangwill:
## Summary

- use the canonical SGLang speculative-decoding counters throughout Miles: `spec_num_correct_drafts`, `spec_num_proposed_drafts`, `spec_verify_ct`, and `completion_tokens`
- carry `Sample.SpecInfo` through the shared session v1/v2 samples wire
- compute acceptance metrics from summed counters instead of averaging per-request ratios
- for session v2, count every committed generation once even when returned leaves share prefixes or selection/truncation excludes a generated branch

## Why

The old Miles field names no longer matched the SGLang response contract, and averaging per-request ratios gives each request equal weight regardless of its number of drafts or verification steps. The session samples codec also dropped `SpecInfo` before it reached rollout metrics.

Session v2 needs one additional boundary: a tree can materialize multiple samples that share generation nodes. Summing leaf-local `SpecInfo` would count shared nodes more than once. The session server therefore publishes one aggregate over committed generation nodes, and rollout metrics deduplicate that carrier by session ID. Samples without a carrier continue to contribute their local `SpecInfo`, which supports batches containing per-sample generator overrides. This path is only enabled when speculative decoding is configured and a non-empty sample reply is returned.

Historical serialized samples using the old Miles field names remain readable; newly serialized data uses the canonical names.

## Validation

- Miles pre-commit hooks on all modified Python files
- `git diff --check`
- 224 passed, 1 skipped across rollout metrics, session v1/v2 samples, agentic v2, dashboard, type, codec, and mock-SGLang tests


guapisolo:
session server v1 / v2 not covered. my agent is fixing. will push to this pr

guapisolo:
I’m about to publish `843cf4ca3ee630a7bc092370a0806a4afe234c07` to this PR’s head. It carries `Sample.spec_info` through the shared session v1/v2 samples wire so speculative metrics do not decode as zero; the candidate-bound suite passed with 171 tests passed and 10 skipped.

guapisolo:
https://github.com/modal-projects/miles/pull/3 added

nanjiangwill:
merged

guapisolo (approved):
need code owner's review.

miles/ray/rollout/metrics.py
```diff
@@ -265,11 +266,23 @@ def _is_zero_std(samples: list[Sample]):
 def _compute_spec_metrics(args, all_samples: list[Sample]):
     if args.sglang_speculative_algorithm is None:
         return {}
-    num_samples = len(all_samples)
-    metrics = {}
-    metrics["spec_accept_rate"] = sum(sample.spec_info.spec_accept_rate for sample in all_samples) / num_samples
-    metrics["spec_accept_length"] = sum(sample.spec_info.spec_accept_length for sample in all_samples) / num_samples
-    return metrics
+    if args.use_session_server == "v2":
```
guapisolo: codex cmt

[P2] Preserve speculative counters from non-session training samples

With a global agentic generator and `use_session_server == "v2"`, a training sample can still select `single_turn.generate` through `generate_function_path`: the per-sample override in `generate_and_rm` also applies during training. These samples correctly populate `sample.spec_info` but do not produce a session carrier. Aggregating only carriers based on the global v2 setting therefore omits the ordinary-generation portion of a mixed training batch; when every sample in a batch uses ordinary generation, both metrics become zero.

I reproduced this through an HTTP mock engine and the production `generate_and_rm_group(evaluation=False)` → per-sample `single_turn.generate` → `log_rollout_data` path. The generated counters were correct=3, proposed=5, verify=2, completion=5, so the expected values were `rollout/spec_accept_rate=0.6` and `rollout/spec_accept_length=2.5`; both logged values were `0.0`. This affects training observability as well as evaluation metrics. The reproduction used mock inference, without a real SGLang GPU run.

Please deduplicate session carriers by `session_id` while also retaining `spec_info` from ordinary-generation samples without a carrier, so the global v2 setting does not exclude their valid counters.

nanjiangwill: Good catch. `use_session_server` selects the default generator, not the provenance of every sample, because `Sample.generate_function_path` can override it during training. Addressed in `ac647b0e0`: the aggregator now deduplicates carrier-bearing v2 samples by `session_id` and includes `sample.spec_info` for samples without a carrier. The updated test covers both a mixed batch and an all-ordinary batch under global v2 configuration.

guapisolo:
failed at tito checks, existing ci failure

guapisolo (approved):

yueming-yuan (approved):

PR #2601, opened 2026-08-18
fix(weight-sync): always resume rollout health monitoring

nanjiangwill:
## Summary

- resume rollout health monitoring after weight synchronization
- restore monitoring from `finally` for both Ray train-group implementations
- expose the existing resume operation through the rollout-manager actor API

## Root cause

Weight synchronization pauses rollout health monitoring to avoid checking engines while their weights are being updated. The pause was not paired with a resume, so monitoring remained disabled until another rollout or evaluation explicitly resumed it. If weight synchronization failed, it could remain disabled indefinitely.

## Impact

Health monitoring is restored after every weight-update attempt while preserving the original weight-sync exception.

## Validation

- `git diff --check`
- Attempted `uv run pytest tests/fast/ray/test_actor_group_shared_ppo.py tests/fast/ray/train/test_group.py -q`; collection could not start in the clean worktree because `torch` is not installed.

### DIFF
--- a/miles/ray/actor_group.py
+++ b/miles/ray/actor_group.py
@@ -126,8 +126,10 @@ class RayTrainGroup:
 
         info = await self.rollout_manager.get_updatable_engines_and_lock.remote()
         await self.rollout_manager.health_monitoring_pause.remote()
-
-        await self._broadcast("update_weights", info=info)
+        try:
+            await self._broadcast("update_weights", info=info)
+        finally:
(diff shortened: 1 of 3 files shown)


nanjiangwill:
Closing after reviewing the existing health-monitor lifecycle: Miles intentionally keeps monitoring paused until the next generate/eval boundary, so unconditional resume after weight sync is not generally correct (notably with rollout offload).

PR #2608, opened 2026-08-18
Refuse a second init in one trainer worker process

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The one-shot guard runs before distributed initialization, and trainer calls are serialized by the default concurrency group. No correctness or regression issue remains at deliver-2.

PR #2609, opened 2026-08-18
Fail a mismatched rpc result in the process that produced it

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue was found. Serialization mismatches now raise within the existing server-side exception boundary, and the target deliver retains this behavior.

PR #2610, opened 2026-08-18
Send the trainer's argparse namespace as a marked pickle hatch

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No production correctness or regression blocker was found. The tagged pickle codec and all trainer initialization annotations remain present at the target safe deliver.

PR #2611, opened 2026-08-18
Declare a method's concurrency group once for both wires

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the specified deliver-2 safe point.

PR #2612, opened 2026-08-18
Hand the driver a rollout data pack rather than a mapping

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the specified deliver-2 safe point.

PR #2613, opened 2026-08-18
Type the trainer surface for the wire

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains at the deliver-2 safe point. The intermediate RPC/Ray object-reference serialization gap is repaired before f5a18edf2aa9fade59d0cff05d865c310fd52e7a.

PR #2614, opened 2026-08-18
Cut the eval fleet along the process seam

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Under the requested safe-deliver criterion, no correctness blocker introduced by this op remains at f5a18edf2aa9fade59d0cff05d865c310fd52e7a. The isolated controller-unreachable regression is repaired by f20c32896f before that deliver.

PR #2615, opened 2026-08-18
Skip one eval point when the controller cannot be reached

fzyzcjy:
Part of #1837


fzyzcjy (commented):
Under the requested safe-deliver criteria, no blocking correctness or regression issue remains at deliver-2. The transient controller-handle lifetime concern is resolved before the target deliver.

PR #2616, opened 2026-08-18
Pick the worker wire independently of who launches it

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker remains at deliver-2. The target deliver contains the downstream worker wiring needed to honor the new Ray/RPC selection, making the standalone gap in this op transient.

PR #2617, opened 2026-08-18
Build a worker handle on the side that will call it

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker introduced by this op remains in the target deliver-2 state.

PR #2618, opened 2026-08-18
Serve a ray-launched worker over rpc

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at deliver-2; later-repaired issues are transient and excluded as requested.

PR #2619, opened 2026-08-18
Support RPC communication mode for Ray object store with cloudpickle

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression issue introduced by this op remains at the target safe deliver after excluding restored transient and test-only issues as requested.

PR #2620, opened 2026-08-18
Fence ray to the launcher and check both wires agree

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The patch only changes test and static-scan code, and unit-test issues are excluded from this review. No in-scope production correctness regression persists at deliver-2.

PR #2621, opened 2026-08-18
Refuse a ray handle that is not of the generation described

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The generation is consistently passed through Ray handle resolution and mismatched generations are rejected. The guard remains intact at the specified safe deliver, and no correctness or regression blocker was found.

PR #2622, opened 2026-08-18
Notice a ray worker that died without being stopped

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/ray_worker_manager.py
+++ b/miles/utils/workers/ray_worker_manager.py
@@ -21,6 +21,7 @@ from miles.utils.workers.backend_capability.base import BackendCapability, Defer
 from miles.utils.workers.backend_capability.ray import RayBackendCapability
 from miles.utils.workers.command_actor import CommandActor
 from miles.utils.workers.naming import compute_cell_id, compute_worker_name
+from miles.utils.workers.ray_worker_handle import RayWorkerHandle
 from miles.utils.workers.rpc.common.metadata import declared_concurrency_groups
 from miles.utils.workers.serving.serve_actor import ServeActor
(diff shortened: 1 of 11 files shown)


miles/utils/workers/ray_worker_manager.py
```diff
@@ -76,22 +79,25 @@ async def init(
...
-        cells = [cell for cell_id in cell_ids if (cell := self._find_cell(cell_id)).actors is None]
-        try:
-            await _gather_or_raise([c.launch_actors() for c in cells])
-            await _gather_or_raise([c.alloc_ports() for c in cells])
-            await _gather_or_raise([c.post_setup() for c in cells])
-        except Exception:
-            logger.error(f"Starting cells {[c.cell_id for c in cells]} failed, rolling back", exc_info=True)
-            await asyncio.gather(*[c.stop() for c in cells], return_exceptions=True)
-            raise
+        async with self._membership_lock:
```
fzyzcjy, line 86: [P1] Let stop_cells interrupt a stalled start

When a resumed actor remains pending scheduling or its constructor/RPC startup never answers, `start_cells` holds this manager-wide lock across the unbounded `alloc_ports` and `post_setup` awaits. Both `stop_cells` and the liveness teardown require the same lock, so they cannot kill the pending actor—or another cell to free resources—and one failed resume wedges every membership operation indefinitely; make startup bounded/cancellable or avoid holding the lock while waiting on workers.

fzyzcjy (commented):
The newly introduced membership lock is held over unbounded remote startup waits, preventing explicit or automatic teardown from recovering a stalled resume. The same blocker remains in deliver-2.

PR #2625, opened 2026-08-18
Namespace training metrics by an optional model id

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/metrics.py
+++ b/miles/ray/rollout/metrics.py
@@ -11,6 +11,7 @@ from miles.utils.metric_utils import (
     compute_statistics,
     dict_add_prefix,
     has_repetition,
+    namespace_metrics,
 )
 from miles.utils.tracking_utils import tracking
 from miles.utils.types import Sample
@@ -69,7 +70,9 @@ def log_eval_skip(rollout_id, args, reason: str):
     tracking.log(args, log_dict, step_key="eval/step")
 
 
-def log_rollout_data(rollout_id, args, samples, rollout_extra_metrics, rollout_time):
+def log_rollout_data(
+    rollout_id, args, samples, rollout_extra_metrics, rollout_time, trainer_model_id: str | None = None
+):
     if (x := args.custom_rollout_log_function_path) is not None:
         custom_log_func = load_function(x)
         if custom_log_func(rollout_id, args, samples, rollout_extra_metrics, rollout_time):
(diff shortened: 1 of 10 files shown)


fzyzcjy (commented):
Deliver-2 still leaves the namespace context incomplete in two supported paths: one-model configurations split producer namespaces, and per-model batch-size overrides emit conflicting values for one step key. The resulting metric names and axes are incorrect.

miles/ray/rollout/metrics.py
```diff
@@ -69,7 +70,9 @@ def log_eval_skip(rollout_id, args, reason: str):
     tracking.log(args, log_dict, step_key="eval/step")
 
 
-def log_rollout_data(rollout_id, args, samples, rollout_extra_metrics, rollout_time):
+def log_rollout_data(
+    rollout_id, args, samples, rollout_extra_metrics, rollout_time, trainer_model_id: str | None = None
```
fzyzcjy, line 74: [P2] Propagate the model ID for one-model configurations

At deliver-2, a supported one-model `--megatron-config` runs through `train.py`/`train_async.py`; `compute_trainer_args` sets the trainer copy's `trainer_model_id`, but those drivers still call `rollout_executor.get(rollout_id)` without this argument, and only `train_multi_policy.py` defines policy metric groups. This mixes `alpha/train/*` and `alpha/perf/*` with unprefixed rollout metrics, while WandB does not bind the namespaced trainer series to their namespaced step keys. Propagate the sole model ID and define its groups, or keep the whole path unprefixed.

miles/ray/rollout/metrics.py
```diff
@@ -87,9 +90,13 @@ def log_rollout_data(rollout_id, args, samples, rollout_extra_metrics, rollout_t
...
         )
     logger.info(f"perf {rollout_id}: {log_dict}")
-    step = compute_rollout_step(args, rollout_id)
-    log_dict["rollout/step"] = step
-    tracking.log(args, log_dict, step_key="rollout/step")
+    log_dict, step_key = namespace_metrics(
+        log_dict,
+        trainer_model_id=trainer_model_id,
+        step_name="rollout/step",
+        step=compute_rollout_step(args, rollout_id),
```
fzyzcjy, line 97: [P2] Use the selected policy's batch size for rollout steps

At deliver-2, `PER_POLICY_ARGS` allows `global_batch_size` to differ by model, but this computes the rollout-side step from the shared executor's base `args`. With `--wandb-always-use-train-step`, a policy whose GBS is half the base value publishes these points at N while its trainer-side rollout/perf logs publish the same `<id>/rollout/step` at 2N, interleaving one namespace on incompatible axes. Compute or pass the step from the selected policy's resolved arguments.

PR #2626, opened 2026-08-18
Add a megatron config naming every trainer

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/megatron_config.py
+++ b/miles/backends/megatron_utils/megatron_config.py
@@ -1,0 +1,98 @@
+import logging
+from typing import Any
+
+import pydantic
+import yaml
+
+from miles.utils.file_arg_utils import resolve_file_arg
+from miles.utils.pydantic_utils import FrozenStrictBaseModel
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------- constants -----------------------------
+
+ACTOR_ROLE = "actor"
+CRITIC_ROLE = "critic"
+DEFAULT_MODEL_ROLE = ACTOR_ROLE
+
+
+# ---------------------------- raw config -----------------------------
+
+
+class _RawMegatronTrainerConfig(FrozenStrictBaseModel):
+    model_id: str
+    role: str = DEFAULT_MODEL_ROLE
+    trainer_id: str | None = None
+    overrides: dict[str, Any] = {}
+
+
+class _RawMegatronConfig(FrozenStrictBaseModel):
+    trainers: list[_RawMegatronTrainerConfig] = pydantic.Field(
+        validation_alias=pydantic.AliasChoices("trainers", "megatron")
+    )
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


miles/backends/megatron_utils/megatron_config.py
```diff
@@ -0,0 +1,98 @@
...
+    role: str
+    overrides: dict[str, Any]
+
+    @classmethod
+    def resolve(cls, raw: _RawMegatronTrainerConfig) -> "MegatronTrainerConfig":
+        return cls(
+            trainer_id=raw.trainer_id if raw.trainer_id is not None else f"{raw.model_id}-{raw.role}",
+            model_id=raw.model_id,
+            role=raw.role,
+            overrides=raw.overrides,
```
fzyzcjy, line 55: [P1] Recompute derived arguments after trainer overrides

When a trainer changes a topology field such as `tensor_model_parallel_size`, deliver-2 applies this mapping only after the base `Namespace` has passed `set_default_megatron_args()` and Megatron validation. On eight GPUs, a base TP of 1 leaves `data_parallel_size=8`; overriding TP to 2 creates DP=4 process groups, but `init_num_microbatches_calculator()` still receives 8, silently halving the effective global batch. `_apply_critical_derived_overrides()` only repairs the tokenizer, so this persists; recompute and revalidate dependent fields per trainer or reject topology overrides, per [.claude/rules/general-code-style.md:22-28](.claude/rules/general-code-style.md#L22-L28).

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -0,0 +1,98 @@
...
+class MegatronTrainerConfig(FrozenStrictBaseModel):
+    trainer_id: str
+    model_id: str | None
+    role: str
+    overrides: dict[str, Any]
+
+    @classmethod
+    def resolve(cls, raw: _RawMegatronTrainerConfig) -> "MegatronTrainerConfig":
+        return cls(
+            trainer_id=raw.trainer_id if raw.trainer_id is not None else f"{raw.model_id}-{raw.role}",
```
fzyzcjy, line 52: [P2] Enforce the trainer pool-name budget

When the default-role `model_id` is 16 characters, this derivation produces a 22-character trainer ID. In deliver-2, `compute_trainer_controller_pool_id()` expands it to 41 characters, while Helm `PoolEntry.name` is capped at `POOL_NAME_MAX_LENGTH=40`, so the launch fails before a trainer starts. The later ID validator checks only DNS syntax, so this persists; validate resolved trainer IDs against the controller pool budget, including the longer synthesized `-critic` suffix.

fzyzcjy (commented):
Two configuration contracts remain unsafe at deliver-2: per-trainer overlays can leave Megatron-derived state inconsistent, and accepted identifiers can exceed the deployment pool-name budget. These can respectively corrupt training batch semantics or prevent launch.

PR #2627, opened 2026-08-18
Coerce a mapping into typed argument values

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/argv_utils.py
+++ b/miles/utils/workers/argv_utils.py
@@ -2,7 +2,7 @@ import argparse
 import dataclasses
 import json
 from collections.abc import Callable, Mapping, Sequence
-from typing import TypeVar
+from typing import Any, NamedTuple, TypeVar
 
 from miles.utils.pydantic_utils import FrozenStrictBaseModel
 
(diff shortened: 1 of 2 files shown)


miles/utils/workers/argv_utils.py
```diff
@@ -170,3 +173,88 @@ def _boolean_option_string(action: argparse.Action, *, value: bool) -> str:
...
+    values: Mapping[str, Any], *, parser: argparse.ArgumentParser, allowed_names: frozenset[str], context: str
+) -> dict[str, Any]:
+    dest_of_option_name = _compute_dest_of_option_names(parser)
+    arg_specs = _compute_arg_specs(parser)
+    allowed_dests = frozenset(dest_of_option_name.get(name, name) for name in allowed_names)
+    return {
+        (dest := dest_of_option_name.get(name, name)): _coerce_value(
+            value, dest=dest, spec=arg_specs.get(dest), allowed_dests=allowed_dests, context=context
+        )
+        for name, value in values.items()
```
fzyzcjy, line 212: [P2] Reject aliases that overwrite the same destination

When `values` contains both a destination and an accepted option alias, such as `add_bias_linear` and `disable_bias_linear`, both resolve to `add_bias_linear`, so this comprehension silently keeps whichever value appears last. This remains a blocker in deliver-2, where `--megatron-config` overrides pass through this unchanged helper and Megatron exposes that alias; reject duplicate resolved destinations rather than silently applying an order-dependent model setting.

fzyzcjy (commented):
The coercion path silently collapses distinct accepted keys onto one argparse destination. Deliver-2 can therefore apply an order-dependent trainer override instead of rejecting an ambiguous configuration.

PR #2628, opened 2026-08-18
Validate the trainers a megatron config declares

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/megatron_config.py
+++ b/miles/backends/megatron_utils/megatron_config.py
@@ -1,11 +1,14 @@
+import argparse
 import logging
-from typing import Any
+import re
+from typing import Any, Literal
 
 import pydantic
 import yaml
 
 from miles.utils.file_arg_utils import resolve_file_arg
 from miles.utils.pydantic_utils import FrozenStrictBaseModel
+from miles.utils.workers.argv_utils import coerce_dict_to_args
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


miles/backends/megatron_utils/megatron_config.py
```diff
@@ -15,14 +18,67 @@
 ACTOR_ROLE = "actor"
 CRITIC_ROLE = "critic"
 DEFAULT_MODEL_ROLE = ACTOR_ROLE
+TrainerRole = Literal["actor", "critic"]
+MODEL_ID_PATTERN = re.compile(r"\A[a-z0-9]([a-z0-9-]*[a-z0-9])?\Z")
+
+PER_POLICY_ARGS: frozenset[str] = frozenset(
```
fzyzcjy, line 25: [P1] Admit every model-definition override

When a policy selects a model definition such as `qwen2.5-7B` or `qwen3-30B-A3B`, deliver-2's `compute_model_args_overrides()` emits destinations such as `untie_embeddings_and_output_weights`, `num_experts`, and `moe_layer_freq`; `_resolve_overrides()` rejects them because this set omits them, while the coercer also rejects list-valued model flags. The multi-policy path therefore cannot carry those models' complete architectures, so the allowlist and coercion must cover the full model-definition surface.

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -15,14 +18,67 @@
...
+        "ffn_hidden_size",
+        "num_attention_heads",
+        "group_query_attention",
+        "num_query_groups",
+        "kv_channels",
+        "add_qkv_bias",
+        "qk_layernorm",
+        "swiglu",
+        "normalization",
+        "layernorm_epsilon",
```
fzyzcjy, line 41: [P2] Accept the legacy normalization destination

When `DEPRECATED_MEGATRON_COMPATIBLE=1` runs with the legacy parser, `--norm-epsilon` resolves to `norm_epsilon`—the compatibility branch in `hf_validate_args` reads that exact attribute—but this set permits only `layernorm_epsilon`. Consequently generated policy configs based on the usual model scripts are rejected by `_resolve_overrides`; accept or normalize both parser destinations. Deliver-2 leaves this unchanged.

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -96,3 +168,70 @@ def _resolve_raw_megatron_config(value: str | None) -> "_RawMegatronConfig | Non
...
+def _assert_valid_trainer_ids(trainer_ids: list[str]) -> None:
+    assert len(set(trainer_ids)) == len(trainer_ids), (
+        f"--megatron-config trainer ids must be unique, got {trainer_ids}; a trainer id addresses one trainer "
+        f"controller and its engine pool, so two entries sharing it would land in the same pool"
+    )
+    _assert_valid_ids(trainer_ids, kind="trainer")
+
+
+def _assert_valid_ids(ids: list[str], *, kind: str) -> None:
+    bad_ids = [identifier for identifier in ids if MODEL_ID_PATTERN.match(identifier) is None]
```
fzyzcjy, line 202: [P2] Enforce the composed pool-name length limit

With the default trainer ID, a 16-character regex-valid model ID passes this check, but deliver-2 forms `trainer-controller-<model_id>-actor`, which is 41 characters while `PoolEntry.name` is capped at 40. Kubernetes-path launches therefore fail later during values construction instead of rejecting the config here; validate both model and trainer IDs against the composed pool-name budget. Deliver-2 leaves this validator unchanged.

fzyzcjy (commented):
The allowlist prevents supported model definitions and the legacy parser from resolving, while ID validation accepts names that the Kubernetes values model rejects. All three production issues remain in deliver-2.

PR #2629, opened 2026-08-18
Extract the checkpoint load fallback into a function

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The helper preserves the original fallback predicate, mutations, and call position. The target safe deliver retains this behavior, and no correctness regression was found.

PR #2630, opened 2026-08-18
Compute every trainer's arguments through one function

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/megatron_config.py
+++ b/miles/backends/megatron_utils/megatron_config.py
@@ -1,7 +1,9 @@
 import argparse
+import copy
 import logging
 import os
 import re
+from argparse import Namespace
 from typing import Any, Literal
 
 import pydantic
@@ -162,7 +164,23 @@ def _compute_trainers(args) -> list[MegatronTrainerConfig]:
         trainers = [MegatronTrainerConfig.resolve(raw=t) for t in raw.trainers]
         assert trainers, "--megatron-config must declare at least one trainer"
 
+    if getattr(args, "use_critic", False) and not any(trainer.role == CRITIC_ROLE for trainer in trainers):
+        assert (
+            len({trainer.model_id for trainer in trainers}) == 1
+        ), "training several policy models does not support --use-critic"
+        trainers = [*trainers, _compute_critic_trainer(policy=trainers[0])]
+
     return trainers
+
+
+def _compute_critic_trainer(*, policy: MegatronTrainerConfig) -> MegatronTrainerConfig:
+    model_id = policy.model_id
+    return MegatronTrainerConfig(
+        trainer_id=CRITIC_ROLE if model_id is None else f"{model_id}-{CRITIC_ROLE}",
+        model_id=model_id,
+        role=CRITIC_ROLE,
+        overrides=dict(policy.overrides),
+    )
 
 
 def _resolve_raw_megatron_config(value: str | None) -> "_RawMegatronConfig | None":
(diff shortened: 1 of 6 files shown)


miles/backends/megatron_utils/megatron_config.py
```diff
@@ -181,6 +199,22 @@ def _assert_no_declared_critic(raw: "_RawMegatronConfig") -> None:
...
+
+def compute_trainer_args(args: Namespace, trainer: MegatronTrainerConfig) -> Namespace:
+    ans = copy.deepcopy(args)
+
+    for key, value in trainer.overrides.items():
+        assert hasattr(ans, key), (
+            f"--megatron-config trainer {trainer.trainer_id!r} overrides {key!r}, which this run's argument "
+            f"parser does not know"
+        )
+        setattr(ans, key, value)
```
fzyzcjy, line 295: [P1] Recompute dependent trainer arguments after overlays

When a permitted overlay changes an input with parse-time dependents, this loop replaces only the source value after Miles and Megatron derivation and validation have run. For example, a `vocab_size` override leaves `padded_vocab_size` tied to the base model even though `model_provider` sizes embeddings from it, while TP/PP overrides leave `data_parallel_size` and VPP tied to the base topology. Recompute and validate dependent fields from the effective namespace, or reject unsupported overrides, per `.claude/rules/general-code-style.md:22-28`.

miles/ray/placement_group.py
```diff
@@ -137,7 +137,7 @@ async def create_training_models(
     capability = get_backend_capability(args)
 
     actor_model = create_trainer_controller_handle(capability=capability, role="actor")
-    actor_start_rollout_ids = await actor_model.init(args)
+    actor_start_rollout_ids = await actor_model.init(compute_actor_args(args))
```
fzyzcjy, line 140: [P1] Give rollout preprocessing the effective trainer arguments

When a config overrides an option also consumed during rollout preprocessing—such as `global_batch_size`, dynamic-batch settings, or `advantage_estimator`—only the actor controller receives the effective namespace here. `RolloutExecutor` still trims, normalizes rewards, and builds the DP/microbatch schedule with base `self.args`, so it can schedule the old step size or normalize for the wrong algorithm. Register each trainer's effective arguments with the rollout side, or reject these overrides; `.claude/rules/general-code-style.md:22-28` requires one coherent override point.

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -162,9 +164,25 @@ def _compute_trainers(args) -> list[MegatronTrainerConfig]:
         trainers = [MegatronTrainerConfig.resolve(raw=t) for t in raw.trainers]
         assert trainers, "--megatron-config must declare at least one trainer"
 
+    if getattr(args, "use_critic", False) and not any(trainer.role == CRITIC_ROLE for trainer in trainers):
```
fzyzcjy, line 243: [P2] Reject PPO overlays when no critic is provisioned

When the base estimator is non-PPO but a permitted overlay sets `advantage_estimator: ppo`, this condition reads base `use_critic=False`, so it creates no critic and the overlaid actor retains the stale false flag. Its first batch nevertheless enters PPO and calls `get_advantages_and_returns_batch` with `values=None`, failing at `len(values_list)`. Reject this overlay or derive both the flag and trainer set from the effective estimator, as required by `.claude/rules/general-code-style.md:22-28`.

fzyzcjy (commented):
The patch leaves effective trainer arguments internally inconsistent and out of sync with rollout and critic orchestration for explicitly allowed overrides. These defects remain at safe deliver f5a18edf2aa9fade59d0cff05d865c310fd52e7a, so they are blockers rather than transient chain issues.

PR #2631, opened 2026-08-18
Absorb the critic argument overrides into the trainer config

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The critic checkpoint, schedule, and neutralization overrides preserve the effective initialization behavior while moving it into the trainer configuration. No correctness regression persists at the target safe deliver.

PR #2632, opened 2026-08-18
Support several policy models in compute_trainer_args

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/backends/megatron_utils/megatron_config.py
+++ b/miles/backends/megatron_utils/megatron_config.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 from argparse import Namespace
+from pathlib import Path
 from typing import Any, Literal
 
 import pydantic
@@ -22,6 +23,7 @@ ACTOR_ROLE = "actor"
 CRITIC_ROLE = "critic"
 DEFAULT_MODEL_ROLE = ACTOR_ROLE
 TrainerRole = Literal["actor", "critic"]
+TRAINER_CHECKPOINT_DIRNAME = "trainers"
 MODEL_ID_PATTERN = re.compile(r"\A[a-z0-9]([a-z0-9-]*[a-z0-9])?\Z")
 
 PER_POLICY_ARGS: frozenset[str] = frozenset(
@@ -215,7 +217,9 @@ def _assert_no_declared_critic(raw: "_RawMegatronConfig") -> None:
 
 
 def compute_trainer_args(args: Namespace, trainer: MegatronTrainerConfig) -> Namespace:
+    megatron_config = resolve_megatron_config(args)
     ans = copy.deepcopy(args)
+    ans.trainer_model_id = trainer.model_id if megatron_config.is_multi_policy else None
 
     for key, value in trainer.overrides.items():
         assert hasattr(ans, key), (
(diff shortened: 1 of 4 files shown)


miles/backends/megatron_utils/megatron_config.py
```diff
@@ -224,13 +227,35 @@ def compute_trainer_args(args: Namespace, trainer: MegatronTrainerConfig) -> Nam
         )
         setattr(ans, key, value)
 
+    _apply_critical_derived_overrides(ans, base=args, trainer=trainer)
+
+    if trainer.model_id is not None:
+        ans.save = _compute_trainer_checkpoint_dir(base_dir=ans.save, trainer_id=trainer.trainer_id)
+        ans.load = _compute_trainer_checkpoint_dir(base_dir=ans.load, trainer_id=trainer.trainer_id)
```
fzyzcjy: [P1] Preserve a root cursor for event-log rollback

When a `--megatron-config` run resumes after producing events beyond its last checkpoint, these lines put every model tracker at `<root>/trainers/<trainer_id>/latest_checkpointed_iteration.txt`, but `event_logger.checkpoint.snapshot()` still stores the rollback snapshot under `<root>/iter_N` and `restore()` reads `<root>/latest_checkpointed_iteration.txt`. No code writes that root tracker, including at the target deliver, so restore returns early and stale post-checkpoint events remain alongside replayed events, invalidating event-derived analysis and recovery state. Write a global cursor or restore against the trainer/global checkpoint state.

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -224,13 +227,35 @@ def compute_trainer_args(args: Namespace, trainer: MegatronTrainerConfig) -> Nam
...
+        ans.save_hf = _compute_trainer_checkpoint_dir(base_dir=ans.save_hf, trainer_id=trainer.trainer_id)
```
fzyzcjy: [P1] Route snapshot eval to the namespaced HF export

When `--megatron-config` names a single trainer and snapshot evaluation reuses periodic `--save-hf` without `--eval-hf-dir`, the worker writes `.complete` under `<base.format(...)>/trainers/<trainer_id>` because of this rewrite, while `EvalDispatcher.dispatch()` probes the unchanged base `args.save_hf.format(...)`. The marker is therefore never found and every scheduled checkpoint evaluation is silently reported as `ckpt_missing`; pass the resolved trainer path to the dispatcher.

miles/backends/megatron_utils/megatron_config.py
```diff
@@ -224,13 +227,35 @@ def compute_trainer_args(args: Namespace, trainer: MegatronTrainerConfig) -> Nam
...
+    if args.megatron_config is not None:
+        resolve_args_checkpoint_load(ans)
+
     return ans
 
 
+def _apply_critical_derived_overrides(ans: Namespace, *, base: Namespace, trainer: MegatronTrainerConfig) -> None:
+    # TODO: most derived defaults are still computed from the base args; revisit after the arguments refactor
+    if "hf_checkpoint" in trainer.overrides and base.tokenizer_model == base.hf_checkpoint:
+        ans.tokenizer_model = ans.hf_checkpoint
```
fzyzcjy, line 340: [P2] Preserve explicit tokenizer choices equal to the base checkpoint

When the caller explicitly passes `--tokenizer-model` with exactly the same value as the base `--hf-checkpoint` to share that tokenizer across policies, this equality test mistakes the explicit choice for the parse-time default and rewrites it to each trainer's `hf_checkpoint`. A policy can then tokenize with an unintended vocabulary despite an explicit CLI setting; track whether the default was applied instead of inferring provenance from value equality.

fzyzcjy (commented):
The per-trainer path rewrites are not propagated to existing global checkpoint consumers, breaking snapshot evaluation and event-log rollback for configuration-backed runs. The tokenizer default detection also overwrites a valid explicit choice.

PR #2633, opened 2026-08-18
Enumerate trainer controllers by trainer id

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/specs/train.py
+++ b/miles/ray/specs/train.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from miles.backends.megatron_utils.megatron_config import (
     ACTOR_ROLE,
     CRITIC_ROLE,
+    MegatronTrainerConfig,
     compute_trainer_args,
     resolve_megatron_config,
 )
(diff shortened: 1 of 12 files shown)


miles/ray/specs/train.py
```diff
@@ -36,51 +37,53 @@
...
-    role: str,
+    config: MegatronTrainerConfig,
     with_ref: bool,
     with_opd_teacher: bool,
-    drives_inference: bool,
 ) -> ServeWorkerSpec:
+    trainer_id = config.trainer_id
     return ServeWorkerSpec(
-        name=compute_trainer_controller_pool_id(role),
+        name=compute_trainer_controller_pool_id(trainer_id),
```
fzyzcjy, line 91: [P2] Bound trainer IDs before constructing pool names

When a Kubernetes run uses an accepted `trainer_id` longer than 21 characters, this produces a controller spec name exceeding the 40-character `PoolEntry.name` limit, so Helm values construction fails before deployment. `_assert_valid_ids` accepts arbitrary-length DNS-label IDs, including default `<model_id>-actor` IDs derived from model IDs longer than 15 characters, and deliver-2 retains this construction; enforce a prefix-aware limit or map the ID to a bounded pool name.

fzyzcjy (commented):
The patch allows valid trainer IDs to generate pool names rejected by the Kubernetes launcher, and this defect remains in deliver-2. Intermediate multi-policy issues repaired by the target deliver are not reported.

PR #2634, opened 2026-08-18
Give every policy its own slice of the trainer placement group

fzyzcjy:
Part of #1837


fzyzcjy (commented):
No correctness or regression blocker attributable to this op remains at deliver-2. The temporary multi-policy integration gaps are resolved before the target deliver.

PR #2635, opened 2026-08-18
Refuse a resume where the actor and its critic disagree

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The change extends rollout consistency validation to both the actor and critic without changing valid start-rollout behavior. The target deliver preserves this invariant, and no correctness or regression blocker remains.

PR #2636, opened 2026-08-18
Build one training model per trainer id

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/placement_group.py
+++ b/miles/ray/placement_group.py
@@ -6,6 +6,7 @@ import ray
 from ray.util.placement_group import PlacementGroup, placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
+from miles.backends.megatron_utils.megatron_config import compute_trainer_args
 from miles.ray.rollout.router_manager import resolve_router_addrs, wait_session_server_ready
 from miles.ray.specs.inference import (
     SESSION_SERVER_POOL_ID,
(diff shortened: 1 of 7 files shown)


miles/ray/placement_group.py
```diff
@@ -138,31 +132,55 @@ def create_placement_groups(args) -> dict[str, PlacementGroupInfo]:
...
     args, rollout_executor: BaseWorkerHandle
 ) -> tuple[BaseWorkerHandle, BaseWorkerHandle | None]:
-    capability = get_backend_capability(args)
-
-    actor_model = create_trainer_controller_handle(capability=capability, trainer_id=ACTOR_ROLE)
-    actor_start_rollout_ids = await actor_model.init(compute_actor_args(args))
+    trainer_configs = compute_trainer_configs(args)
+    [actor_config] = [config for config in trainer_configs if config.role == ACTOR_ROLE]
+    actor_info = await create_training_model(
+        compute_trainer_args(args, actor_config), trainer_id=actor_config.trainer_id
```
fzyzcjy, line 158: [P2] Propagate trainer IDs to actor fault injection

When `--megatron-config` assigns a non-role trainer ID such as `alpha-actor` and `--ci-ft-test-actions` targets that trainer's real cell, this passes the ID only to the controller; `compute_trainer_args()` gives workers only `trainer_model_id`, while `MegatronTrainRayActor.init()` still derives its action cell from `compute_trainer_pool_id(role)`. It consequently compares against `trainer-engine-actor-*` rather than `trainer-engine-alpha-actor-*`, so `crash_before_allreduce` is silently skipped. Propagate `trainer_id` to the worker and use it for the cell identity (`.claude/rules/pool-cell-worker-names.md:10-16`); this mismatch remains at `f5a18edf2a`.

fzyzcjy (commented):
The configured trainer ID is not propagated into train-worker cell identity, leaving actor-level fault injection broken for non-role trainer IDs through the target safe deliver.

PR #2637, opened 2026-08-18
Key rollout state by the sample's policy model

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/ray/rollout/rollout_executor.py
+++ b/miles/ray/rollout/rollout_executor.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import time
+from collections import defaultdict
 from collections.abc import Sequence
 from typing import Any
 
@@ -63,9 +64,10 @@ class RolloutExecutor(NodeProbeMixin):
         configure_logger(args, source=SimpleProcessIdentity(component="rollout_executor"))
 
         self.args = args
-        # set by the training actor after each weight update
-        self.weight_version: int | None = None
-        self._rollouts_since_weight_version_publish = 0
+        # set by the training actor after each weight update, keyed by trainer model id (None for one policy)
+        self._weight_versions_of_model_id: dict[str | None, int] = {}
+        self._rollouts_since_publish_of_model_id: dict[str | None, int] = defaultdict(int)
+        self._train_parallel_configs_of_model_id: dict[str | None, dict[str, Any]] = {}
         self._router_providers = router_providers
         self._session_server_provider = session_server_provider
         self._inference_controller_provider = inference_controller_provider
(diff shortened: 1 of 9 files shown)


miles/ray/rollout/rollout_executor.py
```diff
@@ -160,7 +166,9 @@ async def get(self, rollout_id: int) -> RolloutDataPack:
         if self.args.delay_split_train_data_by_dp:
             data_ref = object_store.get_instance().put(value=data, value_spec=ROLLOUT_DATA_VALUE_SPEC)
         else:
-            data_ref = split_train_data_by_dp(self.args, data, self.train_parallel_config)
+            data_ref = split_train_data_by_dp(
+                self.args, data, self._train_parallel_configs_of_model_id[trainer_model_id]
```
fzyzcjy, line 170: [P1] Use each policy's rollout conversion arguments

When `--megatron-config` uses the supported per-policy overrides for `global_batch_size`, `micro_batch_size`, `max_tokens_per_gpu`, or `advantage_estimator`, this call selects the policy's parallel layout but still supplies the shared driver `self.args`; `postprocess_rollout_data` and `convert_samples_to_train_data` likewise use those shared values. For example, a GBS-128/MBS-2 policy can receive a precomputed GBS-256/MBS-1 schedule that its trainer consumes verbatim, silently ignoring its configuration. This is unchanged in safe deliver `f5a18edf2a`, so store/pass policy-specific rollout arguments with the layout or reject these divergent overrides.

miles/ray/rollout/rollout_executor.py
```diff
@@ -243,7 +251,11 @@ async def _get_rollout_data(self, rollout_id):
                 data = await asyncio.to_thread(
                     call_rollout_function,
                     self.generate_rollout,
-                    RolloutFnTrainInput(rollout_id=rollout_id, weight_version=self.weight_version),
+                    RolloutFnTrainInput(
+                        rollout_id=rollout_id,
+                        weight_version=self._weight_versions_of_model_id.get(trainer_model_id),
+                        trainer_model_id=trainer_model_id,
```
fzyzcjy, line 259: [P1] Provide generation state per policy checkpoint

When valid per-policy `hf_checkpoint` values point to checkpoints with different tokenizers or processors, this still invokes the one rollout object constructed from base args, and `FullyAsyncRolloutFn` consequently owns only one base `GenerateState`. The safe-deliver solver/verifier function calls `single_turn_generate` for both model routers, so the second policy's prompt is encoded with the base tokenizer/chat template before its IDs are sent to and trained by the second policy, corrupting the sequence whenever the vocabularies differ. Safe deliver `f5a18edf2a` has neither keyed generation states nor a compatibility check; provide states per policy or reject incompatible checkpoints.

fzyzcjy (commented):
The patch isolates versions and parallel layouts, but leaves policy-dependent generation state and rollout conversion arguments global. Supported heterogeneous checkpoints and batch configurations can therefore produce or train incorrect data, and both blockers remain in the designated safe deliver.

PR #2638, opened 2026-08-18
Give every policy model its own output queue

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/rollout/fully_async_data_buffer.py
+++ b/miles/rollout/fully_async_data_buffer.py
@@ -14,6 +14,7 @@ from argparse import Namespace
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 
+from miles.backends.megatron_utils.megatron_config import resolve_megatron_config
 from miles.rollout.filter_hub.base_types import MetricGatherer, call_dynamic_filter
 from miles.utils.misc import load_function
 from miles.utils.types import Sample
@@ -23,6 +24,9 @@ logger = logging.getLogger(__name__)
 # A finished group is list[Sample], or list[list[Sample]] when a generate function
 # returns multiple samples per trajectory (e.g. multi-agent).
 Group = list[Sample | list[Sample]]
+
+
+# =================================== shared ===================================
 
 
 def iter_samples(group: Group) -> Iterator[Sample]:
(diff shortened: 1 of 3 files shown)


miles/rollout/fully_async_data_buffer.py
```diff
@@ -190,3 +201,64 @@ def _staleness(group: Group, current_version: int | None) -> int | None:
...
+
+    def __init__(self, input: DataBufferConstructorInput):
+        self._inners: dict[str, DataBuffer] = {
+            model_id: DefaultDataBuffer(input) for model_id in resolve_megatron_config(input.args).model_ids
+        }
+
+    async def put(self, input: DataBufferInput) -> None:
+        # TODO: a full inner blocks the one producer for every policy; give each policy its own dispatcher
+        for trainer_model_id, entry in _split_by_trainer_model_id(input).items():
+            await self._inner_of(trainer_model_id).put(entry)
```
fzyzcjy, line 214: [P1] Dispatch policy queues without cross-policy blocking

When per-policy filters accept at different rates, a fast policy can enter the target deliver's save/eval `Parker` while another is still collecting a batch. This serial await lets the parked policy's bounded queue fill and then stops the only producer before it can route more completions to the lagging policy; that policy cannot reach the same barrier, so the run hangs until `PARK_TIMEOUT_SECONDS` and fails. `f5a18edf` retains this loop, so each inner needs an independent dispatcher.

miles/rollout/fully_async_rollout.py
```diff
@@ -197,7 +206,7 @@ async def _drain(self, input: RolloutFnTrainInput) -> RolloutFnTrainOutput:
         if self._sample_filter is not None:
             self._sample_filter(args, data)
 
-        return RolloutFnTrainOutput(samples=data, metrics=self._output.get_metrics())
+        return RolloutFnTrainOutput(samples=data, metrics=self._output.get_metrics(input.trainer_model_id))
```
fzyzcjy, line 209: [P1] Preserve the no-argument get_metrics plugin contract

When a single-policy run uses an existing `--custom-async-data-buffer-path` implementation with the pre-change `get_metrics(self)` signature, this new positional argument is always supplied as `None`, so the first completed step raises `TypeError`; multi-policy custom buffers with that interface fail likewise. The target deliver still makes this call, and changing a documented plugin contract without a compatibility path conflicts with `.claude/rules/general-code-style.md:41-43`; retain the no-argument path or add a backward-compatible context mechanism.

fzyzcjy (commented):
The per-policy queue composition can deadlock target-deliver synchronization when policy acceptance rates diverge. It also breaks the established custom-buffer metrics interface, including for otherwise unchanged single-policy runs.

PR #2639, opened 2026-08-18
Choose every policy's data buffer class from the command line

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/rollout/fully_async_data_buffer.py
+++ b/miles/rollout/fully_async_data_buffer.py
@@ -10,8 +10,8 @@ Every group-level decision lives here — what to keep, what to hand to
 import asyncio
 import logging
 from abc import ABC, abstractmethod
-from argparse import Namespace
-from collections.abc import Callable, Iterator
+from argparse import ArgumentParser, Namespace
+from collections.abc import Callable, Iterable, Iterator
 from dataclasses import dataclass
 
 from miles.backends.megatron_utils.megatron_config import resolve_megatron_config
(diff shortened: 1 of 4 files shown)


miles/rollout/fully_async_data_buffer.py
```diff
@@ -25,6 +25,25 @@
 # returns multiple samples per trajectory (e.g. multi-agent).
 Group = list[Sample | list[Sample]]
 
+DATA_BUFFER_PATH_PER_MODEL_FLAG = "--custom-async-data-buffer-path-per-model"
+
+
+def add_data_buffer_arguments(parser: ArgumentParser) -> None:
+    parser.add_argument(
+        DATA_BUFFER_PATH_PER_MODEL_FLAG,
```
fzyzcjy, line 34: [P2] Do not ignore mappings in single-policy runs

When a fully-async command is single-policy (no config or a one-model `--megatron-config`), this argument is still registered and parsed, but `FullyAsyncRolloutFn.__call__` selects `DefaultDataBuffer`; only `DefaultMultiDataBuffer` reads the mapping. Consequently even malformed entries are accepted and the requested class is silently ignored, so either apply the mapping for the one model or reject the option in this mode.

miles/rollout/fully_async_data_buffer.py
```diff
@@ -214,8 +233,14 @@ class DefaultMultiDataBuffer(DataBuffer):
...
     def __init__(self, input: DataBufferConstructorInput):
+        paths = _parse_data_buffer_paths(input.args.custom_async_data_buffer_path_per_model)
+        model_ids = resolve_megatron_config(input.args).model_ids
+        assert not (unknown := sorted(set(paths) - set(model_ids))), (
+            f"{DATA_BUFFER_PATH_PER_MODEL_FLAG} names {unknown}, which train no policy of this run "
+            f"({sorted(model_ids)})"
+        )
         self._inners: dict[str, DataBuffer] = {
-            model_id: DefaultDataBuffer(input) for model_id in resolve_megatron_config(input.args).model_ids
+            model_id: (load_function(paths.get(model_id)) or DefaultDataBuffer)(input) for model_id in model_ids
```
fzyzcjy, line 233: [P2] Forward the policy ID to inner metric collection

When a class selected here uses the `trainer_model_id` parameter from the `DataBuffer.get_metrics` contract, the composite later invokes it as `get_metrics()` and drops the ID that `_drain` supplied; `get()` does forward the same context. Such a class works via the global multi-policy custom-buffer path but can fail or attribute metrics to the wrong policy when selected per model, so pass `trainer_model_id` through (`.claude/rules/general-code-style.md:14-15`).

fzyzcjy (commented):
The new option can be silently ignored in single-policy runs, and custom per-policy buffers do not receive the complete metrics context. Both defects remain in the target safe deliver.

PR #2640, opened 2026-08-18
Scope a weight update to one policy's engines

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/test_utils/comparisons/inference_engine_checksums.py
+++ b/miles/utils/test_utils/comparisons/inference_engine_checksums.py
@@ -4,6 +4,7 @@ from miles.utils.audit_utils.event_analyzer.rules import inference_engine_weight
 from miles.utils.audit_utils.event_analyzer.rules.checksum_compare import ChecksumMismatchIssue, compare_flat_dicts
 from miles.utils.audit_utils.event_logger.logger import read_events
 from miles.utils.audit_utils.event_logger.models import InferenceEngineWeightChecksumEvent
+from miles.utils.audit_utils.process_identity import TrainerControllerProcessIdentity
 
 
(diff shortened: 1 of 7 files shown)


miles/utils/test_utils/comparisons/inference_engine_checksums.py
```diff
@@ -19,41 +20,48 @@ def compare_inference_engine_checksums(baseline_dir: str, target_dir: str) -> No
...
+        key = (_compute_model_id(event), event.rollout_id)
+        assert key not in by_model_and_rollout, f"Duplicate InferenceEngineWeightChecksumEvent for {key}"
+        assert event.engine_checksums, f"No engine checksums for {key}"
+        by_model_and_rollout[key] = event.engine_checksums[0]
+    return by_model_and_rollout
+
+
+def _compute_model_id(event: InferenceEngineWeightChecksumEvent) -> str | None:
+    source = event.source
+    return source.model_id if isinstance(source, TrainerControllerProcessIdentity) else None
```
fzyzcjy, line 64: [P1] Carry the policy ID in checksum events

At deliver-2, `placement_group._maybe_log_inference_engine_weight_checksums` emits every policy's event from the orchestration logger, whose source is `SimpleProcessIdentity(component="main")`, so this expression returns `None` for every policy. When two policies emit the same rollout ID, `_checksums_by_model_and_rollout_id` encounters duplicate `(None, rollout_id)` keys and aborts instead of comparing them. Carry `trainer_model_id` explicitly in the event rather than recovering it from the process identity.

fzyzcjy (commented):
Deliver-2 still loses policy identity when checksum events are emitted from the orchestration process, causing multi-policy checksum comparison to fail for overlapping rollout IDs.

PR #2641, opened 2026-08-18
Control health monitoring per policy model

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The model-scoped health-monitoring state is wired consistently through rollout servers and cells. The target deliver passes the relevant model ID for multi-policy rollouts, while existing single-policy and fleet-wide operations preserve their previous behavior.

PR #2642, opened 2026-08-18
Train several policy models in one run

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/multi_policy/utils.py
+++ b/miles/utils/multi_policy/utils.py
@@ -1,0 +1,107 @@
+import logging
+from dataclasses import dataclass
+
+from miles.backends.megatron_utils.megatron_config import MegatronConfig, compute_trainer_args, resolve_megatron_config
+from miles.backends.sglang_utils.sglang_config import resolve_sglang_config
+from miles.ray.placement_group import create_training_model
+from miles.ray.specs.train import compute_trainer_configs
+from miles.utils.arguments import validate_async_off_policy_correction
+from miles.utils.tracking_utils.tracking import define_step_key_metric_group
+from miles.utils.workers.worker_handle import BaseWorkerHandle
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class TrainerInfo:
+    model_id: str
+    start_rollout_id: int
+    handle: BaseWorkerHandle
+
+
+async def create_trainers(args, *, rollout_executor: BaseWorkerHandle) -> dict[str, TrainerInfo]:
+    trainers: dict[str, TrainerInfo] = {}
+    for trainer_config in compute_trainer_configs(args):
+        model_id = trainer_config.model_id
+        assert model_id is not None, f"{trainer_config} carries no policy model id"
+        created = await create_training_model(
+            compute_trainer_args(args, trainer_config), trainer_id=trainer_config.trainer_id
+        )
+        assert model_id not in trainers, f"{trainer_config} shares its model id with an already created trainer"
+        trainers[model_id] = TrainerInfo(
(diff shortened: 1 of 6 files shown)


miles/utils/multi_policy/utils.py
```diff
@@ -0,0 +1,107 @@
...
+            compute_trainer_args(args, trainer_config), trainer_id=trainer_config.trainer_id
+        )
+        assert model_id not in trainers, f"{trainer_config} shares its model id with an already created trainer"
+        trainers[model_id] = TrainerInfo(
+            model_id=model_id, start_rollout_id=created.start_rollout_id, handle=created.handle
+        )
+
+    for model_id, trainer in trainers.items():
+        await rollout_executor.set_train_parallel_config(
+            await trainer.handle.get_train_parallel_config(), trainer_model_id=model_id
```
fzyzcjy, line 37: [P1] Propagate each policy's effective batch configuration

Only the trainer receives `compute_trainer_args(...)`; the rollout executor receives topology here and continues postprocessing and DP scheduling with the base `self.args`. With base `global_batch_size=64` and one policy overriding it to 32, that policy's 64 samples are scheduled as one 64-sample step instead of two 32-sample steps, silently ignoring the supported override. This base-args path remains in target `f5a18edf2a`, so pass effective rollout settings per `model_id` or reject these overrides.

miles/utils/multi_policy/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    assert megatron_config.is_multi_policy, (
+        f"train_multi_policy.py trains several policies, but --megatron-config names "
+        f"only {megatron_config.model_ids}; run train_async.py instead"
+    )
+    assert not args.colocate, "multi policy training does not support --colocate"
+    assert args.fully_async, (
+        f"multi policy training {megatron_config.model_ids} is only supported for --fully-async: every other rollout "
+        f"mode drives one policy per rollout round"
+    )
+    assert not args.use_critic, "multi policy training does not support --use-critic"
```
fzyzcjy, line 55: [P1] Reject PPO after applying each policy override

When the base estimator is GRPO but one trainer overrides `advantage_estimator` to `ppo`, global `args.use_critic` remains false and this assertion passes. The trainer is then created for PPO without a critic, so its first advantage computation receives no values and fails. Target `f5a18edf2a` still checks only the base flag; validate each effective trainer configuration or prohibit PPO overrides.

train_multi_policy.py
```diff
@@ -0,0 +1,107 @@
...
+    megatron_config = resolve_megatron_config(args)
+    validate_multi_policy_args(args, megatron_config=megatron_config)
+    configure_logger(args, source=SimpleProcessIdentity(component="main"))
+    maybe_start_periodic_pyspy_dump()
+    init_tracking(args)
+    define_policy_metric_groups(megatron_config)
+    _worker_manager = launch_worker_manager(args)
+    object_store.init_instance(args, contribute_segment=False)
+
+    inference_controller, rollout_executor, num_rollout_per_epoch = await create_rollout_components(args)
```
fzyzcjy, line 37: [P2] Reject the stock generator for multi-policy runs

With the normal data source and no custom multi-model generation hook, the rollout created here uses stock single-turn generation, which calls only the default router and leaves ordinary samples' `trainer_model_id` as `None`. `DefaultMultiDataBuffer` then asserts on the first produced group, although validation accepted the configuration and launched all components. Target `f5a18edf2a` retains this path; fail fast or provide a default dispatcher that routes and stamps every policy sample.

fzyzcjy (commented):
The new multi-policy path accepts configurations that schedule policies with base batch settings, instantiate PPO without a critic, or fail immediately with the stock generator. All three defects remain in the target safe deliver, so they are blockers rather than transient issues.

PR #2643, opened 2026-08-18
Coordinate a global checkpoint across policies

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/multi_policy/utils.py
+++ b/miles/utils/multi_policy/utils.py
@@ -1,11 +1,13 @@
 import logging
 from dataclasses import dataclass
+from pathlib import Path
 
 from miles.backends.megatron_utils.megatron_config import MegatronConfig, compute_trainer_args, resolve_megatron_config
 from miles.backends.sglang_utils.sglang_config import resolve_sglang_config
 from miles.ray.placement_group import create_training_model
 from miles.ray.specs.train import compute_trainer_configs
 from miles.utils.arguments import validate_async_off_policy_correction
+from miles.utils.multi_policy.checkpoint_state import MultiPolicyCheckpointState
 from miles.utils.tracking_utils.tracking import define_step_key_metric_group
 from miles.utils.workers.worker_handle import BaseWorkerHandle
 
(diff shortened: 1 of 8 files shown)


miles/utils/multi_policy/utils.py
```diff
@@ -37,11 +39,61 @@ async def create_trainers(args, *, rollout_executor: BaseWorkerHandle) -> dict[s
...
+    await rollout_executor.load(leader_rollout_id)
 
     return trainers
 
 
+def _assert_global_rollout_state_exists(args, *, leader_rollout_id: int) -> None:
+    if leader_rollout_id < 0 or not args.rollout_global_dataset or args.load is None:
+        return
+
+    path = Path(args.load) / "rollout" / f"global_dataset_state_dict_{leader_rollout_id}.pt"
```
fzyzcjy, line 53: [P2] Delegate validation to the configured data source

When a multi-policy run uses a custom `--data-source-path` whose required `save()`/`load()` implementation stores state under another filename or backend, this hard-coded check raises before `rollout_executor.load()` can perform the valid restore. Fully-async multi-policy training requires `rollout_global_dataset`, so the source cannot bypass this check; delegate validation to the configured data-source contract rather than requiring the built-in `.pt` path. This preserves the established framework contract required by [general-code-style.md:14-15](.claude/rules/general-code-style.md#L14-L15).

miles/utils/multi_policy/utils.py
```diff
@@ -37,11 +39,61 @@ async def create_trainers(args, *, rollout_executor: BaseWorkerHandle) -> dict[s
...
+            f"policies that already trained on them"
+        )
+        return
+
+    state_dir = args.load or args.save
+    assert state_dir is not None, (
+        f"the leader policy {leader_model_id!r} restored rollout {leader_rollout_id} without --load or "
+        f"--save, so where the other policies stood cannot be read back"
+    )
+    state = MultiPolicyCheckpointState.load(Path(state_dir), leader_rollout_id=leader_rollout_id)
```
fzyzcjy, line 76: [P2] Restore each policy at its recorded iteration

When `--ckpt-step` selects an older global checkpoint and a follower stood at another rollout, every trainer has already received the same base `ckpt_step` before this mapping is loaded. The follower therefore attempts the leader's iteration rather than its recorded one, causing initialization or the equality check to abort, so historical asynchronous checkpoints cannot be restored. Load this record before trainer initialization and apply its per-policy iterations, or explicitly reject the existing selector as required for intentional public APIs by [general-code-style.md:41-43](.claude/rules/general-code-style.md#L41-L43).

fzyzcjy (commented):
The implementation rejects valid custom-data-source restores and cannot restore historical asynchronous checkpoints selected with `--ckpt-step`. Both issues remain in deliver-2.

PR #2644, opened 2026-08-18
Add the capabilities the multi policy example needs

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/workers/argv_utils.py
+++ b/miles/utils/workers/argv_utils.py
@@ -1,6 +1,7 @@
 import argparse
 import dataclasses
 import json
+import shlex
 from collections.abc import Callable, Mapping, Sequence
 from typing import Any, NamedTuple, TypeVar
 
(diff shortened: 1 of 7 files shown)


miles/utils/workers/argv_utils.py
```diff
@@ -178,6 +179,21 @@ def _boolean_option_string(action: argparse.Action, *, value: bool) -> str:
...
+    assert not unknown, f"the argument parser does not declare {unknown} of {text!r}"
+
+    action_by_option_string = parser._option_string_actions
+    dests = []
+    for token in tokens:
+        if not token.startswith("--"):
+            continue
+        assert token in action_by_option_string, f"the argument parser does not declare {token!r}"
+        dests.append(action_by_option_string[token].dest)
+    return {dest: getattr(namespace, dest) for dest in dests}
```
fzyzcjy, line 212: [P1] Reset omitted model options instead of inheriting the leader

When the leader's model script enables a `store_true` architecture option and another policy relies on that option's `False` parser default, this list contains no destination for the omitted option, so the second policy silently keeps the leader's value. The shipped pair demonstrates it: `qwen2.5-0.5B.py` sets `--add-qkv-bias`, `qwen3-0.6B.py` omits it, and `compute_trainer_args()` deep-copies the leader args before applying this map, leaving the verifier at `add_qkv_bias=True`; an args-built/non-bridge verifier then has the wrong QKV-bias parameterization for its checkpoint. The same return logic remains in safe deliver `f5a18edf`, so the generated map must include relevant defaults or explicit resets.

fzyzcjy (commented):
The required-argument parsing failure introduced here is fixed before safe deliver `f5a18edf` and is therefore transient. However, the generated per-policy override map still cannot reset leader-enabled options to another model's defaults, leaving a blocker in the safe deliver.

PR #2645, opened 2026-08-18
Add a solver-verifier example on gsm8k

fzyzcjy:
Part of #1837

### DIFF
--- a/examples/multi_policy/solver_verifier.py
+++ b/examples/multi_policy/solver_verifier.py
@@ -1,0 +1,134 @@
+# See tests/e2e/short/test_multi_policy_solver_verifier_gsm8k.py for an end-to-end run of this example.
+import dataclasses
+import re
+from enum import Enum
+
+from miles.backends.megatron_utils.megatron_config import resolve_megatron_config
+from miles.rollout.base_types import GenerateFnInput, GenerateFnOutput
+from miles.rollout.generate_hub.single_turn import generate as single_turn_generate
+from miles.utils.types import Sample
+
+_AGREE_MARKER = "AGREE"
+_WRONG_MARKER = "WRONG"
+_VERDICT_PREFIX = "VERDICT:"
+
+_VERIFIER_PROMPT_TEMPLATE = (
+    "Another model was asked to solve a math problem, and you must check its work.\n\n"
+    "Question:\n{question}\n\n"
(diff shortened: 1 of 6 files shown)


examples/multi_policy/solver_verifier.py
```diff
@@ -0,0 +1,132 @@
...
+    assert len(model_ids) == 2, (
+        f"examples/multi_policy/solver_verifier.py pairs one solver policy with one verifier policy, but "
+        f"--megatron-config names {model_ids}"
+    )
+    solver_model_id, verifier_model_id = model_ids
+
+    solver_output = await single_turn_generate(input, url=_compute_router_url(args, model_id=solver_model_id))
+    solver_sample = solver_output.samples
+    assert isinstance(solver_sample, Sample), f"{solver_sample=}"
+    assert solver_sample.status != Sample.Status.ABORTED
```
fzyzcjy, line 47: [P1] Return aborted solver samples instead of asserting

When SGLang returns `Sample.Status.ABORTED`, for example during an in-place generation pause, this assertion escapes the custom generate call. The fully asynchronous worker does not recover from that exception, so one recoverable aborted request terminates the producer and fails the run instead of letting the data buffer discard or retry it. Return the unrewarded solver sample stamped with its policy; this remains unfixed in deliver-2.

fzyzcjy (commented):
The patch can terminate the fully asynchronous rollout producer on a recoverable aborted generation and registers a CI label that the comment-command allowlist refuses. Both issues remain in the target safe deliver.

tests/ci/labels.py
```diff
@@ -32,6 +32,7 @@
     "ft-long": "Fault-tolerance trainer soak tests (random-crash survival, realistic-gsm8k convergence)",
     "weight-update": "Weight update tests",
     "fully-async": "Fully-async rollout tests",
+    "multi-policy": "Multi policy training tests (several policy models in one run)",
```
fzyzcjy, line 35: [P2] Expose the new CI label to the comment command

When a maintainer requests `/add-label run-ci-multi-policy`, `resolve_policy` rejects it because `.github/workflows/policies/comment-command-access.json` has an explicit `allowed_labels` list and this patch updates only `KNOWN_LABELS`. Add `run-ci-multi-policy` to that policy as well; deliver-2 still lacks it, so the suite cannot be selected through the repository's comment-command trigger.

PR #2646, opened 2026-08-18
Verify every trainer rank sees its policy's arguments

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The verifier's intermediate missing-rank coverage gap is fixed by 6e0c8ee6c9 before the target deliver. No correctness or regression blocker remains at deliver-2.

PR #2647, opened 2026-08-18
Cover every policy's trainers in trainer fault tolerance

fzyzcjy:
Part of #1837


fzyzcjy (commented):
The op correctly exposes every multi-policy trainer through the fault-tolerance API and scopes audit analysis by model. No correctness regression introduced by this commit remains in deliver-2.

PR #2648, opened 2026-08-18
Let the leader model decide when a multi policy run ends

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/async_utils.py
+++ b/miles/utils/async_utils.py
@@ -13,7 +13,7 @@ __all__ = [
     "run",
     "submit",
     "wait_futures",
-    "wait_cancelling_pending_on_first_exception",
+    "wait_cancelling_pending_on_first_completion",
     "eager_create_task",
 ]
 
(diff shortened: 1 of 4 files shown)


miles/utils/async_utils.py
```diff
@@ -81,8 +81,8 @@ def wait_futures(futures: Sequence[concurrent.futures.Future]) -> list[Any]:
     return results
 
 
-async def wait_cancelling_pending_on_first_exception(tasks: Sequence[asyncio.Task]) -> None:
-    _, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
+async def wait_cancelling_pending_on_first_completion(tasks: Sequence[asyncio.Task]) -> None:
+    _, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
fzyzcjy, line 88: [P1] Close split update windows before cancelling follower tasks

When the leader finishes while a follower is broadcasting weights, switching to `FIRST_COMPLETED` cancels that follower immediately. At deliver-2, `placement_group.update_weights()` may already have called `InferenceController.start_update_weights()` and detached its `ContextLock`; cancellation then skips the matching `end_update_weights()`, so `train_multi_policy()` hangs forever in `inference_controller.dispose()`. Make this window cancellation-safe or stop followers only at a boundary outside it.

fzyzcjy (commented):
The new first-completion shutdown can cancel a follower while it owns the inference-controller update window. This remains unresolved at the target deliver and can hang normal teardown indefinitely.

PR #2649, opened 2026-08-18
Teach the dashboard and the CI checkers the policy namespace

fzyzcjy:
Part of #1837

### DIFF
--- a/miles/utils/tracking_utils/ci_history.py
+++ b/miles/utils/tracking_utils/ci_history.py
@@ -59,6 +59,13 @@ TARGET_METRIC_KEYS: tuple[str, ...] = (
 RECORD_DIR_ENV = "MILES_CI_GATE_RECORD_DIR"
 
 
+def _compute_logged_key(metrics: dict[str, Any], key: str) -> str | None:
+    if key in metrics:
+        return key
+    suffix = f"/{key}"
+    return next((logged for logged in metrics if logged.endswith(suffix)), None)
+
+
 def _encode_value(value: float) -> float | str:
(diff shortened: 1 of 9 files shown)


fzyzcjy (commented):
The CI-history namespace handling collapses multiple policies into one series, making standard all-step gates fail on duplicate steps. The issue remains in the target deliver.

miles/utils/tracking_utils/ci_history.py
```diff
@@ -104,9 +111,9 @@ def log(self, metrics: dict[str, Any], step: int | None = None, **kwargs) -> Non
...
             captured: list[tuple[str, float]] = []
             for key in TARGET_METRIC_KEYS:
-                if key not in metrics:
+                if (logged_key := _compute_logged_key(metrics, key)) is None:
                     continue
-                value = metrics[key]
+                value = metrics[logged_key]
                 if not isinstance(value, (int, float)) or isinstance(value, bool):
                     message = f"CI history metric {key!r} must be int or float, got {type(value).__name__}"
                     logger.error("%s", message)
```
fzyzcjy, line 122: [P2] Preserve policy identity in CI history

In a fresh multi-policy CI run, this finds each policy's namespaced metric but records every value under the same canonical `key`. `tests/ci/ci_utils.py::_merge_attempt_records` then concatenates both policies' step-0 points, and `selection._select_all` rejects the duplicate step, so the gate reports an error instead of checking either policy. Preserve the policy in the recorded coordinate or explicitly select one policy; deliver-2 still has this behavior.

PR #2651, opened 2026-08-18
docs: fix stale paths, flags, env vars and metric names

yueming-yuan:
Audit of `docs/` against `main`. This PR only fixes claims that are **wrong** — commands that fail as written, symbols that no longer exist, and values that disagree with the code. Coverage gaps (flags that exist but are undocumented) are deliberately left out; see the follow-ups at the bottom.

## Commands that fail as written

| Page | Was | Now |
|---|---|---|
| `models/glm/glm4-5`, `models/glm/glm4-7-flash` | `python3 scripts/model_args.py <type>` | `python3 miles/utils/external_utils/model_args_utils.py <type>` — `scripts/model_args.py` does not exist; the other 17 conversion snippets already use the right path |
| `user-guide/cli-reference` | `on_policy_distillation` listed as an `--advantage-estimator` value | removed. `arguments.py:1422` has no such choice, and its own help says OPD is enabled *on top of any estimator* via `--use-opd` |
| `ci/02-docker-build`, `developer/versions` | `--variant rocm-mi300` / `rocm-mi350` | `rocm700-mi30x` / `rocm700-mi35x` / `rocm720-mi35x` — the real keys in `docker/build.py` and the `variant` choice list in `docker-build.yml`. `rocm700-mi35x` was missing from both tables |
| `models/thinkingmachines/inkling` §5.5 | `--optimizer-state-nvme-dir`, `--optimizer-state-nvme-chunk-mb` | neither flag exists. `scripts/run_inkling.py` passes `--offload-train-target disk --offload-train-disk-dir` when colocated and `--stream-optimizer-state-to-disk --offload-train-disk-dir --offload-train-disk-chunk-mb 256` under `--fully-async`. Also split the grad-reduce line, which is `--grad-reduce-in-bf16` vs `--accumulate-allreduce-grads-in-fp32` on the same condition |
| `cli-reference`, `training-backend`, `argument-groups`, `models/kimi/kimi-k2` | `--sglang-enable-ep-moe`, `--sglang-enable-deepep`, `--sglang-enable-deepep-moe`, `--sglang-enable-overlap-schedule` | all four are gone from the pinned sglang `ServerArgs`. Replaced with what the recipes actually pass: `--sglang-ep-size` (75 uses), `--sglang-moe-a2a-backend` (15), `--sglang-moe-runner-backend` (40), `--sglang-deepep-mode` (13) |
| `examples/infra_features/train_infer_mismatch_helper/README.md` | `--tis-mode` / `--tis-level` / `--tis-*-bound` / `--tis-batch-normalize` / `--rs-*` documented as CLI flags, plus a `--use-rollout-correction` flag | none of those are CLI flags. They are keys in `mis.yaml`, reached through `--custom-config-path` + `--custom-tis-function-path`; the only CLI toggles are `--use-tis` and `--use-rollout-logprobs`. `--use-rollout-correction` does not exist anywhere in the repo and was the header of an algorithm table |

## Symbols and paths that do not exist

- **`miles/ray/rollout.py`** is a package now. `advanced/fault-tolerance` (4×), `advanced/pd-disaggregation` and `examples/geo3k_vlm/README.md` now point at `rollout_manager.py`, `rollout_server.py`, `server_group.py` and `train_data_conversion.py` respectively. While there, `fault-tolerance` now also mentions that the rollout paths need `rollout` in `--ft-components`, not just `--use-fault-tolerance`.
- **`MILES_HACK_TRAIN_TORCH_DETERMINISTIC`** (`models/deepseek/deepseek-v4-flash`) does not exist. `scripts/run_deepseek_v4.py` sets five env vars unconditionally (`SGLANG_SKIP_CHECKPOINT_LOAD_CHECK`, `SGLANG_DSV4_FP4_EXPERTS`, `SGLANG_HEALTH_CHECK_TIMEOUT`, `SGLANG_DG_CACHE_DIR_PER_PROCESS`, `SGLANG_OPT_FP8_WO_A_GEMM`) and three more only under `--train-deterministic` (`NCCL_ALGO`, `NVTE_ALLOW_NONDETERMINISTIC_ALGO`, `CUBLAS_WORKSPACE_CONFIG`). The V4-Pro page got the same five.
- **`with_transformers_patch()` / `miles/utils/transformers_patch.py`** are both gone; the `deepseek_ref` → `deepseek_v4` handling is an in-place `config.json` rewrite in the launcher.
- **`eval/skipped_unhealthy`, `eval/skipped_pin_violation`** have no emitter. The built-in reasons are `busy`, `export_failed`, `ckpt_missing`, `crashed`, plus whatever a `CheckpointEvalFn` passes to `EvalSkip(reason)`.
- **The `mis_*` / `mismatch_*` metric names** in the rollout-correction example match nothing in `mis.py`. Replaced with the names actually emitted: `tis_weight_before_bound`, `tis_truncate_fraction`, `tis_clip_fraction_low|high`, `tis_mask_fraction_low|high`, `rs_mask_fraction_low|high`, `rs_catastrophic_token_fraction`, `rs_catastrophic_seq_fraction`, `is_ratio_mean_after_tis_rs`, `batch_norm_factor`, `is_ratio_{mean,min,max}_final`, and the unprefixed `training_log_ppl` / `rollout_log_ppl` / `kl` / `k3_kl` / `log_ppl_diff` / `ppl_ratio` / `chi2_token` / `chi2_seq`.
- **`miles/utils/wandb_utils.py`** → `miles/utils/tracking_utils/wandb_utils.py`.

## Wrong values and premises

- **`--update-weight-buffer-size` defaults to 512 MB**, not 1 GB (`arguments.py:773` is `512 * 1024**2`).
- **`user-guide/argument-groups` opened with "Miles launch scripts are bash arrays."** `scripts/` holds 34 Python launchers and zero `.sh` files, and the page contradicted `user-guide/launch-script`, which already describes the Python layout correctly. Rewritten around the `<group>_args` locals the launchers actually build, with the previously-missing `misc_args` group (35 launchers use it) documented and the architecture flags explained as coming from `scripts/models/<megatron_model_type>.py` via `execute_train`. The `<a id>` anchors are unchanged, so incoming links still resolve; `concepts`, `training-backend`, `user-guide/index`, `developer/contributor-guide` and `models/kimi/kimi-k2.5` were updated to the same names.
- **`developer/architecture`** sent new reward types to `miles/rollout/sglang_rollout.py`; the `rm_type` dispatch lives in `miles/rollout/rm_hub/__init__.py`, and `rm_hub/` was missing from the package tree.
- **`--rm-type` enumerations** (`cli-reference`, `customization`) omitted `gemma_math` — which `scripts/run_gemma_4_*.py` and `models/gemma/gemma-4` both use, so readers would take it for an invalid value — plus `deterministic_random` and the `boxed_` prefix modifier. `--rm-type` is also a free `str`, not an enum.
- **The canonical CI label list** (`ci/01-label`) listed 15 of the 19 entries in `KNOWN_LABELS`, omitting `eval`, `fully-async`, `miles-plugin` and `amd`. `ci/00-stage` already tells readers how `run-ci-amd` behaves.

## Verification

- `pre-commit run --all-files` clean (includes the examples→docs mirror check).
- `python scripts/tools/sync_example_docs.py --check` clean; the two generated pages were regenerated from their READMEs rather than edited directly.
- Re-ran the audit passes on the result: 0 broken internal links, 0 missing images, 0 nonexistent repo paths outside the pages that deliberately document unmerged PRs (`kimi-k3` → #1825, `qwen3-8` → #2488), 0 `--sglang-*` flags absent from the pinned `ServerArgs`.

## Deliberately not in this PR

Coverage and consistency work, worth separate PRs:

- `cli-reference` says it lists *every* Miles flag; it covers 91 of 344, and 133 registered flags appear nowhere in the docs — including all of MLflow (`--use-mlflow` + 4), TensorBoard (`--use-tensorboard` + 2) and Prometheus (`--prometheus-port` + 1). `user-guide/monitoring` documents only wandb as a result.
- `customization` says it lists *every* hook; six `load_function` hooks are missing (`--custom-async-data-buffer-path`, `--custom-update-weight-post-write-path`, `--generate-{execute-tool-function,tool-specs}-path`, `--session-sample-{picker,postprocessor}-path`).
- Ascend NPU has `docker/npu_patch/`, `scripts/run_qwen3_4b_npu.py` and `execute_train_npu()`, but no page mentions it; `index` and `installation` list only NVIDIA and AMD, and disagree on `MI350` vs `MI350X`.
- 24 links across 17 pages still point at the removed `/advanced/architecture-support` and survive only on the `docs.json` redirect, with link text that no longer names the destination.
- Several model and advanced pages still illustrate flags with uppercase `SGLANG_ARGS=(...)` shell blocks, a leftover of the same bash-array framing this PR removed from `argument-groups`.
- `low-precision` writes `GroupLinear` / `GroupGEMM` where `int4-qat` writes TE `GroupedLinear`; a stray empty `## TODO` section in `examples/infra_features/low_precision/README.md`; an unrendered `{/* FIGURE PLACEHOLDER */}` in `launch-script`; and five British spellings.

### DIFF
--- a/docs/user-guide/fully-async.md
+++ b/docs/user-guide/fully-async.md
@@ -326,9 +326,8 @@ Every skipped point is logged at the step it would have landed on, with the reas
 | `eval/skipped_busy` | At `--eval-max-in-flight` under `--eval-overflow-policy skip` |
 | `eval/skipped_export_failed` | The snapshot export raised |
 | `eval/skipped_ckpt_missing` | No `.complete` marker in the snapshot directory |
-| `eval/skipped_unhealthy` | The fleet or its router was unreachable |
-| `eval/skipped_pin_violation` | The engines did not all report the expected weight version |
 | `eval/skipped_crashed` | Anything else the eval raised |
+| `eval/skipped_<reason>` | A `CheckpointEvalFn` raised `EvalSkip(reason)` |
 
 For a point that did run, `eval/{dataset}/weight_version/mean == eval/step` and
 `eval/{dataset}/weight_version/mixed_version_ratio == 0` together confirm it measured
--- a/examples/infra_features/train_infer_mismatch_helper/README.md
+++ b/examples/infra_features/train_infer_mismatch_helper/README.md
@@ -92,13 +92,15 @@ Advantages:
 
 ## APIs of Algorithms
 
-You may choose from above algorithms by specifying arguments below:
+You may choose from above algorithms with the two command-line flags below. They are the
+only CLI flags this feature adds; everything in **Configs and Recommended Settings** is a
+key in the YAML file you pass to `--custom-config-path`.
 
 `--use-rollout-logprobs`: True if only use `rollout_log_probs` to compute the loss, bypassing old_log_probs calculated by training engine;
 
-`--use-rollout-correction`: True if apply importance sampling/rejection sampling to loss.
+`--use-tis`: True if apply importance sampling/rejection sampling to loss.
 
-| `use_rollout_logprobs` | `use_rollout_correction` | Algorithm | Policies |Compute old_log_probs | Batch Invariant | Recommended TIS Mode |
+| `use_rollout_logprobs` | `use_tis` | Algorithm | Policies |Compute old_log_probs | Batch Invariant | Recommended TIS Mode |
 |-----------------|-------------|-----------|--------------|---------------|-----------------|----------------------|
 | False | False | Standard PPO (Algorithm 0) | 2 ($\pi_\theta$, $\pi_{\textcolor{blue}{\text{old}}}$)|Yes | No | N/A |
 | True | False | Bypassing PPO (Algorithm 3) | 2 ($\pi_\theta$, $\pi_{\textcolor{red}{\text{SGLang}}}$) |🚀 Skipped | No | N/A |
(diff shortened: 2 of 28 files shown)


nblintao:
@claude review

nblintao (approved):
My agent found 2 mistakes that seems reasonable to me. But I don't have full context. Could you check if it makes sense?

docs/user-guide/fully-async.md
```diff
@@ -326,9 +326,8 @@ Every skipped point is logged at the step it would have landed on, with the reas
 | `eval/skipped_busy` | At `--eval-max-in-flight` under `--eval-overflow-policy skip` |
 | `eval/skipped_export_failed` | The snapshot export raised |
 | `eval/skipped_ckpt_missing` | No `.complete` marker in the snapshot directory |
-| `eval/skipped_unhealthy` | The fleet or its router was unreachable |
-| `eval/skipped_pin_violation` | The engines did not all report the expected weight version |
```
nblintao, line 330: These two do have emitters — `eval_fleet.py` raises `EvalSkip("unhealthy")` / `EvalSkip("pin_violation")` and `metrics.py` logs `eval/skipped_{reason}`, so the literal names never show up in a grep. I'd keep both rows and add the generic one on top.

examples/infra_features/train_infer_mismatch_helper/README.md
```diff
@@ -178,40 +191,44 @@ When rollout log probabilities are available, MILES automatically tracks compreh
...
+| `tis_weight_after_bound` | IS weights after the `tis_mode` bounding | `use_tis` |
+| `tis_truncate_fraction` | Fraction of weights truncated | `tis_mode: truncate` |
+| `tis_clip_fraction_low` / `tis_clip_fraction_high` | Fraction of weights clipped below / above the bound | `tis_mode: clip` |
+| `tis_mask_fraction_low` / `tis_mask_fraction_high` | Fraction of tokens rejected below / above the bound | `tis_mode: mask` |
+| `rs_mask_fraction_low` / `rs_mask_fraction_high` | Fraction of tokens rejected by rejection sampling | `use_rs` |
+| `rs_catastrophic_token_fraction` | Fraction of catastrophic tokens (below the veto threshold) | `rs_veto_threshold` set |
+| `rs_catastrophic_seq_fraction` | Fraction of sequences holding a catastrophic token | `rs_veto_threshold` set |
+| `is_ratio_mean_after_tis_rs` | Mean IS weight after both TIS and RS | `use_tis` |
+| `batch_norm_factor` | Batch normalization factor applied to weights (1.0 when off) | `use_tis` |
+| `is_ratio_mean_final` / `is_ratio_min_final` / `is_ratio_max_final` | Final IS weight statistics actually multiplied into the loss | `use_tis` |
```
nblintao: These still need the `mis_` prefix: both `compute_mis_weights_with_cp` and `compute_mis_weights_fsdp` end with `result_metrics[f"mis_{key}"]`, so wandb shows `train/mis_tis_truncate_fraction`, `train/mis_is_ratio_mean_final`, etc. Same for the PPL table above. Note `batch_norm_factor` was already correct as `mis_batch_norm_factor` before this PR.

PR #2665, opened 2026-08-19
test(lora): cover multi-adapter training on AMD

Arist12:
Part of #2705.

## Problem

The disaggregated Multi-LoRA path lacks an AMD E2E covering trained adapters with independent lifetimes and checkpoint export.

## Change

Add a manually run eight-GPU Qwen3-4B test with two adapters on unequal step schedules. It asserts the exact TP2 native shard set, finite PEFT artifacts, nonzero trained LoRA-B factors, and distinct adapter weights. Suite assignment remains a reviewer decision.

## Validation

8x MI350X: both adapters reached their configured final steps, produced nonzero/distinct weights, and wrote both checkpoint formats.


Arist12:
Closing: Multi-LoRA training is being re-factored and the code this test covers will be discarded. Removed from the #2705 tracker.

PR #2666, opened 2026-08-19
fix(inkling): support native LoRA on ROCm

Arist12:
Part of #2705.

## Problem

On ROCm, the Inkling launcher selects FA4 and enables multimodal serving for text-only runs, while native LoRA assumes the public `grouped_mm` alias exists. These choices block the existing Inkling-Small LoRA E2E.

## Change

Select SGLang Triton attention on ROCm while retaining FA4 on CUDA, enable multimodal serving only for multimodal ROCm tasks, and lazily fall back to `torch._grouped_mm`.

Dependencies: [SGLang #34702](https://github.com/sgl-project/sglang/pull/34702) and [SGLang #35501](https://github.com/sgl-project/sglang/pull/35501).

## Validation

4x MI350X combined stack: two rollouts and two publishes with `74/74` adapter tensors matching on every TP/EP rank.


Arist12:
Closing: dropping Inkling from the AMD LoRA RL scope for now. Removed from the #2705 tracker; can be reopened if Inkling on ROCm becomes a target again.

PR #2667, opened 2026-08-19
docs: fix LoRA E2E GPU requirement

Arist12:
## Problem

The Qwen2.5 LoRA E2E test docstring identifies 8 GPUs as a requirement, but the intended requirement is 4 GPUs.

## Change

Change the documented GPU requirement from 8 to 4.


Arist12:
@yushengsu-thu same quick fix

PR #2669, opened 2026-08-19
fix(lora): deduplicate native adapter checkpoints

### LINKED ISSUE #2668: [Bug] LoRA native checkpoints duplicate adapter shards across DP ranks
## Problem

`save_lora_checkpoint` writes `adapter_megatron_rank{global_rank}.pt` on every rank. Data-parallel replicas hold identical adapter weights, so they write redundant copies (for example, DP=4 with TP=PP=CP=1 writes four identical files).

## Expected

Write one native adapter shard per unique model-parallel shard and let DP replicas share it, while keeping checkpoint loading backward compatible.

Arist12:
Part of #2705.

## Problem

Single-LoRA checkpoints write one native adapter shard per global rank, so DP and CP replicas duplicate identical data. A writer-side save failure can also leave peer ranks entering later Bridge collectives.

## Reproduction on `main`

Both halves run on CPU with the gloo backend, so no GPU or vendor runtime is on the path.

4 gloo ranks at TP=PP=EP=ETP=1, DP=4 call `save_lora_checkpoint`:

```
4 native shard files, byte-identical adapter tensors, 75% of the bytes redundant
```

An 8-GPU run of `examples/lora/run-qwen3-4b-megatron-lora-result.sh` (no TP/PP, `--save-interval 50`) is DP=8, so it writes eight copies of one shard.

Saving at 4 ranks and resuming the same checkpoint at 8:

```
rank 0-3   load_lora_adapter -> True    adapter restored
rank 4-7   load_lora_adapter -> False   fresh random init
```

The shard name is keyed on the global rank, so half the DP replicas silently start from a different adapter. No crash, no error.

## Change

Write one native shard for each realized `(TP, PP, EP, ETP)` coordinate, retaining distinct EP shards and legacy load compatibility. Share native-save failures across ranks before PEFT export begins.

ETP is part of the shard identity because when `ETP > TP` two ranks share a TP rank while holding different expert slices. The `_etp` filename suffix is emitted only in that case: for `ETP <= TP` the ETP rank is a function of the TP rank, so every filename in use today is byte-identical.

## Validation

- 76 fast tests in `test_lora_utils.py` and `test_multi_lora_checkpoint_naming.py`, including shard naming and completeness under EP and ETP. Full `tests/fast/backends/megatron_utils` green apart from the two `TestUpdateWeightsZeroChunks` failures that reproduce on `main`.
- 4x MI350X: one `adapter_megatron_tp0_pp0.pt` shard with finite tensors
- EP4 full-model runs: four expert-distinct shards with replicated attention tensors

Fixes #2668

## Notes for reviewers

Land this first among the checkpoint PRs: it adds `lora_checkpoint.py`, and #2516 and #2580 each carry their own copy of the collective-failure helper it puts there. See #2705 for the ordering rationale.

### DIFF
--- a/miles/backends/megatron_utils/lora_checkpoint.py
+++ b/miles/backends/megatron_utils/lora_checkpoint.py
@@ -1,0 +1,54 @@
+from pathlib import Path
+
+import torch.distributed as dist
+
+from miles.backends.training_utils.parallel import get_parallel_state
+from miles.utils.distributed_utils import get_gloo_group
+
+_shard_topology: tuple[bool, tuple[tuple[int, int, int], ...]] | None = None
+
+
+def raise_if_any_rank_failed(local_error: Exception | None, operation: str) -> None:
+    message = None if local_error is None else f"{type(local_error).__name__}: {local_error}"
+    if dist.is_initialized():
+        group = get_gloo_group()
+        messages: list[str | None] = [None] * dist.get_world_size(group=group)
+        dist.all_gather_object(messages, message, group=group)
+        message = next((item for item in messages if item is not None), None)
+
+    if message is not None:
+        error = RuntimeError(f"{operation} failed on at least one rank: {message}")
+        if local_error is not None:
+            raise error from local_error
+        raise error
+
+
+def megatron_shard_name(tp_rank: int, pp_rank: int, ep_rank: int, ep_size: int) -> str:
+    name = f"adapter_megatron_tp{tp_rank}_pp{pp_rank}"
+    if ep_size > 1:
+        name += f"_ep{ep_rank}"
+    return name + ".pt"
+
+
+def adapter_shard_topology() -> tuple[bool, tuple[tuple[int, int, int], ...]]:
+    global _shard_topology
+    if _shard_topology is not None:
+        return _shard_topology
+
+    parallel_state = get_parallel_state()
+    coords = (parallel_state.tp.rank, parallel_state.pp.rank, parallel_state.ep.rank)
+    if not dist.is_initialized():
+        _shard_topology = (True, (coords,))
+        return _shard_topology
+
+    current_rank = dist.get_rank()
+    group = get_gloo_group()
+    gathered: list[object] = [None] * dist.get_world_size(group=group)
+    dist.all_gather_object(gathered, (coords, current_rank), group=group)
+    is_writer = current_rank == min(rank for entry_coords, rank in gathered if entry_coords == coords)
+    _shard_topology = (is_writer, tuple(sorted({entry_coords for entry_coords, _ in gathered})))
+    return _shard_topology
+
+
+def all_megatron_checkpoints_exist(step_dir: Path, shard_names) -> bool:
+    return all((step_dir / name).exists() for name in shard_names)
--- a/miles/backends/megatron_utils/lora_checkpoint.py
+++ b/miles/backends/megatron_utils/lora_checkpoint.py
@@ -1,0 +1,88 @@
+"""Naming and topology helpers shared by the single- and multi-LoRA checkpoint paths."""
+
+from pathlib import Path
+
+import torch.distributed as dist
+
+from miles.backends.training_utils.parallel import get_parallel_state
+from miles.utils.distributed_utils import get_gloo_group
+
+# Cached by adapter_shard_topology(); the topology is fixed for the run.
+_shard_topology: tuple[bool, tuple[tuple[int, int, int, int], ...]] | None = None
+
+
+def raise_if_any_rank_failed(local_error: Exception | None, operation: str) -> None:
+    """Raise on every rank when any rank reports ``local_error``, so failures stay collective."""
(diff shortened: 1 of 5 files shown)


jhinpan (commented):
I reviewed the current head (`c2cac7eecb8a083e2252bf5a4d5e62ea60453f0d`) and ran distributed topology probes on AMD Instinct MI355X GPUs.

**Blocker — `(TP, PP, EP)` is not always a unique adapter-shard identity.**

With four ranks configured as TP1 / PP1 / EP2 / ETP2, Megatron produced these distinct coordinates and tensor values:

- rank 0: `(tp=0, pp=0, ep=0, etp=0)`, value 0
- rank 1: `(tp=0, pp=0, ep=0, etp=1)`, value 1
- rank 2: `(tp=0, pp=0, ep=1, etp=0)`, value 10
- rank 3: `(tp=0, pp=0, ep=1, etp=1)`, value 11

`adapter_shard_topology()` selected only ranks 0 and 2 as writers and produced two files containing values 0 and 10. The ETP-distinct values 1 and 11 were silently discarded. Names and shapes still match on load, so #2733 cannot detect this corruption.

The invariant should be that every deduplicated coordinate contains byte-identical adapter tensors. Either include ETP in shard identity/naming when it is independent, or reject any realized topology where one `(tp, pp, ep)` maps to multiple ETP ranks.

**Legacy loading is accepted when the code itself says it is invalid.**

The TP/PP fallback is loaded unconditionally under `ep_size > 1`, despite warning that it is only valid when `EP <= TP`. For example, TP1 / EP4 can make all EP ranks read the same shard. Require an unambiguous realized mapping and a global all-rank existence decision; otherwise reject the legacy layout.

**Rank-local filesystem failures are not fully coordinated or atomic.**

`save_path.mkdir()` runs before the first barrier/error consensus, and `_save_native_adapter_checkpoint` writes directly to the final filename. A rank-local mkdir failure can strand peers at the barrier; interruption during `torch.save` leaves an existing but truncated shard. Coordinate directory creation, write a unique temporary file, then atomically replace the final shard only after successful serialization.

jhinpan (commented):
Follow-up review on the unchanged head (`c2cac7eecb8a083e2252bf5a4d5e62ea60453f0d`): this is distinct from the first review's ambiguous legacy-fallback finding.

**Existing but topology-incompatible native shards silently become a fresh adapter.**

`_native_adapter_shard_path` uses the current run's EP size. A checkpoint written with EP4 contains files such as `adapter_megatron_tp0_pp0_ep0.pt`, while an EP1 resume searches for `adapter_megatron_tp0_pp0.pt`. The global-rank legacy file is absent, and the TP/PP legacy branch is skipped because the current `ep_size` is 1. The loader therefore returns `False, None`; `checkpoint.py` only warns and continues with freshly initialized adapter weights.

EP narrowing cannot be solved by choosing another filename because expert slices would need a defined merge. The correct result is a hard incompatibility error, not silent training from a new adapter.

After the normal selection ladder is exhausted, check whether the directory contains any `adapter_megatron*.pt` shards. If it does, report that native checkpoint data exists but no complete compatible layout can be selected. A shard manifest/completion marker would make this decision unambiguous and also distinguish an interrupted save from a directory that was never a native checkpoint.

Arist12:
Thanks — the ETP point was a genuine bug and is fixed; the rest I'd rather keep out of this PR, with reasoning below.

**Adopted — ETP is part of the shard identity.** With `expert_tensor_parallel_size != tensor_model_parallel_size`, two ranks that share `(tp, pp, ep)` hold different expert shards, so naming by `(tp, pp, ep)` alone made them overwrite each other. `megatron_shard_name` now takes all four coordinates and `adapter_shard_topology` gathers 4-tuples.

Reviewing that change surfaced a second instance of the same mistake: the global writer (the rank that promotes the temp directory) was still elected on `(tp, pp, ep) == (0, 0, 0)`, so at `etp_size > tp_size` two ranks both qualified and raced on `shutil.rmtree(final_dir)` / `os.replace(...)`. Now elected on all four coordinates.

**Deferred — the legacy `tp/pp` filename fallback under `ep_size > 1`.** Correct that the fallback can match a file that a differently-sharded run wrote. It is pre-existing behaviour and removing it breaks resume from every checkpoint written before this PR, so it wants its own change with a migration note rather than being folded into the dedup fix.

**Deferred — mkdir / atomic-write coordination and the EP-narrowing silent fresh start.** Both are real, both are the same class of problem as the load-ladder consensus: the fix is to route the whole save/load ladder through `lora_checkpoint.raise_if_any_rank_failed` and agree on the topology before touching the filesystem. This PR adds that helper and the topology query; using them everywhere is the follow-up that #2733 and #2516 also point at. Keeping this one to "one writer per shard" so the dedup is reviewable on its own.

**Deferred — `_tmp_step_{iteration}` reuse.** Pre-existing; a rerun at the same iteration reuses a stale temp directory. Unrelated to writer election.

Validation: `tests/fast/backends/megatron_utils` — 268 passed, plus the two `TestUpdateWeightsZeroChunks` failures that reproduce on `main`. pre-commit clean.

PR #2673, opened 2026-08-20
Bump Megatron-LM to miles-main-20260819 (latest NVIDIA dev)

yueming-yuan:
ci-megatron-pr: miles-main-20260819

## What

Bump the Megatron-LM that miles depends on from the stale 2026-02-14 fork point to latest
NVIDIA/Megatron-LM `dev` (`3f49de20d`, 2026-08-19). Supersedes #1466 (June bump, never merged).

**Scope: base bump only.** DeepSeek-V4 and true-on-policy are explicitly disabled and land in
dedicated follow-up PRs (DSv4 first, then true-on-policy).

## Megatron side — `radixark/Megatron-LM:miles-main-20260819`

nvidia/dev HEAD + 20 commits: 14 rebased from the June bump branch (`miles-main-20260622`) +
6 newly ported miles-main commits (#21 witness, #63 NVMe streaming, #68 fp32 MoE, #74 ROCm
dist-ckpt, #72 base64 te-precision-config, #82 LayerWise expert ownership).

Dropped as upstreamed into dev since June (verified feature-by-feature):
- #9 INT4 fake QAT (`transformer_engine_int4_fake_qat.py`, same env-var interface)
- #17/#18 TMS no-cpu-backup (now `DistributedDataParallelConfig` fields)
- #51 Nemotron RL (absorbed into dev's `models/hybrid/` refactor: pattern-MTP, `is_mtp_layer`,
  `mtp_use_repeated_layer`, hybrid MTP block specs)
- #60 (June's re-applied #24 already covers it)

## Miles side (this PR)

- Compat fixes re-derived on main: gloo arg rename (4 sites), drop `moe_use_legacy_grouped_gemm`,
  MTP `transformer_layer`→`mtp_model_layer` in all converters, MTP loss tracker
  `reduce_loss_in_tracker`, `name=` kwarg in custom attention, `mtp_detach_heads` adaptation,
  dumper anchors → `_forward_mlp_output_with_bias`, LoRA buffer patch → `ddp_config` flags.
- Explicit disables: megatron backend + `--true-on-policy-mode` raises NotImplementedError;
  `test_model_provider_true_on_policy.py` skipped; DSv4 4-layer e2e disabled via
  `register_cuda_ci(disabled=...)`. fsdp true-on-policy is unaffected.
- `docker/Dockerfile`: `MEGATRON_BRANCH=miles-main-20260819`.

## Follow-ups (not this PR)

1. DSv4 PR: the dsv4 attention variant + `--dsv4-impl {miles,megatron}` dual backend + R3
   threshold sign-off.
2. True-on-policy PR: the SGLang-matching megatron backend + `miles_megatron_plugins` packaging.

### DIFF
--- a/.github/workflows/_build-pr-ci-image.yml
+++ b/.github/workflows/_build-pr-ci-image.yml
@@ -21,6 +21,7 @@ jobs:
     outputs:
       needs_image: ${{ steps.decide.outputs.needs_image }}
       rebuild: ${{ steps.decide.outputs.rebuild }}
+      force_rebuild: ${{ steps.prepare.outputs.force_rebuild }}
     env:
       SAME_REPO: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
     steps:
@@ -28,27 +29,20 @@ jobs:
         with:
           fetch-depth: 2
           persist-credentials: false
-      # Reading the inputs label off the published tag needs registry auth. Forks get no
-      # secrets and cannot push anyway, so they skip straight to the released image.
-      - name: Login to Docker Hub
-        if: env.SAME_REPO == 'true'
-        uses: docker/login-action@v3
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - id: decide
+      - id: compare
         shell: bash
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha || '' }}
-          PR_NUMBER: ${{ github.event.pull_request.number || '' }}
-          FORCE_REBUILD: ${{ contains(github.event.pull_request.labels.*.name, 'rebuild-ci-image') }}
         run: |
           set -euo pipefail
-          emit() { echo "needs_image=$1" >> "$GITHUB_OUTPUT"; echo "rebuild=$2" >> "$GITHUB_OUTPUT"; }
+          emit() {
+            echo "eligible=$1" >> "$GITHUB_OUTPUT"
+            echo "current=$2" >> "$GITHUB_OUTPUT"
+          }
 
           # Cron and manual runs have no PR diff and never own a PR image.
           if [ -z "$BASE_SHA" ]; then
-            emit false false
+            emit false ""
             exit 0
           fi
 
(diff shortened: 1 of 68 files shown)


guapisolo:
@codex review

guapisolo:
@claude review

Zhichenzzz:
@yueming-yuan did you test the dist_muon path for the new branch? worthy having it

guapisolo:
[P2] Avoid re-scaling per-token MTP loss by the microbatch count

At `miles/backends/megatron_utils/model.py:804-809`, `MTPLossLoggingHelper.reduce_loss_in_tracker()` already computes `sum(loss) / sum(tokens)` when `calculate_per_token_loss` is enabled. The pinned Megatron training path correspondingly uses `mtp_loss_scale = 1.0` for that mode, but Miles still unconditionally multiplies the result by `1 / num_microbatches[step_id]`.

With `--enable-mtp-training --calculate-per-token-loss` and N microbatches, the reported `mtp_loss` and the value passed to `check_mtp_loss` are therefore N times too small. Please apply the microbatch averaging only in the non-per-token mode.

guapisolo:
Prospective tip `59cc3c11f3ca851db37c81fc316f6facbeda1318` will stop applying microbatch averaging after `MTPLossLoggingHelper` has already produced a per-token-normalized MTP loss, while preserving legacy averaging. The candidate passed the target file's full pre-commit hooks on `miles-h200-dev-2` and the synced AST check (`per_token=1.0`, `legacy=0.25` for four microbatches).

guapisolo (approved):

guapisolo:
novita 2 has been fixed. re-kick off ci

Zhichenzzz (approved):

guapisolo:
## Codex review

Reviewed against Miles PR head `c292bee42b219fa4921e33a540cdbd55fa73aa18` and `radixark/Megatron-LM:miles-main-20260819` at `235952df607b3820716e5e67728a5ab470ca33ae`.

### Review 2 — [P1] Keep the shared checkpoint merge helper independent of training globals

`4c725b138` added an unconditional `get_args().low_memory_resume` read to `megatron/core/transformer/utils.py::cat_with_oom_fallback`. This function is a shared MCore `ShardedTensorFactory.merge_fn`, used by SSM, QKV attention, and GLU MLP checkpoint factories and reached from the public distributed-checkpoint load path. A standalone MCore, Megatron-Bridge, or checkpoint-conversion caller is not required to initialize `megatron.training` globals, so it now fails before attempting `torch.cat` with:

```text
AssertionError: args is not initialized.
```

This is introduced relative to the new upstream baseline: the parent helper had no training-global dependency. Old `miles-main` had a similar read only in an MLP-private callback; `4c725b138` expands it into the generic helper. Targeted devbox runs reproduce the regression in both existing cases:

- `test_ssm_split_tensor_factory_oom_is_handled[gated_delta_net]`
- `test_ssm_split_tensor_factory_oom_is_handled[mamba_mixer]`

Please make an uninitialized args state equivalent to `low_memory_resume=False`, preserving the normal merge and existing OOM fallback. Longer term, the force-CPU decision should be passed through a core-scoped checkpoint context instead of introducing a `megatron.core -> megatron.training` dependency.

### Review 3 — [P1 if reverse-picked; non-blocking for PR #2673] Do not apply `95e3a6d` to old `miles-main`

`95e3a6d` is correct on the dated branch because upstream `ceaa6c5d7e` rebuilds `model_param_group_index_map` after the optimizer's fp32-first parameter order has been installed. Old `miles-main` does not contain `ceaa6c5d7e`; its only correct rebuild is the implementation introduced by `cf014f903` inside `_build_model_and_main_param_groups`.

Reverse-picking `95e3a6d` onto old produces content conflicts. Resolving those conflicts by accepting the dated deletion removes old's only rebuild. Mixed fp32 and bf16/fp16 parameters in one optimizer group would then use grad-buffer order to index an fp32-first optimizer group, causing checkpoint save/load to read another parameter's main parameter or Adam state. Different shapes normally crash; equal shapes can silently associate state with the wrong parameter. The later `479bad6` restores only the fp32 leaf detach and does not restore this map.

No change is required for PR #2673 for this item: its dated target retains the upstream post-allocation rebuild. This is a backport guard—skip `95e3a6d` when moving dated follow-ups onto old `miles-main`, unless `ceaa6c5d7e` is ported first in the same atomic change.

PR #2675, opened 2026-08-20
[CI] rebuild a PR's docker image only when its build inputs change

yueming-yuan:
## Summary

Reuse a PR-scoped CI image until its Docker build inputs change.

## Motivation

A Dockerfile-touching PR previously rebuilt the same multi-arch image on every rerun or source-only push because CI keyed the build on changed paths. That repeated an expensive build even when the image inputs were unchanged.

## Usage

Push Docker-input changes normally; CI publishes or reuses `radixark/miles:pr-<num>`. After the repository's `rebuild-ci-image` label is provisioned, apply it to rebuild unchanged inputs after a moved base image, floating dependency, or corrupt push; a successful forced build removes the label.

## Design Notes

- **Mental model:** `docker/image_inputs.py` hashes the files consumed by the cu13 build, `docker/build.py` stamps that hash on the PR tag, and `docker-decide` compares the current hash with the published label. A current tag sets `tag_available`, so every GPU suite keeps using the PR image even when the present run skips its build.
- **Rerun safety:** The workflow reads `rebuild-ci-image` from live PR labels, carries that captured decision through the build, and consumes it only after a successful forced build. A completed labeled run therefore reuses its tag when rerun, while a failed build leaves the label available for retry.
- **Registry boundary:** Input comparison happens before Docker Hub authentication. Non-Docker PRs stay on `dev` without touching the registry; same-repository Docker PRs authenticate only to inspect or publish their PR tag, and fork PRs cannot publish.
- **Input boundary:** `Dockerfile.rocm` remains excluded because this workflow builds cu13; moved base images and other floating remote state remain explicit manual-rebuild cases.

## Verification

- `pytest -q tests/ci/test/test_run_suite.py tests/fast/test_docker_image_inputs.py`: 98 tests passed against prospective tip `34f7e6f1`, covering workflow seams, input hashing, and image-label parsing.
- `pre-commit run --files .github/workflows/_build-pr-ci-image.yml tests/ci/test/test_run_suite.py`: every applicable repository hook passed.
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/_build-pr-ci-image.yml .github/workflows/pr-test.yml`: exited 0 with no diagnostics.

## Review Focus

- `docker/image_inputs.py`: verify the declared input set matches the cu13 build context and accepted manual-rebuild boundary.
- `.github/workflows/_build-pr-ci-image.yml`: verify the captured live-label decision across success, failure, and rerun paths.
- `.github/workflows/pr-test.yml`: verify `tag_available` selects a reused PR image without changing fork or non-Docker behavior.

### DIFF
--- a/.github/workflows/_build-pr-ci-image.yml
+++ b/.github/workflows/_build-pr-ci-image.yml
@@ -21,6 +21,7 @@ jobs:
     outputs:
       needs_image: ${{ steps.decide.outputs.needs_image }}
       rebuild: ${{ steps.decide.outputs.rebuild }}
+      force_rebuild: ${{ steps.prepare.outputs.force_rebuild }}
     env:
       SAME_REPO: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
     steps:
@@ -28,27 +29,20 @@ jobs:
         with:
           fetch-depth: 2
           persist-credentials: false
-      # Reading the inputs label off the published tag needs registry auth. Forks get no
-      # secrets and cannot push anyway, so they skip straight to the released image.
-      - name: Login to Docker Hub
-        if: env.SAME_REPO == 'true'
-        uses: docker/login-action@v3
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - id: decide
+      - id: compare
         shell: bash
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha || '' }}
-          PR_NUMBER: ${{ github.event.pull_request.number || '' }}
-          FORCE_REBUILD: ${{ contains(github.event.pull_request.labels.*.name, 'rebuild-ci-image') }}
         run: |
           set -euo pipefail
-          emit() { echo "needs_image=$1" >> "$GITHUB_OUTPUT"; echo "rebuild=$2" >> "$GITHUB_OUTPUT"; }
+          emit() {
+            echo "eligible=$1" >> "$GITHUB_OUTPUT"
+            echo "current=$2" >> "$GITHUB_OUTPUT"
+          }
 
           # Cron and manual runs have no PR diff and never own a PR image.
           if [ -z "$BASE_SHA" ]; then
-            emit false false
+            emit false ""
             exit 0
           fi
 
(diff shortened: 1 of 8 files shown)


.github/workflows/_build-pr-ci-image.yml
```diff
@@ -6,54 +6,103 @@ on:
...
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - id: decide
         shell: bash
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha || '' }}
+          PR_NUMBER: ${{ github.event.pull_request.number || '' }}
+          FORCE_REBUILD: ${{ contains(github.event.pull_request.labels.*.name, 'rebuild-ci-image') }}
```
guapisolo: [P2] Read the live force-rebuild label instead of the triggering payload

`github.event.pull_request.labels` is frozen to the webhook payload for this workflow run. After a successful labeled run removes `rebuild-ci-image`, using Re-run all jobs creates another attempt of that same run, so this expression still evaluates true and rebuilds the full multi-arch image again. That breaks both the documented one-shot behavior and the rerun-reuse goal. Please query the current live labels for the PR at decision time. Gating only on `github.run_attempt == 1` is not sufficient because a first attempt that failed before consuming the label must still rebuild on retry.

.github/workflows/_build-pr-ci-image.yml
```diff
@@ -6,54 +6,103 @@ on:
...
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 2
           persist-credentials: false
-      - id: diff
+      # Reading the inputs label off the published tag needs registry auth. Forks get no
+      # secrets and cannot push anyway, so they skip straight to the released image.
+      - name: Login to Docker Hub
+        if: env.SAME_REPO == 'true'
```
guapisolo: [P2] Defer Docker Hub login until an image is actually needed

This login now runs for every same-repository PR before the `CURRENT == BASE` early exit. A source-only PR therefore depends on Docker Hub credentials and registry availability even though this job should select `dev`; an authentication or registry outage can fail the GPU pipeline before the unchanged-input fast path runs. Please compute `CURRENT` and `BASE` first, then authenticate only when the workflow needs to inspect the existing PR image tag.

yushengsu-thu (approved):

PR #2677, opened 2026-08-20
feat(openenv): report episode end reason

### LINKED ISSUE #2592: openenv agent metrics do not report why an episode ended (no end_reason / truncated flag)
## What is missing

The openenv tbench2 agent loop can end an episode for four different reasons, and the metrics it returns cannot tell them apart:

- the policy said `TASK_COMPLETE`
- the policy emitted no command
- `OPENENV_MAX_TURNS` ran out
- a turn came back `finish_reason="length"` (new in #2587)

All four surface only as a `turns` count:

https://github.com/radixark/miles/blob/e4152f613/examples/experimental/openenv/openenv_agent_function.py#L361-L370

## Why it matters

#2587's own framing is the argument. The post-truncation ghost tail — 22.7% of all episode wall-clock on a GB300 16-node GLM-5.2 run (144.7 h of 638 h) — went unnoticed for as long as it did *because it was invisible*: the dashboard's batch anatomy derives spans from request gaps, so tool/gen time after the final policy call cannot be recovered and the tail renders as a long "queue/waiting" band.

Making truncation terminal fixes the waste but leaves the same blind spot: after #2587, a truncation-ended episode and a `TASK_COMPLETE` episode are distinguishable only by a shorter `turns`. That gates several things we want to watch:

- how often episodes end by truncation at all, per step — the direct readout of whether the per-turn cap is set sanely
- whether #2588's `rollout_max_response_len` 16384 → 8192 changed that rate (the claim is that completed samples never touched the cap; a counter proves it instead of arguing it)
- the population overlap between truncation-ended episodes and the failed-sample loops that #2588 identifies as the collapse carrier (truncated, negative advantage over 16k+ shared near-deterministic scaffolding tokens)

## Suggested shape

Add an end reason to the metrics dict the loop already returns, e.g. `"end_reason": "task_complete" | "no_command" | "max_turns" | "length"`, set at each `break` site. It rides the existing `agent_metrics` path into `Sample.metadata`, so W&B/dashboard aggregation needs no new plumbing. A plain `truncated: bool` would cover the immediate need, but the four-way split costs the same and answers "are we running out of turns or running out of tokens?" too.

If #2591 gets a session-token counter in the loop, `end_reason` is where its stop would report itself as well.

## Related

- #2587 — makes truncation terminal
- #2588 — halves the per-turn cap; wants this counter to verify its premise
- #2591 — the session-budget boundary, same reporting hook

MinhaoLi0318:
## Summary

  - report why each OpenEnv agent episode ended
  - distinguish `task_complete`, `no_command`, `max_turns`, and `length`
  - add offline test coverage for all four termination paths

  ## Testing

  - `python -m pytest --confcutdir=tests/fast/examples/experimental/openenv tests/fast/examples/experimental/openenv -q`
    - 106 passed, 3 skipped
  - Ruff, Black, isort, and autoflake checks passed

  Closes #2592

cc @nblintao @yueming-yuan

### DIFF
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -259,13 +259,15 @@ async def multi_turn(
     ends when the policy stops emitting a command, says TASK_COMPLETE, or hits
     OPENENV_MAX_TURNS. Scoring is the standard ``evaluate`` action: the server
     runs the task's tests/test.sh and reports the verdict (see the module
-    docstring for the required server contract).
+    docstring for the required server contract). The returned agent metrics
+    label the stop as ``task_complete``, ``no_command``, ``max_turns``, or
+    ``length``.
     """
     action_cls = classes["action"]
     task_id = metadata.get("task_id") or metadata.get("task_name")
     max_turns = int(os.getenv("OPENENV_MAX_TURNS", "30"))
 
-    async def body(env: Any) -> tuple[float | None, int, list[float], list[float], float, float]:
+    async def body(env: Any) -> tuple[float | None, int, str, list[float], list[float], float, float]:
         # Per-turn wall-clock timings. gen_times[i] is turn i's policy generation
         # latency; tool_times[i] is turn i's env.step(exec) latency. reset_time and
         # eval_time bracket the one-off reset() and the final evaluate() env steps.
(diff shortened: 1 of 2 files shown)


nblintao:
@claude review

nblintao (approved):

PR #2678, opened 2026-08-20
test(ci): collect wandb init timeout tests

### LINKED ISSUE #2557: Move miles/utils/test_wandb_utils.py under tests/ so CI collects it
`miles/utils/test_wandb_utils.py` sits next to the code it tests, which is outside every root CI scans (`tests/fast`, `tests/fast-gpu`, `tests/e2e`, `tests/ci`), so nothing collects it. Its two tests — that both wandb init call sites pass the extended init timeout — have never run in CI.

It is not disabled and never was: the file has lived at that path since the CI-stage refactor (#1149) era and through the baseline import, always outside the scanned roots.

## Why it was left out of the sweep

The change that moves the example test suites under `tests/fast/examples/` deliberately stops at the agentic-env suites. Where `tracking_utils`' tests belong is a call for whoever is restructuring them: **#2505** (`tom/refactor-miles/op21-5`, open) is adding `tests/fast/utils/tracking_utils/` with `__init__.py`, `test_base.py` and a `test_wandb_utils.py` of its own, while leaving this file where it is.

Note the two files are not duplicates. #2505's covers `_compute_config_for_logging` and the launch record; this one covers the init timeout at both call sites.

## Suggested move

`tests/fast/utils/tracking_utils/test_wandb_init.py` — the directory mirrors `miles/utils/tracking_utils/` and matches where #2505 is putting its tests, and the filename says what the file covers instead of colliding with the `test_wandb_utils.py` that PR adds. Checked while preparing the sweep: the `__init__.py` both sides would add is a zero-byte blob, so the two changes do not conflict.

Nothing else is needed — `tests/fast/` auto-registers as CPU CI, and `wandb` is already in `requirements.txt`, so the test should simply start running in stage-a-cpu.

## Cleanup that comes with it

`tests/ci/test/test_ci_discovery_coverage.py` enforces "no `test_*.py` outside `tests/`" and carries this file as its single `_KNOWN_ORPHANS` entry. Moving the file makes `test_no_known_orphan_outlives_its_reason` fail until that entry is deleted — one line, and the exception list is empty again.

Mantissagithub:
### Summary

- Move the WandB initialization timeout tests into `tests/fast/` so CPU CI collects them.
- Remove the obsolete orphan exception from the CI discovery coverage test.

### Testing

- `PYTHONPATH=. python -m pytest --confcutdir=tests/fast/utils/tracking_utils tests/fast/utils/tracking_utils/test_wandb_init.py -q` (2 passed)
- `PYTHONPATH=. python -m pytest --confcutdir=tests/ci/test tests/ci/test/test_ci_discovery_coverage.py::test_no_known_orphan_outlives_its_reason -q` (1 passed)
- `PYTHONPATH=. python tests/ci/run_suite.py --hw cpu --suite stage-a-cpu --match-all-labels --list-only` (moved test discovered in `stage-a-cpu`)
- `pre-commit run --files tests/fast/utils/tracking_utils/test_wandb_init.py tests/ci/test/test_ci_discovery_coverage.py`
- `git diff --check`

The full `pytest tests/fast` collection is blocked locally because the installed SGLang version does not export `encoding_dsv4`; the targeted CPU tests pass independently.

Closes #2557

cc @nblintao


nblintao:
@claude review

nblintao (approved):

Mantissagithub:
cc @yueming-yuan @guapisolo

behroozazarkhalili:
Verified on top of current main (2f9bb0248); the branch is a pure rename plus the four-line orphan-list cleanup, and it merges with no conflicts.

- tests/fast/utils/tracking_utils/test_wandb_init.py and tests/ci/test/test_ci_discovery_coverage.py on the merge: 5 passed.
- miles/utils/test_wandb_utils.py is gone, and run_suite --hw cpu --suite stage-a-cpu --list-only now lists the moved file, so the two wandb init-timeout tests run on every PR from here.
- ruff/black/isort at the pre-commit pins are clean.

Nothing else to ask for; this closes #2557 as written.

cc @fzyzcjy @Ying1123

PR #2683, opened 2026-08-21
fix: convert base weights with the bridge on the disaggregated path

### LINKED ISSUE #634: SFT Post Training for "Qwen3-VL-8B-Instruct" fails.
@claude @fzyzcjy @Lyken17 @Ying1123 @yushengsu-thu , I am currently doing post training SFT/Eval on top of "Qwen3-VL-8B-Instruct" using megatron as the backend. I get the following error. I was under the impression that Qwen3 VLM series is supprted in Miles. Please let me know how to resolve this issue. Thanks.

File "/root/miles/miles/backends/megatron_utils/update_weight/update_weight_from_distributed.py", line 139, in _update_weight_from_distributed
    converted_named_tensors += convert_to_hf(self.args, self.model_name, name, param, self.quantization_config)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miles/miles/backends/megatron_utils/megatron_to_hf/__init__.py", line 24, in convert_to_hf
    converted_named_tensors = _convert_to_hf_core(args, model_name, name, param)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miles/miles/backends/megatron_utils/megatron_to_hf/__init__.py", line 47, in _convert_to_hf_core
    converted_named_tensors = convert_qwen2_to_hf(args, name, param)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miles/miles/backends/megatron_utils/megatron_to_hf/qwen2.py", line 71, in convert_qwen2_to_hf
    raise ValueError(f"Unknown parameter name: {name}")
ValueError: Unknown parameter name: module.module.vision_model.patch_embed.proj.weight

+++++++
SFT_ARGS=(
   --rollout-function-path miles.rollout.sft_rollout.generate_rollout
   --prompt-data /root/datasets/${DATASET_LOCAL_NAME}/miles_sft_final_100.jsonl
   --input-key messages
   --rollout-shuffle
   --num-epoch 100
   --rollout-batch-size 96
   --transformer-impl transformer_engine
   
   --num-steps-per-rollout 1
   
   --loss-type sft_loss
   --calculate-per-token-loss
   --disable-compute-advantages-and-returns
   --tensor-model-parallel-size 4
   --eval-prompt-data evalset /root/datasets/${DATASET_LOCAL_NAME}/miles_sft_final_20_eval.jsonl
   --eval-interval 5        # Run eval every 5 iterations
   --eval-iters 1           # Number of batches to run during each eval
   --rollout-num-gpus 4
   --rollout-num-gpus-per-engine 1
)

### LINKED ISSUE #2679: Weight sync fails for bridge-built models (Qwen3.5/3.6/3.8 GDN, Qwen3-VL): --spec is ignored in bridge mode but megatron_to_hf expects the spec's names
## Summary

For any model whose Megatron parameter names come from `megatron.bridge` rather than from a `--spec` plugin, weight sync on the disaggregated path fails with `ValueError: Unknown parameter name`. Qwen3.5 / Qwen3.6 / Qwen3.8 hit this on every parameter of every GDN layer, so RL on them is not currently possible with a remote rollout engine.

The underlying reason is that `--megatron-to-hf-mode bridge` and `miles/backends/megatron_utils/megatron_to_hf/` disagree about who names the parameters:

- Loading an HF checkpoint **requires** bridge mode (`checkpoint.py:180`: `assert args.megatron_to_hf_mode == "bridge", "Only bridge mode is supported for loading HF checkpoint"`).
- In bridge mode the model is built by `AutoBridge.to_megatron_provider()` (`model_provider.py:163`), and `args.spec` is **silently ignored** — it is only read in the non-bridge branch at `model_provider.py:223`.
- But the disaggregated weight sync still converts with the hand-written `convert_qwen3_5_to_hf` (`update_weight_from_distributed/mixin.py:124` → `megatron_to_hf/__init__.py`), which only recognizes the names the *ignored* `--spec` plugin would have produced.

So `scripts/models/qwen3.5-27B.py` ships `--spec miles_plugins.models.qwen3_5 get_qwen3_5_spec`, whose `Qwen3_5GatedDeltaNet` nests everything under `self_attention.linear_attn.*`, while the model that actually gets built is the bridge's, which uses `self_attention.in_proj` / `out_norm` / `dt_bias` and zero-centered norms. The converter's accepted list and the built model have almost no names in common.

This looks like the same root cause as #634 (Qwen3-VL-8B-Instruct, still open), generalized: that report also fails in `update_weight_from_distributed` → `convert_to_hf`. Qwen3.5 adds the GDN parameters on top of the VL nesting.

## Repro

No GPU, no cluster and no checkpoint required — this only feeds parameter names through the converter. The names are copied verbatim from `megatron.bridge`'s own Qwen3.5 mapping table (the `megatron_param=` entries in `megatron/bridge/models/qwen/qwen35_bridge.py:177-231`), which is what defines the names a bridge-built model exposes.

```python
"""repro_qwen35_weight_sync.py"""
from types import SimpleNamespace

import torch

from miles.backends.megatron_utils.megatron_to_hf import _convert_to_hf_core

# Qwen3.8-27B / Qwen3.5-27B shapes, matching scripts/models/qwen3.5-27B.py.
ARGS = SimpleNamespace(
    num_experts=None, vocab_size=248320, mtp_num_layers=None, num_layers=64,
    hidden_size=5120, kv_channels=256, num_attention_heads=24, num_query_groups=4,
)
PARAM = torch.zeros(1)

BRIDGE_NAMES = [
    "module.module.decoder.layers.0.self_attention.in_proj.weight",
    "module.module.decoder.layers.0.self_attention.in_proj.layer_norm_weight",
    "module.module.decoder.layers.0.self_attention.out_norm.weight",
    "module.module.decoder.layers.0.self_attention.out_proj.weight",
    "module.module.decoder.layers.0.self_attention.A_log",
    "module.module.decoder.layers.0.self_attention.dt_bias",
    "module.module.decoder.layers.0.self_attention.conv1d.weight",
    "module.module.language_model.embedding.word_embeddings.weight",
    "module.module.vision_model.patch_embed.proj.weight",
]

for name in BRIDGE_NAMES:
    try:
        got = _convert_to_hf_core(ARGS, "qwen3_5", name, PARAM)
        print(f"ok         {name} -> {got[0][0] if got else '<dropped>'}")
    except ValueError as e:
        print(f"ValueError {e}")
```

Output — 9/9 rejected:

```
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.in_proj.weight
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.in_proj.layer_norm_weight
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.out_norm.weight
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.out_proj.weight
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.A_log
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.dt_bias
ValueError Unknown parameter name: module.module.decoder.layers.0.self_attention.conv1d.weight
ValueError Unknown parameter name: module.module.language_model.embedding.word_embeddings.weight
ValueError Unknown parameter name: module.module.vision_model.patch_embed.proj.weight
```

### End-to-end repro

Two-node fully-async RL (trainer and rollout engines on different hosts) on `Qwen/Qwen3.8-27B`, launched through `train_async.py` with `--megatron-to-hf-mode bridge --update-weight-transfer-mode broadcast --tensor-model-parallel-size 1`. Training itself is fine — the model builds, rollouts generate, and the first optimizer step completes. It dies at the first `update_weights`:

```
File "miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py", line 124,
    in _gather_and_update_non_expert_weights
    converted_named_tensors += convert_to_hf(...)
File "miles/backends/megatron_utils/megatron_to_hf/qwen3_5.py", line 193, in convert_qwen3_5_to_hf
    raise ValueError(f"Unknown parameter name: {name}")
ValueError: Unknown parameter name: module.module.vision_model.patch_embed.proj.weight
```

Patching the names one at a time walks through the set in stages: `vision_model.*`, then `language_model.*`, then the GDN parameters. To get the whole list in one run I instrumented the converter to log-and-drop instead of raising. After renaming `vision_model.* -> model.visual.*`, stripping `language_model.`, and inserting the `linear_attn.` level, exactly three names still raise:

| bridge parameter | HF target | rename enough? |
| --- | --- | --- |
| `self_attention.in_proj.layer_norm_weight` | `input_layernorm.weight` | yes |
| `self_attention.in_proj.weight` | `linear_attn.in_proj_{qkv,z,b,a}.weight` | no — one fused tensor → four (`GDNLinearMappingSeparate`) |
| `self_attention.out_norm.weight` | `linear_attn.norm.weight` | no — zero-centered RMSNorm, needs the `+1` (`RMSNorm2ZeroCenteredRMSNormMapping`) |

Worth noting that "the converter accepted it" is not the same as "the mapping is right". `conv1d.weight`, `out_proj.weight`, `A_log` and `dt_bias` all pass once renamed to `linear_attn.*`, but the bridge routes conv1d through `GDNConv1dMapping`, a `ChunkedMapping` that interleaves the q/k/v components across tensor-parallel shards. That is identity at TP=1, which is what I tested, so a hand-rename happens to work there and would silently diverge at TP>1. That is the main argument for reusing the bridge's table rather than growing the hand-written one.

## Why the colocated path is unaffected

`actor.py:253-257` picks `UpdateWeightFromTensor` when `--colocate` is set, and that path converts through `HfWeightIteratorBridge` (`update_weight_from_tensor.py:105` → `hf_weight_iterator_base.py:10-13`), i.e. it uses the bridge's own mapping table and gets all of the above right for free. The disaggregated path is the only one that reimplements conversion by hand.

That workaround is not reachable for async training, though: `train_async.py:23` asserts `not args.colocate` (`"Colocation is not supported for async training."`). So for a bridge-built architecture, correct weight mapping and fully-async disaggregated rollout are currently mutually exclusive.

## Suggested fix

Reusing the bridge's mapping table on the disaggregated path seems strictly better than extending the hand-written converter, since `HfWeightIteratorBridge` already exists and already handles the splits, reshapes and norm offsets:

1. Let `update_weight_from_distributed` use `HfWeightIteratorBase.create(...)` for `weight_type="base"`, as `update_weight_from_tensor` already does, so `--megatron-to-hf-mode bridge` implies bridge-based conversion on both paths. This would also cover #634.
2. Failing that, at least make the inconsistency loud instead of silent: if `args.spec` is set and ignored because bridge mode is on, warn or assert at startup. Right now `scripts/models/qwen3.5-27B.py` looks supported and fails ~10 minutes into a run.

Happy to put up a PR for (1) if that direction sounds right.

## Environment

- miles `fc04f666c08aebd72b241cef586a3939fdd6fa8e` (main, 2026-08-18)
- `megatron.bridge` with `Qwen35Bridge` / `Qwen35VLBridge` (`qwen35_bridge.py`, `qwen_vl/qwen35_vl_bridge.py`)
- sglang `0.5.17.dev36+gcb05a44f3`, transformers `5.12.1`
- Model: `Qwen/Qwen3.8-27B` (`Qwen3_5ForConditionalGeneration`, 64 layers = 48 `linear_attention` + 16 `full_attention`)
- 2 nodes x 8 MI355X, ROCm 7.2.0 (the failure is name-based and not hardware specific; it reproduces from the snippet above on CPU)

sunxxuns:
Fixes #2679. Also fixes #634, which has the same root cause.

## The problem

In bridge mode the model is built by `AutoBridge`, so its parameter names are the bridge's and not the ones `--spec` would have produced: `model_provider` only reads `args.spec` on the non-bridge branch, so the plugin is silently ignored.

`update_weight_from_distributed` nevertheless converted base weights with the hand-written converters in `megatron_to_hf/`, which only know the `--spec` names. Weight sync therefore dies with `ValueError: Unknown parameter name` on every architecture where the two disagree — Qwen3.5/3.6/3.8 (gated delta net) and Qwen3-VL — on the first sync, after the model has loaded and rollouts have been generated.

Renaming inside the converters would not have been enough. Some of these conversions are more than a rename, and the bridge already implements them: Qwen3.5 alone needs the fused `in_proj` split four ways, a zero-centered RMSNorm offset, and a TP-aware conv1d shard interleave.

## The change

Base weights now go through `HfWeightIteratorBase.create()`, which is what the colocated path in `update_weight_from_tensor` has always done. The two paths now agree on who owns the conversion, and bridge-built models are converted by the bridge that built them.

Because the bridge exports routed experts in the same walk and applies its own TP/EP sharding, it replaces the non-expert and expert gather passes rather than feeding them. Every rank iterates the bridge, since it runs collectives internally, and only the source rank transmits — the same rank handling `_update_lora_weights` already uses.

Scope: only runs passing `--megatron-to-hf-mode bridge` change behavior; `raw` stays on the existing converters. Since `raw` cannot build a model from an HF checkpoint at all (it asserts `--load`), bridge mode is what HF-checkpoint users are already on.

## Verification

2 nodes x 8 MI355X, Qwen3.8-27B, fully-async with TITO. Weight sync used to raise immediately; it now completes in 26.5s cold and ~2s in steady state, and the run trains through to completion.

The synced weights are correct, not merely accepted — the trainer's recomputed logprobs agree with the rollout engine's:

| metric | value |
| --- | --- |
| `train_rollout_logprob_abs_diff` | 0.0077 |
| `train_rollout_kl` | 0.00029 |
| `ess_ratio` | 1.000 |
| `ppo_kl` | 7.6e-09 |

No regression on a standard architecture. Qwen3-4B in bridge mode, new path vs. four runs of the previous behavior:

| | previous | this PR |
| --- | --- | --- |
| `logprob_abs_diff` (step 0) | 0.01113 - 0.01136 | 0.01139 |
| `rollout_kl` (step 0) | 0.000799 - 0.000848 | 0.000821 |
| `ess_ratio` | 1.000 | 1.000 |

Added `tests/fast/backends/megatron_utils/test_update_weight_bridge_base_sync.py` (8 tests) covering the routing decision in both modes, that bridge mode no longer reaches the raw gather paths, non-source ranks iterating without transmitting, and iterator reuse.

`tests/fast/backends/megatron_utils/` is 267 passed / 2 failed; both failures (`TestUpdateWeightsZeroChunks`) reproduce on unmodified `main` when the directory runs as a whole and pass when the file runs alone, so they are a pre-existing isolation issue in that suite and unrelated to this change.

## Note for reviewers

Two things surfaced next to this that are outside the scope of this PR, in case they are worth their own issues:

1. `args.spec` being accepted and silently ignored in bridge mode is what makes this class of bug quiet. Warning or erroring when both are passed would surface it at startup rather than at the first weight sync.
2. Megatron's GDN raises `NotImplementedError: GDN does not support packed sequence for now.`, so linear-attention models need `--qkv-format bshd` and cannot use `--use-dynamic-batch-size` — the same trade-off gpt-oss documents for sink attention. Worth documenting on the Qwen3.5/3.8 pages.

### DIFF
--- a/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
+++ b/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/mixin.py
@@ -95,6 +95,75 @@ class DistBucketedWeightUpdateMixin:
                 is_lora=True,
             )
 
+    @property
+    def _convert_base_via_bridge(self) -> bool:
+        """Whether base weights must be converted by the bridge rather than megatron_to_hf/.
+
+        In bridge mode the model is built by AutoBridge, so its parameter names
+        are the bridge's and not the ones ``--spec`` would have produced --
+        ``args.spec`` is ignored on that path. The hand-written converters only
+        know the ``--spec`` names, so they raise "Unknown parameter name" on
+        architectures where the two disagree (Qwen3.5/3.6/3.8 GDN, Qwen3-VL).
+
+        A property rather than state set in __init__, so it holds for all three
+        transports; p2p does not call _init_lora.
+        """
+        return self.args.megatron_to_hf_mode == "bridge"
+
+    def _base_hf_weight_iterator(self) -> HfWeightIteratorBase:
+        """The base-weight bridge iterator, created on first use.
+
+        Built lazily for the same reason: the transports construct themselves
+        differently, and LoRA runs already build their own adapter iterator in
+        _init_lora, which this must not replace.
+        """
+        iterator = getattr(self, "_hf_weight_iterator", None)
+        if iterator is None:
+            iterator = HfWeightIteratorBase.create(
+                args=self.args,
+                model=self.model,
+                model_name=self.model_name,
+                quantization_config=self.quantization_config,
+                is_lora=False,
+            )
+            self._hf_weight_iterator = iterator
+        return iterator
+
+    def _update_base_weights_via_bridge(
+        self,
+        update_bucket_weight_func: Callable[[list[tuple[str, torch.Tensor]], tqdm | None], None],
+        pbar: tqdm | None = None,
+    ) -> None:
+        """Convert and push base weights using megatron.bridge's mapping table.
+
+        Used whenever the model was built by the bridge, because only the bridge
+        knows that model's parameter names. It also owns conversions that are
+        more than a rename -- Qwen3.5 alone needs a fused ``in_proj`` split four
+        ways, a zero-centered RMSNorm offset, and a TP-aware conv1d shard
+        interleave -- so reimplementing them in megatron_to_hf/ would be
+        duplicating logic that already exists here.
+
+        This replaces both the non-expert and expert passes: the bridge exports
+        routed experts as part of the same walk, and applies its own TP/EP
+        sharding rather than consuming pre-gathered tensors.
+
+        Rank handling follows _update_lora_weights: every rank has to iterate
+        the bridge because it runs collectives internally, but only the source
+        rank transmits.
+        """
+        produced_any = False
+        for hf_named_tensors in self._base_hf_weight_iterator().get_hf_weight_chunks({}, weight_type="base"):
+            produced_any = True
+            if self._is_source:
+                update_bucket_weight_func(hf_named_tensors, pbar)
+
+        if not produced_any:
+            raise RuntimeError(
+                "Base weight sync produced zero chunks from the bridge, so no weights "
+                "reached the rollout engines. This usually means the Megatron-Bridge "
+                "version does not support this architecture."
+            )
+
     def _gather_and_update_non_expert_weights(
         self,
         update_bucket_weight_func: Callable[[list[tuple[str, torch.Tensor]], tqdm | None], None],
@@ -371,10 +440,15 @@ class DistBucketedWeightUpdateMixin:
             if not is_lora:
                 pbar = tqdm(desc=f"[{self._group_name}] Update weights", total=0) if self._is_source else None
 
-                self._gather_and_update_non_expert_weights(self._update_weight_implementation, pbar)
-                dist.barrier(group=get_gloo_group())
-                self._gather_and_update_expert_weights(self._update_weight_implementation, pbar)
-                dist.barrier(group=get_gloo_group())
+                if self._convert_base_via_bridge:
+                    # One pass: the bridge exports non-expert and expert params together.
+                    self._update_base_weights_via_bridge(self._update_weight_implementation, pbar)
+                    dist.barrier(group=get_gloo_group())
+                else:
+                    self._gather_and_update_non_expert_weights(self._update_weight_implementation, pbar)
+                    dist.barrier(group=get_gloo_group())
+                    self._gather_and_update_expert_weights(self._update_weight_implementation, pbar)
+                    dist.barrier(group=get_gloo_group())
 
             # Adapter weights: every iteration.
             if is_lora:
--- a/tests/fast/backends/megatron_utils/test_update_weight_bridge_base_sync.py
+++ b/tests/fast/backends/megatron_utils/test_update_weight_bridge_base_sync.py
@@ -1,0 +1,139 @@
+"""Base weight sync on the disaggregated path must follow --megatron-to-hf-mode.
+
+In bridge mode the model is built by AutoBridge, so its parameter names are the
+bridge's; the hand-written converters in megatron_to_hf/ only know the names
+``--spec`` would have produced. Converting bridge-built weights with them raises
+"Unknown parameter name" on architectures where the two disagree (Qwen3.5 GDN,
+Qwen3-VL), so bridge mode has to convert base weights with the bridge too.
+"""
+
+from argparse import Namespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+import torch
+
+from miles.backends.megatron_utils.update_weight.update_weight_from_distributed.mixin import (
+    DistBucketedWeightUpdateMixin,
+)
+
+_MIXIN_MODULE = "miles.backends.megatron_utils.update_weight.update_weight_from_distributed.mixin"
+
+
+class _Harness(DistBucketedWeightUpdateMixin):
+    """Minimal stand-in for the transports, which build themselves differently."""
+
+    def __init__(self, megatron_to_hf_mode, *, is_source=True, chunks=None):
+        self.args = Namespace(
+            megatron_to_hf_mode=megatron_to_hf_mode,
+            update_weight_buffer_size=1 << 30,
+            pipeline_model_parallel_size=1,
+        )
+        self.model = [MagicMock()]
+        self.model_name = "qwen3_5"
+        self.quantization_config = None
+        self.weight_version = 0
+        self.is_lora = False
+        self.multi_lora_adapters = None
+        self._group_name = "test"
+        self.__is_source = is_source
+        self.transmitted: list[list[tuple[str, torch.Tensor]]] = []
+
+        if chunks is not None:
+            iterator = MagicMock()
+            iterator.get_hf_weight_chunks.return_value = iter(chunks)
+            self._hf_weight_iterator = iterator
+
+    @property
+    def _is_source(self):
+        return self.__is_source
+
+    def _update_weight_implementation(self, named_tensors, pbar=None):
+        self.transmitted.append(named_tensors)
+
+
+_CHUNKS = [
+    [("model.layers.0.linear_attn.in_proj_qkv.weight", torch.zeros(2, 2))],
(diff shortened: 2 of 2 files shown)


Arist12:
Obsolete. #2747 moved distributed weight sync onto the unified iterator, which removes the `--spec`-name conversion this PR was working around: the bridge-mode path no longer routes bridge-built base weights through `convert_to_hf`. Nothing left to fix here.

Arist12:
Obsolete. #2747 moved distributed weight sync onto the unified iterator, which removes the `--spec`-name conversion this PR was working around: the bridge-mode path no longer routes bridge-built base weights through `convert_to_hf`. Nothing left to fix here.

PR #2684, opened 2026-08-21
[CI] Report the replay mismatch fraction as a ci metric

yueming-yuan:
The R3 replay checker only ever reported its mismatch count inside the `AssertionError`, so a passing run recorded nothing:

- zero mismatches returned early and logged nothing at all
- a non-zero count under the threshold logged at most ten per-token lines, never a total

So a check drifting toward its threshold was invisible until it tripped, and two runs that both passed could not be compared.

`BaseReplayManager` now accumulates mismatched/checked token counts, `log_train_metrics` all-reduces them across ranks and emits `ci/r3_mismatch_fraction`, and that key is registered in `TARGET_METRIC_KEYS` so the history gate records it per run. The metric only appears when the checker is on (R3 enabled and `--ci-test`).

Every rank pops the counters even though only rank 0 logs, otherwise the non-logging ranks would accumulate across steps. The checker also runs in the backward stage, whose counts land in the following microbatch; the per-step aggregate is unaffected.

Found while debugging a qwen3.5 routing regression, where two runs differed in mismatch count but only the failing one reported a number.

### DIFF
--- a/miles/backends/training_utils/log_utils.py
+++ b/miles/backends/training_utils/log_utils.py
@@ -11,7 +11,6 @@ from miles.utils import train_metric_utils
 from miles.utils.flops_utils import fwd_tflops_per_gpu
 from miles.utils.ft_utils.process_group_utils import MultiPGUtil
 from miles.utils.metric_utils import compute_rollout_step
-from miles.utils.replay_base import routing_replay_manager
 from miles.utils.tracking_utils.structured_log import log_structured
 from miles.utils.types import RolloutBatch
 
@@ -501,6 +500,7 @@ def log_train_step(
     role: str = "actor",
     extra_metrics: dict[str, float] | None = None,
     should_log: bool | None = None,
+    r3_mismatch_fraction: float | None = None,
 ) -> dict[str, float]:
     """Log training metrics for one step.
 
@@ -516,6 +516,7 @@ def log_train_step(
         role: Role name (e.g., "actor", "critic").
         extra_metrics: Optional extra metrics to log (e.g., learning rates, MTP loss).
         should_log: Optional override for logging condition. If None, uses rank == 0.
+        r3_mismatch_fraction: Optional pre-aggregated routing-replay mismatch fraction.
 
     Returns:
         The formatted log_dict (for CI tests or other uses).
@@ -533,11 +534,8 @@ def log_train_step(
         for key, val in extra_metrics.items():
             log_dict_out[f"train/{role_tag}{key}"] = val
 
-    if routing_replay_manager.enable_check_replay_result:
-        # log_train_step runs on one rank only, so no collective here: the counts are
-        # this rank's, and every rank runs the same layers.
-        mismatched, checked = routing_replay_manager.pop_check_stats()
-        log_dict_out[f"ci/{role_tag}r3_mismatch_fraction"] = mismatched / checked if checked else 0.0
+    if r3_mismatch_fraction is not None:
+        log_dict_out[f"ci/{role_tag}r3_mismatch_fraction"] = r3_mismatch_fraction
 
     log_dict_out["train/step"] = accumulated_step_id
 
--- a/miles/backends/training_utils/log_utils.py
+++ b/miles/backends/training_utils/log_utils.py
@@ -11,6 +11,7 @@ from miles.utils import train_metric_utils
 from miles.utils.flops_utils import fwd_tflops_per_gpu
 from miles.utils.ft_utils.process_group_utils import MultiPGUtil
 from miles.utils.metric_utils import compute_rollout_step
+from miles.utils.replay_base import routing_replay_manager
 from miles.utils.tracking_utils.structured_log import log_structured
 from miles.utils.types import RolloutBatch
 
@@ -531,6 +532,12 @@ def log_train_step(
     if extra_metrics:
         for key, val in extra_metrics.items():
             log_dict_out[f"train/{role_tag}{key}"] = val
+
+    if routing_replay_manager.enable_check_replay_result:
+        # log_train_step runs on one rank only, so no collective here: the counts are
+        # this rank's, and every rank runs the same layers.
+        mismatched, checked = routing_replay_manager.pop_check_stats()
+        log_dict_out[f"ci/{role_tag}r3_mismatch_fraction"] = mismatched / checked if checked else 0.0
 
     log_dict_out["train/step"] = accumulated_step_id
 
--- a/miles/utils/replay_base.py
+++ b/miles/utils/replay_base.py
@@ -4,11 +4,21 @@ import os
 import torch
 import torch.distributed as dist
 
+from miles.utils.ft_utils.process_group_utils import GeneralPGUtil
+
 logger = logging.getLogger(__name__)
 
 
 def _get_rank():
     return dist.get_rank() if dist.is_initialized() else 0
+
+
+def reduce_check_stats(stats: tuple[int, int], group: dist.ProcessGroup) -> tuple[int, int]:
+    """Sum replay-check counters across a rank-symmetric Gloo group."""
+    values = torch.tensor(stats, dtype=torch.long)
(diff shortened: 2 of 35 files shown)


guapisolo:
@claude review

miles/backends/training_utils/log_utils.py
```diff
@@ -532,6 +533,12 @@ def log_train_step(
         for key, val in extra_metrics.items():
             log_dict_out[f"train/{role_tag}{key}"] = val
 
+    if routing_replay_manager.enable_check_replay_result:
+        # log_train_step runs on one rank only, so no collective here: the counts are
+        # this rank's, and every rank runs the same layers.
+        mismatched, checked = routing_replay_manager.pop_check_stats()
```
guapisolo: [P1] Aggregate and reset replay stats from a rank-symmetric path

Removing the collective from this logger avoids the previous hang, but popping only this process makes the metric non-global. Megatron calls `log_train_step()` only on `is_first_replica_megatron_main_rank()`; PP stages own different local layer ranges, so this omits the other stages/token shards and their counters are never drained. In FSDP, every rank calls `log_train_step()`, but only rank 0 publishes, so the recorded value still represents only rank 0's data shard. Snapshot/pop on every participating rank at a guaranteed step/attempt boundary, reduce `(mismatched, checked)` over the correct training group, and only then let the single logging rank publish the aggregate.

miles/utils/tracking_utils/ci_history.py
```diff
@@ -53,6 +53,7 @@
     "rollout/raw_reward",
     "rollout/tito_session_mismatch_rate/v1/assistant_text",
     "rollout/tito_session_mismatch_rate/v2/assistant_text",
+    "ci/r3_mismatch_fraction",
```
guapisolo, line 56: [P1] Declare this metric in the per-test gate specs

Adding the key to `TARGET_METRIC_KEYS` only captures it in the run JSONL. `evaluate_gate()` emits metrics only for `register_ci_gate(...)` specs parsed from each concrete test file; this PR adds neither a default nor an R3 test spec, so `result.metrics` omits the key and the baseline writer never stores it in `metric_values`. Add an explicit `steps`/`constraint` registration to every intended R3 e2e test (putting it only in shared `_common.py` is not seen by the AST parser), plus a persistence regression test.

miles/utils/replay_base.py
```diff
@@ -191,6 +200,8 @@ def check_replay_result(self, old_topk_fn, scores, topk, top_indices, *args, **k
         is_mismatch = (overlap < required) & ~is_padding
 
         mismatch_count = is_mismatch.sum().item()
+        self.checked_tokens += int((~is_padding).sum().item())
```
guapisolo: [P2] Use the same denominator as the checker

This metric excludes all-`-1` padding rows, while the checker below still compares against `threshold * orig_flat.shape[0]`. For example, 50 valid rows + 50 padding rows + 1 mismatch passes the existing 1% checker (`1 > 1` is false) but reports 2%, so the metric does not represent distance to the actual gate and changes with padding ratio. Preserve the checker contract by counting `orig_flat.shape[0]` here as well, and add a padding case.

miles/utils/replay_base.py
```diff
@@ -64,6 +64,15 @@ def __init__(self):
         self.enabled = False
         self.stage = "fallthrough"
         self.register_replay_list_func = None
+        self.checked_tokens = 0
+        self.mismatched_tokens = 0
+
+    def pop_check_stats(self) -> tuple[int, int]:
```
guapisolo, line 80: [P2] Reset counters when a training attempt is discarded

These stats are cleared only by `pop_check_stats()` during normal logging. An independent-DP attempt can run `check_replay_result()`, return `DISCARDED_SHOULD_RETRY`, skip `log_train_step()`, and then call `clear_all()`, which does not clear these counters. The next successful retry therefore includes the discarded attempt and double-counts work. Make the stats attempt-scoped and reset them in guaranteed cleanup on every rank, discarding them for non-`NORMAL` attempts.

guapisolo (commented):
Re-reviewed the current head (`aab3135e`). The previous default-WORLD collective hang has been removed, but the replacement still has four correctness gaps inline. The relevant CUDA/ROCm R3 suites are currently skipped, and this diff adds no regression coverage.

guapisolo:
Publication note for this follow-up commit:

This push will move R3 checker snapshot/reset and reduction to rank-symmetric cell boundaries, align the metric denominator with the checker contract, and register `ci/r3_mismatch_fraction` in the concrete per-test history gates.

Candidate `a7f84c7aca9097c529beca4e97029b74e996317e` passed 156 targeted fast/CI tests plus representative Megatron Qwen, FSDP Qwen, and Megatron GLM47 MTP R3 E2Es on `miles-h200-dev-2`. The Megatron reduction will remain cell-local by design: independent-DP cells can diverge on discarded attempts, so a cross-cell collective at this boundary would be unsafe.

The existing PR description will remain unchanged.

miles/backends/training_utils/log_utils.py
```diff
@@ -532,6 +533,14 @@ def log_train_step(
         for key, val in extra_metrics.items():
             log_dict_out[f"train/{role_tag}{key}"] = val
 
+    if routing_replay_manager.enable_check_replay_result:
+        counts = torch.tensor(
+            routing_replay_manager.pop_check_stats(), dtype=torch.float64, device=torch.cuda.current_device()
+        )
+        dist.all_reduce(counts)
+        mismatched, checked = counts.tolist()
+        log_dict_out[f"ci/{role_tag}r3_mismatch_fraction"] = mismatched / checked if checked else 0.0
```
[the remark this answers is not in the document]
guapisolo: Current head removed the WORLD collective from `log_train_step()`, so the deadlock path described here no longer exists. The remaining global aggregation issue is tracked separately.

miles/backends/training_utils/log_utils.py
guapisolo: Fixed in `a7f84c7ac`. Megatron now pops checker counters on every rank immediately after `train_one_step()`, discards non-`NORMAL` attempt stats without reducing, and reduces successful stats over the rank-symmetric cell-local Gloo group before only the first-replica main rank publishes. FSDP likewise pops/reduces on every rank at the optimizer-step boundary while only rank 0 tracks. Fast regressions cover discarded and non-main paths, and the representative Megatron/FSDP R3 E2Es completed successfully.

miles/utils/tracking_utils/ci_history.py
guapisolo: Fixed in `a7f84c7ac`. `ci/r3_mismatch_fraction` now has an `all`-steps gate default, all 21 concrete routing-R3 E2Es register the metric (with the GLM47 5% override where required), and the integration regression verifies that captured values are persisted as `metric_values` rows.

miles/utils/replay_base.py
guapisolo: Fixed in `a7f84c7ac`. `check_replay_result()` now uses `orig_flat.shape[0]` as `checked_count` for both the accumulated metric denominator and the checker threshold, preserving the existing pass/fail contract. The regression includes one mismatch plus one padding row and asserts `(1, 2)`. Revalidated on the exact seq-4 devbox candidate: the targeted replay/lifecycle suite passed (`36 passed`).

miles/utils/replay_base.py
guapisolo: Fixed in `a7f84c7ac`. Megatron now calls `pop_check_stats()` on every rank immediately after `train_one_step()` and before the outcome branch; `DISCARDED_SHOULD_RETRY` drops those stats without reduction or logging, so retries cannot inherit counters. NORMAL non-main ranks still participate in the cell-local reduction. FSDP likewise pops/reduces at the rank-symmetric optimizer-step boundary. Revalidated on the exact seq-4 devbox candidate: the targeted replay/lifecycle suite passed (`36 passed`).

PR #2686, opened 2026-08-21
fix: treat temperature=0 as greedy selection, untempered logprobs

EazyReal:
Twin of THUDM/slime#2197.

## Problem

`get_responses` scaled logits by `rollout_temperature` whenever it was not 1.0 **and** `> 0`. That silently skips T=0 (avoids divide-by-zero) without stating the contract.

SGLang treats T=0 as greedy token selection, but still returns finite untempered model log-probs. slime/Miles store those as `rollout_log_probs` and build PPO ratios from `exp(log π_θ − log π_old)`. Value heads already skip scaling because they have last-dim 1.

## Change

`apply_rollout_temperature` is the shared rule:

- T in {0, 1}: leave logits unchanged
- otherwise: `logits / T`

Behavior at T=0 matches the previous `> 0` guard; the rule is now explicit, same as SGLang.

## Test plan

- [ ] T=0 and T=1 leave policy logits unchanged; T=0.5 divides
- [ ] Value heads (`last dim == 1`) still skip temperature

### DIFF
--- a/miles/backends/training_utils/loss_hub/logit_processors.py
+++ b/miles/backends/training_utils/loss_hub/logit_processors.py
@@ -8,6 +8,12 @@ from miles.backends.training_utils.loss_hub.math_utils import calculate_log_prob
 from miles.backends.training_utils.parallel import get_parallel_state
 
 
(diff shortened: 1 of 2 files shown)


EazyReal:
try to port the changes from slime, but already handled here.

PR #2691, opened 2026-08-21
fix(deepseekv3): match router dtype for disk-delta export

nanjiangwill:
## Summary

- preserve the FP32 Megatron router for existing weight-transfer modes
- cast the DeepSeek/GLM router to the configured BF16 or FP16 dtype when exporting raw disk-delta bytes
- cover BF16, FP16, non-disk-delta, and expert-bias behavior

## Why

Megatron keeps the MoE router weight in FP32, while the canonical HF checkpoint stores the gate weight in the model dtype. Disk-delta applies byte-level deltas to that canonical checkpoint, so its exported tensor must use the checkpoint layout. The expert correction bias remains FP32 and is intentionally unchanged.

## Test plan

- `python -m pytest tests/fast/backends/megatron_utils/test_deepseekv3_to_hf.py -q -p no:cacheprovider`
- pre-commit hooks on changed files

### DIFF
--- a/miles/backends/megatron_utils/megatron_to_hf/deepseekv3.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/deepseekv3.py
@@ -3,6 +3,16 @@ import re
 import torch
 
 
+def _convert_router_weight(args, param):
+    if getattr(args, "update_weight_transfer_mode", None) != "disk-delta":
+        return param
+    if getattr(args, "bf16", False):
+        return param.to(torch.bfloat16)
+    if getattr(args, "fp16", False):
+        return param.to(torch.float16)
+    return param
+
+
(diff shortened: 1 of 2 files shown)


nanjiangwill:
Closing after rechecking the current Megatron and Miles conversion paths. The router weight already follows `params_dtype`; `moe_router_dtype=fp32` controls routing computation rather than stored parameter dtype, while the expert correction bias intentionally remains FP32. The disk-delta-only, model-specific cast therefore duplicates conversion policy and is a no-op for the GLM BF16 configurations it was intended to fix. PR #2692 provides the correct transport-layer behavior: validate the emitted tensor against the canonical checkpoint dtype, shape, and byte layout and fail before publishing if a real mismatch exists.

PR #2701, opened 2026-08-21
feat(ci): declare supported GPU generations on every CUDA test

yueming-yuan:
**Chain 2/5** — stacked on `b200-ci-runner`. Review that one first.

`register_cuda_ci` gains a required `hardware=[...]` listing the GPU generations whose kernels and precision paths a test supports.

Required rather than defaulted: a default derived from the suite name would stamp "hopper only" on every existing test, which is false for most of them and defers the real judgement to a sweep that never happens.

Each of the 99 CUDA registrations was classified by what it actually exercises:

| | |
|---|---|
| 4 × `["blackwell"]` | nvfp4 / mxfp8 kernel paths |
| 6 × `["hopper"]` | 4× `--sglang-attention-backend fa3` (sm90-only), `ScriptArgs(hardware="H200")` in the v3.2 fp8 test, and the MXFP8 quantizer parked on an H200 reference |
| 89 × `["hopper", "blackwell"]` | no arch-gated code path |

Grepping for signals alone gets three of these wrong:

- `test_inkling.py` matches `NVFP4` only inside the checkpoint name `thinkingmachines/Inkling-Small-NVFP4`, and runs on H200 today.
- `test_mxfp8_quantizer.py` is disabled waiting on an **H200** reference path, not on Blackwell hardware — the opposite conclusion from its sibling `test_nvfp4_quantizer.py`.
- The nvfp4 quantizer and the glm5.2 nvfp4 test needed their home suite moved to `stage-c-8-gpu-b200` to satisfy the home-stage invariant.

`tests/ci/hardware.py` holds the arch vocabulary and the suite → arch map. Validation rejects an empty or unknown arch list, a duplicate, `hardware` on a CPU or ROCm registration, an unknown CUDA suite, and a home stage that is not on the test's first supported arch — so "where a test lives" and "where it runs by default" cannot disagree. The unknown-suite check also closes the hole that let `stage-c-8-gpu-b200` sit in test files for months with no workflow job behind it.

**No behaviour change**: nothing reads `hardware` until 4/5.


guapisolo:
codex cmt

I will publish `3774ed4ccb9a67dd0b2b67f4bf847d0c48c3a816` to resolve the P1 integration failure: CUDA tests added on `main` otherwise lack the required `hardware` declaration, so full CI discovery exits before suite filtering and blocks CPU as well as GPU runs. This publication will preserve current `main` test metadata and carry an exact candidate that passed full discovery (551 files / 593 registrations), a CPU `--list-only` plan, and 177 focused CI registry tests.

guapisolo:
codex cmt

Prospective tip `452b5f9a1668f30bf9dda2b1f3449f6f251f4053` will apply Black 24.3.0's formatting to the three merged CUDA registrations so the repository-wide pre-commit check accepts the merge result. `pre-commit run --all-files --show-diff-on-failure --color=always` passes against this candidate.

guapisolo (approved):

PR #2706, opened 2026-08-21
DeepSeek-V4 support-v2: native megatron implementation as option

yueming-yuan:
ci-megatron-pr: dsv4-dual-backend

## What

Bring DeepSeek-V4 back on the bumped Megatron, and make the training implementation an explicit
choice. Second of the three bump PRs; stacked on #2673 (base bump), which disabled DSv4.

`--dsv4-impl {miles,megatron}` selects a whole path, at config+spec build time:

| | `miles` (default) | `megatron` |
|---|---|---|
| attention variant | `dsv4` (ported PR #28 slice) | `dsv4_hybrid` (dev-native) |
| mHC | miles hyper-connections | native `enable_hyper_connections` |
| kernels | tilelang, BSHD + sparse CP | cudnn / none |
| tensor parallelism | yes | TP must be 1 |

Unsupported combinations raise, they do not fall back — June's `MILES_DSV4_ATTENTION_BACKEND`
silently forced miles on TP>1 and that behaviour is retired.

## Megatron side — `radixark/Megatron-LM:dsv4-dual-backend`

`miles-main-20260819` + 6 commits: the miles DSv4 attention variant, hash-routing under TP + MoE
freeze knobs, and four fixes found by running it:

- `tid2eid` was a buffer, not a parameter — miles' weight sync materializes full tensors from the
  parameter buffer, so it pushed an all-zero routing table into sglang and the engine became a
  different model. The weight checker caught this on the first run.
- the shared-expert SwiGLU clamp opt-out (`activation_func_clamp_shared_expert`) was lost when the
  mHC revert landed, so training clamped what DeepSeek-V4 and the engine leave unclamped.
- rollout routing replay registered on MTP routers, which desyncs the replay streams.
- three sites in `csa.py` read `Optional` `dsa_indexer_loss_coeff` with
  `getattr(..., 0.0)` — whose default only fires when the attribute is *missing* — so `None`
  reached `kl_div * loss_coeff`. Two other sites in the same file already did it correctly.

## Miles side (this PR)

- `--dsv4-impl`, with the arg translation and the three hard-errors, plus its unit test.
- The plugin's weights now carry Megatron's names and layout, so one HF checkpoint loads into both
  impls; mbridge, the quantizers, the weight checker and the converters follow the rename.
- fp32-only params use Megatron's `mark_keep_in_fp32`.
- The DSv4 4-layer e2e is re-enabled (it was `disabled=` in #2673).
- 4-layer conversion runs at PP1 — the bumped Megatron asserts hash-MoE + PP>1 needs an explicit
  `pipeline_model_parallel_layout`.

## Verification (devbox, 8xH200)

**4-layer RL e2e, default thresholds, no skip list, no relaxed gate** — matches the pre-bump stack
(miles `49cac7b52` + megatron `2e64d788b`):

| | pre-bump | this PR |
|---|---|---|
| rollout 0 / 1 train-rollout gap | 0.0103 / 0.0094 | 0.0103 / 0.0099 |
| `train_rollout_logprob_abs_diff` step 0 / 1 | 0.0833 | 0.0858 / 0.0836 |
| `train_rollout_kl` step 0 / 1 | 0.00927 | 0.00973 / 0.00942 |
| `found NaN` | 0 | 0 |
| R3 mismatch tokens | 0 | 0 |
| job | succeeded | succeeded |

**Both impls, same recorded rollout, same args, TP=1:**

| | miles impl | megatron impl |
|---|---|---|
| rollout logprob (input) | -1.6760895252227783 | -1.6760895252227783 |
| train-forward logprob | -1.6852522790431976 | -1.6873207092285156 |
| `train_rollout_logprob_abs_diff` | 0.0837 | 0.0885 |
| job | succeeded | succeeded |

Native mHC and miles mHC agree to 2.1e-3. Converting with `--dsv4-impl megatron` gives
27,390,526,557 parameters, identical to the miles impl.

**Not covered:** every sample in that rollout has reward 0, so GRPO advantage and `grad_norm` are 0
on both impls — the backward ran clean but its numerics are untested. Also untested: TP>1 (rejected
by design), the cudnn kernel backend (needs `flash_mla`, absent from the image), multi-step runs of
the megatron impl.

Full investigation, including the wrong turns, in `bump_docs/05-dsv4-dual-backend.md`.

## 50-step A/B on 8×GB300 (32 GPU, sm103)

Identical config for both runs — TP1 · PP8 (first 4 / last 3 layers) · CP1 · EP4 · DP4,
max-tokens-per-gpu 4096, FP8 blockwise, fp32 optimizer, GRPO on DAPO-math-17k (32 prompts × 8
samples, max response 4096, lr 1e-6, deterministic mode), cudnn kernel backend for conversion and
training — only `--dsv4-impl` differs. megatron = job 2695 (50 steps), miles = job 2703 (54 steps).

**Timing, p50 over steps 2+** (steps 0–1 excluded: per-shape JIT warm-up). Full cycle = wall-clock
between consecutive rollout completions.

| phase | megatron impl | miles impl | Δ |
|---|---|---|---|
| train | 82.6 s | 79.0 s | −4.4% |
| log_probs | 23.2 s | 22.9 s | −1.3% |
| update_weights | 38.4 s | 38.3 s | −0.3% |
| full cycle | 200.5 s | 197.0 s | −1.7% |

| | megatron impl · 2695 | miles impl · 2703 |
|---|---|---|
| steps | 50 | 54 |
| reward, mean | 0.763 | 0.765 |
| `train_rollout_kl`, mean | 0.0270 | 0.0254 |
| logprob diff (train − rollout), mean | −0.0313 | −0.0295 |
| logprob diff, range | −0.058 … −0.011 | −0.066 … −0.012 |

reward

train_rollout_kl


miles_plugins/models/deepseek_v4/ops/v4_indexer.py
```diff
@@ -69,6 +71,11 @@ def __init__(self, config: TransformerConfig, pg_collection=None):
             cp_group=pg_collection.cp,
         )
 
+        # RL rollout-routing-replay (R3) seam for the sparse-attention indexer topk: lets the miles
+        # indexer_replay_manager record (on rollout) / replay (on train forward) the top-k KV picks,
+        # mirroring the MoE routing-replay seam. No-op unless the manager is enabled.
+        indexer_replay_manager.register_to_module(self, "indexer_replay", stream_idx=layer_id)
```
Zhichenzzz, line 77: (maybe later we should opt this indexer replay, since now it is easy to cause host oom / and long-latency. and this will be very help for the following models, e.g., glm, qwen, ...

miles_plugins/mbridge/deepseekv4.py
Zhichenzzz, line 1: rename the file into deepseek_v4?

miles/backends/megatron_utils/megatron_to_hf/deepseekv4.py
```diff
@@ -7,54 +7,90 @@ def get_deepseek_v4_atomic_update_groups():
...
+                    ".self_attention.core_attention.indexer.compressor.linear_wkv.weight",
+                    ".self_attention.core_attention.indexer.compressor.linear_wgate.weight",
                 ),
             ),
         ]
     ]
 
 
-def convert_deepseekv4_to_hf(args, name, param):
+def _packed_alphas(name: str, param, bucket):
```
Zhichenzzz, line 45: claude code says this will harm async train, but i dont agree (maybe quick check

miles/backends/megatron_utils/megatron_to_hf/processors/quantizer_fp8.py
Zhichenzzz, line 96: old names?

scripts/run_deepseek_v4.py
Zhichenzzz, line 2: do we also need to modify amd script into the latest impl?

Zhichenzzz (approved):
Most looks good! just small comments

miles/backends/megatron_utils/megatron_to_hf/processors/quantizer_fp8.py
yueming-yuan: good catch

miles/backends/megatron_utils/megatron_to_hf/deepseekv4.py
yueming-yuan: just checked, this should not harm the training

Issue #2705, opened 2026-08-21
[AMD][LoRA] AMD LoRA RL support tracker


Arist12:
# [AMD][LoRA] AMD LoRA RL support tracker

Bugs hit while bringing LoRA RL up on AMD. Each PR is self-contained — reproduction against `main` and an explanation of the fix live in its own description.

The AMD enablement itself is already merged upstream: [SGLang #34702](https://github.com/sgl-project/sglang/pull/34702) and [#36379](https://github.com/sgl-project/sglang/pull/36379) make the MoE-LoRA align kernel JIT-compile under HIP, and [Megatron-Bridge #28](https://github.com/radixark/Megatron-Bridge/pull/28) supplies the ROCm Transformer Engine grouped-linear contract fields per-expert MoE LoRA needs.

## Checkpoint and resume

- [#2669](https://github.com/radixark/miles/pull/2669) — every DP/CP replica writes its own adapter shard and the filename is keyed on the global rank, so resuming at a different world size silently leaves some ranks on a fresh adapter. Writes one shard per realized `(TP, PP, EP, ETP)` coordinate instead.
- [#2516](https://github.com/radixark/miles/pull/2516) — the distributed optimizer's fp32 masters and Adam moments are never checkpointed, so a resumed run either steps straight into NaN or has the restored adapter overwritten on its first step. Saves and restores the parameter state, and refreshes the masters on every path that writes adapter weights.
- [#2733](https://github.com/radixark/miles/pull/2733) — the native loader reports success for a shard that matches no parameter, is missing tensors, or holds NaN. Validates before the weight copy, and rejects virtual pipeline chunks, whose colliding parameter names the name-keyed shard format cannot represent.
- [#2580](https://github.com/radixark/miles/pull/2580) — the exported PEFT adapter carries none of the `base_model.model.` prefixes PEFT matches on, so external loaders silently get an identity adapter. Delegates serialization to Megatron-Bridge and promotes the staged export weights-first.
- [#2661](https://github.com/radixark/miles/pull/2661) — a run relaunched with `--lora-adapter-path` restarts at rollout 0, re-steps the LR schedule and never finds the dataset cursor, because the LoRA save path writes no `latest_checkpointed_iteration.txt` and the bridge branch repoints `--load`. Treats a standard `iter_*/adapter` path as a resume; anything else stays an explicit warm start.
[...]

PR #2708, opened 2026-08-22
fix(lora): let a resumed disaggregated run reach the rollout engine

Arist12:
Part of #2705.

## Problem

A disaggregated LoRA run resumed with `--lora-adapter-path` can fail before its first rollout: native-only checkpoints cannot be preloaded by SGLang, partial best-effort PEFT output can look preloadable, and a preloaded adapter collides with the mandatory initial publish.

## Change

Normal train drivers start resumed rollout engines from the base model and let the mandatory initial weight publish install the trainer's restored adapter. Publish single LoRA adapters with SGLang's existing `upsert=True` contract so fresh startup and later refreshes use one path without unload churn. `--debug-rollout-only`, which skips the publish, remains a blocker before this PR is ready.

## Validation

- 28 focused fast tests covering complete, partial and native-only resume directories
- Full Qwen3-30B-A3B, 4 trainer + 4 rollout MI350X: restored four EP shards (`384` tensors each), optimizer, scheduler and global-dataset cursor from `iter_29`; continued at rollout 30, completed two optimizer steps, saved `iter_30`, and republished the trained adapter (`grad_norm=0.0152`, train/rollout abs diff `0.0210`-`0.0237`)
- Full Qwen3-4B, 4+4 MI350X: a 30-rollout run resumed at `iter_4` and `iter_14`, restored `288` tensors plus optimizer/scheduler/data cursor, and completed through `iter_29`
- Dense held-out DAPO accuracy `0.711 → 0.789`, truncation `0.211 → 0.078`; train/rollout abs diff mean `0.01265`, max `0.01419`
- Checkpoints at 4/14/29 each contain one finite TP0/PP0 native shard and distinct adapter hashes

### DIFF
--- a/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
+++ b/miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
@@ -127,7 +127,7 @@ class UpdateWeightFromDistributed(DistBucketedWeightUpdateMixin):
             pbar.update(1)
 
     def _update_lora_weight_implementation(self, named_tensors: list[tuple[str, torch.Tensor]]) -> None:
-        """Send adapter metadata over Ray, then broadcast the tensors (src=0).
+        """Upsert adapter metadata over Ray, then broadcast the tensors (src=0).
 
         Reuses the base broadcast group (``self._model_update_groups`` /
         ``self._group_name``); base and adapter syncs are strictly sequential, so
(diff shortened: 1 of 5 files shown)


Arist12:
Closing against the refactor. The `update_weight/` half of this patch no longer exists — #2754/#2756 deleted the unload-then-reload contract it modified and replaced it with register-then-stream. The one issue that survives is unrelated to weight sync: `sglang_engine.py` still passes `--lora-adapter-path` straight through as a PEFT `lora_paths` entry at engine startup, which cannot work for a directory holding miles' native per-rank shards. That is SGLang-LoRA integration code, so it will come back as its own PR once that refactor lands.

PR #2709, opened 2026-08-22
fix(rollout): handle invalid tool arguments

### LINKED ISSUE #2689: fix(rollout): handle malformed and non-object tool arguments before execution
## Problem

At `819ba9650e4951c5001e08f91c972bd7776456e7`, the built-in `multi_turn` path parses `ToolCallItem.parameters` with `json.loads()` immediately before invoking the user-provided executor.

That path has two failure modes for stochastic model output:

- malformed JSON raises an uncaught `JSONDecodeError` and aborts generation;
- valid JSON that is not an object (for example a list or number) is passed to an executor whose argument contract is a parameter mapping.

In both cases the tool-call envelope was recognized, but there is no explicit invalid-call outcome between parsing and execution.

## Minimal Reproduction

Run in a normal Miles development environment:

```python
import asyncio

from sglang.srt.function_call.core_types import ToolCallItem

from miles.rollout.generate_utils.tool_call_utils import _execute_tool_call


async def exercise(raw: str) -> None:
    calls = []

    async def execute_one(name: str, params: dict) -> str:
        calls.append((name, params, type(params).__name__))
        return "ok"

    call = ToolCallItem(tool_index=0, name="bash", parameters=raw)
    try:
        result = await _execute_tool_call(call, execute_one)
        print(raw, "result=", result, "calls=", calls)
    except Exception as exc:
        print(raw, "raised=", type(exc).__name__, str(exc), "calls=", calls)


async def main() -> None:
    for raw in ('{"cmd":"pytest"}', '{"cmd":"pytest"', '[1, 2]', '42'):
        await exercise(raw)


asyncio.run(main())
```

Observed on the exact `main` commit:

| arguments | result |
|---|---|
| `{"cmd":"pytest"}` | executor receives `dict` |
| `{"cmd":"pytest"` | uncaught `JSONDecodeError`; executor is not called |
| `[1, 2]` | executor is called with `list` |
| `42` | executor is called with `int` |

Relevant path:

- [`multi_turn.generate`](https://github.com/radixark/miles/blob/819ba9650e4951c5001e08f91c972bd7776456e7/miles/rollout/generate_hub/multi_turn.py#L68-L75) parses the model text and immediately dispatches calls.
- [`_execute_tool_call`](https://github.com/radixark/miles/blob/819ba9650e4951c5001e08f91c972bd7776456e7/miles/rollout/generate_utils/tool_call_utils.py#L38-L52) assumes `json.loads()` succeeds and yields the executor's mapping.

## Expected Behavior

A recognized call should have an explicit argument-parse outcome before any effect:

- JSON object: invoke the executor with the parsed mapping.
- malformed or non-object JSON: do not invoke the executor and do not let the exception escape the rollout.
- unrecognized tool-call envelope: keep the existing no-tool/assistant-output behavior.

My recommended built-in `multi_turn` policy is to append a deterministic tool-error response for the recognized invalid call, preserve the sampled model output/tokens, and allow the model to correct itself on the next turn.

Would maintainers prefer that behavior, immediate turn termination, or marking the sample aborted? The choice affects the conversation presented to the model, so I wanted to confirm it before proposing a narrow fix.

## Scope

This is separate from #1327, which hardens echoed history only at the chat-template render boundary. A fix here should not introduce render sentinels as executable kwargs and should not change valid-call behavior.

MinhaoLi0318:
## Summary

  - Parse arguments for recognized SGLang and OpenAI tool-call envelopes before execution.
  - Invoke the executor only when the decoded arguments are a JSON object.
  - Return a deterministic tool-error response for malformed or non-object JSON so the rollout can continue and the model can correct the call on the next turn.
  - Preserve existing behavior for valid objects, empty arguments, sampled assistant output/tokens, and envelope recognition.

  ## Behavior

  - **JSON object:** Call the executor with the parsed mapping and return the existing tool response.
  - **Empty string:** Call the executor with an empty mapping and return the existing tool response.
  - **Malformed JSON:** Do not call the executor; return a deterministic tool-error response.
  - **JSON list or number:** Do not call the executor; return a deterministic tool-error response.

  ## Tests

  - Added parameterized regression coverage for both ToolCallItem and ChatCompletionMessageToolCall.
  - Targeted regression cases: 10 passed.
  - Pre-commit, Ruff, Black, isort, autoflake, and py_compile passed.
  - The full CPU suite is left to PR CI because the local macOS environment does not contain the pinned SGLang/PyTorch source stack used by Miles CI.

  Fixes #2689

@yueming-yuan


MinhaoLi0318:
@guapisolo @yueming-yuan  please help take a look when you have time~

PR #2713, opened 2026-08-22
[multi-lora] 1/n - 7, test: pin update_weight_version abort_all_requests=False regression

yushengsu-thu:
PR #2589 changed the engine's weight-version bump to send `abort_all_requests: False` so in-flight requests survive a weight update, but the behavior landed without a test. This adds a fast CPU test that pins the exact `/update_weight_version` wire payload, so a revert or refactor cannot silently reintroduce the aborts. Test-only; no production changes.

Part of the [multi-lora] 1/n merge train: #2273 carries the same (or dependent) content and rebases after this merges.

### DIFF
--- a/tests/fast/backends/sglang_utils/test_sglang_engine.py
+++ b/tests/fast/backends/sglang_utils/test_sglang_engine.py
@@ -1,4 +1,5 @@
 import time
+from types import SimpleNamespace
 
 import pytest
 import requests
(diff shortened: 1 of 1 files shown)


yushengsu-thu:
Closing: not an extraction from #2273 (the stack carries no sglang_utils production diff), so it does not belong in the 1/n merge train.

PR #2714, opened 2026-08-22
Bump sglang to v0.5.18

yueming-yuan:
Rebases the `sglang-miles` stack onto `v0.5.18` and repoints the miles images at it. Targets #2673 so the Megatron and sglang bumps are validated together.

## sglang side

`sglang-miles-v0.5.18` = 35 commits on `v0.5.18`, replayed from `sglang-miles-v0.5.17` rather than from production: production is still the v0.5.16 stack, so going through v0.5.17 halves the upstream drift (736 commits instead of 1317) and carries forward resolutions a green CI already validated. Production gained no commits in the meantime, so nothing is skipped. Rollback points: `sglang-miles-v0.5.16-final` (production) and `sglang-miles-v0.5.17`.

Only 9 of 35 picks conflicted. Where v0.5.18 has since **removed** machinery that the v0.5.17 resolutions leaned on, the newer shape wins and our behaviour was re-expressed on top of it:

- `tokenizer_manager`: v0.5.18 dropped the parallel-sampling `child_rid_to_logical_rid` / `lifecycle_id` / `_remove_req_state` machinery. Our abort-by-prefix gate, pre-dispatch flagging and LoRA lease release were rebuilt on v0.5.18's simpler form; the dispatch paths keep upstream's new CUDA-VMM prepare/cancel structure with our abort gate ahead of the reservation.
- `forward_batch_info`: v0.5.18 split mrope into decode/extend helpers and **already forces the text-only path on `rl_on_policy_target`** — the same intent as ours. Took its structure and added our contract switch at both decision points, since `true_on_policy_contract` is a separate switch from `rl_on_policy_target`.
- `deepseek_v2`: v0.5.18 moved the shared-expert fusion decision into the loader plus a `shared_experts_fusion_disable_reason` classmethod that carries our SBO/TBO and DeepEP reasons, so `determine_num_fused_shared_experts` is upstream's no-arg version.
- `compile_utils`: kept upstream's `envs.SGLANG_DG_CACHE_DIR` default and layered our `SGLANG_DG_CACHE_DIR_PER_PROCESS` suffix on it (colocated RL engines would otherwise race on one JIT cache).
- `glm4_moe`, `dsa_indexer`: each side adds a distinct method, both kept.

Post-pick repairs: restored `_post_load_weights` (the pick dropped the definition but kept its call sites) and `get_exec` in `glm4_moe_nextn`; a mangled `determine_num_fused_shared_experts` remnant was replaced with upstream's. `ruff` F821 is zero across every file the stack touches, and `sys.exit(pytest.main(...))` now matches the repo convention in the parallelism-context test.

## miles side

- `docker/Dockerfile`: `SGLANG_IMAGE_TAG=v0.5.18`, `SGLANG_BRANCH=sglang-miles-v0.5.18` (switched to the stable `sglang-miles` name at merge time).
- `docker/build.py`: release variant pinned to `v0.5.18-cu129`.
- The cu12 `pyproject.toml` rewrite block was re-checked against v0.5.18's upstream Dockerfile — the three sed patterns and the three markers are unchanged, so it still applies as-is.

Rust files are untouched; local rustfmt/clippy hooks fail only for want of `cargo` on this machine.


yueming-yuan:
All CI in `run-ci-imaged` passed, so I take the label off to do some final clean

miles/utils/reloadable_process_group.py
```diff
@@ -261,6 +261,39 @@ def bound_device_id(self, dev):
         self.group.bound_device_id = dev
 
 
+def _forward_remaining_collectives():
```
Zhichenzzz, line 264: LGTM if @yueming-yuan tested this under GB300 nodes

Zhichenzzz (approved):

guapisolo (approved):

PR #2716, opened 2026-08-23
fix: align MLA RoPE types with model configs

zyzshishui:
## Summary

- Use plain RoPE for GLM-4.7-Flash, GLM-5/5.1/5.2, and JoyAI-LLM-Flash MLA.
- Refresh affected argument and launcher snapshots.

## Symptom & Reproduction

`python miles/utils/external_utils/model_args_utils.py glm4.7-flash` previously emitted `--multi-latent-attention` without `--rope-type`. The same omission appeared in the GLM-5 family and JoyAI-LLM-Flash definitions.

When those arguments reached Megatron, Miles filled a missing MLA RoPE type with `yarn`, even though the official checkpoints describe plain/default RoPE: [GLM-4.7-Flash](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/7dd20894a642a0aa287e9827cb1a1f7f91386b67/config.json), [GLM-5](https://huggingface.co/zai-org/GLM-5/blob/c183ef8c61faee82855eca1ed9bb3a9a7ce3b0b2/config.json), [GLM-5.1](https://huggingface.co/zai-org/GLM-5.1/blob/26e1bd6e011feb778d25ae34b09b07074139d92d/config.json), [GLM-5.2](https://huggingface.co/zai-org/GLM-5.2/blob/b4734de4facf877f85769a911abafc5283eab3d9/config.json), and [JoyAI-LLM-Flash](https://huggingface.co/jdopensource/JoyAI-LLM-Flash/blob/8d31d53f43de15f3bd5e34fea03bd8e186611fc8/config.json).

## Root Cause

1. Each affected `model_args()` enabled `--multi-latent-attention` but omitted `--rope-type`.
2. `load_model_args()` propagated that omission unchanged into launcher and conversion commands.
3. `set_default_megatron_args()` maps a missing `rope_type` to `yarn` whenever `multi_latent_attention` is enabled.
4. The MLA implementation branches on `config.rope_type`, so the omission constructs `YarnRotaryEmbedding` rather than `RotaryEmbedding`.

This fallback is appropriate for the DeepSeek and Kimi definitions that use YaRN, but it does not match these GLM and JoyAI checkpoint configs.

## Fix

Add `--rope-type rope` to the six affected base model definitions. Derived pruned and LoRA variants inherit the corrected value without changing the global MLA default or the existing explicit YaRN configurations.

**Note**: GLM-5 `DSAMLASelfAttention.forward()` unconditionally unpacks `(rotary_pos_emb, mscale)`, matching the two-value return contract of `YarnRotaryEmbedding.forward()`. Selecting `--rope-type rope` instead constructs `RotaryEmbedding`, whose `forward()` returns one tensor, so the existing GLM-5 path can fail during unpacking. This attention call site must be adapted together with the flag before the GLM-5 portion is safe to merge.

### DIFF
--- a/scripts/models/glm4.7-flash.py
+++ b/scripts/models/glm4.7-flash.py
@@ -37,6 +37,7 @@ def model_args() -> str:
         "--swiglu "
         "--untie-embeddings-and-output-weights "
         "--position-embedding-type rope "
+        "--rope-type rope "
         "--no-position-embedding "
         "--normalization RMSNorm "
         "--norm-epsilon 1e-5 "
--- a/scripts/models/glm5-744B-A40B.py
+++ b/scripts/models/glm5-744B-A40B.py
@@ -36,6 +36,7 @@ def model_args(n_moe_layers: int = 75) -> str:
         "--swiglu "
         "--untie-embeddings-and-output-weights "
         "--position-embedding-type rope "
+        "--rope-type rope "
         "--no-position-embedding "
         "--normalization RMSNorm "
         "--qk-layernorm "
--- a/scripts/models/glm5-744B-A40B.py
+++ b/scripts/models/glm5-744B-A40B.py
@@ -36,7 +36,6 @@ def model_args(n_moe_layers: int = 75) -> str:
         "--swiglu "
         "--untie-embeddings-and-output-weights "
         "--position-embedding-type rope "
-        "--rope-type rope "
         "--no-position-embedding "
         "--normalization RMSNorm "
         "--qk-layernorm "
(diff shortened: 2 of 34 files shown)


guapisolo:
Letting my agent to do ablation and see the scale of rope issue.

guapisolo:
Commit `b9f61310d421a1ad6bbe0b8ac4cb955ecfe9c91a` will remove the GLM5/5.1/5.2 plain-RoPE launcher and generated snapshot changes from this PR because the current GLM5 adapter has not been validated for that override, while preserving the GLM-4.7 and JoyAI changes. Targeted GLM5 snapshot checks passed (20 model-argument, 28 Python-launcher, and 2 shell-launcher cases), and `git diff --check` passed.

guapisolo (approved):

PR #2289, opened 2026-08-09
[7/7] feat(ray): fence evaluation and weight publication

ashtonchew:
## Summary

- Shares one admission boundary between shared evaluation and trainer weight updates.
- Revalidates prefetched fully async work against the manager-published model version.
- Records a successful update before reopening source and train-batch admission.
- Adds one composed acceptance test for the complete issue #2254 ownership path.

## Context

Training ownership can be correct while evaluation or prefetched rollout work crosses a model update. Evaluation can then observe a changing engine. Training can admit a group using the version snapshot captured before the update.

This final layer uses the lifecycle and checkpoint boundary from #2288. It coordinates shared evaluation, rollout admission, and trainer-owned weight publication through the existing scheduler. Router polling and a second scheduler remain outside the design.

[Issue #2254](https://github.com/radixark/miles/issues/2254) defines the full path through source reservation, terminal execution, trainer admission, checkpoint, evaluation, and weight publication.

The logical parent is [#2288](https://github.com/radixark/miles/pull/2288). All drafts target `main`, so GitHub shows the parent and foundation commits in this PR. Review this layer after #2288. Merge the chain in order. If a parent is squash-merged or rebase-merged, rebase this branch onto the updated `main`.

Review this layer as the shared model-version boundary. A trainer update claims the update fence and waits for active shared evaluation and the rollout frontier. It then publishes engine weights, records the manager-owned version, and releases the exact hold. Shared evaluation retains its admission until its worker and result handling are terminal. A trainer update starts after that admission is released.

This PR adds five commits:

- [`becd2a588`, which invalidates prefetched work after recorded weight updates](https://github.com/ashtonchew/miles/commit/becd2a588)
- [`b34dbd794`, which fences weight updates against shared evaluation](https://github.com/ashtonchew/miles/commit/b34dbd794)
- [`c5b7eaa1c`, which sequences trainer updates through admission holds](https://github.com/ashtonchew/miles/commit/c5b7eaa1c)
- [`678db05c3`, which composes the issue #2254 ownership path in one acceptance test](https://github.com/ashtonchew/miles/commit/678db05c3)
- [`e58a9bb7b`, which exercises the ownership path through a real Ray object reference and remote trainer rank](https://github.com/ashtonchew/miles/commit/e58a9bb7b)

## Description

- Adds an explicit `WeightUpdateAdmissionHold` capability. The existing train-admission hold contract for other rollout plugins stays unchanged.
[...]

Issue #2254, opened 2026-08-07
[rollout] Define lifecycle ownership and recovery for fully async rollout


ashtonchew:
## Summary

[Fully async rollout](https://github.com/radixark/miles/blob/main/docs/user-guide/fully-async.md) lets Miles generate the next prompt group while the trainer works on an earlier batch. This overlap improves throughput. It also creates a period where a group has left the data source before training accepts it. A source group is one group of prompts from the data source.

Current [`main`](https://github.com/radixark/miles/tree/main) stores each part of that work separately in the data source, rollout worker, manager, and trainer. It advances the source cursor when the group leaves the data source. After cancellation, remote failure, or restart, the process cannot always tell whether that group is still active, may be replayed, or is already committed.

For example, the source can return group 42 and advance its cursor to group 43 while group 42 exists only in worker memory. If the process saves the new cursor and then stops, restore can resume at group 43 and lose group 42. Nothing records which attempt a result belongs to, so a late or duplicated result cannot be matched back to the source work it came from.

For fully async runs that use the reservation-capable global data source, this issue adds one ownership path. Reservations, execution receipts, terminal receipts, leases, and trainer admission receipts carry the same source identity through generation, buffering, Ray publication, checkpoint, and model-version changes. Periodic checkpoints make unfinished reservations durable across restart. Miles can replay a group after its current attempt ends and commit it after every required trainer reads the exact train batch.

A lost remote response can cause more than one generation attempt. The ownership path gives one reservation one recorded settlement outcome. A committed group cannot return to replay after a later failure.

## How fully async works on `main` at `f2b7c7929`

Miles reads prompts from a data source in groups. SGLang generates responses for those prompts. The rollout manager converts the responses into train data, and the trainers run optimizer steps. Miles then sends updated model weights back to the rollout engines.

[`FullyAsyncRolloutFn`](https://github.com/radixark/miles/blob/main/miles/rollout/fully_async_rollout.py) overlaps generation with training. Its background worker asks the data source for prompt groups, starts generation, and places each prompt group and result in an in-memory queue. The trainer drains that queue while the worker submits more work.

The [`RolloutDataSource`](https://github.com/radixark/miles/blob/main/miles/rollout/data_source.py) advances its `sample_offset` as soon as `get_samples()` returns a group. Its checkpoint stores the cursor, the epoch, and the running group and sample counters. Unfinished groups that already left the source live only in process memory.

[...]

PR #2245, opened 2026-08-07
fix(ray): allow 120 seconds for router startup

ashtonchew:
## Summary

- Gives the SGLang router up to 120 seconds to become ready.
- Keeps the existing health check, polling interval, and failure behavior.
- Provides an independent startup prerequisite for the fully async lifecycle work in [issue #2254](https://github.com/radixark/miles/issues/2254).

## Context

The rollout manager starts the router while rollout workers and serving processes initialize. The current 30-second deadline can expire before the router health endpoint is available even while startup is progressing.

This PR changes the startup deadline only. Miles keeps the existing health criteria, health-check interval, and post-deadline behavior.

This PR is independent of the two stacked rollout chains in [issue #2254](https://github.com/radixark/miles/issues/2254), so it can merge separately.

Review this change as one commit: [`d3605c9f` fixes the startup deadline](https://github.com/radixark/miles/pull/2245/commits/d3605c9f565891b7747ffc704e2591fc566bcd0b). The review invariant is that only the allowed startup time changes.

## Description

- Changes the router startup limit from 30 seconds to 120 seconds.
- Keeps the one-second health polling interval.
- Keeps the existing timeout error and cleanup path.
- Adds focused coverage for the configured deadline and polling behavior.

## Test plan

- [x] `tests/fast/ray/rollout/test_router_manager.py`: 12 tests passed at exact head `d3605c9f56`.
- [x] An exact composed 8xB200 validation delayed router startup by 35.000 seconds. The router became ready after the former 30-second deadline and before the new 120-second deadline.
- [x] The composed 8xB200 run used Miles ref [`7cd212d81f`](https://github.com/ashtonchew/miles/commit/7cd212d81f6c94f72d67b4443e8c047008e4d517) with validation child [`cf512e6a8a`](https://github.com/GymPod/slime/commit/cf512e6a8a59c7c935e273dc69b9fd84c32e35a3). This evidence covers router startup in the composed run. Isolated PR-head and full training evidence require separate runs.

## Risks and follow-ups

An unhealthy router can now take up to 90 seconds longer to report its startup failure.

PR #2719, opened 2026-08-23
[8/8] fix(ray): settle trainer admissions without blocking the manager loop

ashtonchew:
## Summary
Trainer-admission settlement no longer blocks the `RolloutManager` actor's event loop. `commit_trainer_admission` and `rollback_trainer_admission` are awaited under `asyncio.shield`, `TrainBatchLease` gains `commit_async`/`rollback_async` beside the blocking pair, and the two remaining rollback sites in `_generate` and `_get_rollout_output` await their settlement the same way. No timeouts are added; `get_trainer_admission_status` stays synchronous.

## Context
Last PR of the issue #2254 ownership stack; applies on top of #2289. Found by an adversarial review of the composed stack; the change log in issue #2254 records it.

## Description
`RolloutManager` is a Ray async actor, so its synchronous methods run inline on the actor's event-loop thread. Settlement handed work to the separate rollout lifecycle loop and then blocked on a `concurrent.futures.Future` with no timeout — for that whole duration the manager could serve nothing, including `get_trainer_admission_status`, which `RayTrainerAdmissionAdapter` calls to reconcile a lost commit or rollback response. This contradicts the manager-loop rule in `miles/ray/rollout/eval_fleet.py` ("keep everything here awaiting rather than blocking") and the sanctioned `_submit_lifecycle_coroutine` bridge.

The lease ABC keeps its once-only `_claim_settlement` contract across both entry points; default async implementations delegate to the synchronous ones, so other lease implementations keep working. Shielding is required because an async actor method is cancellable where a sync one was not, and an abandoned settlement would strand the admission in its fail-closed `COMMIT_FAILED`/`ROLLBACK_FAILED` pre-state.

## Test plan
- [x] `test_commit_settles_without_blocking_the_manager_loop` / `..._rollback_...`: settlement gated on an event; the manager loop keeps serving (status reads `COMMIT_FAILED`/`ROLLBACK_FAILED` meanwhile); terminal status after release.
- [x] `test_cancelled_commit_caller_cannot_abandon_settlement` / `..._rollback_...`: cancelling the caller still reaches `COMMITTED`/`ROLLED_BACK`.
- [x] `test_commit_async_settles_across_loops_without_blocking_the_caller`: owner loop stalled on a dedicated thread; the calling loop keeps ticking; acknowledgement lands after release.
- [x] 223 tests passed and 2 skipped across the manager-handoff, fully-async rollout, real-Ray, vertical-slice, trainer-admission, and coordinator suites at this head; 1,108 passed in the wider rollout/ray sweep.
- [x] ruff, black, isort pass on every changed file.
- [ ] `run-ci-fully-async` H100/H200 workflow at this head.

## Risks and follow-ups
- Settlement is now cancellable at the Ray boundary; the shield keeps state transitions atomic, but a cancelled caller must consult `get_trainer_admission_status` (the adapter already does).
- Custom `TrainBatchLease` implementations may override `_commit_async`/`_rollback_async` to avoid blocking; the defaults call the synchronous hooks.

https://claude.ai/code/session_01WZM6myVw1W4B7GF27PRViP


ashtonchew:
Rebased onto upstream `main` at `851ddba2a`; no change to this PR's own commit. New head `4d6e44f50` is the stack tip, so the `run-ci-fully-async` label still belongs here. Details in the [issue comment](https://github.com/radixark/miles/issues/2254#issuecomment-5399596929).

ashtonchew:
Rebased onto upstream `main` at `f2b7c7929`; no change to this PR's own commit. #2245 left the stack (superseded by upstream #2604), so the stack is now 16 PRs and `d3556ffb1` is the tip; the `run-ci-fully-async` label still belongs here. Details in the [issue comment](https://github.com/radixark/miles/issues/2254#issuecomment-5415223288).

PR #2724, opened 2026-08-23
[multi-lora] 1/n - 1, feat: per-token loss-weight/advantage sample channels

yushengsu-thu:
Adds client-supplied per-token channels to the sample data format: `Sample.loss_weights`/`Sample.advantages` (response-aligned like `loss_mask`, float32 on the wire while the binary `loss_masks` stay int32), CP-slice coverage, and observation-span zero-padding in multi-turn merges. `AdapterRef` gains registration-scoped serving-identity fields and `get_reward_value` tolerates arg namespaces without `reward_key`. All additions are inert until a rollout path populates them.

Extracted from #2273; the stack rebases after this merges.


Shi-Dong (approved):
LGTM

PR #2725, opened 2026-08-23
[multi-lora] 1/n - 2, ci: verify import targets in the source-resolution harness step

yushengsu-thu:
Extends `tests/ci/verify_source_resolution.py` — the pre-test step every CPU and GPU CI job already runs — with import-integrity checks: statically resolve every miles-internal import site (module-level, function-local, and relative) across `miles/` and `examples/` against the source tree, walk the optional multi-lora namespaces in full when present (absent packages skip cleanly on main), and import the `update_weight/` lazy-import targets. Failures raise `RuntimeError` with the offending `file:line` and import target; missing non-miles optional deps stay tolerated.

Running this in the harness step instead of a fast test covers every job (CPU and GPU) before pytest even starts, which is broader than the fast suite. Born from the `mixin.py` dead lazy-import incident, where a moved module was only referenced from a function-local import and nothing failed until runtime.

Extracted from #2273; the stack rebases after this merges.


guapisolo:
I think this PR should be narrowed to the third motivation only: protecting `update_weight` function-local/lazy imports from going stale.

The current change mixes three concerns: repo-wide AST linting, optional-namespace import smoke tests, and `update_weight` lazy-import integrity, then runs them in every CPU/CUDA job. The `mixin.py` incident only justifies the last one. The broader resolver is also incomplete (`from package import submodule` checks only the package), while namespace imports add side effects and environment coupling.

Suggested minimal cut:

- statically inspect only `actor.py` and `update_weight/**/*.py` for function-local Miles imports;
- resolve targets without importing backends, so Mooncake/RDMA remains lazy;
- preserve `file:line` and add a negative fixture proving a missing `mixin.py` fails;
- run the check once in one CPU preflight job.

I would remove the repo-wide `miles/` + `examples/` scan, `OPTIONAL_PACKAGES` walk, and all-job fan-out. A general repository import-integrity linter can be a separate PR with explicitly defined import semantics and its own focused tests.

PR #2730, opened 2026-08-24
fix(lora): check distributed adapter tensors

Arist12:
Part of #2705. Requires [SGLang #36190](https://github.com/sgl-project/sglang/pull/36190).

## Problem

`--check-lora-weight-equal` verifies the colocated tensor route but is a no-op for NCCL adapter broadcasts. Sending a checksum field from Miles alone is not enough: SGLang's request decoder is built with `extra_behavior="ignore"`, so a build without the receiver-side check drops the field and still returns success — the flag reads as verified while verifying nothing.

## Change

When the checker is enabled, hash the adapter as sent, forward the manifest to every distributed engine, and require each engine's explicit `checksums_verified` acknowledgement. The manifest itself moves to `update_weight/common.py` so both transports share one hash recipe instead of two copies of the expression. Normal publishes do no hashing and keep their current payload.

## Validation

- 26 fast tests in `tests/fast/backends/megatron_utils/test_lora_weight_sync_validation.py`, covering the checksum payload and a missing acknowledgement from both the attribute-style and Mapping-style engine results; SGLang #36190 carries the receiver-side tests for peer mismatch and receive failure
- 8x MI350X, Qwen3-4B, four trainer + two TP2 rollout engines: TP0 and TP1 on both engines verified `504/504` tensors before rollout 0 and after its optimizer step; rollout, nonzero gradient and checkpoint completed
- 8x MI350X, Qwen3-4B, four trainer + four TP1 rollout engines: every engine verified `504/504` tensors on both the restored initial publish and the post-training publish
- 8x MI350X, Qwen3-30B-A3B, four trainer + four TP1 rollout engines: every engine verified `37,248/37,248` tensors before resumed rollout 30 and after its two optimizer steps

## Reviewer notes

- Failing loudly against an older SGLang is deliberate: a silent pass is the exact failure this closes. The flag is off by default, so only runs that asked for the check are affected.
- Land after #2708. Both insert a keyword argument on the same `load_lora_adapter_from_distributed` call, so they collide on one line in either order; the resolution is to keep both.

### DIFF
--- a/docs/advanced/lora.md
+++ b/docs/advanced/lora.md
@@ -122,7 +122,7 @@ vision towers unadapted. Use the model launcher as the source of truth.
 | `--lora-base-cpu-backup` | off | Colocated mode only: keep a CPU mirror of the frozen SGLang base and avoid re-sending base weights. This trades host RAM for faster and more reliable pause/resume. |
 | `--lora-train-only` | off | Train the adapter while keeping ordinary rollout engines on the frozen base policy. |
 | `--experts-shared-outer-loras` | off | Use shared outer factors for grouped MoE experts. This layout is not checkpoint-compatible with per-expert LoRA. |
-| `--check-lora-weight-equal` | off | On the colocated path, verify each synchronized adapter tensor with SHA-256. |
+| `--check-lora-weight-equal` | off | Verify each synchronized adapter tensor with SHA-256, on both transports. |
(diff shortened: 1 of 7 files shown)


Arist12:
Superseded. #2756 made the LoRA checksum manifest transport-agnostic and moved the expectation onto `end_weight_update(expected_lora_checksums=...)`, which covers the NCCL broadcast case this PR was adding. The engine-side acknowledgement is now present in `sglang-miles` (#36384), so the remaining gap this PR described is closed as well.

PR #2735, opened 2026-08-24
fix(dsv4): fix deepseek v4 hareware branch

guapisolo:
## Summary

Keep the 4-layer H200 smoke test out of the GB300 TP8 rollout profile.

## Symptom & Reproduction

- **Symptom:** `stage-c-4-gpu-h200 (0)` in [nightly run 32743697384, attempt 1](https://github.com/radixark/miles/actions/runs/32743697384/job/97486158879) failed at revision `851ddba2a7dbaef22bb090e12efd12d8a65af2c3` after only `4/8 clients joined`.
- **Reproduction:** `test_deepseek_v4_flash_4layer_ci.py` launches one 4-GPU H200 node but generated `rollout_num_gpus_per_engine=8`/TP8/EP8.

## Root Cause

1. [`010e77993`](https://github.com/radixark/miles/commit/010e77993282b5d95cfb3152189e6fdc3d910da4) in [#2717](https://github.com/radixark/miles/pull/2717) used `num_gpus_per_node == 4` for GB300 TP8.
2. `DeepSeek-V4-Flash-FP8-4layer` also uses four GPUs/node, so four ranks entered TP8/EP8.

## Fix

`is_gb300_profile` now requires four actor GPUs per node and excludes `is_4layer`. Full-model GB300 runs keep TP8/EP8; the 4-layer H200 smoke test returns to TP4/EP4, with all other profiles unchanged.

## Verification

- `tests/fast/launch_scripts/test_run_deepseek_v4.py`: `2 passed`, covering 1x4 4-layer H200 and 8x4 full-model GB300 command generation.
- Targeted `scripts/run_deepseek_v4.py::train` launcher snapshot: `1 passed`.
- `ruff check` and `black --check` on both changed files: passed.

## Review Focus

- `is_gb300_profile`: the 4-layer profile must stay excluded while the 8x4 full-model profile keeps TP8/EP8.

PR #2738, opened 2026-08-24
feat: resolve --hardware from the node in every launcher

yueming-yuan:
## Summary

Make `--hardware` the single source for a launcher's machine profile: it defaults to detecting the node it runs on, and `num_gpus_per_node` follows from it.

Stacked on #2735 — the first three commits are that PR. Review from `ca4fcba01`.

## Problem

Every launcher hardcodes two things about the machine:

```python
num_gpus_per_node: int = 8
hardware: Literal["H200", "B200", "GB300"] = "H200"
```

Two ways to get a silently wrong profile out of that:

1. **Forget `--hardware`.** You get whichever machine the recipe was first written on. #2735 is one instance of the same class — a 4-GPU H200 smoke test took the Grace TP8 rollout profile and hung at `4/8 clients joined`.
2. **Pass `--hardware GB300` but not `--num-gpus-per-node 4`.** The topology is still sized for 8 GPUs per node.

## Fix

`--hardware` defaults to `"auto"`, resolved once in `__post_init__`:

```python
self.hardware = U.resolve_hardware(self)
self.num_gpus_per_node = self.num_gpus_per_node or U.NUM_GPUS_OF_HARDWARE[self.hardware]
```

`detect_hardware()` (added in #2735) maps the node to a `NUM_GPUS_OF_HARDWARE` key: compute capability picks the generation, the host CPU separates Grace from the x86 part carrying the same die, HBM capacity splits H100 from H200. `resolve_hardware()` additionally rejects a hardware the launcher declares no profile for, rather than letting a `match args.hardware` fall through it:

```
AssertionError: ScriptArgs has no verified profile for B200
```

An explicit `--hardware` still overrides, and an explicit `--num-gpus-per-node` still wins over the derivation — that is how a job takes four GPUs of an eight-GPU node.

## Scope

19 launchers declare `hardware`; all of them change. The 33 that do not keep their literal, because there `num_gpus_per_node` is often a deliberate partial node (`run_qwen3_0_6b_fsdp` takes 4, `examples/verifiers` takes 2) rather than the machine's shape.

## Verification

The replaced literals were all consistent — **every launcher that declares `hardware` hardcoded exactly `NUM_GPUS_OF_HARDWARE[its default hardware]`**, 13/13, so this is behavior-preserving at the default:

```
script                                     hw default  gpus  NUM[hw]
scripts/run_deepseek_v32.py                B200           8        8
scripts/run_glm45_355b_a32b.py             GB200          4        4
scripts/run_glm5_744b_a40b.py              H200           8        8
...                                        (13/13 match; 6 more already derived)
```

- `tests/manual/launch_scripts/test_py_launch_scripts.py`: **snapshots unchanged**, 182 passed (the 8 failures and 4 errors are the same set as on `main` in this checkout, all under `scripts/amd/**` and `run_kimi_k25`).
- `tests/fast/utils/test_command_utils.py` + `tests/fast/launch_scripts/test_run_deepseek_v4.py`: 80 passed.
- `pre-commit run --files <changed>`: passed.

Two new tests keep the recording pins honest: every pinned hardware must be one the launcher accepts, and a launcher the frozen default already covers may not be pinned twice.

## Review focus

- `resolve_hardware()` reads the supported set off the field's `Literal` — that is the one piece of reflection here.
- `_HARDWARE_A_RECORDING_REPRESENTS` in the snapshot suite: a launcher that reads the node cannot be recorded on a CPU runner, so the suite states which machine each recording is.
- `detect_hardware()`'s capability table is **not yet verified on real hardware** (sm_103 for B300/GB300, `aarch64` for Grace are derived from the architecture, not measured). Worth one run on a B300 / GB300 / H200 node before merge.

### DIFF
--- a/examples/retool_v2/run_retool_multi_turn.py
+++ b/examples/retool_v2/run_retool_multi_turn.py
@@ -14,7 +14,7 @@ WANDB_GROUP = "sft-multi-turn-batch-32"
 class ScriptArgs(U.ExecuteTrainConfig):
     mode: Literal["normal", "debug_minimal"] = "normal"
     run_id: str = field(default_factory=U.create_run_id)
-    hardware: Literal["H100", "GB200", "GB300"] = "H100"
+    hardware: Literal["auto", "H100", "GB200", "GB300"] = "auto"
     num_gpus_per_node: int | None = None
     use_sft_model: bool = True
     save_path: str = "/root/Qwen3-4B_miles/retool_v2_multi_turn"
@@ -28,6 +28,7 @@ class ScriptArgs(U.ExecuteTrainConfig):
     ref_load: str = field(init=False)
 
     def __post_init__(self):
+        self.hardware = U.resolve_hardware(self)
         self.num_gpus_per_node = self.num_gpus_per_node or U.NUM_GPUS_OF_HARDWARE[self.hardware]
         if self.use_sft_model:
             self.hf_checkpoint = "/root/font-info/qwen3-4b-sft"
--- a/miles/utils/external_utils/command_utils.py
+++ b/miles/utils/external_utils/command_utils.py
@@ -13,6 +13,7 @@ import socket
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
+from typing import get_args
 
 from miles.utils.external_utils.exec_command import exec_command_cpu, exec_command_gpu, exec_command_multi_node
 from miles.utils.external_utils.model_args_utils import shell_safe_model_args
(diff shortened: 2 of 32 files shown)


guapisolo:
Open disucssion: Should we keep the hardware detect -> num gpu per node assumption logic in script level like this PR, or move it into miles launch? (like arguments.py)

Okay I think current logic is better. We shouldn't let the engine know too many logics.

guapisolo (approved):

PR #2739, opened 2026-08-24
feat: dist_muon offloading in megatron

Zhichenzzz:
`--stream-optimizer-state-to-disk` currently only works with Adam. This makes it work with Muon too.

## Why it did not work

The flag binds to `DistributedOptimizer`, which Muon never uses: `set_default_megatron_args` clears `use_distributed_optimizer` for every non-Adam optimizer, and Muon is built through `LayerWiseDistributedOptimizer`. The two asserts guarding the flag were unreachable for Muon by construction — one required the distributed optimizer that had just been cleared, the other required `optimizer == "adam"`.

## Approach

Rather than port the DistOpt-bound store to a different optimizer — which would mean reimplementing its bucket planner and its main→param write-back, and re-deriving the master-weight residency protocol — this reuses the offloader Megatron already runs for Muon.

`Float16OptimizerWithFloat16Params` dispatches its step through an `_optimizer_state_offloader` slot, and `ChunkedOptimizerStateOffloader` fills that slot today under `--chunked-optimizer-state-offload`: chunked restore, the `assert_master_weights_resident` protocol and the prefetch hooks already work with Muon. Checkpointing needed two fixes, below. Its only tie to host memory is one allocator, `_new_cpu_buffer`, returning pinned CPU tensors.

So `setup_muon_state_on_disk` subclasses that offloader and overrides just the allocator, returning tensors backed by files on node-local NVMe. It lives in `nvme_stream.py` beside the Adam store and shares its rank-scoped directory layout and stale-state purge; the two entry points stay separate only because Muon must be installed before the optimizer is built, which is when the offloader is constructed. `copy_` moves data to and from an mmap tensor exactly as for a pinned one, and `is_pinned()` returns False so the inherited code takes its synchronous-copy path. The other two CPU allocations in that file were checked and left alone: one targets the parameter's own device, the other is a bounded per-chunk staging buffer.

The flag's gates are split by optimizer family rather than removed — Adam keeps requiring the distributed optimizer, Muon requires the `dist_` prefix plus the chunk knobs the offloader is driven by.

## Why it matters

File-backed pages land in the cgroup's reclaimable `file` rather than in `anon`, so the kernel can evict them under pressure instead of OOM-killing the run.

Measured on Qwen3.5-35B-A3B, 4x GB300, TP2/EP4, 8k responses, at `--optimizer-state-offload-fraction 1.0` — the setting at which the pinned-CPU path dies:

| backend | outcome |
|---|---|
| pinned CPU (`--chunked-optimizer-state-offload` alone) | host peak 908GB against a 919GB limit, SIGTERM at step 6 |
| disk (this change) | ran to completion, `rc=0` |

With the disk backend the non-reclaimable `anon` stayed flat at 86GB across 4 optimizer steps while `file` grew to 619GB, and `grad_norm` was 0.1166 — inside the 0.088–0.170 band the same model spans without offload.

## Testing

`tests/fast/optimizers/test_nvme_stream.py`, 8 cases, no GPU needed. They cover the silent-failure path: if the allocator override stops returning file-backed storage, the offloader keeps working against pinned host memory while the log line still claims the disk backend is installed. Each assertion was checked against a deliberately broken implementation to confirm it fails.

`tests/e2e/megatron/test_qwen3_4B_muon_offload_disk.py` mirrors the Adam test `test_qwen3_4B_offload_disk_stream.py` with `--optimizer dist_muon`, which is the only route to the layer-wise distributed optimizer the offloader hangs off. Completing the run proves little on its own — a rebind that misses the consuming module falls back to pinned host memory and still converges — so the offloader now logs one line per step and the test asserts all four ranks stepped against file-backed state rather than merely that `install()` ran. Verified on 4x GB300: 4/4 ranks armed, 4/4 file-backed, `rc=0` in 292s.

At that rollout size every sample truncates, so the registered metric gates sit at zero and cannot catch a regression by themselves — the same holds for the Adam test this mirrors. The disk-backed-step assertion is what guards the feature.

## Checkpointing

Review found two ways this went wrong against `adopt_cpu_optimizer_state`, which
reallocates every non-pinned CPU tensor in `optimizer.state`. Ours never report pinned, so
every checkpoint was copying the whole offloaded state into fresh mappings; the stock
pinned path never reaches that branch. Buffers now carry a marker and the allocator returns
them unchanged. Separately, `synchronize_for_checkpoint` msyncs our mappings before the
writer fsyncs its own files, so that cost is ours and attributable rather than left in the
kernel's writeback queue.

Verified on 4x H200 with three saves, `rc=0`. The offloaded total goes 3.38 -> 6.77 GB at
the first save, where the fp32 masters are offloaded for the first time, then holds at 6.77
across the second -- reallocation would have doubled it again. All four ranks agree.

Two things worth knowing. Saving is measurably slower with the disk backend: 381s per save
against 229s for the same run on pinned host memory, an A/B on one box, so roughly +66%
from reading the state back through the mappings. And CI does not cover this: three saves
of this model take 19 minutes, past `run_suite.py`'s 1800s per-file limit, so the e2e test
stays checkpoint-free and the above is a manual check.

## Notes

`--optimizer-cpu-offload` is a separate flag and remains Adam-only. It is not a gap for Muon: `HybridDeviceOptimizer` wraps an optimizer and substitutes the tensors it steps on, which collides with `LayerWiseDistributedOptimizer` resharding `group["params"]` and then wrapping each child in `Float16OptimizerWithFloat16Params` to create fp32 masters — two layers substituting the same tensors. Megatron's own answer is the `_optimizer_state_offloader` slot, which is what this change builds on.


miles/utils/arguments.py
yueming-yuan, line 1: should we default `--optimizer-state-offload-chunk-size-mb` to be a non-0 value, or add a warning? The current default is 0, and in megatron it seems 0 -> non-streaming, so the peak memory cannot be saved. The user might now be aware of this

miles_plugins/optimizers/nvme_stream.py
```diff
@@ -468,7 +483,43 @@ def setup_optimizer_state_streaming(args, optimizer) -> None:
...
+    base = defining_module.ChunkedOptimizerStateOffloader
+    if base.__name__ == "DiskOptimizerStateOffloader":
+        return
+    rank_dir = _purge_rank_dir(_state_dir_root(args))
+
+    class DiskOptimizerStateOffloader(base):
+        state_dir = rank_dir
+        _disk_bytes = 0
+
+        def _new_cpu_buffer(self, tensor: torch.Tensor) -> torch.Tensor:  # type: ignore[override]
```
yueming-yuan, line 541: codex comment, looks reasonable:

Preserve existing mmap state during checkpoint adoption — miles_plugins/optimizers/nvme_stream.py:503
Megatron’s synchronize_for_checkpoint() calls adopt_cpu_optimizer_state(), which treats every non-pinned CPU tensor as foreign and reallocates it through _new_cpu_buffer. Because these mmap tensors intentionally report is_pinned() == False, every checkpoint copies the entire offloaded optimizer state into new mappings and permanently inflates _disk_bytes. For the hundreds-of-GB workloads targeted here, this adds a full-state copy to every checkpoint. Track already-managed mappings or override the adoption behavior.

miles_plugins/optimizers/nvme_stream.py
```diff
@@ -89,6 +95,15 @@ def _rw_full(op, fd: int, offset: int, buf) -> None:
         done += n
 
 
+def _disk_backed_like(tensor: torch.Tensor, directory: str) -> torch.Tensor:
+    nbytes = max(tensor.numel() * tensor.element_size(), 1)
+    fd, path = tempfile.mkstemp(dir=directory, suffix=".bin")
+    os.close(fd)
+    storage = torch.UntypedStorage.from_file(path, shared=True, nbytes=nbytes)
```
yueming-yuan, line 115: codex + claude comment, looks reasonable

UntypedStorage.from_file(..., shared=True) expands the file with ftruncate but does not reserve physical blocks. If optimizer state exceeds available NVMe capacity, mapping succeeds and later writes can terminate the process with SIGBUS instead of producing an actionable ENOSPC error. Preallocate with posix_fallocate, as the existing _allocate_file path already does, before mapping and unlinking the file.

yueming-yuan (commented):
overall LGTM, just add small comments

Zhichenzzz:
@claude review

miles/utils/arguments.py
Zhichenzzz: Megatron has this warning already

yueming-yuan (approved):

Issue #1360, opened 2026-06-17
[refactor] weight update refactor


yueming-yuan:
This issue is written by human.
Feel free to talk to @yueming-yuan on any concerns/suggestions/questions.

## Why this refactor?
The current weight update logic is facing many issues/complaints... Some examples:

1. With the current code, the lifecycle is maintained separately on both sides. Developers generally prioritize developing `update_weight_from_tensor`, so disaggregated mode doesn't support many features, e.g. (1) megatron-bridge (2) lora (and previously also didn't support low-precision, mtp, now fixed/is fixing) 
2. Weight checker is heavy, not suitable for runtime use. And the weight checker is affected by non-trivial post process. Need a lightweight runtime checker to check transferred data is correct, and better to be used for all transfer method.
3. Code duplication: gather megatron tensor -> HF bucket logic is duplicated. There are several different gather methods. The reason is the logic differs between colocate and disaggregate. (colocate gathers PP but disaggregated does not)
4. Some APIs are confusing and easy to miss, e.g. `/post_process_weight`.

### 1. `HfWeightIterator` refactor
A key reason why the above 1,3 happens is the way the colocated and disaggregated modes gather their tensor is different.

In colocate, the workflow is (1) gather all parallelisms, including **PP**, EP, TP, with `HFWeightIterator` (2) send through cuda_ipc
but in disaggregated, the workflow is (1) gather EP, TP with independently written logic (2) send through NCCL/P2P

This placement different causes different bucketing, so colocate and disaggregate to use different codepath and people have to write many features twice.

So the target here is to **unify both colocate and disaggregated to use `HFWeightIterator`**. It uses `WeightUpdatePlacement` to know which parallelism will be gathered. The mental model here is "give me the final placement you want, I gather the correct tensors and create the ready-to-send buckets for you". Bucketing is decoupled from the protocol; the iterator buckets by placement/size only.

The placement is related to both protocol and Iterator type. For example, cuda_ipc always gather across all parallel dimensions; for NCCL/P2P we usually only gather EP+TP, but if we choose `HfWeightIteratorBridge`, it force the placement to gather across all parallel because of restriction of megatron bridge.

```
HfWeightIteratorBase()
  Raw, bridge
  - weight collecting
    Execute according to the received placement from the transfer protocol
    - (1) PP broadcast (only call when collapse_pp = True)
    - (2) EP all_gather (same for distributed; changed from broadcast for colocate.)
    - (3) TP all_gather (same as current, just dedup code)
[...]

PR #2742, opened 2026-08-25
[refactor] HF-namespace atomic groups and unit-stream bucketing

yueming-yuan:
Second stage of the weight-update refactor (#1360), stacked on #2740. This is
the one stage with intended semantic changes; they are enumerated below and
gated by the weight-checker CI plus a large-scale colocate A/B (pending).

### What

- **`iter_hf_base_weights` becomes a base template.** Backends implement
  `_iter_hf_param_units` — a stream of conversion units, one per training-side
  parameter, holding every HF tensor it converted into (weight, or weight +
  quant scales). The shared contract layer
  (`weight_update/bucketing.py`) runs `assemble_atomic_update_groups` +
  `pack_units_by_size` on that stream; no backend re-implements bucketing.
  Unit granularity structurally guarantees weight+scale pairs never split
  across load calls — including pairs no group registry covered before
  (v3.2 indexer fp8 `wk` pair, v4 `wo_a` pair).
- **Atomic groups re-registered in HF namespace**, verified against the sglang
  loaders that motivate them: deepseekv4 `wqkv_a` / `compressor` /
  `indexer.compressor` (loader hard-asserts on a half-arrived pair) and the
  q_lora `fused_qkv_a_proj_with_mqa` pair (silent no-op on a half-arrived
  pair). inkling registers no groups: its fusions happen inside the converter
  (module-global accumulator, batch-independent) and its engine-side loads are
  slice-disjoint and split-safe. The megatron-namespace registry stays for the
  distributed mixin until that path migrates onto the iterator.
- **Colocate EP gathering: broadcast → symmetric all_gather.** Expert metadata
  stays EP-local (no cross-EP metadata merge); the full expert set materializes
  at gather time via an EP all_gather with a name exchange, matching the
  distributed path's semantics. Gather batches split into non-expert/expert,
  expert batches accounted x ep_size.

### Intended behavior changes (colocate)

1. EP collective changes from per-param broadcast to symmetric all_gather.
2. Bucket composition shifts: packing now uses real post-conversion nbytes
   (was: metadata sizes x TP replication), expert tensors stream after
   non-expert (was: name-interleaved), and buckets can span gather batches.
   Bucket boundaries still only fall between conversion units, so weight+scale
   and fused pairs stay intact by construction.

### Verification

- Unit tests: `weight_update/test_bucketing.py` (grouping/packing semantics),
  `test_hf_atomic_update_groups.py` (registry contents, suffix
  non-shadowing, precedence), direct-iterator packing tests updated.
- Weight-checker CI covers numeric equivalence end to end.
- Pending before merge: one large-scale colocate A/B run (deepseek/glm) for
  the EP-collective and bucket-composition changes.


yueming-yuan:
Re-cut by review boundary: the bucketing pipeline is now #2745 and the EP all_gather change is #2746 (identical cumulative code).

PR #2744, opened 2026-08-25
[refactor] Distributed weight sync on the unified iterator

yueming-yuan:
Third stage of the weight-update refactor (#1360), stacked on #2742.

### What

- `gather_pp=False` support in the direct iterator (skip PP metadata exchange,
  PP broadcast, and the global consistency check); `forced_placement` becomes
  the tp/ep lower bound.

- Every distributed protocol (broadcast / p2p / rdt / disk-delta) now streams
  buckets from `iter_hf_base_weights`; the mixin's own gather stack
  (`_gather_and_update_*`, `all_gather_param`, the megatron-namespace atomic
  groups and their registrations) is deleted.
- Protocols decide `is_sender` / `is_lora_sender` once at connect and own their
  pairing: broadcast/delta via the shared `derive_replica_position`
  (`weight_update/transfer.py`), p2p/rdt via their transfer plan. Non-sender
  ranks drive the iterator with `materialize=False` (join collectives, skip
  conversion).
- Distributed paths read the weights-backuper snapshot like colocate
  (previously they silently read live params).
- LoRA over distributed engines no longer requires PP=1: the megatron base
  gathers the adapter across PP.

### Behavior notes

- Distributed bucket composition changes the same way colocate did in #2742
  (real post-conversion sizes, experts after non-experts); sender sets and
  collective patterns per protocol are unchanged.
- Weights-source switch (live params → snapshot) is intended; it matters under
  --rematerialize-param-from-master-weight.

### Verification

Compile + lint per commit; full test pass batched with the stack per the
review plan (weight-checker CI + colocate and distributed A/B before merge).


yueming-yuan:
Superseded by #2747 (rebased onto the re-cut stack; identical content).

PR #2745, opened 2026-08-25
refactor(update-weight): unit-stream bucketing with HF-namespace atomic groups

yueming-yuan:
Weight-update refactor (#1360), stacked on #2740.

One story: bucketing moves behind the iterator API, post-conversion.

- `iter_hf_base_weights` becomes a base template: backends implement
  `_iter_hf_param_units` — one unit per training param, holding every HF
  tensor it converted into (weight, or weight + quant scales). The shared
  layer (`weight_update/bucketing.py`) runs `assemble_atomic_update_groups` +
  `pack_units_by_size`. Unit granularity keeps weight+scale pairs in one load
  call by construction.
- Atomic groups re-registered in HF namespace
  (`weight_update/atomic_groups.py`), verified against the sglang loaders
  that motivate them: deepseekv4's three pairs hard-assert on a half-arrived
  pair; q_lora's `fused_qkv_a_proj_with_mqa` pair is a silent no-op. inkling
  registers none (converter-side fusion; engine loads are split-safe). The
  megatron-namespace registry stays for the distributed mixin until that path
  migrates.
- EP gathering unchanged.

Intended behavior change (colocate): bucket composition — real
post-conversion nbytes, unit granularity, buckets may span gather batches.
Gated by weight-checker CI + the colocate A/B run.


guapisolo:
codex cmt, make sense to me. But not sure about the effort to fix this. ignore if non-blocking

[P2] Flush the pending output bucket before materializing the next gather batch

`pack_units_by_size` asks the upstream generator for the next unit before deciding whether to yield the current bucket. In the direct iterator, producing that unit materializes the entire next gather batch. The current output bucket still retains converted tensors from the previous batch, so when both are near `--update-weight-buffer-size`, transient GPU memory can approach 2x the configured buffer and OOM a tight run.

This does not require one huge atomic group: ordinary units can fill the output bucket, while atomic grouping only changes the indivisible unit size. Please expose or observe gather-batch boundaries and flush before crossing one, or otherwise release the pending bucket before materializing the next batch.

Zhichenzzz (approved):

PR #2747, opened 2026-08-25
refactor(update-weight): distributed weight sync on the unified iterator

yueming-yuan:
Weight-update refactor (#1360), stacked on the EP all_gather PR.

One story: the distributed paths consume the iterator; senders are decided
at connect.

- Every distributed protocol (broadcast / p2p / rdt / disk-delta) streams
  buckets from `iter_hf_base_weights`; the mixin's gather stack and the
  megatron-namespace atomic-group machinery are deleted.
- Protocols decide `is_sender` / `is_lora_sender` once at connect:
  broadcast/delta via the shared `derive_replica_position`
  (`weight_update/transfer.py`), p2p/rdt via their transfer plan. Non-sender
  ranks drive the iterator with `materialize=False` (join collectives, skip
  conversion).
- The direct iterator gains `gather_pp=False` (skip PP metadata exchange, PP
  broadcast, and the global consistency check); `forced_placement` becomes
  the tp/ep lower bound.
- Distributed paths read the weights-backuper snapshot (previously they
  silently read live params — matters under
  --rematerialize-param-from-master-weight).
- LoRA over distributed engines no longer requires PP=1.

Intended behavior changes (distributed): bucket composition (as in the
bucketing PR), the weights source, and nothing else — sender sets and
collective patterns are unchanged per protocol.


guapisolo (approved):

guapisolo (approved):

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
```diff
@@ -75,33 +77,21 @@ def connect_rollout_engines(
...
-        if self._is_source:
-            self._group_name = f"miles-pp_{pp_rank}"
-
-        if self._is_source:
+        # One sender per replica set; one NCCL group (sender + all engines) per shard.
+        placement = self._hf_weight_iterator.placement
+        replica_rank, _ = get_data_replica_rank_and_size(get_parallel_state(), placement)
+        self.is_sender = replica_rank == 0
+        shard = 0 if placement.gather_pp else get_parallel_state().pp.rank
+        self.is_lora_sender = self.is_sender and shard == 0
```
guapisolo, line 85: Human cmt:
Here is a small hardcode, non blocking issue. suggest add a TODO here and fix later for native lora dissagregate case.

Here we assume `is_lora_sender` is always called with `placement.gather_pp=True` (using megatron bridge). If `placement.gather_pp=False` (e.g. native-lora dissagregate placement, banned now), only pp0's lora weight will be sent.

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/delta.py
```diff
@@ -134,11 +136,10 @@ def connect_rollout_engines(
...
         self._group_name = "miles-disk-delta"
-
-    @property
-    def _is_source(self):
-        """If it's the source gpu producing the gathered HF tensors this rank publishes."""
-        return get_parallel_state().intra_dp_cp.rank == 0 and get_parallel_state().tp.rank == 0
+        placement = self._hf_weight_iterator.placement
+        replica_rank, _ = get_data_replica_rank_and_size(get_parallel_state(), placement)
+        self.is_sender = replica_rank == 0
+        self.is_lora_sender = self.is_sender and (placement.gather_pp or get_parallel_state().pp.rank == 0)
```
guapisolo, line 142: Whether is should be ?

```
is_lora_sender = 
gather pp == True and pp rank = 0
or gather pp = False 
```

miles/backends/megatron_utils/update_weight/update_weight_from_distributed/broadcast.py
guapisolo: Whether is should be ?

is_lora_sender = 
gather pp == True and pp rank = 0
or gather pp = False

Zhichenzzz (approved):

PR #2751, opened 2026-08-25
refactor(update-weight): extract the engine session RPCs into session.py

yueming-yuan:
Weight-update refactor (#1360), stacked on #2747.

One story: every engine-side RPC of a weight-update session goes through one
module — `training_utils/weight_update/session.py` (pause/resume, begin/end,
selector, version stamp, sync-result checks). The mixin, colocate, and delta
call sites are rewired; `_check_weight_sync_results` becomes the public
`session.check_weight_sync_results`.

Intended behavior change: the colocate path no longer flushes the KV cache
under `--pause-generation-mode in_place` (the distributed path already
skipped it; in_place exists to preserve that cache).


miles/backends/training_utils/weight_update/session.py
```diff
@@ -0,0 +1,72 @@
...
+def end_weight_update(rollout_engines: Sequence[ActorHandle]) -> None:
+    """Close the session (post-load + quantization post-process on the full model)."""
+    ray.get([engine.end_weight_update.remote() for engine in rollout_engines])
+
+
+def set_weight_version(rollout_engines: Sequence[ActorHandle], weight_version: int) -> None:
+    ray.get([engine.update_weight_version.remote(weight_version=str(weight_version)) for engine in rollout_engines])
+
+
+def weight_update_selector(args: Namespace) -> str:
```
guapisolo, line 44: human cmt

[P2] Let each weight iterator own the selector decision

`weight_update_selector` reopens the raw/bridge distinction inside the generic session layer, even though `get_hf_weight_iterator` has already resolved that distinction to `HfWeightIteratorDirect` or `HfWeightIteratorBridge`. This leaves session guessing from `args.mtp_num_layers` and `megatron_to_hf_mode`; for bridge it always returns `all`, so a bridge model with no exported MTP weights still selects and post-processes the frozen draft—the limited failure that `target` is intended to avoid.

Could we move `weight_update_selector` onto the HF weight iterator interface, have Direct and Bridge each decide and return `target` or `all` for its own conversion path, and leave session to only forward that value?

yueming-yuan: great point! fixed in #2753

Zhichenzzz (approved):

PR #2752, opened 2026-08-25
refactor(update-weight): weight transfer protocol ABC and backend-neutral updater

yueming-yuan:
Part of #1360. Stacked on #2751 (review only the last two commits).

## What

Splits the distributed weight-update mixin into its two roles:

- **`training_utils/weight_update/protocol.py`** — `WeightTransferProtocol`: the transport contract. `connect(...)` makes every pairing decision once (`is_sender` / `is_lora_sender` / channels); the stream surface is `begin_sync` → `send_bucket`* → `after_base_weights` → `finalize`, plus `send_adapter` for LoRA. `get_weight_transfer_protocol(args)` is the factory.
- **`training_utils/weight_update/updater.py`** — `WeightUpdater`: the backend-neutral driver. Owns the iterator (built against `protocol.required_placement`), the engine session frame (pause → begin → … → version → end → resume via `session.py`), the base-bucket stream (senders transmit, other ranks join the gathers), and single/multi LoRA orchestration.

`broadcast` / `disk-delta` / `p2p` / `rdt` become pure protocols in place (file move to `training_utils` is the next PR); `mixin.py` is deleted. The megatron actor constructs `WeightUpdater` for every non-colocate path; colocate stays on `UpdateWeightFromTensor` until the next PR.

## Design notes (deviations from the agreed plan)

1. **`begin_sync(weight_version, iter_buckets) -> bool` is back.** We had agreed delta could ride the standard frame, but its first call only seeds the baseline snapshot: no version bump, no engine session, and the weight-checker CI path stamps version 0 explicitly. A pre-frame hook that can veto the round (collectively) is the honest contract; `iter_buckets` lets delta drive the bucket stream for seeding. Broadcast/p2p/rdt return `True` (p2p uses the hook to register its pinned CPU memory on first sync — a local op, previously done post-pause).
2. **`uses_session_frame = False` for disk-delta.** Today's delta never pauses/begins/ends/resumes the engines — the transport is asynchronous behind the engines' per-host flock, and the reload pauses each engine itself. Forcing the frame on it would be a behavior change, so the frame is a protocol classvar instead.
3. **`rollout_engine_lock` is part of `connect`'s engines info** (broadcast serializes NCCL sends with it; the others ignore it).
4. **RDT's `_finalize_and_resume_engines` override is dropped**: it set the engine weight version and then called the mixin's finalize, which set it again. The updater sets it exactly once.

## Not in this PR

- Colocate (`UpdateWeightFromTensor`) onto the updater — next PR.
- Moving the four protocol files under `training_utils/weight_update/` — final mechanical PR.

Tests: lock-lifecycle and LoRA-orchestration tests retargeted from the mixin to `send_bucket` / `WeightUpdater._send_lora_adapter` / `send_adapter`.


guapisolo:
codex cmt

[P2] Restore unconditional disk-delta round cleanup

Before this refactor, `_encode_delta()` kept the executor and pinned-buffer queue round-local and guaranteed `pool.shutdown()` in `finally`. `WeightUpdater.update_weights()` now calls `after_base_weights()` only after the entire stream succeeds, while that hook is the only path that shuts the disk-delta pool down. An iterator, send, or worker exception therefore leaves executor tasks and pinned buffers live; even a successful round leaves `self._free_q` retaining the preallocated pinned tensors until the next sync. Under disk-delta this can tie up many GiB of host-pinned memory and starve the training process. Please add a protocol cleanup/abort hook invoked from `finally`, and clear `_pool`, `_inflight`, and `_free_q` on both success and failure.

guapisolo:
codex cmt

[P3] Make the progress label part of the protocol contract

`WeightUpdater` reads `protocol._group_name`, but `WeightTransferProtocol` never declares it. The built-in protocols happen to set it during `connect()`, so existing modes are unaffected, but a new implementation can satisfy every declared ABC requirement and still fail on its first base sync with `AttributeError` when the progress bar is created. This is limited to protocol extensibility and is non-blocking, but it makes the new abstraction contract incomplete. Please expose a public `display_name` or `group_name` with a default on the ABC and use it here.

guapisolo (approved):
approve. but codex cmt seems to worthy take a look.

Zhichenzzz (approved):

PR #2753, opened 2026-08-25
refactor(update-weight): colocated transfer as a protocol on the unified updater

yueming-yuan:
Part of #1360. Stacked on #2752 (review only the last commit).

## What

`UpdateWeightFromTensor` becomes a `WeightTransferProtocol`; every megatron path now runs on `WeightUpdater`, and the actor has one construction site. Net −81 lines.

- **Colocate as a protocol**: `required_placement = gather_pp=True`; `connect` keeps the hybrid split (colocated IPC + NCCL tail) and the engine-layout Gloo groups; `send_bucket` = IPC gather + hybrid broadcast; `send_adapter` = the colocated LoRA send (engine-src unload stays internal to the transport). `is_sender` = "this rank is covered by an engine" — the per-engine `gather_object` is collective among group members, so every covered rank sends; placeholder ranks now iterate with `materialize=False` (join the gathers, keep nothing) instead of no-opping inside the send.
- **`needs_base_resync_for_lora` protocol fact**: the old `skip_base_sync` composition lives in the updater (`sync_base = not is_lora or protocol.needs_base_resync_for_lora`); the persistence knowledge lives with the protocol — remote engines keep their base (default False), colocate recomputes at connect (hybrid / CPU backup / persistent GPU copy, weight-checker override).
- **MM towers as an updater hook**: `_mm_tower_bucket()` reads the frozen vision/audio tower tensors once from `hf_checkpoint` and appends them to every base sync — now on **all** protocols, closing the gap where distributed inkling runs never resynced towers.
- **Adapter lifecycle owned by transports**: `unload-before-reload` moved from `WeightUpdater._send_lora_adapter` into `broadcast.send_adapter` (tracks `_lora_loaded` itself). Forced by colocate: its `send_adapter` is a per-engine-group collective and unloads on each engine src internally, so a global updater-side unload would fire once per rank per engine.

## Behavior changes to review

1. **dist-LoRA syncs no longer call `begin/end_weight_update`**: the frame's begin/end brackets base-weight writes and is now gated on `sync_base`. The colocate path already skipped them deliberately when no base bytes land; the mixin called them incidentally around adapter-only syncs.
2. **Colocate now stamps the engine weight version** (`set_weight_version` in the frame). Previously the version only rode the per-bucket send RPC — and adapter-only colocate syncs never stamped it at all.
3. **Colocate + multi-LoRA now raises `NotImplementedError`** (`send_adapter(upsert=True)`); previously it silently did a single-adapter-style sync of whatever `get_hf_lora_weights()` returned.
4. **MM towers on distributed paths**: per-PP-shard senders each send the tower set (duplicates load idempotently).
5. Dead code dropped: `_lora_base_synced` (written, never read), the unused `lora_loaded` parameter of `_send_to_colocated_engine`.

## Not in this PR

Moving the protocol files under `training_utils/weight_update/` — final mechanical PR.

Tests: colocate LoRA pass-through retargeted to `send_adapter` (+ upsert-rejection case); empty-base-iteration retargeted to `WeightUpdater`; unload-before-reload tests moved from the updater orchestration suite into the broadcast transport suite.


yushengsu-thu:
One thing that might be an issue with the new `is_sender = self._ipc_gather_group is not None` gate: if the global source rank is not covered by any colocated engine (multi-model / gpu_offset layouts where engines don't start at GPU 0), it gets `is_sender=False`, so all weight sends are silently skipped while `update_weights` still bumps `weight_version` and finalizes — engines would quietly keep stale weights. The old path called `_send_base_params` unconditionally on every rank. Might be worth a connect()-time assert that the sender set covers the source rank.

yueming-yuan:
> One thing that might be an issue with the new `is_sender = self._ipc_gather_group is not None` gate: if the global source rank is not covered by any colocated engine (multi-model / gpu_offset layouts where engines don't start at GPU 0), it gets `is_sender=False`, so all weight sends are silently skipped while `update_weights` still bumps `weight_version` and finalizes — engines would quietly keep stale weights. The old path called `_send_base_params` unconditionally on every rank. Might be worth a connect()-time assert that the sender set covers the source rank.

good catch! fixed in https://github.com/radixark/miles/pull/2871

Zhichenzzz (approved):

PR #2754, opened 2026-08-25
refactor(update-weight): move the transfer protocols into training_utils/weight_update

yueming-yuan:
Part of #1360. Stacked on #2753 (review only the last four commits). The final PR of the step-5 stack: the whole weight-update engine now lives in `training_utils/weight_update/`.

## What

- **Commit 1 — decouple the colocated protocol's backend imports**: `lora_base_cpu_backup_enabled` (pure args inspection) moves from `megatron_utils/lora_utils.py` to `miles/utils/lora.py` next to `is_lora_enabled`; both callers follow. The `from ..sglang import ...` compat shim is inlined as direct sglang imports (`megatron_utils/sglang.py` stays behind as the megatron actor's dep manager).
- **Commit 2 — the move** (`git mv`, imports/path-constants updated, zero logic diff):

| from `megatron_utils/update_weight/` | to `training_utils/weight_update/` |
|---|---|
| `update_weight_from_distributed/broadcast.py` | `broadcast.py` |
| `update_weight_from_distributed/delta.py` | `delta.py` |
| `update_weight_from_distributed/p2p.py` | `p2p.py` |
| `update_weight_from_distributed/p2p_transfer_utils.py` | `p2p_transfer_utils.py` |
| `update_weight_from_rdt.py` | `rdt.py` |
| `update_weight_from_tensor.py` | `colocate.py` |

  Three test files whose subjects all moved follow into `tests/fast/backends/training_utils/weight_update/` (the lock test is renamed `test_broadcast_lock.py` to match `broadcast.py`).
- Two trailing commits: pre-commit (isort) normalization and one stale docstring path.

What stays in `megatron_utils/update_weight/`: the megatron iterator implementations (`hf_weight_iterator*.py`) and `common.py` — backend code behind the iterator interface.

## Mechanical Move

The transform script reproduces this PR's entire tree from its base commit (`verify_mechanical_refactor` scaffold: applies the transform to a fresh worktree at base, runs pre-commit, diffs against the PR head — verified PASS locally). Script inline below; run from the repo root.

<details>
<summary>transform_move_weight_update_protocols.py</summary>

```python
#!/usr/bin/env python3
"""Reproducible transform: move the weight-transfer protocols into training_utils/weight_update/.

Commit A re-homes the colocated protocol's two backend imports
(lora_base_cpu_backup_enabled -> miles/utils/lora.py; sglang compat imports inlined).
Commit B is the pure directory move plus import-path fixes.

[...]

PR #2756, opened 2026-08-25
refactor(update-weight): rewrite LoRA weight update path

yueming-yuan:
ci-sglang-pr: #36384

Part of #1360 Step 6 (design: `lora_weight_sync_cleanup.md`). Stacked on #2754. Engine counterpart: sgl-project/sglang#36384 — **miles tracks the sglang-miles tip continuously, so this PR must merge only after that change is in sglang-miles (and sglang-miles must keep the legacy routes until this merges).**

## What

The second LoRA send path is deleted; adapter tensors are ordinary entries in the bucketed stream, and adapter identity moves to a control plane.

- **Iterator**: `iter_hf_base_weights` → `iter_hf_weights(weights, *, include_base=True, adapters=(), materialize=True)`. Adapter units join the same unit stream under `{lora_name}:{hf_key}` names via the new `_iter_hf_adapter_units` hook. The megatron base implements it as the existing PP-local export hook (`_export_pp_local_lora`, unchanged in bridge/direct) + the PP gather now **placement-gated** (runs for gather_pp placements — bridge, colocate; disappears for direct+dist) instead of unconditional. The zero-export / no-lora-names guards move into the hook.
- **Updater**: one loop. `_adapters_to_push()` (single: `miles_lora`; multi: `__miles_slot_{i}` sorted) + eager `register_lora_adapter` for new tenants before `begin` (engine validates r ≤ max_lora_rank before any bytes move); the frame is unconditional again with the scope passed as `begin_weight_update(..., sync_base=...)` — reverting the #2753 call gating whose substance (no quant round-trip on adapter-only rounds) now lives in the session scope. `--check-lora-weight-equal` becomes a driver-side sha256 manifest accumulated during the stream and verified engine-side at `end_weight_update` (now on **both** transports; previously colocate-only).
- **Deletions** (net −326): `send_adapter` from the ABC + both implementations, the `is_lora_sender` role from all five protocols, `get_hf_lora_weights` / `_export_lora_named_tensors`, `_send_lora_adapter` / `_send_multi_lora_adapters`, unload-before-reload (+ `session.unload_lora_adapter`), the colocate LoRA IPC branch and the `_repack_onto_fresh_storage` cuMem workaround (adapter tensors now travel in sync-time-allocated `FlattenedTensorBucket`s, which is exactly why the base path never had the problem).

## Behavior changes

1. Single-LoRA becomes upsert-in-place (no unload/reload window; engine-side rollback on failure).
2. Colocate multi-LoRA works (previously `NotImplementedError`).
3. LoRA runs stamp `weight_version` (frame is unconditional); the actor-side CI version assertion can be re-enabled in a follow-up.
4. Adapter placement follows base placement (PP-sharded senders under direct+dist; bridge unchanged).
5. `supports_lora` remains the gate for p2p/rdt/disk-delta (unchanged: unsupported).

Tests: iterator template tests retargeted to adapter units in the stream; transport-era `send_adapter`/orchestration test classes removed (their engine-side successors live in sglang's `test_lora_stream_update.py`).


yushengsu-thu (approved):

guapisolo:
Non-blocking TODO for later: in the distributed path, only one trainer rank actually sends each bucket, so we should eventually gather TP/ETP/EP weights only to that sender instead of all-gathering them onto every trainer rank. The tensor/CUDA-IPC path can keep all-gather. Avoiding the full-tensor broadcast to every rollout rank is a separate shard-aware transport change. This PR does not need to address either.

miles/backends/megatron_utils/update_weight/hf_weight_iterator.py
```diff
@@ -27,13 +28,28 @@ class MegatronHfWeightIteratorBase(HfWeightIteratorBase):
...
         return get_hf_atomic_update_groups(self.model_name, q_lora_rank=self.args.q_lora_rank)
 
-    def _export_lora_named_tensors(self, adapter):
-        # Both megatron exporters gather TP/EP but not PP.
-        return _gather_pp_full_adapter(self._export_pp_local_lora(adapter))
+    def _iter_hf_adapter_units(self, lora_name, adapter, *, materialize):
+        """Both megatron exporters are PP-local after gathering TP/EP; the PP
+        gather runs only where the resolved placement asks for it."""
+        named_tensors = self._export_pp_local_lora(adapter)
+        if self.placement.gather_pp:
```
guapisolo, line 44: human cmt

[P2] Track native-LoRA disaggregate reachability

This placement-gated gather fixes the hardcoded `is_lora_sender = self.is_sender and shard == 0` behavior discussed in #2747: with `gather_pp=False`, adapter tensors can now remain PP-local and ride each shard's normal sender stream. 

Native-LoRA disaggregate is still blocked by the actor-side `megatron_to_hf_mode == "bridge"` assertion, so this path is not production-reachable today and no supported run is affected. 

Please leave a TODO linking this branch to removal of that guard when native-LoRA disaggregate is enabled.

miles/backends/training_utils/weight_update/updater.py
```diff
@@ -105,66 +109,66 @@ def update_weights(self) -> None:
...
+            assert adapters is not None, "actor must set multi_lora_adapters before update_weights"
+            return [(slot_lora_name(adapters[name].slot), adapters[name]) for name in sorted(adapters)]
+        return [(LORA_ADAPTER_NAME, None)]
+
+    def _get_new_lora_registrations(self, adapters: list[tuple[str, object]]) -> list[tuple[str, dict]]:
+        """Adapters not yet registered on the current engine set, with their
+        per-adapter config; eager so the engine validates rank before any bytes move."""
+        new = []
+        for lora_name, adapter in adapters:
+            if lora_name in self._registered_adapters:
```
guapisolo, line 170: codex cmt

[P1] Re-register when a multi-LoRA slot changes tenants

`_registered_adapters` is keyed only by the reusable engine name (`__miles_slot_i`). After tenant A retires and tenant B reuses the same slot, this branch skips registration even though B has a new `registration_id` and may use different `r`/`lora_alpha`. SGLang then applies B’s streamed tensors using A’s stored config: a rank change can fail loading, while an alpha-only change can silently serve the wrong scaling on a supported multi-LoRA path. Cache the tenant identity or config fingerprint per slot and re-register whenever it changes.

miles/backends/training_utils/weight_update/updater.py
```diff
@@ -105,66 +109,66 @@ def update_weights(self) -> None:
         self.weight_version += 1
 
         sync_base = not self.is_lora or protocol.needs_base_resync_for_lora
+        adapters = self._get_updated_adapters()
 
         driver = dist.get_rank() == 0
+        if driver:
+            for lora_name, lora_config in self._get_new_lora_registrations(adapters):
+                register_lora_adapter(protocol.rollout_engines, lora_name=lora_name, lora_config=lora_config)
```
guapisolo: codex cmt

[P2] Leave the partial-recovery ordering to the FT refactor

This FT-only path can corrupt in-flight outputs: after a partial recovery, reconnect clears `_registered_adapters`, so this replays the mutating empty-tensor registration to surviving engines before `pause_engines()` and can temporarily zero a resident LoRA. Since the upcoming FT refactor will touch this lifecycle, please leave a TODO here to move registration inside the quiesced transaction and commit the cache only after every engine RPC succeeds, with rollback/resume on failure.

guapisolo:
After human check, three codex cmt make sense to me. Could you let your agent check?

miles/backends/training_utils/weight_update/updater.py
```diff
@@ -105,66 +109,66 @@ def update_weights(self) -> None:
...
+                        record_lora_checksums(bucket, checksums)
+                    protocol.send_bucket(bucket, self.weight_version)
+                    pbar.update(1)
+            protocol.after_base_weights()
             dist.barrier(group=get_gloo_group())
 
         with timer("finalize_and_resume_engines"):
             protocol.finalize(self.weight_version)
             if protocol.use_weight_update_session and driver:
                 set_weight_version(protocol.rollout_engines, self.weight_version)
```
guapisolo: codex cmt

[P2] Confirm that `weight_version` means a committed update

This publishes the new version before `end_weight_update` verifies and applies the staged LoRA tensors. The mismatch only occurs on the limited failure path (for example, a checksum mismatch or apply rollback), but then an engine can report the new version while still serving the previous adapter. Is `weight_version` intended to mean that transfer started, or that the update committed? If it is a commit marker, please publish it only after a successful `end_weight_update`; the SGLang counterpart should likewise defer its per-bucket version update until the session commits.

miles/backends/training_utils/weight_update/updater.py
yueming-yuan: pending this first, will fix in multi lora PR series

miles/backends/training_utils/weight_update/updater.py
yueming-yuan: fixed

miles/backends/training_utils/weight_update/updater.py
yueming-yuan: fixed

miles/backends/megatron_utils/update_weight/hf_weight_iterator.py
Zhichenzzz, line 27: From claude code: forced_placement = gather_pp=True on the Megatron base changes the sender topology of bridge-mode broadcast versus main. On main the distributed path always streamed base weights PP-locally through the raw path with one sender per PP stage (mixin.py:98-130, group miles-pp_{pp_rank}), and the bridge iterator was only ever driven for LoRA. Here resolve_placement promotes broadcast's gather_pp=False to True, get_data_replica_rank_and_size degenerates to dist.get_rank(), and broadcast.py:52 leaves exactly global rank 0 as sender (miles-pp_0) — while every other PP rank still runs the full bridge export with materialize=False just for the collectives. On PP=N that is an N× drop in base-sync bandwidth plus N-1 discarded full-model exports per sync. Either let the bridge iterator produce PP-local shards for protocols that don't require gather_pp, or state the topology change in the PR description and measure it.

miles/backends/training_utils/weight_update/updater.py
```diff
@@ -105,66 +109,64 @@ def update_weights(self) -> None:
...
-                    self._send_lora_adapter()
+            pbar = tqdm(desc=f"[{protocol._group_name}] Update weights", total=0) if protocol.is_sender else None
+            for bucket in self._hf_weight_iterator.iter_hf_weights(
+                self.weights_getter(),
+                include_base=sync_base,
+                adapters=adapters,
+                materialize=protocol.is_sender,
+            ):
+                if protocol.is_sender:
+                    if driver and checksums is not None:
```
Zhichenzzz, line 137: The manifest is complete only because every reachable LoRA config gathers PP (cuda_ipc requires it; broadcast is forced to it by the bridge iterator, and actor.py:251-257 forbids raw-mode distributed LoRA — your TODO at hf_weight_iterator.py:40 says the same). Worth pinning that assumption where it's consumed: assert self._hf_weight_iterator.placement.gather_pp when checksums is not None, so the day the PP-local branch is enabled this fails loudly instead of verifying shard 0's manifest.

miles/backends/training_utils/weight_update/protocol.py
```diff
@@ -30,7 +31,6 @@ def __init__(self, args: Namespace) -> None:
         self.rollout_engines: Sequence[ActorHandle] | None = None
         self._connection_stale = False
         self.is_sender: bool | None = None
```
Zhichenzzz, line 33: _group_name is read by the updater (L125) but never declared here; today every protocol happens to set it on the sender ranks, so it works. Declaring self.group_name = "miles" (and self.update_weight_metrics: dict = {} instead of the __dict__.pop) in the base makes the contract visible.

Zhichenzzz (approved):
LGTM by human reviews. but claude code raised two suggestions and one nit, plz refer to them

miles/backends/megatron_utils/update_weight/hf_weight_iterator.py
yueming-yuan: This seems to be an issue on main, it uses the `raw` converter for `bridge` models, so it requires the mbridge style converter to be defined for a megatron-bridge model to run with disaggregated mode.. The new version is using megatron-bridge native way, which can be slower than raw, but it is the real `bridge` mode

PR #2760, opened 2026-08-25
feat(tito): support Qwen3.8 27B and Flash-Next templates

guapisolo:
## Summary

Add Qwen3.8-27B and Qwen3.8-Flash-Next TITO families with one fixed renderer.

## Motivation

Qwen3.8 serializes reasoning effort into the system prefix, so Miles needs a family-owned renderer when it supplies pretokenized `input_ids`. The 27B and Flash-Next tokenizers share that serialization contract, while their model architectures still need distinct public TITO family names.

This PR intentionally supports the registered default reasoning mode only. Per-request reasoning customization needs the separate request-argument contract refactor so a cached session prefix cannot silently change rendering mode.

## Usage

Launch Qwen3.8-27B with `--use-session-server --tito-model qwen38small`, or Qwen3.8-Flash-Next with `--use-session-server --tito-model qwen4exp`.

Both families use `preserve_thinking=true` and `reasoning_effort=xhigh`; conflicting nested template kwargs fail validation. Top-level `reasoning_effort` customization is outside this PR.

## Design Notes

- **Mental model:** `--tito-model` selects either public Qwen3.8 name, both resolve to `Qwen38SmallTITOTokenizer`, and its fixed renderer plus Qwen3 boundary merge produces the pretokenized session input.
- **Default boundary:** request preparation forwards the registered `xhigh` kwargs to SGLang, keeping local and remote default rendering aligned without changing rollout/session state.
- **Template sources:** the renderer follows [Qwen3.8-27B `1d4bf0f`](https://huggingface.co/Qwen/Qwen3.8-27B/blob/1d4bf0f2ff6012fd82039f2fa52739d0dd7c60c0/tokenizer_config.json) and [Qwen3.8-Flash-Next `de4b8e4`](https://huggingface.co/Qwen/Qwen3.8-Flash-Next/blob/de4b8e4d43b917e7706784d8bb445c9af86a3540/tokenizer_config.json), removing only the synthetic-prefix no-user guard.
- **E2E runtime:** `qwen38small` uses [Qwen3.8-27B-FP8 `017b9c7`](https://huggingface.co/Qwen/Qwen3.8-27B-FP8/tree/017b9c7af6b5689d5dd426a76e0bc077eb5ca20a) on two H200s with FP8 KV cache and a Mamba ratio of `4.59` for the 5,120-token operating point.
- **Runtime boundary:** full Qwen3.8-Flash-Next execution still requires the separate `sglang-miles-qwen38next` runtime/image.

## Verification

- Qwen3.8/Qwen4 fixed-template and tokenizer selection checks: 23 passed on runtime-equivalent parent `929e3da82`; `aff38c14d` adds only a comment.
- `tests/fast/utils/test_utils/test_session_verify_runner.py`: 24 passed on the same runtime state.
- `uvx pre-commit run --files miles/utils/chat_template_utils/tito_tokenizer.py` and `git diff --check`: passed on the `aff38c14d` candidate.
- Exact-tip two-H200 `test_qwen38small.py`: pending after publication; the earlier run passed on `8145b66f` but is not claimed for this tip.

## Review Focus

- `QWEN38_SMALL | QWEN4_EXP`: two external names intentionally share one serialization implementation.
- `Qwen38SmallTITOTokenizer.FIXED_TEMPLATE`: the FIXME marks fixed `xhigh` as temporary until request-argument precedence is unified.
- `ModelConfig`: FP8 KV and Mamba ratio remain opt-in per model, leaving other family commands unchanged.


Shi-Dong:
@claude review always.

Shi-Dong (approved):
LGTM!

PR #1890, opened 2026-07-28
Update the mechanical-refactor-verify skill to the latest sglang version

fzyzcjy:
Part of #1837

PR #2761, opened 2026-08-26
fix(ci): exempt .claude skill tooling from the test-discovery rule

yueming-yuan:
main is red since #1890: the mechanical-refactor-verify skill update added its internal test suites under `.claude/skills/.../scripts/tests/`, and `test_no_test_file_lives_outside_the_tests_tree` flags all 25 files on every PR's merge ref. These suites are exercised by the skill's own workflow, not by the CI runners this rule guards, so `.claude/` joins `tests/` in the path exemption.


Rockdu (approved):
LGTM

guapisolo (approved):

PR #2765, opened 2026-08-26
feat(metrics): log per-rollout response lengths

Shi-Dong:
## Summary

- Log per-episode response-length metrics in addition to the existing per-sample metrics.
- Group TITO compaction fragments by `(adapter, group_index, rollout_id)` before summing their lengths.
- Report both trainable output length and the full masked-plus-unmasked sample footprint.
- Reuse the existing fallback episode identity for non-TITO samples.
- Document the metrics in the compaction example.

## New metrics

- `rollout/episode_response_length/{mean,median,max,min}` measures total trainable model-output tokens per original rollout after applying loss masks. Shared prefixes across compacted sibling samples are counted once.
- `rollout/episode_total_response_length/mean` sums masked and unmasked response tokens across every sample in an original rollout, then averages those totals across rollouts. Shared prefixes count each time they occur in a sample.
- When every rollout produces exactly one sample, every `episode_response_length/*` statistic equals its corresponding `rollout/response_len/*` statistic.

## Multi-LoRA and filtered samples

- Multi-LoRA batches skip the episode-length metrics entirely: adapter-scoped length aggregation is not defined yet, while `rollout/response_len/*` still reports per-sample lengths. Episode rewards remain adapter-scoped as of #2778.
- Rows with `remove_sample=True` contribute zero trainable tokens to `episode_response_length` but keep their full footprint in `episode_total_response_length`.

## Validation

- `30 passed` in `tests/fast/ray/rollout/test_metrics.py` on the branch merged with latest main, including the #2778 adapter-scoping tests running through the shared episode-identity helper.
- All repository pre-commit hooks passed for the changed files.


miles/ray/rollout/metrics.py
```diff
@@ -126,17 +126,51 @@ def _compute_metrics_from_samples(args, samples):
...
+
+def _compute_episode_response_length_metrics(samples: list[Sample]) -> dict[str, float]:
+    """Aggregate trainable and total response tokens per original rollout.
+
+    Session compaction can split one rollout into several training samples.
+    Sibling samples share a rollout ID, so their lengths are summed before
+    computing batch-level statistics. Effective lengths count only trainable
+    tokens; total lengths count both masked and unmasked tokens in every sample.
+    """
+    effective_lengths_by_rollout: dict[tuple[str, int | None, int], int] = {}
```
guapisolo: [P1] Skip per-rollout length metrics for Multi-LoRA

Each Multi-LoRA adapter owns an independent `RolloutDataSource`, so different adapters reuse the same `group_index`/`index`. The current key then merges distinct episodes and silently reports incorrect length statistics. Until the key is adapter-aware, skip the whole metric batch:

```suggestion
    if any(sample.adapter is not None for sample in samples):
        return {}

    effective_lengths_by_rollout: dict[tuple[str, int | None, int], int] = {}
```

Please also note in the README that these metrics are not emitted for Multi-LoRA workloads; Multi-LoRA training itself is unaffected.

Shi-Dong:
@claude review always

miles/ray/rollout/metrics.py
```diff
@@ -126,17 +126,51 @@ def _compute_metrics_from_samples(args, samples):
...
+    total_lengths_by_rollout: dict[tuple[str, int | None, int], int] = {}
+    for position, sample in enumerate(samples):
+        rollout_key = _get_rollout_key(sample, position)
+        effective_lengths_by_rollout[rollout_key] = (
+            effective_lengths_by_rollout.get(rollout_key, 0) + sample.effective_response_length
+        )
```
guapisolo, line 165: [P2] Exclude removed samples from trainable episode length

When `--rollout-sample-filter-path` sets `sample.remove_sample = True`, this metric is recorded before `convert_samples_to_train_data()` clears the loss mask. The dashboard therefore reports trainable tokens for a sample that contributes zero loss. Keep the total response length unchanged, but count its effective length as zero:

```suggestion
        effective_response_length = 0 if sample.remove_sample else sample.effective_response_length
        effective_lengths_by_rollout[rollout_key] = (
            effective_lengths_by_rollout.get(rollout_key, 0) + effective_response_length
        )
```

Please add a focused test where `remove_sample=True`, asserting effective length `0` while total length remains unchanged.

guapisolo (approved):

maocheng23 (approved):

PR #2769, opened 2026-08-26
Fix Adam and Muon optimizer state resets

vedant-harmonic:
The existing `reset_optimizer_states` implementation has two bugs:
1 - It does not reset state for muon
2 - it is incorrect for some adam implementations, e.g. CPUAdam, where the step parameter is stored in `optimizer.state` instead of `optimizer.param_groups`.

This PR supports these optimizers correctly, and fails loudly for other optimizers instead of silently not resetting state

### DIFF
--- a/miles/backends/megatron_utils/model.py
+++ b/miles/backends/megatron_utils/model.py
@@ -28,6 +28,7 @@ from megatron.training.training import get_model
 from miles.backends.megatron_utils.ft.indep_dp import allreduce_grads_and_losses_across_replicas
 from miles.backends.megatron_utils.ft.types import TrainStepOutcome
 from miles.backends.megatron_utils.local_weight_checksum import dump_local_weight_checksums
+from miles.backends.megatron_utils.optimizer_state_reset import reset_optimizer_states
 from miles.utils.audit_utils.witness.allocator import WitnessInfo
 from miles.utils.audit_utils.witness.module import witness_dump_and_clear_stale
 from miles.utils.dumper_utils import DumperMegatronUtil, DumperPhase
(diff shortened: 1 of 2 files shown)


guapisolo:
codex cmt

[P1] Dispatch on the normalized `muon` optimizer name — `miles/backends/megatron_utils/optimizer_state_reset.py:89`

Megatron validation rewrites `dist_muon` to `muon` before `train()` passes `args.optimizer` here, so this branch is unreachable for both supported Muon spellings. As a result, every Muon run with `--reset-optimizer-states` raises `NotImplementedError` at the first train call instead of clearing `momentum_buffer`; this breaks the PR’s core Muon path and blocks merge. Match the normalized `muon` value (or dispatch from the concrete optimizer type) and add a parser-to-reset regression test.

Shi-Dong:
@claude review always

Zhichenzzz:
Thanks for taking this on — the direction is right (cover Muon, reset CPUAdam's per-param `step`, fail closed on unknown state). I reviewed it and ran it on an H200 devbox against the shipped image (TE 2.17, emerging_optimizers 0.3.0), and three things stop it from working as-is:

1. **The `dist_muon` branch can never run.** Megatron's `validate_args` rewrites `args.optimizer` from `dist_muon` to `muon` (and sets `use_layer_wise_distributed_optimizer`) before training starts, and miles calls it unconditionally. By the time `train()` runs, `args.optimizer == "muon"`, neither branch matches, and every `--optimizer dist_muon --reset-optimizer-states` job raises `NotImplementedError("... does not support optimizer 'muon'")` at its first rollout — the configuration this PR targets. `--optimizer muon` / `adaptive_muon` are rejected the same way even though they build the same `TensorParallelMuon` leaves.
2. **Partial CPU offload regresses.** `HybridDeviceOptimizer` with any `--optimizer-offload-fraction < 1.0` (the AMD DSv4 recipe uses 0.75) has a FusedAdam GPU child, and the new code raises `TypeError` on it. The inline loop this replaced handled that case, because `HybridDeviceOptimizer.state` aliases the children's per-param dicts.
3. **stage-a-cpu breaks.** The new module-level `from megatron.core.optimizer import Adam ...` (plus two submodule imports) runs under `tests/fast/backends/megatron_utils/test_model_initialize.py`'s stub of `megatron.core.optimizer`, which exports none of those names, so both tests there fail with `ImportError` — and every GPU stage is gated on stage-a. This PR's own CI never got that far (its stage-a shard died on a stale-base test that main has since fixed).

I opened #3091 with a fix: your commit is kept unchanged as the first commit, and two commits on top dispatch on each leaf optimizer's class instead of the name, walk `chained_optimizers` / `sub_optimizers` / `.optimizer` recursively, and fix the fixture stubs, plus unit tests. Verified there end to end on Qwen3-4B for both `dist_muon` and `adam + --optimizer-cpu-offload --optimizer-offload-fraction 0.75`, with every moment tensor nonzero before the reset and zero after.

PR #2774, opened 2026-08-27
feat(session): configure session server workers explicitly

guapisolo:
## Summary

Add `--session-server-workers` so the starting port and instance count are configured independently.

## Motivation

Encoding the instance count as the end of `--session-server-port START END` overloads one flag with two independent settings. Separate starting-port and worker-count inputs let operators change concurrency without recomputing an endpoint.

## Usage

```bash
--use-session-server \
--session-server-port 5000 \
--session-server-workers 16
```

This launches 16 independent session-server processes on ports 5000 through 5015. Omitting `--session-server-workers` selects the 32-worker default; an explicit value overrides it.

## Design Notes

- **Mental model:** `--session-server-port` and `--session-server-workers` flow through `_resolve_session_server_ports` to launch one independent server on each consecutive port with the existing per-session affinity.
- The old two-value port form is rejected instead of retained as a compatibility path, so the starting port and instance count each have one owner.
- In-repo example launchers explicitly request 32 workers so their migration does not silently depend on the new default.
- Generated example docs mirror the source READMEs' session ports 30000-30031 and router port 31000.
- Non-positive worker counts fail before any session-server process is spawned.

## Verification

- Session-worker production checks on the unchanged Python candidate at `97498d439`: exact argparse and resolver source harnesses passed.
- `uvx ruff check` across every changed Python file at `97498d439`: passed.
- README source comparison at `97498d439`: both network guides match the launchers' session ports 30000-30031 and router port 31000.
- `uvx --from pre-commit pre-commit run sync-example-docs --all-files`: passed on `97498d439`.
- `uvx pytest --confcutdir=tests/fast/doc tests/fast/doc/test_sync_example_docs.py::test_docs_examples_matches_the_readmes -q`: passed on `97498d439`.
- Full local pytest and real session-server startup remain unverified because the local environment lacks project runtime dependencies; remote CI remains required.

## Review Focus

- `_resolve_session_server_ports` in `miles/ray/rollout/router_manager.py`: consecutive-port expansion and the intentional removal of the old end-port form.
- Example source and generated network guides: required port reachability stays synchronized with launcher configuration.
- Automatic start selection probes only the starting port before deriving the remaining consecutive ports; review conflict behavior for that derived block.

### DIFF
--- a/examples/experimental/terminus-compaction/run.py
+++ b/examples/experimental/terminus-compaction/run.py
@@ -173,6 +173,7 @@ def _agent_args(args: ScriptArgs) -> str:
         "--use-session-server v2 "
         f"{bind_arg}"
         "--session-server-port 30000 "
+        "--session-server-workers 32 "
     )
 
 
--- a/examples/swe-agent-harbor-docker/run.py
+++ b/examples/swe-agent-harbor-docker/run.py
@@ -174,6 +174,7 @@ def execute(args: ScriptArgs):
         "--tito-model glm47 "
         "--use-session-server "
         "--session-server-port 30000 "
+        "--session-server-workers 32 "
     )
 
     misc_args = (
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -2546,7 +2546,7 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
             parser.add_argument(
                 "--session-server-workers",
                 type=int,
-                default=1,
+                default=32,
                 help="Number of standalone session servers to launch on consecutive ports.",
             )
             parser.add_argument(
--- a/miles/utils/arguments.py
+++ b/miles/utils/arguments.py
@@ -2540,11 +2540,14 @@ def get_miles_extra_args_provider(add_custom_arguments=None):
             parser.add_argument(
                 "--session-server-port",
                 type=int,
-                nargs="+",
                 default=None,
-                help="Port(s) of the standalone session servers. One value: a single server on "
-                "that port. Two values: a half-open range [start, end), one server per port. "
-                "Auto-allocates a single port if not set.",
+                help="Starting port for standalone session servers. Auto-allocated if not set.",
+            )
+            parser.add_argument(
+                "--session-server-workers",
+                type=int,
+                default=1,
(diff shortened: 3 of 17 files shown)


guapisolo:
@claude review

miles/utils/arguments.py
```diff
@@ -2540,11 +2540,14 @@ def add_session_arguments(parser):
...
                 default=None,
-                help="Port(s) of the standalone session servers. One value: a single server on "
-                "that port. Two values: a half-open range [start, end), one server per port. "
-                "Auto-allocates a single port if not set.",
+                help="Starting port for standalone session servers. Auto-allocated if not set.",
+            )
+            parser.add_argument(
+                "--session-server-workers",
+                type=int,
+                default=32,
```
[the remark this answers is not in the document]
guapisolo: Thanks — the 32-worker default is intentional. A single session-server process leaves substantial fixed overhead, so defaulting to 1 would make operators manually opt into the intended efficient configuration. In e9ea5e8c0, all in-repo launchers that pass `--session-server-port` now also pass `--session-server-workers 32` explicitly, so their migration is visible rather than silent; external callers that omit the flag intentionally adopt the new 32-worker default.

Shi-Dong:
@claude review always

Shi-Dong (approved):

examples/experimental/terminus-compaction/run.py
Shi-Dong, line 1: Can you make sure the README in this example is up-to-date?

examples/swe-agent-harbor-docker/run.py
Shi-Dong, line 1: Also this one.

examples/experimental/terminus-compaction/run.py
guapisolo: Updated in 0cdda4e87: the README now states that the launcher uses ports 30000-30031 for its 32 session-server workers and port 31000 for the SGLang router.

examples/swe-agent-harbor-docker/run.py
guapisolo: Updated in 0cdda4e87: this README now documents the same 30000-30031 session-server range and port 31000 for the SGLang router.

PR #2778, opened 2026-08-27
fix(metrics): scope episode rewards by adapter

Shi-Dong:
## Summary

- scope rollout reward aggregation by `AdapterRef`
- prevent unrelated Multi-LoRA rollouts with reused local IDs from sharing one reward bucket
- add a regression test covering unequal compaction-fragment counts across adapters

## Why

Each Multi-LoRA adapter owns an independent rollout data source, so `(group_index, rollout_id)` is not globally unique. Without adapter identity in the key, `rollout/episode_raw_reward` can silently weight one adapter more heavily when its rollout produces more training samples.

This follows up on the pre-existing issue identified in [PR #2765](https://github.com/radixark/miles/pull/2765#discussion_r3869035608).

## Validation

- `24 passed` in `tests/fast/ray/rollout/test_metrics.py`
- Ruff check passed for both changed files
- direct real-code probe now reports the correct per-rollout mean (`0.5` instead of `0.75`)

### DIFF
--- a/miles/ray/rollout/metrics.py
+++ b/miles/ray/rollout/metrics.py
@@ -14,7 +14,7 @@ from miles.utils.metric_utils import (
 )
 from miles.utils.misc import load_function
 from miles.utils.tracking_utils import tracking
-from miles.utils.types import Sample
+from miles.utils.types import AdapterRef, Sample
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Shi-Dong:
@claude review always

guapisolo (approved):
Good fix.

maocheng23 (approved):

PR #2777, opened 2026-08-27
Qwen3.8-Flash-Next RL training support

Zhichenzzz:
ci-sglang-pr: sglang-miles-qwen38next
ci-megatron-pr: #89

Adds RL training support for Qwen3.8-Flash-Next: a 48-layer hybrid MoE with hyper-connections (HC), a frozen host-resident PLE n-gram embedding, and QSA sparse attention. docker image is `docker pull radixark/miles:qwen38next`.
image
image

PR #2785, opened 2026-08-27
docs: add the Qwen3.8-Flash-Next RL recipe page

Zhichenzzz:
Qwen3.8-Flash-Next is the GDN + QSA hybrid MoE preview of the Qwen4 architecture, not a variant of the dense Qwen3.8-27B, so it gets its own page beside Qwen3-Next rather than a section under Qwen3.8.

The page names the three components that sit outside a stock Megatron layer and the part of the recipe each one forces: hyper-connections in place of the block layernorms, with the hc_head_contraction spec slot they needed in Megatron-LM#89; QSA sparse attention, reimplemented training-side with a Triton forward and backward; and the frozen host-resident PLE table. Then the image, the checkpoint conversion, the launch shapes and the metrics CI gates.

Implementation is radixark/miles#2777, still open, pinned together with its SGLang and Megatron branches in docker.io/radixark/miles:qwen38next.


Shi-Dong:
@claude review always

Shi-Dong (approved):

yushengsu-thu (approved):

PR #2786, opened 2026-08-27
GLM 5.3 Flash RL Support

Zhichenzzz:
ci-sglang-pr: sglang-miles-glm53next
ci-megatron-pr: #89

RL training support for zai-org/GLM-5.3-Flash (model_type `glm5_next`): 45 layers = 34 KDA linear-attention + 11 DSA sparse-attention layers, 288-expert MoE (sigmoid top-8), mHC hyper-connections, NoPE MLA, kpool-compressed lightning indexer. docker image is in `docker pull radixark/miles:glm53next`

- **customized kernels** (`miles_plugins/models/glm5_next/`): per-layer KDA/DSA block spec — KDA on fla `chunk_kda` + `fused_kda_gate`, DSA as no-rope absorbed SparseMLA with a triton kpool indexer, mHC matching the GLM-5.3 reference conventions.
- **bridge** (`miles_plugins/mbridge/glm5_next.py`): HF<->Megatron weight mapping; conversion and per-rollout weight sync verified with `--check-weight-update-equal`.
- **routing replay** wired end to end; indexer-topk replay implemented, off by default.

 
IMG_1061

IMG_1062


QQontheMoon (approved):

QQontheMoon:
accidental approval, please ignore

miles_plugins/models/glm5_next/dsa.py
```diff
@@ -0,0 +1,197 @@
...
+            kv_compressed, group=parallel_state.get_context_parallel_group()
+        )
+
+        query = query.contiguous()
+        key = kv_compressed.contiguous()
+
+        q_compressed = q_compressed.detach()
+        hidden_states = hidden_states.detach()
+
+        index_q, _ = self.wq_b(q_compressed)
```
HJSang, line 108: `q_compressed` here is un-normalized: `q_layernorm` is `IdentityOp` and the q RMSNorm (HF `q_a_layernorm`) is fused inside `linear_q_up_proj`. The reference feeds the indexer the normalized tensor. In HF `modeling_glm5_next.py` that's `q_resid = self.q_a_layernorm(self.q_a_proj(x))`, then `self.wq_b(q_resid)`, and SGLang's `forward_mla` / `forward_mha` do the same. The 1/rms factor doesn't change the top-k, but γ does. γ is mostly near-uniform, with a 4–9% spread, but some channels are near zero (min 0.09 in layer 23, −0.008 in layer 43), so the trainer can select differently from the engine past `index_topk`. Suggest mirroring the kv path:
```python
q_normed = F.rms_norm(q_compressed.float(), (q_compressed.shape[-1],),
                      weight=self.linear_q_up_proj.layer_norm_weight.float(),
                      eps=self.config.layernorm_epsilon).to(q_compressed.dtype)
index_q, _ = self.wq_b(q_normed.detach())
```
The GLM-5 base (`glm5.py`) has the same pattern.

yueming-yuan:
LGTH, please check codex comment (the first one is because of my just-merged refactor; only need to solve this naming conflict. The second one might need checking)


[P1] Import the existing HF config module — [glm5_next.py:15](https://github.com/radixark/miles/pull/2786/files)
miles.utils.hf_config does not exist at the PR head; the loader lives in miles.utils.hf_utils.config. Importing the new model spec raises ModuleNotFoundError: No module named 'miles.utils.hf_config', blocking both checkpoint conversion and training.
[P1] Preserve the checkpoint’s SwiGLU clamp — [glm5.3-flash.py:39](https://github.com/radixark/miles/pull/2786/files)
Both supported checkpoints specify swiglu_limit=10, which SGLang applies to dense, shared, and routed MLPs. The trainer enables ordinary SwiGLU and never sets activation_func_clamp_value, so outputs and gradients diverge whenever activations exceed the limit. A CPU counterexample with gate/up values of 20 produces output 400 versus ≈100, and gradients 20 versus 0. Carry the checkpoint’s clamp through every MLP path.

miles_plugins/models/glm5_next/glm5_next.py
```diff
@@ -0,0 +1,158 @@
...
+    assert (
+        text_config.index_kpool > 1
+        and getattr(text_config, "index_kpool_compress", False)
+        and getattr(text_config, "index_kpool_always_select_tail", False)
+    ), "GLM-5.3 kpool indexer expects index_kpool>1 with compress and always_select_tail"
+
+    config.glm5_next_full_attn_layers = full_attn_layers(text_config)
+
+
+def _patch_mean_output_contract() -> None:
```
yueming-yuan, line 59: can we use the `hc_head_contraction` in Megatron#89 to avoid the patch here?

miles/utils/reloadable_process_group.py
yueming-yuan, line 1: Because I'm not sure about the context of this change, qq here: did we check that the changes here are correct and needed? I see Qwen 3.8-next's PR also has this change; can we make it a separate PR?

yueming-yuan (approved):

PR #2794, opened 2026-08-28
dashboard: scroll the token strip instead of paging through it

Shi-Dong:
## Summary

The token strip in the sample view showed a 1024-token window at a time. Reading a long response meant clicking `◀` / `▶` through it page by page, and the `window` / `start` inputs were the only way to reach a position in the middle. The strip now renders the whole sequence into a scroll box, so it is one continuous scroll and the scrollbar reaches any position directly.

| before | after |
| --- | --- |
| <a href="https://raw.githubusercontent.com/radixark/miles/f99090f48b337ec2b085a895fa6fce456de39635/token-scroll/before.png"></a> | <a href="https://raw.githubusercontent.com/radixark/miles/f99090f48b337ec2b085a895fa6fce456de39635/token-scroll/after.png"></a> |

Same 25,257-token sample. On the left, `tokens 0–1024 of 25257` with the paging widgets; reaching the end takes 24 clicks. On the right the paging widgets are gone, and the strip is scrolled deep into the sequence.

## Why rendering everything is free

The metric chart under the strip already fetched the full range (`start=0, end=total`). The windowed strip request was therefore asking the server for a slice of bytes it was already sending. Measured on a 64,006-token sample:

| | cost |
| --- | --- |
| full-range `/tokens` fetch | ~920 ms — already paid today, by the chart |
| DOM build, all 64,006 spans | 66 ms |
| heap for the built strip | ~29 MB |

So the windowing was never buying render time; the strip was cheap all along. Folding the strip onto the chart's payload takes the pane from three `/tokens` calls (probe + window + full range) down to one, and the whole pane now renders 1.9 s after the tab is clicked on that 64k sample.

I also measured a delegated single-container `mousemove` against the current per-span handlers: 60 ms vs 66 ms at 64k spans. That is not worth the regression risk, so the tooltip code is untouched.

## What changed

- `style.css` — `.tokens` becomes `max-height: 480px; overflow-y: auto`, matching the `.tablewrap` scroll box the tables already use. `position: relative` makes the box the offsetParent of its tokens, so a token's `offsetTop` is directly the scroll offset that brings it to the top.
- `views_tokens.js` — one full-range read replaces the probe and the windowed read; `◀` / `▶` / `window` / `start` are deleted, leaving `N tokens (prompt P)` and `color by`.

Two things follow from having the whole sequence in hand, and both are improvements the paging design could not offer:

- **The color scale is computed once over the full response** instead of per window. Previously the scale was rebuilt from whatever slice was on screen, so the same value drew a different color depending on which page you were on.
- **Spans are built once and only recolored** when `color by` changes, so switching the metric no longer discards where you had scrolled to.

The box opens at the first generated token. An agentic prompt runs to thousands of tokens of chat history, none of it carrying a per-token metric, and landing at position 0 would put the reader that far from anything worth reading. The prompt is one scroll up.

## Verification

No JS test harness exists in the repo (no `package.json`; the Python tests under `tests/fast/dashboard/` never load the static files), so this was verified by driving real Chrome via Playwright against dumps built by `tests/fast/dashboard/dummy_dump.py` through the real dump pipeline.

Behavior, on a 64,006-token sample:

| check | result |
| --- | --- |
| `/tokens` requests for the pane | **1** (was 3) |
| spans rendered | 64,006 — the whole sequence |
| box geometry | `clientHeight` 480, `scrollHeight` 29,378, `overflow-y: auto` |
| scrolling the box | box moves, `window.scrollY` unchanged |
| tooltip after scrolling | shows all 9 stats for the hovered token |
| `color by` switch | recolors, `scrollTop` 315 → 315, span count unchanged |
| chart below the box | still drawn and reachable |
| page errors | none |

Cases, each opened in a fresh browser context:

| case | result |
| --- | --- |
| 5,000-token prompt + 800-token response | opens at `scrollTop` 2186 (clamped from the response start at 2299 — the response is on screen), paging buttons `[]`, number inputs `0` |
| eval view (`?eval=1`) | only `rollout_log_probs` offered, 800 tinted, no masked spans — no train-side columns exist |
| sample with no conversation sidecar | tokens-only layout, strip scrolls the same |
| removed sample (all-zero loss mask) | 10 masked spans, **0 tinted** — masked positions stay dim and uncolored |
| 10-token sample | `scrollHeight == clientHeight`, no scrollbar — short samples look exactly as before |

The masked case is the one that matters for review: this PR replaces the inline mask check with an `isMasked` helper called from both the build and the repaint, and that case confirms both callers agree.

## Test plan

`pre-commit` covers Python only and no Python file is touched. The dashboard's Python tests are unaffected — this is static frontend only, and no API, schema, or reader behavior changes.

### DIFF
--- a/miles/dashboard/README.md
+++ b/miles/dashboard/README.md
@@ -55,8 +55,9 @@ Views:
   (`zero_std`), average weight-version staleness, eval tab.
 - **sample view** — a `conversation` tab (role-tagged turns with thinking /
   tool calls, from the trajectory sidecar) and a lazily-loaded `tokens` tab
-  (per-token metric strips + a selectable metric-vs-position chart,
-  loss-masked regions dimmed).
+  (the whole sequence in one scrollable metric-colored strip, opened at the
+  first generated token, + a selectable metric-vs-position chart, loss-masked
+  regions dimmed).
 
 ## Develop
 
--- a/miles/dashboard/static/style.css
+++ b/miles/dashboard/static/style.css
@@ -81,7 +81,12 @@ table.data tr.flagged td { color: var(--bad); }
 .tablewrap { max-height: 480px; overflow: auto; }
 
 .tabs { display: flex; gap: 4px; margin-bottom: 10px; }
-.tokens { line-height: 1.9; font-family: ui-monospace, monospace; font-size: 13px; word-break: break-all; }
+/* position: relative makes the box the offsetParent of its tokens, so a token's
+   offsetTop is the scroll offset that brings it to the top of the box */
+.tokens {
+  line-height: 1.9; font-family: ui-monospace, monospace; font-size: 13px; word-break: break-all;
+  max-height: 480px; overflow-y: auto; position: relative;
+}
 .tok { border-radius: 2px; padding: 1px 0; position: relative; }
 .tok.prompt { color: var(--muted); }
 .tok.masked { opacity: 0.35; }
--- a/miles/dashboard/static/views_tokens.js
+++ b/miles/dashboard/static/views_tokens.js
@@ -110,7 +110,11 @@ export async function renderTokens(view, meta, route) {
 
   const conversationPane = renderConversation(conversationRow);
   const tabs = el("div", { class: "tabs" });
-  const body = el("div");
+  // both panes stay mounted and are toggled with `hidden`, because detaching a
+  // pane destroys its layout and the browser resets the token strip's scroll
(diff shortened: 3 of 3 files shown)


Shi-Dong:
@claude review always

Zhichenzzz:
Great fix! also claude code's review is worthy i think?

Zhichenzzz (approved):

PR #2802, opened 2026-08-28
agent functions: one failure contract for the OpenEnv, NeMo Gym and Harbor legs

nblintao:
Stacked on #2801.

## Purpose

The OpenEnv, NeMo Gym and Harbor agent functions ended failed episodes three different ways that all trained as `reward 0` without saying why — and OpenEnv's comment claimed returning `None` discards the sample, which it never did. Give the three legs one failure contract.

## What

One rule, applied in place in each leg (no shared module yet — this makes the copies identical so a later extraction is mechanical):

- **Discard (`InfraAbort`) only what the policy cannot have caused**: sandbox create failed past the retry budget (`SandboxUnavailable`), agent/env server unreachable (`ServerUnreachable`), a tbench2 server without the canonical scoring contract (`NonCanonicalVerifier`).
- **Score 0 with a named `exit_status` for everything else**, wall-clock timeouts included (`TimeLimitExceeded`) — a discarded timeout is an exit the policy can learn to take.
- One `exit_status` vocabulary across the legs (Harbor's set plus the three discard causes), documented next to the who-can-cause-what table.

## Behaviour change

- New discards: an unreachable agent/env server after the policy started, and OpenEnv episodes scored by a non-canonical server — both previously trained as false-negative 0s.
- Samples that were already `reward 0` stay `reward 0`, now with an `exit_status` the dashboards can split on.
- OpenEnv's `exit_status` strings change (`timeout` → `TimeLimitExceeded`, `completed` → `Submitted`).


nblintao:
@claude review

nblintao:
@claude review always

nblintao:
On the openenv_sandbox_common.py finding: not changing it. A connect or reset failure happens before the first model call, so the session has no recorded turns and `agentic_tool_call` already marks the sample ABORTED (`NoModelCalls`) and drops its group; the `AgentError` dict is never trained on. Raising `InfraAbort` there would only change the label.

Issue #2804, opened 2026-08-28
Agentic rollout: shared miles-side layer and Harbor in-process follow-ups


nblintao:
Tracking issue for the refactor sequence around the agent-function legs (Harbor, OpenEnv, NeMo Gym). Context: #2800, #2801, #2802.

Two lines of work that do not wait on each other:

**Line 1 — independent of #2801/#2802 (base: main)**
- [ ] #2805 — shared layer (`session.py`, `credentials.py`)
- [ ] #2806 — in-process Harbor agent function; #2836 — its launcher; #2837 — docs
- [ ] #2838 — Daytona recipe onto the in-process path (**do not merge before the in-process Daytona smoke**)
- [ ] Live validation of #2806 on AgentENV (public `harbor[e2b]` + mini-swe-agent, then terminus-2 on the fork), posting results to the PR and README
- [ ] **Shared golden/smoke harness (its own PR, written during that validation)**: one driver taking `(leg, backend, task set)`, running a golden episode and asserting the reward, credentials via the key-file contract — replacing the per-leg ad-hoc scripts (`scan_golden`, `eval_tbench2_via_api`, `eval_nemogym_via_api`, ...) with one that keeps the legs at feature parity. Manual / nightly, real resources.

**Line 2 — after #2801/#2802 land**
- [ ] `miles/rollout/agentic/episode.py` + `results.py`: hoist the aligned failure semantics (`run_for_training` skeleton, `exit_status` vocabulary, result schema) out of the three legs; the Harbor example's local result mapping joins them.
- [ ] Harbor in-process: raise `InfraAbort` for the platform-side Harbor exceptions (`EnvironmentStartTimeoutError` and friends) once the contract is in.
- [ ] Stall guard: with discards in place a persistent infra failure turns into silent regeneration (sync over-sampling never fills; fully-async `retry` loops). Fail the step when the dropped-group ratio (from `rollout/aborted/drop_*`) exceeds a threshold.

**Later / needs a second user first**
- [ ] `HARNESS_BINDINGS` (dialect / placement / endpoint shape / model-name prefix) into the shared layer when the NeMo Gym leg becomes its second user.
- [ ] Launcher skeleton and `agent_metrics` aggregation shared across legs.
- [ ] (maybe) Agent server as a thin uvicorn shell over the in-process function — not planned currently: the agent server and the fork stay as they are; the in-process path just has to work without them.
[...]

PR #2805, opened 2026-08-28
rollout: shared miles-side layer for agent-function legs (session URL, sandbox credentials)

nblintao:
Moves the logic that the OpenEnv, NeMo Gym and Harbor agent-function legs had in identical copies into a shared package, `miles/rollout/agentic/`, and gives it direct unit tests that need no provider account or real key. Zero behaviour change; independent of #2801/#2802 (base: `main`). Tracking: #2804.

- `session.py` — `resolve_session_url` (append `/v1`, rewrite the host to `MILES_ROUTER_EXTERNAL_HOST`): three byte-identical copies become one; the legs import it.
- `credentials.py` — the sandbox-provider credential contract from the OpenEnv launcher: `PROVIDER_CREDENTIALS`, `sandbox_key_supply` (forward the key file's *path*, never the value — the value would ride ray's `runtime_env` in plaintext), `forward_address`, `preflight_sdk`, and the worker-side `resolve_provider_api_key`.

The package stays torch-free (asserted by a test): the NeMo Gym adapter must load on CPU-only hosts.

## Tests — no real keys

Everything here is env vars + files, so `tests/fast/rollout/agentic/` covers the whole contract with tmp files and monkeypatch: the spec table's shape, forwarded vars never credential-shaped, userinfo-URL rejection, every `sandbox_key_supply` branch (incl. empty/unreadable key files and Modal's token pair), `preflight_sdk` both outcomes, `resolve_provider_api_key` all four supplies, session-URL shaping, torch-free import. The OpenEnv example keeps only its own obligations (every registered backend has a spec naming a real launcher arg).

## Mechanical move — reproducible

8 commits per `.claude/skills/mechanical-refactor-verify`: the 3 relocations are `mechanical_provable` with byte-for-byte reproduce proofs (chain verdict **PASS**, 3/3); the 5 `non_mechanical_provable` commits are small and single-purpose (two pre-move renames, the dedupe that points the legs at the shared functions, a message-prefix cleanup, the tests).

Proof folder: https://gist.github.com/nblintao/6e3d73bdbf74044244fd1d2cdea1f31d — re-run with:

```bash
gh gist clone 6e3d73bdbf74044244fd1d2cdea1f31d proof && mkdir -p proof/repro_scripts && mv proof/[0-9a-f]*.py proof/repro_scripts/
python3 .claude/skills/mechanical-refactor-verify/scripts/mechanical_refactor_reproduction_cli.py \
    --base main --branch tao/agentic-shared-layer --proof proof
```

Two proofs are hand-written `Repro`s (the generator mis-inferred those recipes); both compose only whitelisted primitives. One repo fact the proofs surfaced: `[tool.isort] ignore_whitespace = true` means isort never inserts the section blank line before first-party imports, so the new imports use the adjacent form the tooling reproduces.

### DIFF
--- a/examples/experimental/nemo-gym/nemogym_agent_function.py
+++ b/examples/experimental/nemo-gym/nemogym_agent_function.py
@@ -35,9 +35,9 @@ import logging
 import os
 import random
 from typing import Any
-from urllib.parse import urlparse, urlunparse
 
 import httpx
+from miles.rollout.agentic.session import resolve_session_url
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


Shi-Dong:
@claude review always

Shi-Dong (approved):
LGTM!

PR #2806, opened 2026-08-28
examples: in-process Harbor agent function

nblintao:
Stacked on #2805. Tracking: #2804.

## Purpose

Let a rollout worker run a Harbor trial itself when the task sandbox is on a cloud backend (E2B / AgentENV, Daytona, Modal, ...). The agent server exists to put `Trial.run()` next to a Docker daemon; with a cloud backend the sandbox is a network call away and no server is needed. The agent-server path is untouched and stays the way to run the local `docker` backend.

## What

One new example module plus its tests — the launcher is #2836 and the README/docs are #2837, so each is reviewable alone. The module docstring here is the operational reference (env vars, what the fork branch provides); #2837 links to it rather than restating it.

- `examples/experimental/harbor/harbor_agent_function.py` — `run()` builds a `TrialConfig` from the sample, runs `Trial.run()` under a wall-clock cap, and returns `{reward, exit_status, eval_report, agent_metrics}` (the agent server's result mapping). `HARBOR_ENV_TYPE` goes straight into Harbor's `EnvironmentType`; backend settings ride `HARBOR_ENV_KWARGS`. No per-backend branch in Miles.
- `HARNESS_BINDINGS` — what each harness needs to be handed the session URL (terminus-2, claude-code, opencode, mini-swe-agent, plus a default for other OpenAI-chat agents). Ported from the fork's `_agent_connection_config`, which stays the agent-server copy; the table header names that function as the sync anchor for harbor-pin bumps, and the tests assert why each line exists (terminus-2 must abort on truncation, opencode must not resolve the `openai` provider id, ...).
- Tests: unit tests against a faked `harbor` package (bindings, pass-through, result/exception mapping, timeout), and contract tests against the real package (`importorskip`) that validate the built config with Harbor's own pydantic models — no sandbox, no credentials.

## Behaviour

New path only; nothing existing changes. Two deliberate differences from the agent server: an unknown `HARBOR_ENV_TYPE` raises instead of silently running docker, and configuration errors (missing task dir, bad env vars) raise instead of scoring 0 — they fail every sample, and a loud stop beats training on silent all-zero rewards. Episode failures keep the agent-server semantics (no verdict → reward 0 with a named `exit_status`); wiring `InfraAbort` waits for #2801.

**Validated on the real platform (2026-09-02)**: the sandbox smoke (#2868) runs this path end to end with Harbor's oracle agent on Terminal-Bench-2's `fix-git` — harbor × e2b **PASS** and harbor × daytona **PASS** (reward 1.0; config building, template build, sandbox lifecycle, verifier round trip all live). Still pending: a real harness through the session server and one GRPO step — that is #2876 (registered disabled, run manually; first run deferred until this stack is polished).

### DIFF
--- a/examples/experimental/harbor/harbor_agent_function.py
+++ b/examples/experimental/harbor/harbor_agent_function.py
@@ -159,7 +159,10 @@ def _claude_code_kwargs(
 def _claude_code_env(session_url: str, api_key: str) -> dict[str, str]:
     env = {
         "ANTHROPIC_API_KEY": api_key,
-        "ANTHROPIC_BASE_URL": session_url,
+        # The Anthropic SDK appends /v1/messages itself, and the session
+        # server's Anthropic route hangs off the session root -- so hand it
+        # the root, not the /v1-suffixed URL the OpenAI-style clients get.
+        "ANTHROPIC_BASE_URL": session_url.removesuffix("/v1"),
         # Claude Code's server-side Tool Search cannot be forwarded to the backend.
         "ENABLE_TOOL_SEARCH": "false",
     }
--- a/examples/experimental/harbor/harbor_agent_function.py
+++ b/examples/experimental/harbor/harbor_agent_function.py
@@ -95,19 +95,29 @@ def _allowed_hosts(var: str) -> list[str]:
 # What Miles must know per harness to hand it the session URL: which env vars /
 # kwargs carry the endpoint and key, and what the model is called on its wire.
 # Everything else about running the harness is Harbor's.
+#
+# Ported from harbor-framework/harbor agent_server/trial_runner.py
+# ``_agent_connection_config`` at harbor-miles-v0.20.0 (53a6e92a). That copy
+# keeps serving the agent-server path and keeps evolving; when bumping the
+# harbor pin, diff that function against this table. The tests assert WHY each
+# line exists (terminus-2 must abort on truncation, opencode must not resolve
+# the ``openai`` provider id, ...), so a re-sync knows which lines carry a
+# constraint and which just follow the harness.
 
 
 @dataclass(frozen=True)
 class HarnessBinding:
-    # (session_url, api_key, sampling_params, model) -> AgentConfig.kwargs
-    kwargs: Callable[[str, str, dict[str, Any], str], dict[str, Any]]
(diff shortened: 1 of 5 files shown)


nblintao:
@claude review

examples/experimental/harbor/harbor_agent_function.py
```diff
@@ -0,0 +1,427 @@
...
+    kwargs: dict[str, Any] = {"disallowed_tools": "WebSearch,WebFetch"}
+    if max_turns := _env_int("HARBOR_AGENT_MAX_ITERATIONS"):
+        kwargs["max_turns"] = max_turns
+    return kwargs
+
+
+def _claude_code_env(session_url: str, api_key: str) -> dict[str, str]:
+    env = {
+        "ANTHROPIC_API_KEY": api_key,
+        "ANTHROPIC_BASE_URL": session_url,
```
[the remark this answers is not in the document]
nblintao: fixed

Shi-Dong:
@claude review always

examples/experimental/harbor/harbor_agent_function.py
```diff
@@ -0,0 +1,452 @@
...
+  not used      ``agent_server/``, the dashboard, the docker egress-control
+                and network-prune patches, the daytona create jitter, the
+                critic agents, S3 upload, adapters.
+
+Env vars (read on the rollout worker):
+  HARBOR_TASKS_DIR       directory holding one Harbor task dir per
+                         ``metadata.instance_id`` (required)
+  HARBOR_ENV_TYPE        Harbor ``EnvironmentType`` value (required; no default
+                         -- it decides whose quota a run spends)
+  HARBOR_ENV_KWARGS      JSON object passed as ``EnvironmentConfig.kwargs``
```
Shi-Dong, line 32: Suppose that we run Daytona with this agent function, do we still need things like `{"auto_stop_interval_mins": 540, "auto_delete_interval_mins": 1440}` to ensure that Daytona sandboxes are automatically cleaned up?

Shi-Dong (approved):

examples/experimental/harbor/harbor_agent_function.py
nblintao: Yes. I have a separate PR https://github.com/radixark/miles/pull/2838 (in draft, not tested yet) for supporting daytona in this path.

PR #2808, opened 2026-08-29
openenv e2b backend: send the template name under the field the server reads (AgentENV interim)

nblintao:
Interim client-side fix so `Template.build` works against a self-hosted AgentENV: an opt-in env var, `OPENENV_E2B_TEMPLATE_NAME_FIELD`, that makes the e2b SDK also send the template name under the field the server reads (`name`).

## Why

Every e2b Python SDK release through 2.10.2 submits a template build with the name in `alias`; AgentENV's `POST /v3/templates` reads `name` and rejects the request (`400 template name must be provided`), so `ensure_task_template` cannot bake a single task image and the whole e2b/agentenv backend stops at step one. Reproduced 2026-08-28 against the dev sandbox-service cluster (AgentENV v0.1.3); with the field added, the build takes 7 s and the rest of the path (create from alias, `commands.run`, `get_host`) works unchanged.

The proper fix is server-side — [radixark/AgentENV#8](https://github.com/radixark/AgentENV/issues/8), PR to upstream in flight. This module goes away once a server that accepts `alias` is deployed; the env var is off by default, so E2B Cloud and any other provider are untouched.

## What

- `examples/experimental/openenv/e2b_template_name_compat.py`: patches the SDK's `TemplateBuildRequestV3.to_dict` to copy `alias` into the named field. Idempotent; a no-op when the variable is unset.
- `tb2_sandbox_e2b.ensure_task_template`: applies it right before `Template.build`.
- `openenv_launch_common`: forwards the variable to rollout workers like the other `E2B_*` settings.
- `examples/experimental/agentenv/README.md`: the export line, with the reason and the removal condition.
- Tests: the shim under a fake SDK model (copies the field, idempotent, no-op when unset).

## Verification

`tests/fast/examples/experimental/openenv/` (37 tests) pass in `radixark/miles:latest`. End to end from a laptop against the dev sandbox-service cluster with `OPENENV_E2B_TEMPLATE_NAME_FIELD=name`: `scan_golden.py --tasks fix-git` scores 1.0 (35.7 s including the template bake).

### DIFF
--- a/examples/experimental/agentenv/README.md
+++ b/examples/experimental/agentenv/README.md
@@ -59,6 +59,9 @@ export E2B_API_URL=http://<server>:8000       # control plane
 export E2B_SANDBOX_URL=http://<server>:8000   # data plane (envd proxy)
 # plain-HTTP deployments only; the default is https
 export OPENENV_E2B_URL_SCHEME=http
+# the SDK sends the template name as `alias`; AgentENV reads `name`
+# (radixark/AgentENV#8). Drop this once the deployed server accepts `alias`.
+export OPENENV_E2B_TEMPLATE_NAME_FIELD=name
 # required, but any well-formed key passes: AgentENV does not check it, while
 # recent SDKs do validate the format client-side
 export E2B_API_KEY=e2b_0000000000000000000000000000000000000000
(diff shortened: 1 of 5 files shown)


nblintao:
Closing: this shim turned out to be unnecessary. The 400 only reproduces with e2b ≤2.10.2, which is what pip resolves on Python 3.9; current releases (2.46.0, Python ≥3.10 — Miles' floor) send `name` and work against AgentENV unchanged. Verified without this branch: e2b 2.46.0, an unbaked task, golden reward 1.0 on the dev sandbox-service cluster.

PR #2811, opened 2026-08-29
openenv: raise the tbench2_env contract floor to OpenEnv#1025

nblintao:
## Why

huggingface/OpenEnv#1025 (merged 2026-08-29, `38b2a3135`) closes the gap @Shi-Dong spotted: a canonical `tests/test.sh` run whose verifier never wrote `reward.txt` used to come back as `reward=0.0` **with** the `harness="tests/test.sh"` marker — byte-for-byte a genuine test failure. Now the server reports it through `observation.error` / `reward=None`.

Nothing on the miles side worked around that bug (nothing could — the reply looked valid), so there is nothing to delete. The fix lands in the existing guard's `reward is None or eval_error` branch and, in #2802, in the `eval_error` → `no_verdict_reason` mapping; neither changes. The `harness != "tests/test.sh"` branch stays: it covers pre-#1012 installs and the pytest fallback, which #1025 did not touch.

## What

- **Launcher preflight** (`openenv_launch_common.py`): the source probe of the installed `tbench2_env` now also requires `_require_canonical_verdict`. A pre-#1025 install is the one case the per-episode marker guard cannot catch (the fake 0 carries the marker), so the install-time probe is the only place to reject it. All three sandbox backends bake this install into their task images, so one probe covers them.
- **Contract floor** in the README and the four agent-function module docstrings: `#1012 (04d259ea6)` → `#1025 (38b2a3135)`.
- **Guard comment** in `multi_turn`: the `reward=None / error` bullet now names "test.sh never wrote reward.txt" alongside toolkit timeout / staging I/O.

## Not in this PR

- The remote docker-host env server behind `--openenv-env-url` cannot be preflighted; it needs a redeploy from a ≥ #1025 checkout to actually stop emitting the fake 0.
- #1025 routes both "uvx download failed" (infra, droppable) and "in-shell `timeout` fired" (possibly policy-caused) through the same error path, distinguishable only by the log tail. #2802 currently maps every `eval_error` to `VerifierError`/0; whether miles should split that by error text is a #2802 decision.

## Validation

No test changes: the guard and its tests are unchanged; the preflight has no unit test today (it imports the real `tbench2_env`). Fast tests run in CI (no torch locally).

### DIFF
--- a/examples/experimental/openenv/README.md
+++ b/examples/experimental/openenv/README.md
@@ -51,11 +51,11 @@ few tasks at a time if it does.
 
 Whichever provider you pick, install `tbench2_env` **editable**: the recipe
 bakes the installed source into each task image, so that install must carry
-the `>=` #1012 server contract. The launcher preflights the installed source
+the `>=` #1025 server contract. The launcher preflights the installed source
 and fails fast on an older one.
 
 ```bash
-git clone https://github.com/huggingface/OpenEnv.git   # >= the #1012 merge (04d259ea6, the full canonical contract for both modes); pin that sha if you need frozen reward semantics across a long run
+git clone https://github.com/huggingface/OpenEnv.git   # >= the #1025 merge (38b2a3135: canonical contract for both modes, and a missing verdict is an error rather than reward 0); pin that sha if you need frozen reward semantics across a long run
 pip install -e OpenEnv/envs/tbench2_env
 ```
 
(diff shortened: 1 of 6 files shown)


Shi-Dong:
@claude review always

Shi-Dong (approved):
LGTM!

PR #2812, opened 2026-08-29
openenv: stop the episode when the session crosses --max-seq-len

### LINKED ISSUE #2591: openenv agent loop keeps generating past the session-level --max-seq-len cut, whose tokens collect_samples discards
## What happens

The openenv tbench2 agent loop has no idea when a session has crossed `--max-seq-len`, and keeps generating turns that sample assembly will throw away.

`--max-seq-len` is enforced **after the episode**, inside the session server: `collect_samples` calls `truncate_samples_by_total_tokens`, which strips the overshoot off the last sample's output, marks it `Status.TRUNCATED`, and `break`s out of the sample list.

https://github.com/radixark/miles/blob/e4152f613/miles/rollout/session/samples/merge.py#L195-L221

https://github.com/radixark/miles/blob/e4152f613/miles/rollout/generate_hub/agentic_tool_call.py#L149-L157

Nothing signals this to the agent while the episode runs. So an episode that crosses the session budget at, say, turn 12 goes on to turn 30 (or the 3600 s wall clock), and every token after the cut is discarded — exactly the waste #2587 measured at the per-turn boundary (22.7% of all episode wall-clock on a GB300 16-node GLM-5.2 run), at a different boundary.

## Why #2587 does not cover it

The two cuts are reached by different mechanisms, and only one of them is observable in-loop:

| | per-turn cap (`--rollout-max-response-len`) | session budget (`--max-seq-len`) |
|---|---|---|
| where enforced | SGLang, during generation | session server, in `collect_samples` after the episode |
| in-loop signal | `finish_reason="length"` on the response | **none** |
| v2 tree server | later request on that node gets 409 `TruncatedGenerationError` ([session_state.py#L69](https://github.com/radixark/miles/blob/e4152f613/miles/rollout/session/v2/session_state.py#L64-L72)) | nothing — `node.truncated` is set from `finish_reason`, which never fires here |
| fixed by | #2587 | nothing yet |

So #2587's `break` cannot see this one: there is no `finish_reason` to read, and no error comes back either.

## Why exposure just went up

#2588 lowers the recipe's session budget from a hard-coded 131072 to `--max-seq-len 65536` (default). With 30 turns × up to 8192 generated tokens plus the env output fed back each turn, crossing 65k is realistic where 131k left roughly 2x headroom. The change is well justified on its own terms (CP4 activation sizing, and the 8192 per-turn cap that completed samples never touched), but it halves the distance to a boundary that has no stop-loss.

## Possible fixes

1. Track the session's accumulated token count in the agent loop and end the episode as it approaches `--max-seq-len` — same shape as #2587, one boundary over. Needs the budget plumbed to the agent function (it is currently only known to `agentic_tool_call` / the session server) and a token count the loop can trust (`usage.total_tokens` off each response is the cheap approximation).
2. Or have the session server refuse / flag a request once the stored session would exceed the budget, so the loop learns the same way it learns about a per-turn cut.

Either way the loop should stop rather than keep spending decode on tokens sample assembly has already decided to drop.

## Related

- #2587 — the same waste at the per-turn boundary, with the wall-clock measurement
- #2588 — lowers the session budget to 65536

chethanuk:
## Problem

`--max-seq-len` is enforced after the episode, inside `collect_samples`. The agent loop gets no
signal when a session crosses it — no `finish_reason`, no error — so an episode that crosses at
turn 12 keeps going to turn 30 or the wall clock, and everything past the cut is discarded. Same
waste #2587 removed at the per-turn boundary, one boundary over.

## What changes

`multi_turn` reads `usage.total_tokens` off each completion and breaks when it reaches the budget.

Assignment, not accumulation: every turn re-sends the whole conversation, so the request's
`total_tokens` (prompt + completion) already *is* the session total. It lags by one env response —
turn N's output is appended after the completion and only priced into turn N+1's prompt — so the
loop stops just after crossing, never before. That direction is deliberate; stopping early would
throw away a rollout that was still inside the budget.

The budget arrives via `metadata["max_seq_len"]`, which `agentic_tool_call.generate` already puts
there. If a turn reports no `usage`, the loop does not stop on a stale count from an earlier turn.

Two new metrics in `agent_metrics`: `session_tokens` and `stopped_on_budget`, so the frequency is
visible in a run rather than inferred.

## Testing

- `python -m pytest --confcutdir=tests/fast/examples/experimental/openenv tests/fast/examples/experimental/openenv -q`
  - 110 passed, 3 skipped
- Six new parametrized cases: crossing mid-episode, crossing on the first turn, staying under
  budget, no `max_seq_len` in metadata, a turn that reports no `usage`, and `finish_reason="length"` taking precedence.
- `ruff`, `black`, `isort` clean on both changed files

Closes #2591

cc @nblintao

### DIFF
--- a/examples/experimental/openenv/openenv_agent_function.py
+++ b/examples/experimental/openenv/openenv_agent_function.py
@@ -266,13 +266,19 @@ async def multi_turn(
     action_cls = classes["action"]
     task_id = metadata.get("task_id") or metadata.get("task_name")
     max_turns = int(os.getenv("OPENENV_MAX_TURNS", "30"))
+    raw_budget = metadata.get("max_seq_len")
+    try:
+        max_seq_len = int(raw_budget) if raw_budget is not None else None
+    except (TypeError, ValueError):
+        max_seq_len = None
 
-    async def body(env: Any) -> tuple[float | None, int, str, list[float], list[float], float, float]:
+    async def body(env: Any) -> tuple[float | None, int, str, list[float], list[float], float, float, int]:
         # Per-turn wall-clock timings. gen_times[i] is turn i's policy generation
         # latency; tool_times[i] is turn i's env.step(exec) latency. reset_time and
         # eval_time bracket the one-off reset() and the final evaluate() env steps.
         gen_times: list[float] = []
         tool_times: list[float] = []
+        session_tokens = 0
 
         t0 = time.monotonic()
         reset_result = await (env.reset(task_id=task_id) if task_id else env.reset())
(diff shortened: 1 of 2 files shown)


behroozazarkhalili:
Verified on top of current main (2ed189c0d); the branch merges with no conflicts.

- tests/fast/examples/experimental/openenv on the merge: 103 passed, 3 skipped.
- Same suite with openenv_agent_function.py reverted to main: the 6 new budget cases fail, the other 97 pass, so the tests pin the fix.
- ruff/black/isort at the pre-commit pins are clean on both files.
- Plumbing: agentic_tool_call.generate puts args.max_seq_len into metadata (agentic_tool_call.py L58-62) and the session server forwards upstream usage on the chat response (session/core.py L114-115), so usage.total_tokens is available in-loop. Daytona/Modal/E2B run through the same multi_turn (openenv_sandbox_common.py run_episode), so one change covers all four backends.

Looks ready for a maintainer look. cc @nblintao

PR #2814, opened 2026-08-30
fix(rollout): drop groups with missing rewards

guapisolo:
> **Depends on:** #3094

## Summary
Reject rollout groups with missing rewards before they reach training conversion.

## Symptom & Reproduction
- **Symptom:** A group containing `None` rewards can reach reward tensor construction and stop the batch before the optimizer. Guarding only the logger, as proposed in [#989](https://github.com/radixark/miles/pull/989), moves rather than fixes the failure.
- **Reproduction:** Complete a rollout group with a raw `None` reward or `{"score": None}` selected by `reward_key`, with no custom dynamic filter. The old collector counts it toward `rollout_batch_size`; the expected behavior is to reject it and refill or drop it according to the collector.

## Root Cause
1. Rollout collector paths such as `generate_rollout_async` accepted groups without mandatory reward validation.
2. `Sample.get_reward_value` could return an accepted `None` reward for conversion.
3. `torch.tensor` failed before optimization.

## Fix
`apply_preput_filters` evaluates `apply_aborted_filter`, `apply_missing_reward_filter`, then the configured custom filter. This preserves aborted retry/drop behavior while rejecting raw, selected, and nested missing rewards before conversion. Synchronous collectors refill rejected capacity; fully async, Multi-LoRA, and Verifiers retain their existing drop, no-enqueue, or resampling actions. A missing selected `reward_key` still raises `KeyError`.

The stock filters follow the parent's `apply_xxx_filter` naming convention. Existing dynamic sampling dotted paths remain available through deprecated wrappers inherited from #3094; their signatures, structured results, and exceptions are preserved without runtime deprecation warnings.

## Verification
- On `927a847f50909f4b0cdea83d23337a53c3219c5d`, `python -m pytest --confcutdir=tests/fast/rollout tests/fast/rollout/test_filters.py -o addopts= -q`: 21 passed, covering filter precedence, missing rewards, and wrapper compatibility; repository-wide integration fixtures were not loaded.
- Direct buffer probes passed for both legacy dotted paths, aborted recycling, reward-variance rejection, valid enqueue/get, and missing-reward rejection before custom filtering without recycling.
- The rebased tree equals the original child result plus the parent's wrapper change; independent replay preserved both source commits' raw author, committer, and message bytes.
- Ruff `0.14.7`, isort `5.13.2`, Black `24.3.0`, compilation, and `git diff --check` passed for the changed Python files.
- Full collector integration and GPU CI were not rerun locally.

## Review Focus
- `apply_preput_filters`: `aborted -> missing reward -> custom` ordering.
- Collector acceptance boundaries reject invalid rewards before conversion while retaining each collector's buffer action.
- Legacy dynamic filter paths remain compatible through the parent's deprecated wrappers.

### DIFF
--- a/examples/experimental/verifiers/verifiers_rollout.py
+++ b/examples/experimental/verifiers/verifiers_rollout.py
@@ -37,7 +37,8 @@ from miles.rollout.base_types import (
     RolloutFnTrainInput,
     RolloutFnTrainOutput,
 )
-from miles.rollout.filter_hub.base_types import MetricGatherer, call_dynamic_filter
+from miles.rollout.filter_hub.base_types import MetricGatherer
+from miles.rollout.filter_hub.common_filters import apply_preput_filters
 from miles.rollout.generate_utils.prefill_logprobs import recompute_samples_rollout_logprobs_via_prefill
 from miles.utils.lora import LORA_ADAPTER_NAME, is_lora_enabled
 from miles.utils.types import Sample
@@ -767,6 +768,7 @@ class VerifiersRolloutFn(BaseRolloutFn):
         all_groups = []
         all_traces = []
         metrics = MetricGatherer()
+
         semaphore = asyncio.Semaphore(self.max_concurrent)
         pending: set[asyncio.Task] = set()
         async with self.env.serving():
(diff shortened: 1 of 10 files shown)


miles/rollout/filter_hub/base_types.py
```diff
@@ -1,18 +1,32 @@
...
+def iter_samples(group: list[Sample | list[Sample]]) -> Iterator[Sample]:
+    for sample in group:
+        if isinstance(sample, list):
+            yield from sample
+        else:
+            yield sample
+
+
+def call_dynamic_filter(fn, args, samples: list[Sample | list[Sample]], **kwargs):
+    if any(sample.reward is None or sample.get_reward_value(args) is None for sample in iter_samples(samples)):
```
yueming-yuan: Can we move the check out of this entrypoint function `call_dynamic_filter`? This will catch some other bad types too early; e.g., when the sample is aborted, it cannot reach the corresponding filter and the error will just be "group_has_missing_reward"

yueming-yuan:
qq, when the reward function is incorrect and all groups give a None reward, will it just hang because all samples are dropping?

miles/rollout/filter_hub/base_types.py
guapisolo: good finding. I am working on refac filter behavior. This actually mix the motivation of filters.

yueming-yuan (approved):

PR #2817, opened 2026-08-30
dashboard: give a step with no samples the summary schema

Shi-Dong:
## Summary

Follow-up to #2815, which stopped the Rollouts tab from *landing* on an unusable step but left the underlying crash in place. This fixes the crash.

A step can be dumped with no samples at all. `summary()` inferred its schema from an empty row list, which yields a frame with **zero columns** rather than zero rows — so every view built on top of it died on a missing column instead of reporting an empty step.

## Blast radius

Measured by blanking one step's samples in a real dummy dump and calling each consumer directly:

| consumer | before | why it matters |
| --- | --- | --- |
| `groups()` | `ColumnNotFoundError: unable to find column "group_index"` | `_translate_errors` has no handler for it, so the rollout page returns **500** |
| `summary()["sample_index"]` | `ColumnNotFoundError: "sample_index" not found` | the trajectories endpoint indexes this |
| `step_aggregates()` | same `ColumnNotFoundError` | walks **every** step, so one empty step anywhere takes down the whole dump-derived metrics page, not just its own |

That third one is the reason this is worth fixing at the reader rather than patching `groups()`: a single empty step in the middle of a run poisons the metrics page for every other step.

## What changed

`summary()` supplies the declared schema when there are no rows. All three consumers then read an *empty* table instead of a broken one — `groups()` returns no groups, and `step_aggregates()` reports the step honestly as `n_samples: 0` with null aggregates while its neighbours keep their real values:

```
{'rollout_id': 0, 'n_samples': 8, 'reward_mean': 0.25,  ...}
{'rollout_id': 1, 'n_samples': 0, 'reward_mean': None,  ...}   <- the empty step
{'rollout_id': 2, 'n_samples': 8, 'reward_mean': 0.5,   ...}
```

`SUMMARY_COLUMNS` has to be declared because an empty row list carries no schema to infer. Since it only shapes the no-sample case, nothing else would notice it drifting away from `_summary_row`, so a test pins it against the columns a real summary produces — adding a column without updating it fails there. Dtypes are deliberately not pinned: they are data-dependent (`spec_accept_rate` infers as `Null` when every value is `None`, `Float64` otherwise), so the empty frame uses `Null`, which casts cleanly in the aggregations downstream.

## Verification

Three tests added to `tests/fast/dashboard/test_dump_reader_views.py`, all run locally.

The empty step is placed in the **middle** of a 3-step dump specifically so `step_aggregates()` has to walk past it and the neighbouring steps can be asserted intact. A separate test covers the parquet cache, because the all-null frame is written to disk and re-read only on the *second* call — a schema that failed to round-trip would otherwise pass on the first call and break in production.

Confirming the tests actually catch the bug, not just the new constant: reverting **only** the `summary()` behaviour while keeping `SUMMARY_COLUMNS` in place turns both behaviour tests red with the real `ColumnNotFoundError`, while the drift test stays green (correctly — it guards a different thing).

| run | result |
| --- | --- |
| `test_dump_reader_views.py`, with fix | 18 passed, 1 skipped |
| same file, `summary()` behaviour reverted | **2 failed**, 16 passed, 1 skipped |
| whole `tests/fast/dashboard/` | 245 passed, 4 skipped, 1 failed |
| `black==24.3.0`, `isort==5.13.2`, `ruff` | clean |

The one failure is `test_core_integration.py::test_engine_topology_gpu_range_logic`, which fails identically on a clean `origin/main` checkout with my changes stashed — pre-existing and unrelated to this diff.

The skipped test is `test_realdata_views`, which needs `MILES_DASHBOARD_REALDATA_DIR`.

## Test plan

Covered by the three new tests above. No API, schema, or frontend change — the empty step simply renders as an empty table rather than an error page.

### DIFF
--- a/miles/dashboard/advisory.py
+++ b/miles/dashboard/advisory.py
@@ -236,6 +236,8 @@ def _rollout_advisories(reader: DumpReader, args: dict) -> list[Advisory]:
             out.append(Advisory(level="warning", message=message))
 
     summary = reader.summary(rollout_id)
+    if summary.height == 0:
+        return out
     truncated_frac = summary["truncated"].cast(float).mean()
     if truncated_frac is not None and truncated_frac >= TRUNCATED_FRAC_WARN:
         message = f"{truncated_frac:.0%} of samples in rollout {rollout_id} are truncated"
--- a/miles/dashboard/dump_reader.py
+++ b/miles/dashboard/dump_reader.py
@@ -303,7 +303,7 @@ class DumpReader:
     FRESH_SECONDS: ClassVar[float] = 60.0
 
     # bump to invalidate summary parquet caches when their columns change
-    SUMMARY_VERSION: ClassVar[int] = 4  # v4: per-sample staleness
+    SUMMARY_VERSION: ClassVar[int] = 5  # v5: no-sample steps get the declared schema (v4 cached them columnless)
 
     # Column order of summary(). Only needed to give a step with no samples the
     # same shape as any other step; with rows present the schema comes from
--- a/miles/dashboard/dump_reader.py
+++ b/miles/dashboard/dump_reader.py
@@ -305,6 +305,43 @@ class DumpReader:
     # bump to invalidate summary parquet caches when their columns change
     SUMMARY_VERSION: ClassVar[int] = 4  # v4: per-sample staleness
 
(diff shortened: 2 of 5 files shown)


Shi-Dong:
@claude review always

Zhichenzzz (approved):

PR #2824, opened 2026-08-31
Temporarily revert PR 1313

fzyzcjy:
Part of #1837


miles/utils/arguments.py
```diff
@@ -837,12 +837,10 @@ def add_rollout_arguments(parser):
             )
             parser.add_argument(
                 "--update-weight-transfer-mode",
-                choices=["broadcast", "p2p", "disk-delta", "rdt"],
+                choices=["broadcast", "p2p", "disk-delta"],
```
fzyzcjy, line 867: [P1] Restore RDT before deliver-1

Existing launchers that pass `--update-weight-transfer-mode rdt` now fail during argument parsing because this choice and the supporting backend were removed. The designated deliver-1 commit `d3507a9438` still exposes only these three choices and contains no production RDT implementation, so this regression persists at the safe merge point and must be restored on the refactored interfaces.

fzyzcjy (commented):
The op removes an existing RDT weight-sync mode, but the designated safe deliver still does not restore it. Existing RDT configurations therefore cannot start at deliver-1.

PR #2825, opened 2026-08-31
fix: chunk true-on-policy logprob computation

zyzshishui:
## Summary
Honor logprob chunking in the true-on-policy loss path.

## Symptom & Reproduction
- **Symptom:** An 8192-token true-on-policy response can OOM while computing entropy because all response rows are processed together.
- **Reproduction:** Set a positive `log_probs_chunk_size`; the true-on-policy path still runs one full-response `log_softmax` instead of bounded row chunks.

## Root Cause
1. `calculate_log_probs_and_entropy` omitted `chunk_size` from the true-on-policy call.
2. `_calculate_log_probs_and_entropy_true_on_policy` materialized full-response softmax and entropy tensors.

## Fix
Pass the configured chunk size into the true-on-policy helper and process response rows in bounded chunks after TP vocabulary gathering. Sampling masks still affect only selected logprobs; entropy remains normalized over the full vocabulary.

## Verification
- `pytest -q --confcutdir=tests/fast/utils tests/fast/utils/test_true_on_policy_logprobs.py tests/fast/backends/training_utils/test_sampling_mask.py`: 19 passed.

## Review Focus
- `_calculate_log_probs_and_entropy_true_on_policy`: chunk response rows, never vocabulary columns.
- Sampling-mask path: preserve full-vocabulary entropy and its gradient behavior.

### DIFF
--- a/miles/backends/training_utils/loss_hub/math_utils.py
+++ b/miles/backends/training_utils/loss_hub/math_utils.py
@@ -978,6 +978,7 @@ def calculate_log_probs_and_entropy(
             tp_group,
             with_entropy=with_entropy,
             entropy_requires_grad=entropy_requires_grad,
+            chunk_size=chunk_size,
             vocab_size=vocab_size,
             sampling_mask=sampling_mask,
         )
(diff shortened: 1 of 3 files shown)


zyzshishui:
@claude review

adrenaline21 (approved):

guapisolo (approved):
checked and add ci

yueming-yuan (approved):

Issue #2207, opened 2026-08-05
[RFC] Load compute kernels from the Hugging Face Hub


adarshxs:
# [RFC] Load compute kernels from the Hugging Face Hub

## Overview

This RFC proposes adopting [huggingface/kernels](https://github.com/huggingface/kernels) as an optional kernel source for miles. Kernels are published as Hub repos with one prebuilt variant per `(torch, CUDA, C++ ABI, arch, OS)`; the client resolves a variant at load time and imports it from the HF cache. No compiler is needed on the target machine.

miles's `docker/Dockerfile` currently installs `flash_attn`, `flash_attn_3`, `apex`, `fast_hadamard_transform`, `causal_conv1d`, `mamba_ssm` and `transformer_engine` from prebuilt wheels produced out of band, plus FlashQLA from source. Image builds might be very long and any kernel change requires a rebuild.

Scope is the FSDP backend (`--train-backend fsdp`), opt-in by default. `kernels` is already present in every miles image: SGLang v0.5.16 declares `kernels>=0.14.1,<0.15` as a dependency, and ships a Hub FlashAttention-3 in [sgl-project/sglang#20796](https://github.com/sgl-project/sglang/pull/20796).

## Phase 0: attention, no code changes. Works already

`--attn-implementation` is a free-form `str` with no `choices`, passed directly to `from_pretrained` for both the policy and the reference model. transformers 5.12.1 resolves Hub repo IDs there via `is_kernel()` in `integrations/hub_kernels.py`.

```bash
python train.py --train-backend fsdp --attn-implementation kernels-community/flash-attn2@v2
```

- `kernels-community/flash-attn2` exposes `bwd` and `varlen_bwd`, so it covers the training backward pass.
- The packed-sequence path is unchanged: `cu_seqlens` is derived by transformers in `prepare_fa_kwargs_from_position_ids`, and `load_and_register_attn_kernel` binds transformers' own `flash_attention_forward` whenever the kernel exposes `flash_attn_varlen_func`.

[...]

PR #2829, opened 2026-08-31
feat(fsdp): load module-level compute kernels from the Hub

adarshxs:
Implements the module level kernels of #2207.

`--kernel-backend hub` resolves three kernels from `kernels-community` repos instead of the image's wheels. one prebuilt variant per (torch, CUDA, ABI, arch, OS), no compiler on the node.

| Slot | Repo | Feeds |
|---|---|---|
| `gated_delta_rule` | `kernels-community/fla` v1 | GatedDeltaNet recurrence (Qwen3-Next / 3.5 / 3.6) |
| `causal_conv1d` | `kernels-community/causal-conv1d` v1 | GatedDeltaNet short causal conv |
| `flash_attn_varlen` | `kernels-community/flash-attn2` v2 | NemotronH attention varlen path |


Zhichenzzz:
@claude review

behroozazarkhalili:
Ran this on top of current main (df0e677f6); 32 commits behind but merges with no conflicts.

- tests/fast/backends/test_fsdp_hub_kernels.py + test_fsdp_hub_kernels_distributed.py on the merge: 43 passed, with no `kernels` package installed, so the mocked paths are self-contained.
- Whole tests/fast/backends: 1509 passed, 13 skipped, 6 xfailed.
- With actor.py and both arguments.py restored from main the new tests cannot import, so they pin the plumbing.
- ruff/black/isort at the pre-commit pins are clean on all 10 Python files.

One thing worth a look before merge. `resolve_slot` (kernels/hub.py L159) calls `load_module_kernels(args)` again at model-build time, and `load_module_kernels` re-invokes the mapping provider on every call (L96-97). `prefetch_hub_module_kernels` gathers the mapping across ranks and checks it matches (`_collect_mapping`, L236-249), but the spec a model actually binds comes from that second call. With the built-in preset this is a fixed dict, so nothing changes; with `--kernel-mapping-path` a provider that reads anything non-deterministic can hand one rank a spec no rank agreed on, and `_resolve_module` would then download on a non-leader rank. Caching the agreed mapping from prefetch and having `resolve_slot` read that cache closes it in a few lines.

Everything else I checked holds: the default is `native`, `validate_kernel_backend_args` keeps hub kernels out of the bit-exact modes, non-strict failures log and keep the native kernel, strict re-raises with an actionable message, and the mapping errors are gathered and raised together across ranks.

cc @Zhichenzzz @yueming-yuan

Zhichenzzz:
> Ran this on top of current main ([df0e677](https://github.com/radixark/miles/commit/df0e677f6dd51fa551d48d37860812ece904cd8f)); 32 commits behind but merges with no conflicts.
> 
> * tests/fast/backends/test_fsdp_hub_kernels.py + test_fsdp_hub_kernels_distributed.py on the merge: 43 passed, with no `kernels` package installed, so the mocked paths are self-contained.
> * Whole tests/fast/backends: 1509 passed, 13 skipped, 6 xfailed.
> * With actor.py and both arguments.py restored from main the new tests cannot import, so they pin the plumbing.
> * ruff/black/isort at the pre-commit pins are clean on all 10 Python files.
> 
> One thing worth a look before merge. `resolve_slot` (kernels/hub.py L159) calls `load_module_kernels(args)` again at model-build time, and `load_module_kernels` re-invokes the mapping provider on every call (L96-97). `prefetch_hub_module_kernels` gathers the mapping across ranks and checks it matches (`_collect_mapping`, L236-249), but the spec a model actually binds comes from that second call. With the built-in preset this is a fixed dict, so nothing changes; with `--kernel-mapping-path` a provider that reads anything non-deterministic can hand one rank a spec no rank agreed on, and `_resolve_module` would then download on a non-leader rank. Caching the agreed mapping from prefetch and having `resolve_slot` read that cache closes it in a few lines.
> 
> Everything else I checked holds: the default is `native`, `validate_kernel_backend_args` keeps hub kernels out of the bit-exact modes, non-strict failures log and keep the native kernel, strict re-raises with an actionable message, and the mapping errors are gathered and raised together across ranks.
> 
> cc @Zhichenzzz @yueming-yuan

Thanks for the contribution! could you also profile the speedup compared with the default kernels in e2e RL?

adarshxs:
Thank you @Zhichenzzz @behroozazarkhalili i'll profile e2e RL runs with kernels and report back

adarshxs:
@Zhichenzzz profiled it on 2x B200 (sm_100), torch 2.13+cu130, transformers 5.12.1, kernels 0.14.1. Caveats up front: shared host with other tenants on the GPUs, so the noise is wider than ideal. 

**Speedup vs the default kernels in e2e RL: none, and that's expected.**

GRPO, 2 FSDP ranks colocated with SGLang, same seed/data, 2 runs per arm, first 3 steps dropped. `actor_train` throughput, hub vs native (mean / median):

| model | hub vs native | native vs native (noise floor) |
|---|---|---|
| Qwen3.5-2B, gsm8k, 30 rollouts | -0.2% / -1.0% | -1.7% / -5.1% |
| Nemotron-3-Nano-4B, dapo-math, 20 rollouts, ref model bound | +2.0% / -3.0% | ~10% |

The Hub builds are the same upstream kernels as the wheels, so there's nothing to gain. A profiled `actor_train` micro-batch on Qwen3.5-2B spends ~7.4% of GPU time in the GDN recurrence and ~2% in causal-conv, and `actor_train` + `log_probs` are ~3.8s of a ~17.6s step, so even an infinitely fast kernel caps out around 2% of step time.

Kernel microbench (T = 8k-32k, fwd and fwd+bwd): `fla` within ±6% of pip 0.5.2, Hub `flash-attn2` consistently 2.6-5.8% slower than the flash_attn 2.7.4.post1 wheel, Hub `causal-conv1d` faster at 8k but 12-22% slower fwd+bwd at 32k.

**Where it matters: a node without the wheels.**

Same Qwen3.5-2B run with `flash-linear-attention` and `causal-conv1d` hidden from the trainer, so transformers binds its torch fallbacks (the ones that ignore `cu_seqlens` / take no `seq_idx`):

| | `actor_train` tok/s | train/rollout logprob abs diff | train/rollout KL |
|---|---|---|---|
| wheels missing, native | 2,985 | 0.119 | 0.055 |
| wheels missing, `--kernel-backend hub --kernel-strict` | 22,277 | 0.0074 | 0.00028 |
| wheels present, native (reference) | 19,382 | 0.0073 | 0.00028 |

So 7.5x, but that's against the fallback, not against the wheels. One packed 16k row fwd+bwd on one GPU: 5041 ms / 172 GiB (torch fallback) vs 387 ms / 112 GiB (hub).

Conclusion: Hub == native numerically.

also, @behroozazarkhalili's mapping point is fixed in d6e1bcbf. Thanks!

adarshxs:
@alexnails @Zhichenzzz

PR #2834, opened 2026-08-31
fix(session): use incremental R3 for all non-retract cases, and fix asserts

guapisolo:
## Summary
Use incremental session R3 for every non-retract pause mode.

## Symptom & Reproduction
- **Report:** Shi Dong reported `partial_rollout`, rather than disaggregated `abort`, as the incompatible session-server case.
- **Symptom:** `SessionServer` returned full R3 for `abort`, and argument validation rejected disaggregated session-server `abort`, even though completed R3 rows remain append-only.
- **Reproduction:** Constructing `SessionServer` with `pause_generation_mode="abort"` selected full-R3 behavior instead of incremental offsets.

## Root Cause
1. `SessionServer.__init__` enabled additional R3 only for `in_place`.
2. `miles_validate_args` rejected `abort` instead of the incompatible `partial_rollout` behavior.

## Fix
Derive `use_addition_r3` directly from `pause_generation_mode != "retract"`, so `abort` works for both colocated and disaggregated serving. Reject `partial_rollout` for every session-server version. Keep `retract` on full R3 and warn that its payloads can become large and R3 from retract-mode weight updates has known issues in SGLang. The v1 and v2 offset tests cover `abort`, `in_place`, and `retract`, while test fixtures now provide the required pause-mode argument explicitly.

## Verification
- `git diff 2975c576320875bc02ecf86793e8b4384b16bcb1 420acf35a94947c4ea0df543d706eb731c90cc70 --check`: passed.
- `black==24.3.0 --check`: passed for all nine changed Python files.
- `ruff==0.14.7 check`: passed for all nine changed Python files.
- `python -m compileall`: passed for all nine changed Python files.
- Focused pytest did not collect because the local environment lacks `sglang`; `ld` also found no compatible Running devbox.

## Review Focus
- `SessionServer.use_addition_r3`: only `retract` requests full R3.
- `miles_validate_args`: session serving rejects `partial_rollout`, independent of `abort` topology.
- Retract warning: payload-size and SGLang R3 risks remain explicit.

### DIFF
--- a/docs/user-guide/agentic-rollout.md
+++ b/docs/user-guide/agentic-rollout.md
@@ -145,9 +145,7 @@ Whether a replayed message counts as "the same" as the stored one is decided by
 `--session-message-matcher` (default `strict`); see
 [Choose replay matching](#choose-replay-matching).
 
-The v1 wrapper returns one `Sample`. The v2 wrapper returns a `list[Sample]`, one
-for each selected tree leaf. Both versions reject `--pause-generation-mode=abort`
-and `--partial-rollout`, and use in-place weight update as instead to avoid harness pause.
+The v1 wrapper returns one `Sample`. The v2 wrapper returns a `list[Sample]`, one for each selected tree leaf. Both versions reject `--partial-rollout`. With R3 replay, colocated `abort` and `in_place` request only additional R3 rows, while `retract` returns full R3 data on every turn and emits a warning because the payloads can become very large. Non-colocated session servers cannot use `abort`.
 
 Set `--max-seq-len` to cap the context length. Miles also includes this value in the
 metadata passed to your agent so an external environment can stop early.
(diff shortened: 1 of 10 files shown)


Shi-Dong:
@claude review always

guapisolo:
> Beyond the reported findings, I checked the switch from `getattr(args, "pause_generation_mode", None)` to a direct `args.pause_generation_mode` access in `server.py` — `--pause-generation-mode` is a defined argparse flag with `default="retract"`, so the attribute is always present and this is not a regression.
> 
> Extended reasoning...

we ignore getattr

guapisolo:
/rerun-test tests/e2e/sglang/test_session_server_addition_r3.py

Shi-Dong (approved):
LGTM!

PR #2839, opened 2026-08-31
[multi-lora v2 0/7] Remove the multi-LoRA v1 implementation

yueming-yuan:
Co-authored with @yushengsu-thu.

Remove the dataset-driven multi-LoRA v1 driver, per-adapter data sources and scheduling, adapter controller, and legacy checkpoint flow.

Retain the slot primitives under `miles/backends/megatron_utils/lora/`: model instrumentation, shard addressing, per-slot optimizer state and gradient retention, slot-aware batch packing, and adapter transport. The Tinker command executor builds on these primitives in #2840.

The removal preserves the prefill batching assertions and updates the multi-LoRA documentation.

Merge current main while preserving the rollout session-metrics import and removing the retired AdapterRef import.

Validation: `pre-commit run --all-files` and the complete component `git diff --check` passed on this head. This update adds no test cases. Current validation and limitations are recorded on #2846.

PR #2840, opened 2026-08-31
[multi-lora v2 1/7] LoRA model extension and slot actor

yueming-yuan:

Expose command-level training on `MultiLoRATrainRayActor` in `megatron_utils/lora/actor.py`: `forward_backward` accumulates gradients across calls; `optim_step(adam_params_by_slot)` steps only the requested slots; `forward_only_logprobs` scores tokens; and `load_slot` / `unload_slot` manage slot weights, gradients, and optimizer state.

`lora/model.py` extends the shared model execution through `run_forward_backward`, which shares `run_forward_backward_pass` and `setup_train_iteration_config` with the existing training path. Adam parameters are applied per request. Successful steps and nonfinite-gradient skips return per-slot outcomes; unknown preparation, optimizer, and collective failures propagate to the existing TrainerCell failure boundary. Slot optimizer construction isolates the base optimizer factory while preserving LayerWise wrapping and FP32 state. The command dispatcher requires one live trainer cell. Existing optimizer tests cover nonfinite skips and propagation of unknown execution failures.

Slot optimizer construction passes the model chunks to LayerWiseDistributedOptimizer so equal-size adapter parameters use Megatron's rank-independent ownership tie-breaker.

Multi-LoRA pool initialization returns the model with no pool optimizer or scheduler. Only per-slot optimizers own training state; scheduler construction is skipped at this layer.

Stacked on #2839.

The multi-LoRA actor inherits the common initialization lifecycle and owns the slot optimizer registry. `MegatronTrainRayActor._init_training_state()` initializes the ordinary weight updater; the multi-LoRA override initializes slot state. The worker spec selects the subclass only for Megatron multi-LoRA actor workers. Full-model, single-LoRA, FSDP, and critic selection remain unchanged.

Validation:
- `pre-commit run --all-files` and the incremental `git diff --check` passed on this component head.
- Cascade verification checks that moved methods retain their behavior and that the final trees match the reviewed code. Formatting is committed separately.
- No new test files were added. The existing worker-class selection test now covers the multi-LoRA actor and preserves the critic's backend selection.
- Fresh CPU/GPU CI is pending. Combined validation and limitations are recorded on #2846 and #3198.


yushengsu-thu (approved):
checked

miles/backends/megatron_utils/lora/executor.py
```diff
@@ -0,0 +1,96 @@
...
+            group["weight_decay"] = adam_params["weight_decay"]
+
+
+def load_slot(model: Sequence[DDP], optimizer: MegatronOptimizer, slot: int, rank: int, alpha: float) -> None:
+    from megatron.bridge.peft.multi_lora_layers import init_adapter_slot
+
+    init_adapter_slot(model, slot, rank=rank, alpha=alpha)
+    zero_adapter_slot_grads(model, slot)
+    zero_optimizer_state_for_adapter(optimizer, model, slot)
+    optimizer.reload_model_params()
```
nanjiangwill: `optimizer` is the outer `LayerWiseDistributedOptimizer`, so `reload_model_params()` walks every child optimizer. Each BF16 child copies its model parameter back into its FP32 master. Loading or unloading slot A therefore rounds the FP32 masters for every live slot, changing their later Adam updates. Reload only `for child in _slot_children(optimizer, slot)` here and in `unload_slot`, and add a two-slot regression test where slot B’s FP32 master differs from its BF16 model value.

miles/backends/megatron_utils/lora/optimizer.py
```diff
@@ -160,37 +164,63 @@ def step_adapter_slots(
...
-        slot_norm = get_grad_norm_fp32(grads_for_norm, grad_stats_parallel_group=None)
-        if clip_grad > 0.0 and slot_params:
-            clip_grad_by_total_norm_fp32(slot_params, clip_grad, slot_norm, False)
-        grad_norms[slot] = float(slot_norm)
-
-        for child in children:
-            child.step_with_ready_grads()
-
+        try:
+            outcomes[slot] = _step_one_slot(optimizer, slot, batch_size, clip_grad)
```
nanjiangwill: `_step_one_slot()` can fail on one rank during `prepare_grads()` or batch-size scaling before `get_grad_norm_fp32()` enters the world all-reduce. That rank catches the exception here and moves to the Gloo outcome gather while healthy ranks block in the norm all-reduce, turning per-slot recovery into a collective timeout. Add rank consensus after local preparation and before the norm collective, or treat any pre-collective failure as cell-fatal. Cover this with a two-rank test that fails before `get_grad_norm_fp32()`.

PR #2841, opened 2026-08-31
[multi-lora v2 2/7] Per-batch loss selection and Tinker losses

yueming-yuan:

Select `loss_fn` and `loss_fn_config` per work unit. Add cross-entropy, importance-sampling, PPO, CISPO, and DRO losses; client loss inputs provide normalization, including custom-loss gradients represented as `weights = -dL/dlogprob`.

Forward/backward and forward-only calls return each datum's sample index, logprobs, and loss through the pipeline's microbatch results. There is no module-global output collector, so loss recomputation cannot append duplicate outputs. Add coverage for recomputation, detached outputs, and isolation between calls. Optimizer requests also apply `grad_clip_norm` per slot. `loss_normalizer` is separate from the execution batch size, so Tinker accumulation retains raw sums.

Remove the gateway gradient-reset RPC; slot initialization, optimizer steps, and unload retain their internal gradient cleanup.

The per-batch API extends `lora/model.py:run_forward_backward` and the commands in `lora/actor.py`; the subclass retains its slot optimizer registry.

Stacked on #2840.

Validation:
- Preserve native/custom `per_datum` metrics; a targeted before/after probe confirmed the metric stays in scalar logging. Add 14 nonzero objective/gradient cases covering IS, PPO, CISPO, DRO, clipping/beta overrides, masking, per-datum sums, and recomputation. CPU CI execution is pending.
- `pre-commit run --all-files` and the incremental `git diff --check` passed on this component head.
- Cascade verification checks that moved methods retain their behavior and that the final trees match the reviewed code. Formatting is committed separately.
- No new test files were added. The existing worker-class selection test now covers the multi-LoRA actor and preserves the critic's backend selection.
- Fresh CPU/GPU CI is pending. Combined validation and limitations are recorded on #2846 and #3198.


miles/backends/training_utils/loss_hub/tinker_losses.py
```diff
@@ -0,0 +1,173 @@
+"""Tinker protocol losses.
```
yushengsu-thu: Suggestion:
https://github.com/radixark/miles/pull/3145

miles/backends/megatron_utils/lora/executor.py
```diff
@@ -39,34 +41,45 @@ def forward_backward(
...
 
 def optim_step(
-    args: Namespace,
     model: Sequence[DDP],
     optimizer: MegatronOptimizer,
     adam_params_by_slot: dict[int, dict],
-) -> dict[int, float]:
+) -> dict[int, dict]:
+    outcomes: dict[int, dict] = {}
     for slot, adam_params in adam_params_by_slot.items():
```
nanjiangwill: With DP > 1, this loop turns one optimizer barrier into K full-pool synchronizations. Every `step_adapter_slots({slot: ...})` call performs `_merge_outcomes_across_ranks()` and then `LayerWiseDistributedOptimizer.allgather_params()`, whose parameter lists contain all adapter slots. Stepping K of N slots therefore moves O(KN) adapter parameters—O(N²) when K=N. Pass per-slot clip values into one `step_adapter_slots` call and perform the outcome merge and parameter all-gather once.

tests/fast/backends/training_utils/loss/test_tinker_loss_outputs.py
```diff
@@ -0,0 +1,72 @@
...
+    assert [[output["sample_index"] for output in outputs] for outputs in completed] == [[7], [11]]
+    for outputs in completed:
+        assert len(outputs) == 1, "loss recomputation must not append another datum output"
+        assert outputs[0]["loss"].item() == 1.75
+        assert outputs[0]["logprobs"].tolist() == [-0.5, -0.25]
+        assert not outputs[0]["loss"].requires_grad
+        assert not outputs[0]["logprobs"].requires_grad
+
+
+@pytest.mark.parametrize("loss_fn", sorted(tinker_losses.TINKER_LOSS_FUNCTIONS))
```
nanjiangwill, line 55: This parametrized test cannot validate the four new objectives because every result is multiplied by an all-zero mask; even an implementation that always returns zero passes. Only cross-entropy has a nonzero expected loss and gradient case. Add nonzero loss and gradient checks for importance sampling, PPO, CISPO, and DRO, including default and overridden clip/beta settings.

miles/backends/training_utils/loss_hub/tinker_losses.py
```diff
@@ -0,0 +1,157 @@
...
+    log_probs: list[torch.Tensor],
+    per_datum_losses: list[torch.Tensor],
+) -> tuple[torch.Tensor, dict]:
+    if per_datum_losses:
+        loss = torch.stack(per_datum_losses).sum()
+    else:
+        # a microbatch with no supervised tokens still needs the graph alive; fp32 sum avoids fp16 inf -> nan
+        loss = logits.sum(dtype=torch.float32) * 0
+    per_datum = [
+        {"sample_index": index, "logprobs": log_prob.detach().cpu(), "loss": sample_loss.detach().cpu()}
```
nanjiangwill, line 59: Non-blocking API note: keep this result extensible. `logprobs` is enough for the current Tinker losses, but future client-side losses may also request multi-target logprobs, values, or entropy. No additional output is needed in this PR; just avoid treating `{sample_index, logprobs, loss}` as a closed schema.

miles/backends/training_utils/loss.py
```diff
@@ -215,10 +192,12 @@ def loss_function(
         if apply_megatron_loss_scaling:
             loss = loss * parallel_state.cp.size
 
+    per_datum = log.pop("per_datum", None)
```
nanjiangwill, line 194: `log.pop("per_datum")` also runs for native and custom miles losses, where `batch["loss_fn"]` is `None`. `get_loss_function()` still supports `--custom-loss-function-path`, and that metric namespace does not reserve `per_datum`, so an existing metric with this name now disappears from logging. Extract protocol outputs only on the Tinker path—for example, when `batch.get("loss_fn") is not None`—or return protocol outputs separately from the metrics dict.

yushengsu-thu (approved):

PR #2842, opened 2026-08-31
[multi-lora v2 3/7] Per-slot checkpoints with optimizer state

yueming-yuan:

Save and restore one adapter slot independently. Checkpoints contain adapter weight shards, per-rank Adam state and step counters, and FP32 master parameters; loading optimizer state requires the same parallel topology.

Checkpoint load/save execution failures, including filesystem errors, propagate to `TrainerCell` and terminate the current single-cell Tinker service. `lora/actor.py` and controller commands expose `save_slot` and `load_slot(..., ckpt_path=..., load_optimizer=...)`. A TODO in `checkpoint_io.py` records deferred IO failure isolation.

Checkpoint names are relative symlinks to immutable version directories. The shared writer completes shards and metadata before atomically replacing the public link. A process crash leaves the old or new complete checkpoint reachable; older versions remain on disk for readers that already resolved them. Legacy directory checkpoints remain readable, but overwriting them is unsupported: save under a new name. Successful phases retain their collective barriers; execution errors still propagate to the trainer cell.

Stacked on #2841.

Validation:


yushengsu-thu (commented):
Suggestion:
https://github.com/radixark/miles/pull/3143

miles/backends/megatron_utils/lora/checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+        shards["weights"] = torch.load(checkpoint_dir / _weight_shard_name(), map_location="cpu", weights_only=True)
+        if load_optimizer:
+            optim_state = torch.load(checkpoint_dir / _optim_shard_name(), map_location="cpu", weights_only=True)
+            _check_optimizer_slot_state(optimizer, slot, optim_state)
+            shards["optim"] = optim_state
+
+    def apply_shards():
+        loaded = load_adapter(model, slot, shards["weights"])
+        assert loaded > 0, f"loaded 0 adapter tensors from {checkpoint_dir / _weight_shard_name()}"
+        optimizer.reload_model_params()
```
nanjiangwill: `optimizer` is the outer LayerWise/Chained optimizer, so this reload walks every slot’s child and copies BF16 model weights into FP32 masters. Loading slot S therefore rounds the masters of every live slot; only S is restored from `masters` below. The preceding `lora_executor.load_slot()` call does the same. Add a slot-local reload helper and use it in both paths, with a two-slot test that keeps slot B’s FP32 master different from its BF16 model value.

miles/backends/megatron_utils/lora/checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def load_slot(model: Sequence[DDP], optimizer: MegatronOptimizer, slot: int, path: str, load_optimizer: bool) -> None:
+    from megatron.bridge.peft.multi_lora_layers import load_adapter
+
+    checkpoint_dir = Path(path)
+    shards: dict = {}
+
+    def read_shards():
+        shards["weights"] = torch.load(checkpoint_dir / _weight_shard_name(), map_location="cpu", weights_only=True)
```
nanjiangwill: The read phase validates only optimizer list lengths. It does not validate master/moment tensor shapes or adapter weight keys and shapes. `load_adapter()` copies matching tensors first and reports missing or unused keys afterward, so a bad shard can mutate part of the live slot before raising. Validate the complete weight and optimizer tensor layout in `read_shards()` before any copy, or add a validation-only Bridge helper.

miles/backends/megatron_utils/lora/checkpoint.py
```diff
@@ -0,0 +1,193 @@
...
+    def publish_dir():
+        # write-then-rename so readers never see a partial checkpoint; on overwrite
+        # the old version survives (as _old_<name>) until the replacement is in place
+        if _rank() == 0:
+            if final_dir.exists():
+                old_dir = final_dir.parent / f"_old_{final_dir.name}"
+                if old_dir.exists():
+                    shutil.rmtree(old_dir)
+                os.replace(final_dir, old_dir)
+                os.replace(tmp_dir, final_dir)
```
nanjiangwill: This code moved; the current comment is here: https://github.com/radixark/miles/pull/2842#discussion_r3983558646

tests/fast/backends/megatron_utils/test_lora_slot_checkpoint_load.py
```diff
@@ -0,0 +1,44 @@
...
+def _write_weight_shard(tmp_path):
+    with patch.object(ckpt, "get_parallel_state") as parallel_state:
+        parallel_state.return_value = MagicMock(
+            tp=MagicMock(rank=0), pp=MagicMock(rank=0), ep=MagicMock(rank=0, size=1)
+        )
+        torch.save({"layer.adapter.weight": torch.zeros(2)}, tmp_path / ckpt._weight_shard_name())
+        return ckpt._weight_shard_name()
+
+
+def test_a_missing_optim_shard_leaves_the_slot_untouched(tmp_path):
```
nanjiangwill: These tests cover read-phase failures only. They never apply a valid checkpoint, so they do not verify FP32 masters, Adam moments and step counters, weights-only loading, or equivalence of the next Adam update. Add a CPU round-trip test for exact continuation. Include a second live slot whose FP32 master differs from its BF16 model value so the test also catches cross-slot reloads.

miles/backends/training_utils/checkpoint_io.py
```diff
@@ -0,0 +1,64 @@
...
+            tmp_dir.mkdir(parents=True)
+
+    def publish_dir():
+        if _rank() != 0:
+            return
+        if final_dir.exists():
+            old_dir = final_dir.parent / f"_old_{final_dir.name}"
+            if old_dir.exists():
+                shutil.rmtree(old_dir)
+            os.replace(final_dir, old_dir)
```
nanjiangwill, line 56: `write_checkpoint_dir()` now rolls back a caught failure from the second rename, but the published path is still absent between `final_dir -> old_dir` and `tmp_dir -> final_dir`. A process exit in that window leaves the last valid checkpoint only under `_old_<name>`, so readers lose the public path. Also, `shutil.rmtree(old_dir)` still runs inside the publish phase, so cleanup failure reports the save as failed after the replacement is committed. Use immutable version directories plus an atomically replaced pointer/exchange for crash atomicity, and make old-version cleanup best-effort.

yushengsu-thu (approved):

HJSang:
`checkpoint_io.run_local_io_collective`: only `OSError` is caught before the `all_gather_object`. In `make_tmp_dir` / `publish_dir` rank 0 is the only rank doing work, so if it raises anything else (e.g. `TypeError` from `json.dumps(metadata)`, `shutil.Error`), rank 0 exits the function while ranks 1..N block in the Gloo gather until timeout. That turns a local failure into a collective hang, the same pre-collective divergence pattern flagged on the optimizer path.

Suggest: catch `Exception` for the purpose of reaching consensus, gather the error, then classify after the gather on every rank: `OSError` → `CheckpointIOError`, anything else → re-raise (cell-fatal, but on all ranks at once). A test where rank 0 raises a non-OSError inside `step` and asserts every rank raises rather than hangs would pin it.

PR #2843, opened 2026-08-31
[multi-lora v2 4/7] Publish immutable adapter snapshots to disk

yueming-yuan:

Publish Tinker sampler adapters through `WeightPublisher` in `training_utils/weight_update/snapshot_publisher.py`. Every trainer rank joins the shared HF iterator's gathers; rank 0 writes the selected slot's rank-trimmed tensors, adapter configuration, and metadata into an immutable directory. `AdapterSpec(slot, rank, alpha)` identifies the export, and an existing version cannot be overwritten.

Tensor materialization, safetensors serialization, file writes, and directory publication all propagate execution errors through the common `TrainerCell` boundary. `export_slot` no longer converts checkpoint IO failures into result dictionaries. The checkpoint helper from #2842 retains successful-write barriers and complete-directory publication.

Tinker engines load the same frozen HF base at startup. Their resolved configuration sets `update_weights=False`, allowing the existing server-cell lifecycle to register them as ready without a trainer push; explicit weight updates and dummy base loading are rejected. The runtime in #2845 carries the snapshot name and disk path with sampling requests for SGLang's existing on-demand loading.

Full-model and single-LoRA training retain their ordinary weight updater. Remove the Tinker staged-push path, its engine-session error classes, and its update-window changes from the shared updater, actor, controller, and drivers. `MultiLoRATrainRayActor._init_training_state()` initializes the publisher independently of the transport; its `export_slot` command owns the export boundary. Remove staged-publication-specific tests and align the existing shared-updater tests with the restored path. The FT harness uses the current `end_update_weights(snapshot_cell_id_to_hashes)` API; an empty snapshot releases the window without marking cells ready.

Stacked on #2842.

Validation:

Multi-LoRA engines default `max_loaded_loras` to twice their effective `max_loras_per_batch` when unset. Explicit CLI/group overrides are preserved. Full pre-commit passed; no new tests were added.


yushengsu-thu (commented):
Suggest change to this:
[Async lora rollout and weight sync]
[[LoRA] Publish fresh streamed versions alongside generation by yushengsu-thu · Pull Request #38165 ·](https://github.com/sgl-project/sglang/pull/38165)
[[LoRA] Publish immutable adapter versions alongside generation by yushengsu-thu · Pull Request #3127](https://github.com/radixark/miles/pull/3127)

miles/backends/training_utils/weight_update/hf_weight_iterator/__init__.py
```diff
@@ -89,9 +93,11 @@ def iter_hf_weights(
         """
         hf_param_units = self._iter_hf_param_units(weights, materialize=materialize) if include_base else iter(())
         for lora_name, adapter in adapters:
-            hf_param_units = itertools.chain(
-                hf_param_units, self._iter_hf_adapter_units(lora_name, adapter, materialize=materialize)
+            prefixed = (
+                [(f"{lora_name}:{name}", tensor) for name, tensor in unit]
+                for unit in self._iter_hf_adapter_units(adapter, materialize=materialize)
             )
+            hf_param_units = itertools.chain(hf_param_units, prefixed)
```
joyliu-q: Hi, I think there is a bug in this code because generators are lazily evaluated.

I simplified your code a bit to give the following example:
```
import itertools

def build(adapters):
    units = iter(())
    for lora_name, tensors in adapters:
        prefixed = (f"{lora_name}:{t}" for t in tensors)
        units = itertools.chain(units, prefixed)
    return list(units)  # consumption happens here, after the loop

print(build([("A", ["w1", "w2"]), ("B", ["w1", "w2"])]))
# prints ['B:w1', 'B:w2', 'B:w1', 'B:w2']
# even though we want ['A:w1', 'A:w2', 'B:w1', 'B:w2']
```

If you add the following test to `TestIterHfWeightsTemplate`, it will also fail

```
class TestIterHfWeightsTemplate:
    ...

    def test_multiple_adapters_keep_their_own_prefixes(self):
        iterator = _StubIterator(SAMPLE_LORA_WEIGHTS)
        names = self._names(
            iterator.iter_hf_weights(None, include_base=False, adapters=[("slot_0", None), ("slot_1", None)])
        )
        assert names == [f"slot_0:{name}" for name, _ in SAMPLE_LORA_WEIGHTS] + [
            f"slot_1:{name}" for name, _ in SAMPLE_LORA_WEIGHTS
        ]

    ...
```

joyliu-q: example of a similar issue with better explanation: https://go.dev/blog/loopvar-preview

miles/backends/training_utils/weight_update/updater.py
```diff
@@ -94,53 +84,73 @@ def pop_metrics(self) -> dict[str, float]:
...
+            protocol,
+            self.args,
+            staged=staged,
+            sync_base=sync_base,
+            selector=self._hf_weight_iterator.weight_update_selector,
+            registrations=registrations,
+            lora_path=lora_path,
+        )
+        with session:
+            self._registered_adapters.update(lora_name for lora_name, _ in registrations)
```
nanjiangwill, line 124: `_registered_adapters` is updated before the staged stream and commit complete. If either fails, SGLang aborts and removes the pending registration, but a retry sees the cached name and skips registration; the second call sends only begin/end. Reusing a successfully published name also skips `defer_publish` freshness validation. Do not cache staged publication names—register each one and let SGLang enforce uniqueness—and add failed-retry and duplicate-name tests.

miles/backends/megatron_utils/actor.py
```diff
@@ -782,6 +790,53 @@ def _get_actor_weights(self):
...
+
+    @with_logs
+    def unload_adapter(self, info: "UpdatableEngines", lora_name: str) -> None:
+        assert self.args.multi_lora, "unload_adapter is a multi-LoRA slot command"
+        self._heartbeat.bump()
+        # rank 0's RPC failure must fail every rank together, not strand a barrier
+        failure = [None]
+        if dist.get_rank() == 0:
+            try:
+                async_utils.wait_futures(
```
nanjiangwill, line 853: `wait_futures()` only reports transport or task exceptions. SGLang can return HTTP 200 with `{"success": false, "error_message": ...}`, and this code discards those results before broadcasting success. Pass the results through `check_weight_sync_results(..., is_lora=True)`, as registration does, and then broadcast any failure.

miles/ray/train/group.py
```diff
@@ -417,9 +417,35 @@ async def load_slot(
...
         await self._execute_slots("unload_slot", slot=slot)
 
+    @asynccontextmanager
+    async def _updatable_engines(self):
+        """Release the inference controller's update lock even when the slot command fails."""
+        info = await self._inference_controller.start_update_weights()
+        try:
+            yield info
+        finally:
+            await self._inference_controller.end_update_weights(
```
nanjiangwill, line 433: Non-blocking lifecycle question: recovery from a failed warm push assumes that abort leaves the engine clean. `EngineWeightUpdateSession.__exit__()` logs and suppresses abort failures, while `_updatable_engines()` still marks the engine ready. If the engine retains an open update session or pending adapter name, disk backfill under that name may fail. Add an abort-failure test or define that this case requires replacing the engine.

yushengsu-thu (approved):

PR #2844, opened 2026-08-31
[multi-lora v2 5/7] Tinker gateway protocol face

yueming-yuan:

Add the gateway's protocol and scheduling layers under `miles/tinker/core` and `miles/tinker/server`.

- Per-model streams order and deduplicate requests. The planner packs compatible datums or merges ready optimizer barriers.
- A terminal forward/backward, optimizer-step, or load failure ends the affected model stream and fails its queued requests. Completed futures retain their results. Authenticated content-validation failures with a valid sequence settle at that position, so later training cannot skip a failed command.
- JSON and protobuf boundaries validate SDK datums and render future results. Authentication accepts `X-API-Key` and bearer tokens.

A sampler save exports the current adapter and returns its path after the immutable directory commits. It makes no inference-engine calls and neither applies nor discards accumulated gradients. Sampling resolves ownership and base-model compatibility from the saved metadata whether or not the training model still exists. There is no separate in-memory publication registry.

Known request-local forward-only and sampling failures leave the training stream available. Checkpoint load/save/export execution failures terminate the shared trainer cell and service; `load_state` rejects `sampler_weights` before metadata lookup or backend execution because the training loader requires Megatron DCP. `weights_info` remains available for both checkpoint kinds. Failed sampling does not invalidate a saved snapshot. Unknown trainer and background-task failures reach the serving supervisor. This implements failure handling and isolation; automatic recovery and stream replay are not implemented.

Model creation rechecks its record under the trainer lock. Model closure fails its creation future and only unloads initialized slots. Creation, training, and sampling duplicates whose results expired receive a stable terminal failure without replaying backend work. Unknown sampling sessions are user errors; foreign-tenant sessions remain ownership errors.

Futures, sessions, leases, deduplication, and unsaved accumulation are process-local. Saved checkpoints retain ownership and adapter metadata. These protocol layers do not import `miles.ray` or `miles.backends`; runtime wiring follows in #2845.

Request completion is centralized in `_finish_request`; batch and barrier execution share its failure policy. Checkpoint helpers live in `core/utils.py`, request validation stays in the service, and retry-result retention belongs to `FutureStore`. The dispatcher cancels and awaits its model-creation, sampling, and lease tasks before exiting.

Saving preserves the public checkpoint link; loading resolves it once before metadata validation and backend execution. Legacy checkpoints can be loaded, while legacy-directory overwrites are rejected before reaching the trainer.

Stacked on #2843.

Validation:

Session lease correction: models, sampling sessions, and in-flight sampling requests track their owning session. Expiry reclaims only that session’s resources; direct sessionless sampling is unaffected. The sweep decides expiry after acquiring the trainer lock and settles sampling futures before cancellation.

Validation: a real-service/fake-backend before/after probe reproduced stale model retention and confirmed selective slot reclamation. No regression tests are added for this change. Full pre-commit passed; CPU CI execution is pending.

Protobuf loss configuration: prefer v2 number/text values when present, otherwise decode the legacy numeric map. SDK 0.26.2 protobuf probes cover legacy, v2 precedence, v2-only values, zero, empty text, and malformed values. Full pre-commit passed; no regression tests were added.


yueming-yuan:
Planner rewritten to the final architecture (no interim FCFS-passthrough): per-model streams compile command runs into commutative windows; the planner packs ready datums across models into WorkUnits by (kind, loss_fn, config) equivalence class under a token budget — cross-tenant packing and large-request splitting are the same datum-granular bin-packer; ready optim barriers of different models merge into one BarrierUnit (`step_adapter_slots` takes the slot dict). Completion accounting reassembles per-datum outputs back to per-request promises; a request may span units and a unit may span requests.

In-memory fake-executor smoke covers packing, barrier merge, seq ordering across barriers, out-of-order arrival, idempotent resubmit, and tenant ownership.

yushengsu-thu (approved):

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+        return session_id
+
+    def heartbeat(self, tenant: str, session_id: str) -> None:
+        session = self.sessions.get(session_id)
+        if session is not None and session["tenant"] == tenant:
+            session["last_heartbeat"] = time.monotonic()
+
+    def create_model(self, tenant: str, payload: dict) -> tuple[str, str]:
+        """Two-phase like every command: allocate now, initialize the slot behind the future."""
+        base_model = payload["base_model"]
```
nanjiangwill, line 81: `CreateModelRequest` includes `session_id` and `model_seq_id`, but this method ignores both and always allocates a random model and slot. Tinker 0.26.2’s generated POST client retries up to 10 times by default with the same request body, so a lost response can run `load_slot()` twice and consume two slots. The model is also leased by tenant, so another live session for the same API key keeps it alive. Validate the session, key creation by `(tenant, session_id, model_seq_id)`, store the owning session, and return the existing future/model on retry.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+                )
+
+    async def _run_create_model(self, request_id: str, record: ModelRecord) -> None:
+        try:
+            async with self._backend_lock:
+                await self.backend.load_slot(record.slot, record.lora_rank, record.lora_alpha)
+        except Exception as error:
+            self.models.pop(record.model_id, None)
+            self.planner.remove_stream(record.model_id)
+            self.free_slots.add(record.slot)
```
nanjiangwill: `load_slot()` can fail after `init_adapter_slot`, gradient reset, or optimizer reset has already mutated some ranks, but this exception path returns the slot directly to `free_slots`. The test fails before any mutation, so it does not cover that state. Run `unload_slot()` before reuse and return the slot only if cleanup succeeds; otherwise quarantine it.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+        """None -> the HTTP layer answers 410 and the SDK resubmits."""
+        return self.futures.get(request_id, tenant)
+
+    # -------- sampling plane (future-based but never queues) --------
+
+    def create_sampling_session(self, tenant: str, payload: dict) -> str:
+        sampling_session_id = f"sampling-{uuid.uuid4().hex}"
+        self.sampling_sessions[sampling_session_id] = {
+            "tenant": tenant,
+            "model_path": payload.get("model_path"),
```
nanjiangwill: `CreateSamplingSessionRequest` carries `session_id`, `sampling_session_seq_id`, and a `base_model`/`model_path` binding, but this stores only `model_path`. Later SDK sample requests normally send only `sampling_session_id`, so a session created with `base_model="other"` is accepted and then samples this gateway’s configured base model. The stable `(session_id, sampling_session_seq_id)` is also ignored, so a retry creates another session. Validate the parent session and model selection, store the full binding, and deduplicate creation by that pair.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+            if session["tenant"] != tenant:
+                raise OwnershipError("sampling session does not belong to this tenant")
+            model_path = model_path or session["model_path"]
+        if payload.get("num_samples", 1) > self.config.max_samples_per_request:
+            raise UserInputError(
+                f"num_samples {payload['num_samples']} exceeds max_samples_per_request="
+                f"{self.config.max_samples_per_request}"
+            )
+        lora_name, lora_path = self._resolve_sampler(tenant, model_path) if model_path else (None, None)
+        future = self.futures.create(model_path or "base", tenant)
```
nanjiangwill: `decode_sample_request()` drops `seq_id`, and `submit_sample()` always creates a new future and task. Tinker 0.26.2 calls `execute_with_retries()` with the same request ID, so a lost response runs generation again and returns different IDs. Preserve `seq_id`, deduplicate by `(sampling_session_id, seq_id)` before scheduling, return the retained result on retry, and fail terminally after retention expires.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def submit(self, tenant: str, op: str, payload: dict) -> str:
+        """Admit a decoded command; content errors settle its future as a user failure."""
+        try:
+            op = CommandOp(op)
+        except ValueError:
+            raise UserInputError(f"unknown command op {op!r}") from None
+        model_id = payload["model_id"]
+        self.get_model(tenant, model_id)
+        seq_id = payload["seq_id"]
```
nanjiangwill: Non-blocking: `ModelStream` advances only from watermark 0 to sequence 1, but `submit()` accepts `seq_id <= 0` and non-integers. Those entries can never match `watermark + 1`, so their futures remain pending forever. Reject non-integers and values below 1 before creating the future; protobuf’s missing value is 0, so cover that boundary.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+    async def _retire_after_failed_step(self, stream, pending, error: str) -> None:
+        """The slot may hold a half-applied or rank-divergent step; it must not keep serving."""
+        message = f"model unloaded after a failed optimizer step ({error}); restore from a checkpoint"
+        self.futures.fail(pending.command.request_id, message, "server")
+        stream.finish(pending)
+        await self._evict_model(stream.model_id, message, "server")
+
+    async def _fail_if_poisoned(self, stream, pending) -> bool:
+        """Fail the next optimizer step when a batch discarded its accumulated gradients."""
+        poison = self._poisoned_slots.pop(stream.slot, None)
```
nanjiangwill: `_poisoned_slots.pop()` runs before `zero_grads()`. If `zero_grads()` raises, `_run_barrier()` fails the current optimizer request, but the poison has already been forgotten; the next `optim_step` can run on ambiguous gradients. Keep the poison until cleanup succeeds. If cleanup fails, retire or quarantine the model instead of continuing to serve it.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,686 @@
...
+            outcomes = await self.backend.optim_step(
+                {stream.slot: pending.command.payload["adam_params"] for stream, pending in entries}
+            )
+        except Exception as error:  # noqa: BLE001  can fail after some slots already stepped
+            logger.exception("the optimizer step failed at the backend level")
+            for stream, pending in entries:
+                await self._retire_after_failed_step(stream, pending, f"{type(error).__name__}: {error}")
+            return
+        for stream, pending in entries:
+            outcome = outcomes[stream.slot]
```
nanjiangwill: Non-blocking: validate the complete outcome map before settling any future. This loop can resolve and remove one stream, then hit a missing or malformed outcome for a later slot. `_run_barrier()` then calls `_fail_barrier()` for every entry, and the second `stream.finish()` on the already removed entry raises `ValueError`, terminating `service.run()`. A validation pass before settlement avoids the partial state.

miles/tinker/server/proto_codec.py
```diff
@@ -0,0 +1,121 @@
...
+            tokens.extend(np.frombuffer(chunk.encoded_text.tokens, dtype=np.int32).tolist())
+        inputs = {name: _decode_tensor(name, tensor) for name, tensor in datum.loss_fn_inputs.items()}
+        datums.append(build_datum(tokens, inputs, index))
+
+    decoded = {
+        "model_id": message.model_id,
+        "seq_id": message.seq_id,
+        "datums": datums,
+        "loss_fn": message.loss_fn,
+        "loss_fn_config": dict(message.loss_fn_config),
```
nanjiangwill, line 50: Non-blocking compatibility note: this decodes only the legacy `loss_fn_config`. Tinker 0.26.2 dual-writes numeric values to both maps but writes strings only to `loss_fn_config_v2`, so string settings are silently dropped. Prefer `loss_fn_config_v2` when present and fall back to the legacy map.

PR #2845, opened 2026-08-31
[multi-lora v2 6/7] Tinker gateway runtime and entry point

yueming-yuan:

Connect the protocol service to Miles through `MilesBackend` and `serve_tinker.py`.

Training commands place a work unit in the object store, let workers split it, and reassemble per-datum outputs by sample index. Forward-only runs the requested loss without backward. Explicit actor failure results pass through to the gateway; unknown trainer failures retain the common `TrainerCell` boundary and terminate the current single-cell service through its supervisor.

The entry point launches the worker manager, inference controller, trainer, service loop, and HTTP server. It requires the trainer to load the same frozen HF checkpoint as the inference engines. The runtime sends sampling requests through the router with the immutable adapter name and disk backfill path, without owning an inference controller or a weight-update window. Save and export commands forward gateway metadata to the worker's directory transaction. Checkpoint execution errors propagate through the same trainer-cell and supervisor shutdown path.

The gateway exposes its port, base-model name, and checkpoint root, and derives its adapter layout from normalized HF target names. Both standard and canonical Megatron target names are supported. Per-slot optimizer ownership is established in #2840.

Stacked on #2844.

Validation:
- Reject a missing checkpoint root before worker initialization. A before/after startup probe confirmed no worker launch is reached on invalid configuration.

Controller lifecycle: match `train.py` by disposing the inference controller, then the trainer, after normal completion of the serving tasks. No `AsyncExitStack` or initialization/exception cleanup guarantees are introduced. Full pre-commit passed; no regression tests were added.

Router dependency: install the router wheel and standalone binary from the existing `cu130-torch213-x86_64` / `cu130-torch213-aarch64` rolling releases in `yueming-yuan/miles-wheels`, using the normal `/tmp/wheels` installation path. Both architectures were rebuilt on CPU devboxes from `radixark/sgl-router-for-miles@b6342a3515e259bbd1eb4db728035a4efbe4fdf9` and verified to forward `return_indexer_topk` and `lora_backfill_paths`. No commit-specific router release is required. The existing `pr-2849` Docker image has not been rebuilt by this change.

Router-pin validation: full pre-commit passed; the Docker RUN shell syntax and image-input hash change were checked. Both architectures' wheel and binary downloads match the four previously tested SHA256 digests, and the source checkout resolves to the exact pinned commit. The existing native HTTP probes passed for both artifacts on both architectures. No new tests were added. The new Miles PR image and hosted CI have not finished yet.


yushengsu-thu (approved):

serve_tinker.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    service = TinkerService(MilesBackend(trainer, router_url, dp_size=dp_size), config)
+
+    server = uvicorn.Server(
+        uvicorn.Config(build_app(service), host="0.0.0.0", port=args.tinker_server_port, log_level="info")
+    )
+    logger.info(f"tinker gateway serving {config.base_model} on :{args.tinker_server_port}")
+    service_task = asyncio.create_task(service.run())
+    try:
+        await server.serve()
```
nanjiangwill: Addressed: `service.run()` and Uvicorn are now supervised together, and completion or failure of either task tears down the other.

miles/tinker/runtime.py
```diff
@@ -0,0 +1,197 @@
...
+    async def push_slot(
+        self, slot: int, lora_name: str, rank: int, alpha: float, lora_path: str | None = None
+    ) -> None:
+        await self.trainer.push_slot(slot=slot, lora_name=lora_name, rank=rank, alpha=alpha, lora_path=lora_path)
+
+    async def _run_batch(self, method: str, batch_id: int, train_data: dict) -> list:
+        store = object_store.get_instance()
+        data_ref = store.put(value=train_data, value_spec=ROLLOUT_DATA_VALUE_SPEC)
+        try:
+            return await getattr(self.trainer, method)(batch_id=batch_id, data_ref=data_ref)
```
nanjiangwill: `TrainerController._execute_slots()` calls `TrainerCell.execute()` with the default `kill_on_failure=True`; any actor exception marks the only trainer cell errored and kills its workers. `TinkerService` instead treats batch failures as recoverable and calls `zero_grads()` before continuing, so cleanup targets a dead cell and can stop dispatch. Create and load failures likewise leave the service alive with no usable trainer. Align the contracts: either stop the gateway when a trainer RPC kills the cell, or classify recoverable request errors before that boundary. Keep CUDA/NCCL failures cell-fatal.

miles/tinker/runtime.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def _to_sequence(response: dict) -> dict:
+    output_token_logprobs = response["meta_info"]["output_token_logprobs"]
+    finish = response["meta_info"]["finish_reason"]["type"]
+    return {
+        "sequence_id": f"seq-{uuid.uuid4().hex}",
+        "tokens": [entry[1] for entry in output_token_logprobs],
+        "logprobs": [entry[0] for entry in output_token_logprobs],
+        "stop_reason": "length" if finish == "length" else "stop",
```
nanjiangwill, line 207: `_to_sequence()` maps every finish reason except `length` to `stop`. SGLang returns `type="abort"`, and miles’ normal rollout path maps that to `Sample.Status.ABORTED`; here it becomes a successful partial sample that the client can train on. Raise on `abort`—the Tinker wire enum has only `stop` and `length`—so `_run_sample()` settles the request as a server failure.

nanjiangwill (commented):
Runtime integration review: three correctness/availability concerns and two explicitly non-blocking lifecycle suggestions.

serve_tinker.py
```diff
@@ -0,0 +1,86 @@
...
+        args=args,
+        role="actor",
+        with_ref=False,
+        with_opd_teacher=False,
+        inference_controller=inference_controller,
+        rollout_executor=None,
+    )
+    await trainer.init()
+
+    checkpoint_root = args.tinker_checkpoint_root or (args.save and f"{args.save}/tinker")
```
nanjiangwill, line 46: Non-blocking: compute and validate `checkpoint_root` before `launch_worker_manager()` and controller initialization. A missing `--tinker-checkpoint-root`/`--save` currently starts the worker and GPU stack before hitting this assertion.

serve_tinker.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    logger.info(f"tinker gateway serving {config.base_model} on :{args.tinker_server_port}")
+    service_task = asyncio.create_task(service.run())
+    try:
+        await server.serve()
+    finally:
+        # service.run() loops forever; cancel it so the gateway can exit
+        service_task.cancel()
+        with suppress(asyncio.CancelledError):
+            await service_task
```
nanjiangwill: Non-blocking: dispose `trainer` and `inference_controller` in this `finally` block. Their disposal stops watchers, tickers, and inference server clients; cancelling only `service_task` leaves those resources live until process exit.

PR #2846, opened 2026-08-31
[DO NOT MERGE] stacked multi-lora full PR

yueming-yuan:
Combined view of the multi-LoRA v2 stack against `main`. **Do not merge this PR; land the component PRs in order.**

| Part | PR | Content |
|---|---|---|
| 0/7 | #2839 | Retire multi-LoRA v1 and retain slot primitives |
| 1/7 | #2840 | LoRA model extension, dedicated actor, and selective optimizer stepping |
| 2/7 | #2841 | Per-batch Tinker losses and per-datum outputs |
| 3/7 | #2842 | Per-slot checkpoints and atomic directory publication |
| 4/7 | #2843 | Immutable disk snapshots through WeightPublisher |
| 5/7 | #2844 | Gateway protocol, scheduling, and tenant ownership |
| 6/7 | #2845 | Miles runtime and frozen-base serving entry point |
| 7/7 | #2849 | Example, gateway tests, and launcher snapshots |

Tinker training runs through ordered per-model streams and per-slot optimizers. `lora/actor.py` extends `MegatronTrainRayActor` with the slot commands; `lora/model.py:run_forward_backward` extends the shared model execution. The worker spec selects the subclass only for Megatron multi-LoRA actor workers.

Saving sampler weights exports the current adapter into an immutable directory containing tensors, adapter config, and ownership/base-model metadata. Saving neither applies nor discards pending gradients. Engines load the same frozen HF base at startup and load adapter snapshots from disk on demand; sampler saves make no inference-engine calls.

`training_utils/weight_update/snapshot_publisher.py` reuses the shared HF iterator and checkpoint directory transaction. The multi-LoRA actor initializes the publisher; the general actor initializes the ordinary updater for full-model and single-LoRA training. Checkpoint load/save/export execution errors, including filesystem errors, propagate through TrainerCell and stop the current single-cell service. `load_state` rejects sampler snapshots before reaching the Megatron checkpoint loader; `weights_info` retains metadata support for both checkpoint kinds. Checkpoint names are relative symlinks to immutable version directories. The shared writer completes shards and metadata before atomically replacing the public link. A process crash leaves the old or new complete checkpoint reachable; older versions remain on disk for readers that already resolved them. Legacy directory checkpoints remain readable, but overwriting them is unsupported: save under a new name. Successful phases retain their collective barriers; execution errors still propagate to the trainer cell. A TODO records deferred IO failure isolation.

A terminal training failure ends the affected model stream, including queued commands, while completed futures retain their results. Known request-local forward-only and sampling failures preserve healthy training resources; a failed sample leaves its snapshot usable. Futures, leases, deduplication, and unsaved accumulation remain process-local. Automatic training recovery and replay are not implemented.

The combined branch and #2849 have the same head: `5ccde8dcde766c2290b721545ce6f0758e51dae2`. Checkpoint publication belongs to #2842, the FT harness correction to #2843, gateway validation and checkpoint path handling to #2844, and documentation/tests to #2849. The changes cascade through #2845 and downstream #3198.

Request completion is centralized in `_finish_request`; checkpoint utilities are stateless, validation remains in the service, and the dispatcher owns background-task shutdown. The final #3198 tree exactly matches reviewed commit `de7a5565e82cb6c89e26174d5a2b26980648e086`, including the existing single-LoRA checkpoint changes. All updated branches preserve their previous heads as ancestors.

Validation:
- This isolates checkpoint/Adam continuation with identical post-reduction gradients; it does not validate forward/backward numerics, topology changes, MoE/TP/PP layouts, server restart, or failure recovery. Fresh hosted CPU/NVIDIA CI is pending; broad GPU selection remains on #3198. The Tinker cookbook and multi-tenant acceptance tests are now enabled under `run-ci-multi-lora`; their first GPU validation is pending.
- Launchers and command snapshots are unchanged by these fixes. The prior local launcher gate has not passed (362 passed, 30 failed, 122 errors), including macOS/Linux assumptions and existing launcher mismatches.

Router dependency: install the router wheel and standalone binary from the existing `cu130-torch213-x86_64` / `cu130-torch213-aarch64` rolling releases in `yueming-yuan/miles-wheels`, using the normal `/tmp/wheels` installation path. Both architectures were rebuilt on CPU devboxes from `radixark/sgl-router-for-miles@b6342a3515e259bbd1eb4db728035a4efbe4fdf9` and verified to forward `return_indexer_topk` and `lora_backfill_paths`. No commit-specific router release is required. The existing `pr-2849` Docker image has not been rebuilt by this change.

Tinker GPU CI: add `tests/e2e/lora/test_tinker_multi_tenant_gateway.py` and enable the cookbook gateway test. Both carry the `multi-lora` domain, selected by the repository label `run-ci-multi-lora`, on 8 H200 GPUs. They share Qwen3-4B gateway setup (4 training GPUs, TP2/DP2; 4 sampling GPUs; rank-8 adapters). The multi-tenant client trains four distinct markers concurrently, exports each adapter, and checks loss decrease and its own sampled marker.

Validation: the real CI selector selects exactly both enabled tests for `run-ci-multi-lora`; mocked entrypoint checks confirm both client commands and launcher cleanup. Full pre-commit passed. GPU convergence, cookbook execution, and end-to-end tenant isolation have not yet been validated.


yushengsu-thu:
- Engine-side adapter versions are never unloaded: TrainerController.unload_adapter (miles/ray/train/group.py:416) has no caller; lease expiry only clears the trainer-side slot. Every save_weights_for_sampler pushes another adapter name into each engine's host RAM; max_loaded_loras is unset, so growth is unbounded. 

-  Checkpoint ownership lives only in memory: load_state requires the source model to exist in self.models (miles/tinker/core/service.py:369), so after a restart or lease expiry every load raises OwnershipError.

yushengsu-thu (commented):
#Optimization direction 1 - (Async Lora Rollout) Per-LoRA Weight Updates Block Concurrent Rollouts:
- Problem: when you push slot for LoRA A pauses every rollout and clears global KV cache, so updating one LoRA still stops unrelated LoRAs (B/C), causing recompute, jitter, and possible stuck pause on failure; in-place replace can also mix weights with old KV.2)
- Solution: versioned names; stage load A@new; publish only after all engines succeed; no global pause, no global KV flush; drain and unload old versions after in-flight requests finish; rollback on failure.

yushengsu-thu:
Follow-up to my review https://github.com/radixark/miles/pull/2846#pullrequestreview-5107661735 (Optimization direction 1, async LoRA rollout): https://github.com/radixark/miles/pull/3104 implements the minimal version against this branch — push_slot no longer pauses/flushes the fleet, and sampler_version is published only after every engine acked the push. Verified on Qwen3-0.6B (2 trainer + 2 engine GPUs) against base 34ffdb864: pause/flush/resume rounds 3×2 → 0, in-flight samples on another adapter unaffected during a push. Hunk homes in the stack: mixin.push_slot → #2843, TinkerService → #2844. Details and numbers in the PR body.

yushengsu-thu (commented):
Suggestion:
[Async lora rollout and weight sync]
[[LoRA] Publish fresh streamed versions alongside generation by yushengsu-thu · Pull Request #38165 ·](https://github.com/sgl-project/sglang/pull/38165)
[[LoRA] Publish immutable adapter versions alongside generation by yushengsu-thu · Pull Request #3127](https://github.com/radixark/miles/pull/3127)

miles/tinker/server/app.py
```diff
@@ -0,0 +1,166 @@
...
+                op, payload = decode_command(route_op, await request.json())
+                request_id = service.submit(_tenant(request), op, payload)
+                return {"request_id": request_id, "model_id": payload["model_id"]}
+
+            return command
+
+        app.post(route)(_make(route_op))
+
+    @app.post("/api/v1/retrieve_future")
+    async def retrieve_future(request: Request):
```
mathewjhan, line 141: tinker uses long polling instead of immediately returning with their clients, that way multiple tinker clients won't send too many requests at once

some examples:

- https://github.com/NovaSky-AI/SkyRL/blob/4650b643dabf09da37a2ba76d6cf36805cf2857d/skyrl/tinker/api.py#L1492
- https://github.com/verl-project/verl-recipe/blob/ee3aef1690d6bb5e6448052c4842e8efb7a3f76c/verl_tinker/src/verl_tinker/tinker_router.py#L720

miles/tinker/runtime.py
```diff
@@ -0,0 +1,185 @@
...
+def _build_train_data(slot_datums: list) -> dict:
+    datums = [datum for _, datum in slot_datums]
+    train_data = {
+        "tokens": [datum["tokens"] for datum in datums],
+        "loss_masks": [[1] * datum["target_len"] for datum in datums],
+        "response_lengths": [datum["target_len"] for datum in datums],
+        "total_lengths": [len(datum["tokens"]) for datum in datums],
+        "sample_indices": list(range(len(datums))),
+        "adapter_slots": [slot for slot, _ in slot_datums],
+        "dynamic_global_batch_size": len(datums),
```
mathewjhan: I might be wrong here, but I think this won't work correctly in cases where datums not divisible by dp size? I think before @yushengsu-thu original code had some way to pad to dp size, but I am not sure if that code is still in this refactored version

i asked an AI for a possible way to repro:

1. Launch the documented gateway unchanged:
      - 4 training GPUs
      - TP=2
      - PP=1
      - CP=1
      - therefore DP=2

2. Change the smoke client to submit only one datum, for example:

     datums = build_datums(tokenizer, marker)[:1]

3. Call:

     await training.forward_backward_async(datums, loss_fn="cross_entropy")

4. Await the returned future.
5. The trainer should report a server failure, with logs showing division by zero around miles/backends/training_utils/data.py:469.

yueming-yuan: This is fixed in commit 4cf86ee27e. thanks!!

yushengsu-thu: already resolved in the PR.

miles/backends/megatron_utils/lora/optimizer.py
```diff
@@ -0,0 +1,216 @@
...
+        config.timers = None
+        base_optimizers = _build_slot_base_optimizers(
+            config, model, slot_params, use_gloo_process_groups=args.use_gloo_process_groups
+        )
+        assert base_optimizers, f"adapter slot {slot} produced no optimizer children"
+        self._inner = LayerWiseDistributedOptimizer(
+            base_optimizers,
+            config,
+            ProcessGroupCollection.use_mpu_process_groups(),
+            init_state_fn_list=[_adam_init_state_fn] * len(base_optimizers),
```
nanjiangwill, line 127: found by codex when checking megatron side multi-lora

Please pass `model_chunks=list(model)` to `LayerWiseDistributedOptimizer` here. On the decoupled path used by multi-LoRA, the pinned Megatron implementation uses `model_chunks` to build a rank-independent tie-breaker for ping-pong ownership. Without it, equal-numel parameters fall back to rank-local insertion order; Megatron notes that this can make a parameter double-owned or unowned across DP ranks. LoRA has many equal-shaped parameters, so this path is likely to encounter those ties. Please add a DP>1 test that verifies exactly one owner per adapter parameter and completes one optimizer step/all-gather.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,691 @@
...
+
+    def _observe_background_task(self, task: asyncio.Task) -> None:
+        self._create_tasks.discard(task)
+        if not task.cancelled() and (error := task.exception()) is not None:
+            if self._background_error is None:
+                self._background_error = error
+            self._wake.set()
+
+    async def _run_create_model(self, request_id: str, record: ModelRecord) -> None:
+        async with self._backend_lock:
```
nanjiangwill: There is a race between `create_model` and lease eviction here. The model record is published before this task acquires `_backend_lock`, so the sweeper can remove the record and free the slot first. I reproduced the ordering `unload_slot -> load_slot`: the create future succeeded, but the model was absent and its slot was marked free. The real backend may also try to unload an optimizer that has not been created yet. Please track the record as allocating/live/retired and re-check it after acquiring the lock before loading or resolving the future. Eviction should cancel an in-progress creation instead of unloading an uninitialized slot.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,691 @@
...
+    async def _save_state(self, record: ModelRecord, pending, payload: dict) -> dict:
+        """Save parameters and optimizer state; call after optim_step to persist accumulated training work."""
+        name = payload["name"] or f"checkpoint-{pending.command.seq_id:06d}"
+        _validate_checkpoint_segment(name)
+        checkpoint_dir = self._checkpoint_dir(record.model_id, "weights", name)
+        if not payload["overwrite"] and os.path.exists(checkpoint_dir):
+            raise UserInputError(f"checkpoint {name!r} already exists; pass overwrite=True to replace it")
+        if (failure := await self.backend.save_slot(record.slot, checkpoint_dir)) is not None:
+            return failure
+        if (failure := self._stamp_checkpoint_meta(checkpoint_dir, record)) is not None:
```
nanjiangwill: `META.json` needs to be part of the same directory transaction as the tensor shards. `write_checkpoint_dir` publishes the tensor directory first, then `_stamp_checkpoint_meta` writes into the final directory here. I injected a metadata-write failure: the save reported failure but left a final checkpoint without `META.json`; retrying with `overwrite=False` then failed because that directory already existed. Please write the metadata into the temporary directory before the final rename and add failure-injection tests around metadata writing, both renames, and old-directory cleanup.

serve_tinker.py
```diff
@@ -0,0 +1,91 @@
...
+    await inference_controller.init()
+
+    trainer = TrainerController(
+        args=args,
+        role="actor",
+        with_ref=False,
+        with_opd_teacher=False,
+        rollout_executor=None,
+    )
+    await trainer.init()
```
Zhichenzzz, line 49: codex review (zhichen revised)

[P1] No sampling workers are available after startup

The HTTP server starts immediately after initialization without marking the engines' weights ready. With the default `update_weights=True`, engines remain in `StatePendingWeights` and are not registered with the router. In the tested commit `c295325`, only the first successful adapter publication calls `mark_weights_ready`, so sampling the base model directly or sampling an existing checkpoint after a restart fails. Complete engine readiness and router registration before accepting requests.

Head `0a3fee6` switches publication to disk snapshots, but `serve_tinker.py` still does not complete this readiness flow. Removing `push_slot` also removes the previous publication-time call that marked weights ready.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,701 @@
...
+            request_id = self._retained_request_id(request_id, model_id, tenant)
+            session["models_by_seq"][model_seq_id] = (request_id, model_id)
+            return request_id, model_id
+        base_model = payload["base_model"]
+        if base_model != self.config.base_model:
+            raise UserInputError(f"this gateway serves {self.config.base_model!r}, not {base_model!r}")
+        lora_config = payload.get("lora_config") or {}
+        self._reject_unsupported_lora_config(lora_config)
+        rank = lora_config.get("rank", 32)
+        if rank > self.config.max_lora_rank:
```
Zhichenzzz: codex review (zhichen revised)

[P1] `rank=0` can shut down the entire gateway

This validates only the upper bound; the pinned `tinker==0.26.2` SDK schema does not enforce a lower bound either. I reproduced zero reaching `load_slot` and triggering the underlying `0 < rank <= max_rank` assertion. The exception fails the trainer cell and terminates the dispatcher, after which the task supervision in `serve_tinker.py` shuts down the HTTP server, affecting other tenants. Reject any rank that is not a positive integer as a user input error before allocating a slot or submitting trainer work. Head `0a3fee6` still checks only the upper bound here.

miles/tinker/runtime.py
```diff
@@ -0,0 +1,234 @@
...
+        self, slot: int, rank: int, alpha: float, path: str, metadata: dict | None = None
+    ) -> dict | None:
+        return _slot_failure(
+            await self.trainer.export_slot(slot=slot, rank=rank, alpha=alpha, path=path, metadata=metadata)
+        )
+
+    async def push_slot(
+        self, slot: int, lora_name: str, rank: int, alpha: float, lora_path: str | None = None
+    ) -> dict | None:
+        async with update_weight_window(self.inference_controller) as info:
```
Zhichenzzz: codex review (zhichen revised)

[P2] Health checks remain paused after the first adapter publication (historical finding on `c295325`)

Entering `update_weight_window` calls `start_update_weights`, which pauses health monitoring; on exit, `end_update_weights` only releases the lock. The Tinker path does not call `prepare_rollout` / `prepare_eval`, which restore monitoring in the regular training flow. I reproduced a successful publication leaving the health checker's `active=False`, so subsequent engine hangs would no longer be detected by this checker. For this publication path, restore health monitoring when the weight-update window ends.

Version note: by the time this comment was posted, head `0a3fee6` had switched to disk-snapshot publication and removed these `push_slot` / `update_weight_window` calls. This comment records the finding on the tested version; it is not an outstanding issue on that head.

Zhichenzzz (commented):
codex review (zhichen revised)

These three findings are based on the tested commit `c29532571c45aa22fa7af060a1587329206d1694`; the inline locations refer to that version. Before posting, I rechecked the updated head, `0a3fee6c64436a1b063b169f652ac53353f928a2`: the gaps described in findings 1 and 2 remain, while the `push_slot` weight-update path involved in finding 3 has been removed. Finding 3 is retained as a historical observation, not an outstanding issue on that head.

HJSang:
Admission never range-checks token ids. `build_datum` / `_validate_batch_payload` check lengths and loss inputs, but an id >= vocab_size goes straight into the Megatron embedding lookup, raises a device-side assert, and that's an unclassified actor exception: the only trainer cell is killed, `trainer_dead()` trips, and the dispatch loop + HTTP server exit. One client's tokenizer mismatch takes every tenant down.

Suggestion: carry `vocab_size` into `GatewayConfig` at serve time (the HF config has it) and reject out-of-range ids in `build_datum` and `decode_sample_request` as `UserInputError`, so it settles as an ordered user failure on that stream only. A test submitting a datum with id == vocab_size and asserting a co-tenant's stream is untouched would pin it.

miles/tinker/server/encoding.py
```diff
@@ -0,0 +1,203 @@
...
+    """One decoded datum (token list + loss_fn_inputs lists) -> internal datum."""
+    unknown = set(inputs) - set(LOSS_INPUT_KEYS) - {"target_tokens"}
+    if unknown:
+        raise UserInputError(f"datum {index}: unknown loss_fn_inputs {sorted(unknown)}")
+    for name, values in inputs.items():
+        if any(isinstance(value, (list, tuple)) for value in values):
+            raise UserInputError(
+                f"datum {index}: loss_fn_inputs[{name!r}] must be 1-D; multi-target inputs are not supported"
+            )
+    targets = [int(t) for t in inputs["target_tokens"]]
```
HJSang: Admission never range-checks token ids. `build_datum` / `_validate_batch_payload` check lengths and loss inputs, but an id >= vocab_size goes straight into the Megatron embedding lookup, raises a device-side assert, and that's an unclassified actor exception: the only trainer cell is killed, `trainer_dead()` trips, and the dispatch loop + HTTP server exit. One client's tokenizer mismatch takes every tenant down.

Suggestion: carry `vocab_size` into `GatewayConfig` at serve time (the HF config has it) and reject out-of-range ids here and in `decode_sample_request` as `UserInputError`, so it settles as an ordered user failure on that stream only. A test submitting a datum with id == vocab_size and asserting a co-tenant's stream is untouched would pin it.

miles/backends/training_utils/checkpoint_io.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def publish_dir():
+        if _rank() != 0:
+            return
+        if metadata is not None:
+            (tmp_dir / "META.json").write_text(json.dumps(metadata, indent=2))
+        version_dir = final_dir.parent / f"_version_{final_dir.name}_{uuid4().hex}"
+        os.replace(tmp_dir, version_dir)
+        tmp_dir.symlink_to(version_dir.name, target_is_directory=True)
+        os.replace(tmp_dir, final_dir)
```
HJSang, line 54: Nice: the versioned dir + relative symlink swap closes the two-rename window from the earlier thread.


- A crash between `os.replace(tmp_dir, version_dir)` and this final `os.replace` leaves an orphan `_version_*` dir that no link targets and nothing reclaims. The same sweep would cover it: any `_version_<name>_*` not targeted by `<name>`.

miles/tinker/runtime.py
```diff
@@ -0,0 +1,217 @@
...
+        candidates = position or []
+        candidate_token_ids = [entry[1] for entry in candidates][:k]
+        candidate_logprobs = [float(entry[0]) for entry in candidates][:k]
+        token_ids.append(candidate_token_ids + [0] * (k - len(candidate_token_ids)))
+        logprobs.append(candidate_logprobs + [float("nan")] * (k - len(candidate_logprobs)))
+    return {"token_ids": token_ids, "logprobs": logprobs}
+
+
+def _to_sequence(response: dict) -> dict:
+    output_token_logprobs = response["meta_info"]["output_token_logprobs"]
```
HJSang, line 208: `_to_sequence` indexes `response["meta_info"]["output_token_logprobs"]` unconditionally, but the router returns an error body (no `meta_info`) for e.g. a prompt longer than the engine context or an unsupported sampling param. `sample()` only catches `httpx.HTTPError`, so this is a `KeyError` out of `_run_sample`, which `_observe_background_task` turns into a dispatcher exit: one tenant's oversized prompt stops training for every tenant.

Suggest: check the response shape here (`"meta_info" not in response` → return `{"error": response.get("error") or "engine returned no generation"}`), and in `TinkerService._run_sample` catch `Exception`, fail that future as `server`, and log, reserving the dispatcher exit for `trainer_dead()`.

miles/tinker/core/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+def read_checkpoint_metadata(checkpoint_dir: str, tenant: str, shown_path: str) -> dict:
+    meta_file = Path(checkpoint_dir) / "META.json"
+    if not meta_file.exists():
+        raise UserInputError(f"unknown checkpoint {shown_path!r}")
+    try:
+        meta = json.loads(meta_file.read_text())
+    except (OSError, json.JSONDecodeError) as error:
+        raise UserInputError(f"cannot read checkpoint {shown_path!r}: {error}") from error
+    if meta["tenant_digest"] != _tenant_digest(tenant):
```
HJSang, line 52: `meta["tenant_digest"]` here and `meta[key]` in `validate_checkpoint_compatibility` raise `KeyError` on a `META.json` missing a field (older format, hand-edited, or truncated write since nothing fsyncs). On the `load_state` path this happens inside `_dispatch_barrier_op`, and `_run_barrier` catches only `UserInputError`/`OwnershipError`, so it escapes `run()` and the gateway exits for all tenants. Trigger: a client passing an old checkpoint path.

Suggest: validate the metadata schema in `read_checkpoint_metadata` (required keys present, types right) and raise `UserInputError("checkpoint ... has unreadable metadata")`; and have `_run_barrier` catch `Exception` → fail the barrier's entries as `server` instead of exiting.

miles/tinker/core/service.py
```diff
@@ -0,0 +1,598 @@
...
+
+
+class TinkerService:
+    def __init__(self, backend, config: GatewayConfig) -> None:
+        self.backend = backend
+        self.config = config
+        self.futures = FutureStore()
+        self.planner = Planner(config.batch_token_budget)
+        self.models: dict[str, ModelRecord] = {}
+        self.sessions: dict[str, dict] = {}
```
HJSang, line 41: Non-blocking, for a follow-up after the stack lands: `ModelRecord` and `Command` are dataclasses, but sessions, sampling sessions, and every backend outcome are string-keyed dicts. The two crash paths flagged elsewhere (`response["meta_info"]` in `_to_sequence`, `meta["tenant_digest"]` in `read_checkpoint_metadata`) are this pattern: a missing key fails at the read site inside the dispatcher instead of at construction.

Suggest `SessionRecord`, `SamplingSession`, and an `Outcome` (ok | error + category + payload) dataclass, with the backend `Protocol` typed to return `Outcome`. Then `_finish_request` becomes the only place that interprets failure, and the fake backend and `MilesBackend` can't drift on result shape. Happy to leave this for a later PR.

miles/backends/megatron_utils/model.py
```diff
@@ -413,68 +402,15 @@ def _zero_grads(model: Sequence[DDP], optimizer: MegatronOptimizer | None, disab
...
-        reset_grad_metadata_keep_grads(model)
-    else:
-        _zero_grads(model, optimizer, disable_optimizer)
-
-    if args.custom_megatron_before_train_step_hook_path:
-        from miles.utils.function_registry import load_function
-
-        custom_before_train_step_hook = load_function(args.custom_megatron_before_train_step_hook_path)
-        custom_before_train_step_hook(args, rollout_id, step_id, model, optimizer, opt_param_scheduler)
+def run_forward_backward_pass(
```
HJSang, line 405: Not multi-LoRA specific, but since this splits the hot path every run goes through: `train_one_step` → `run_forward_backward_pass` plus the actor `_init_training_state` extraction are mechanical, and every semantic fork I could find is gated on `is_multi_lora_enabled` / `batch.get("loss_fn")`, so full-param should be numerically unchanged. It would be good to have that on record rather than by inspection.

Suggest a quick sanity job before the stack lands: one existing full-param recipe (e.g. the dense Qwen3-4B or GPT-OSS e2e config) run on `main` and on this head with identical seeds, comparing loss, grad norm, and a handful of logged metrics for the first N steps. `mechanical-refactor-verify` (rdxa_skills) is the fitting harness for the `run_forward_backward_pass` extraction itself. Broad GPU CI on #2849 covers a lot of this already; the ask is just a side-by-side on one recipe so the "unchanged for full-param" claim is in the PR rather than implied.

miles/tinker/core/future.py
```diff
@@ -0,0 +1,84 @@
...
+    state: str = PENDING
+    result: dict | None = None
+    error: str | None = None
+    error_category: str | None = None
+    finished_at: float | None = None
+    # long-poll wakeup: set when the future settles
+    settled: asyncio.Event = field(default_factory=asyncio.Event)
+
+
+class FutureStore:
```
HJSang: low priority and not blocker.
Maybe we can rename them to `RequestFuture` and `RequestFutureStore` to avoid naming collision in ` Future from core.future`

serve_tinker.py
```diff
@@ -0,0 +1,97 @@
...
+        base_model=args.tinker_base_model or args.hf_checkpoint,
+        n_slots=args.multi_lora_n_adapters,
+        checkpoint_root=checkpoint_root,
+        vocab_size=vocab_size,
+        lora_alpha=args.lora_alpha,
+        max_lora_rank=args.lora_rank,
+        trains_attn=bool(target_modules & {"q_proj", "k_proj", "v_proj", "o_proj"}),
+        trains_mlp=bool(target_modules & {"gate_proj", "up_proj", "down_proj"}),
+        trains_unembed="lm_head" in target_modules,
+    )
```
nanjiangwill, line 69: `max_tokens_per_datum` should come from the trainer limit. the example runs with `--max-tokens-per-gpu 8192`, but this keeps the 32768 default. `get_minimum_num_micro_batch_size()` does not split or reject one oversized sequence, so an 8193–32768-token datum is admitted and placed alone in an over-budget microbatch. can we derive this from `args.max_tokens_per_gpu` and the model context limit, and add a boundary test?

serve_tinker.py
```diff
@@ -0,0 +1,97 @@
...
+
+    target_modules = set(convert_target_modules_to_hf(args.target_modules))
+    config = GatewayConfig(
+        base_model=args.tinker_base_model or args.hf_checkpoint,
+        n_slots=args.multi_lora_n_adapters,
+        checkpoint_root=checkpoint_root,
+        vocab_size=vocab_size,
+        lora_alpha=args.lora_alpha,
+        max_lora_rank=args.lora_rank,
+        trains_attn=bool(target_modules & {"q_proj", "k_proj", "v_proj", "o_proj"}),
```
nanjiangwill: these booleans collapse partial and model-specific target layouts into a contract tinker cannot represent. for example, `--target-modules linear_qkv` sets `trains_attn=true` even though `o_proj` is not trained, while an mla-only attention layout reports false. can we canonicalize the targets once and reject layouts that are not exactly representable by `train_attn`, `train_mlp`, and `train_unembed`, with tests for partial and mla layouts?

PR #2847, opened 2026-08-31
Fix targeted reruns for source-only CI labels

fzyzcjy:
## Summary

- parse the checked-out source snapshot's literal `KNOWN_LABELS` dictionary with trusted AST-only code
- validate source registrations against that schema while keeping runner, suite, timeout, path, symlink, and disabled-test policy under default-branch control
- document and test targeted reruns when a pull request introduces a new CI label

## Testing

- 96 focused tests passed
- PR #2542 source snapshot resolver exited successfully
- independent security and test-weakening review: LGTM


fzyzcjy:
(vibed, not check yet)

PR #2850, opened 2026-09-01
Honour MILES_HOST_IP for the router and session server bind addresses

nblintao:
`--miles-host-ip 0.0.0.0` — which the Harbor recipes document and pass — stopped working when the #2305 stack moved the router and the session servers onto the worker manager. This restores it for both.

On `main` both bind an address `get_host_info()` computes, which reads `MILES_HOST_IP` first: the router directly, the session servers by inheriting `sglang_router_ip`. Under the worker manager the bind address comes from the ray placement instead, so the flag has no effect: a deployment that needs these services to accept connections forwarded from another host silently gets them bound to the pod address only.

The bind address is not the address peers reach a worker on, so the fix stays local — both launch commands fall back to `MILES_HOST_IP` before the placed address, and for the session servers the explicit `--session-server-ip` still wins.

## Scope

Independent in code of #2851 (external addresses for off-cluster agents) — different concept, no shared lines — but a functional prerequisite for it: an external address is useless if the server only listens on the pod interface. Both are based on `tom/refactor-miles/op13-162`.

## Tests

Not executed — no torch on this machine. `ruff`, `black` and `py_compile` are clean. New cases pin the fallback order for the session server (flag → env → placement) and the env fallback for both router flavours; an autouse fixture clears ambient `MILES_HOST_IP` so the module's pre-existing host assertions stay environment-independent.


nblintao:
Obsolete: `main` now reads `MILES_HOST_IP` inside `NodeProbeMixin._get_node_ip` (via #2051), so every worker-manager service — router and session servers included — binds and advertises it again, which is what this PR restored on the old stack base. The `--session-server-ip` override this PR also covered survives on `main` independently.

PR #2851, opened 2026-09-01
Let off-cluster agents reach session servers on more than one host

nblintao:
Answers the multi-host question on #2305 ([comment](https://github.com/radixark/miles/pull/2305#issuecomment-5432764485)) by making the combination work rather than forbidding it.

## What is missing

#2305 lets the session servers sit on more than one host. Reaching them from outside the cluster was never built for that: the off-cluster legs (Harbor, OpenEnv, NeMo Gym) each rewrite the host of the URL they are handed to `MILES_ROUTER_EXTERNAL_HOST` — **one address, which can only ever point at one machine**. Spread the instances over several nodes and no single value reaches them all: the rewrite has replaced the one field that tells them apart. The interim answer on `main` pins every session server back onto the head node, fencing off the capability #2305 added.

## The approach: record the external view at the source

Each node reports the address that reaches it, read from `MILES_NODE_EXTERNAL_IP` **in that node's own environment** — the probe runs on the node, so every node reports its own value, the way a kubernetes node carries its own `ExternalIP`. Nothing has to hold a node-to-node table, which a dynamically scheduled pod could not fill in anyway: its address only exists after placement.

- `HostAndPort` gains `external_host`, falling back to the placed address — a deployment whose node addresses already route from the sandbox network needs no configuration at all.
- `SessionServerInstance` (#3108) gains `external_addr`; the tracer splits `base_url` (the driver's own calls) from `agent_base_url` (handed to the agent). Both views come from the one chosen record, so the two never name different instances, and the driver's traffic never leaves the cluster network.
- The legs stop rewriting: the URL they receive already names the instance as they reach it. `MILES_ROUTER_EXTERNAL_HOST` is removed; `resolve_session_url` keeps only the shared `/v1` join, renamed `openai_session_url`.
- With the rewrite gone the pin has nothing left to guard, so the session servers follow `--pin-rollout-manager-to-head` again instead of being pinned unconditionally.

The launcher flag is now `--node-external-ip`: one value for every node at once, correct only while the instances share a host, as its comment says. Its Harbor default changed from `socket.gethostname()` to empty, so it no longer overrides what a deployment sets per pod.

## Stacked on

#3108 (the session servers as one list of instance records). This PR adds the `external_addr` field to that record; as separate lists the external view would have been a third structure to keep aligned.

## Tests

**E2E smoke — passed (2026-09-04)**, on this stack: two-node ray cluster (8×H200 head + CPU-only second node), each node reporting its own tailnet address via `MILES_NODE_EXTERNAL_IP`; openenv e2b leg against the internal AgentENV sandbox service, GLM-4.7-Flash GRPO. Ray spread the 32 session servers across both nodes; the driver published each instance with its own node's external address; agents dialed those external addresses across nodes — 600+ session calls in the final run, zero `SessionNotFoundError` (3,000+ across all runs); a full train step completed with real rewards from the sandbox verifier (`rollout/raw_reward 0.375`, `train_step outcome=NORMAL valid_step=true`). Every failure along the way was test-rig or sandbox-provider territory (asymmetric node envs, orphaned processes between reruns, AgentENV#12/#15) — no defect surfaced in this stack's code.

Not executed — no torch on this machine. `ruff`, `isort`, `black` and `py_compile` are clean; the record and tracer paths were exercised under python 3.12:

```
OK record carries both views; one choice, never split
```

New cases cover the record's external field and its fallback, the worker manager threading each node's reported address into its workers, the driver publishing per-node external addresses, and the env-var contract.

### DIFF
--- a/examples/experimental/nemo-gym/run.py
+++ b/examples/experimental/nemo-gym/run.py
@@ -52,10 +52,11 @@ class ScriptArgs(U.ExecuteTrainConfig):
 
     # NeMo Gym settings
     nemo_gym_url: str = os.environ.get("NEMO_GYM_URL", "http://localhost:12000")
-    # Trainer address reachable from the NeMo Gym host; only needed when that
-    # host cannot resolve the trainer's hostname (e.g. it dials back over a
-    # tailnet).
-    router_external_host: str = os.environ.get("MILES_ROUTER_EXTERNAL_HOST", "")
+    # Trainer address reachable from the NeMo Gym host; only needed when that host
+    # cannot resolve the trainer's hostname (e.g. it dials back over a tailnet).
+    # Sets MILES_NODE_EXTERNAL_IP for every node at once, so on a multi-node job
+    # leave it empty and have the deployment set that variable per pod instead.
+    node_external_ip: str = os.environ.get("MILES_NODE_EXTERNAL_IP", "")
 
 
 def cleanup():
@@ -183,8 +184,8 @@ def execute(args: ScriptArgs):
         "PYTHONPATH": f"{args.megatron_path}:{SCRIPT_DIR}:{U.repo_base_dir}",
         "NEMO_GYM_URL": args.nemo_gym_url,
     }
-    if args.router_external_host:
-        extra_env_vars["MILES_ROUTER_EXTERNAL_HOST"] = args.router_external_host
+    if args.node_external_ip:
+        extra_env_vars["MILES_NODE_EXTERNAL_IP"] = args.node_external_ip
 
     U.execute_train(
         train_args=train_args,
--- a/miles/ray/specs/inference.py
+++ b/miles/ray/specs/inference.py
@@ -108,7 +108,7 @@ def spec_session_server(args) -> CommandWorkerSpec:
             num_workers_per_cell=1,
             num_gpus_per_worker=0,
             num_cpus_per_worker=0,
-            pin_to_head=True,
+            pin_to_head=args.pin_rollout_manager_to_head,
         ),
         launch_command=_compute_launch_command,
     )
--- a/miles/utils/misc.py
+++ b/miles/utils/misc.py
@@ -69,6 +69,21 @@ def get_current_node_ip():
     return address
 
 
+MILES_NODE_EXTERNAL_IP_ENV = "MILES_NODE_EXTERNAL_IP"
+
+
+def get_node_external_ip() -> str | None:
+    """The address off-cluster peers reach this node on (a DNS name also works),
+    or None when its own address already routes from there.
+
+    The deployment supplies this per node -- a tailscale sidecar or an init
+    container writing the interface it joined. Passing it through ray's
+    runtime_env instead would give every node one shared value, which only
+    addresses the right instance while all of them sit on one host.
+    """
+    return os.environ.get(MILES_NODE_EXTERNAL_IP_ENV) or None
+
+
 def get_free_port(start_port=10000, consecutive=1):
     # find the port where port, port + 1, port + 2, ... port + consecutive - 1 are all available,
     # scanning upwards from start_port and wrapping around once the ports run out
@@ -119,6 +134,10 @@ class NodeProbeMixin:
     @staticmethod
     def _get_node_ip() -> str:
         return os.getenv(MILES_HOST_IP_ENV) or get_current_node_ip()
+
+    @staticmethod
+    def _get_node_external_ip() -> str | None:
+        return get_node_external_ip()
 
     @staticmethod
     def _get_free_port_block(*, start_port: int, count: int) -> int:
--- a/docs/examples/swe-agent-harbor-docker.md
+++ b/docs/examples/swe-agent-harbor-docker.md
@@ -108,7 +108,7 @@ python examples/swe-agent-harbor-docker/run.py \
     --num-rollout 200 \
     --save-interval 20 \
     --agent-server-url http://<agent-server>:30000 \
-    --router-external-host <trainer-host-reachable-from-agent-server> \
+    --node-external-ip <trainer-host-reachable-from-agent-server> \
     --miles-host-ip 0.0.0.0 \
     --save-traces-dir /path/to/traces
 ```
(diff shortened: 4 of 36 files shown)


nblintao:
@claude review

nblintao:
@claude review

miles/utils/misc.py
```diff
@@ -120,6 +135,10 @@ class NodeProbeMixin:
     def _get_node_ip() -> str:
         return os.getenv(MILES_HOST_IP_ENV) or get_current_node_ip()
 
+    @staticmethod
+    def _get_node_external_ip() -> str | None:
+        return get_node_external_ip()
```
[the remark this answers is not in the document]
nblintao: fixed

nblintao:
/rerun-test tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py

miles/ray/specs/inference.py
```diff
@@ -107,7 +107,7 @@ def _compute_launch_command(ctx: LaunchCommandContext) -> str:
             num_workers_per_cell=1,
             num_gpus_per_worker=0,
             num_cpus_per_worker=0,
-            pin_to_head=True,
+            pin_to_head=args.pin_rollout_manager_to_head,
```
guapisolo, line 110: cc cmt

[P1] With the head pin gone, the shared `--node-external-ip` mode sends agents to the wrong node on multi-node runs.

`--node-external-ip` goes into the job runtime env, so every node reports the same `MILES_NODE_EXTERNAL_IP`. That address is right only while all session servers sit on one host, which the hard `pin_to_head=True` guaranteed on main. Here the pin follows `--pin-rollout-manager-to-head`, which defaults to False and no launcher sets. The session-server actors reserve 0 CPUs and 0 GPUs, so Ray places them on any node; the two-node smoke in the description saw them spread.

Example: two nodes, head A reachable at `100.64.0.1`, launched with `--node-external-ip 100.64.0.1`. Instance 1 lands on node B at `10.0.0.2:30001`, but its record gets `external_addr=100.64.0.1:30001`. The driver creates the session over `10.0.0.2`, so startup and readiness pass. The agent then dials A on port 30001, where nothing listens, and every episode routed to a non-head instance fails or scores 0. The same launch works on main.

This is P1 because it silently breaks a multi-node configuration that works today (e.g. `run_glm52_lora_tb2_daytona.py` with the flag), with no startup error. Single-node runs and the per-node `MILES_NODE_EXTERNAL_IP` mode are unaffected.

Either change would close it:
- the launchers pass `--pin-rollout-manager-to-head` whenever `node_external_ip` is set, restoring main's placement for the shared-address mode; or
- `wait_session_server_ready` fails fast when instances on different placed hosts share one external host, so the misconfiguration surfaces at startup.

Evidence: static reading of this head (00307917) plus the placement seen in the PR's own smoke. I have not run the shared-flag multi-node case.

examples/experimental/nemo-gym/run.py
```diff
@@ -52,10 +52,12 @@ class ScriptArgs(U.ExecuteTrainConfig):
...
-    # Trainer address reachable from the NeMo Gym host; only needed when that
-    # host cannot resolve the trainer's hostname (e.g. it dials back over a
-    # tailnet).
-    router_external_host: str = os.environ.get("MILES_ROUTER_EXTERNAL_HOST", "")
+    # The trainer address the NeMo Gym host reaches, needed only when that host cannot resolve the
+    # trainer's hostname (e.g. it dials back over a tailnet). It is set as MILES_NODE_EXTERNAL_IP on
+    # every node, so it is only right while the session servers share a host. The default is not read
+    # from that variable: in the launcher's environment it holds this node's own address, and
+    # forwarding it would overwrite every other node's.
+    node_external_ip: str = ""
```
guapisolo, line 60: cc cmt

[P2] An existing `MILES_ROUTER_EXTERNAL_HOST` export is now ignored without any signal.

The old NeMo Gym README told users to `export MILES_ROUTER_EXTERNAL_HOST=<trainer host/IP>` before running this launcher, and this launcher (like the harbor, openenv, terminus-compaction and swe-agent-harbor-docker ones) took that variable as its default. After this change nothing reads it: `node_external_ip` defaults to `""` and the legs no longer rewrite the session URL. A user who keeps the old export gets no error, and the NeMo Gym host receives the placed cluster address. When that address does not route from there (the tailnet case the README described), every episode fails at its first policy call.

This is P2: it only hits deployments that relied on the rewrite and kept the old export, and passing `--node-external-ip` works around it, but the failure shows up mid-training instead of at launch. The description lists the removal under user-visible changes; what is missing is a runtime signal.

Suggested fix: fail fast in `U.execute_train`, which all of these launchers call, when `MILES_ROUTER_EXTERNAL_HOST` is set, naming `--node-external-ip` / `MILES_NODE_EXTERNAL_IP` as the replacement.

nblintao:
/rerun-test tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py

examples/experimental/nemo-gym/run.py
nblintao: Fixed.

miles/ray/specs/inference.py
nblintao: Fixed a bit differently: the shared address is now its own train arg, --session-server-external-host, and setting it pins the session servers to the head, as on main. MILES_NODE_EXTERNAL_IP stays per-node only, and startup now fails if instances on different hosts would share one external host.

guapisolo:
@claude review always

guapisolo:
cc cmt

[P1] I will push 7564a2afe9 onto this branch to switch `examples/experimental/nemo-gym-workspace-assistant/workplace_agent.py` (added on main by #3658) from the removed `resolve_session_url` to `openai_session_url`, because at 4e1525a4 the module fails to import (`ImportError: cannot import name 'resolve_session_url'`), so `run_nemotron35_workplace.py` cannot start any episode and no test catches it. Verified by importing the module before (ImportError) and after (imports cleanly, the URL ends in `/sessions/<id>/v1/chat/completions`) with black and isort clean; the example itself was not run end to end.

guapisolo:
cc cmt

I will force-push (with lease on 7564a2af) this branch to 1b7141c9f8, a history rewrite that replays 4e1525a4 and 7564a2af unchanged onto the rebased #3108 (8d3754c286, on main b2c9e463 with #3695's fix for the unrelated `test_shared_ppo_lifecycle.py` CPU failure) and adds 1b7141c9f8, which says in the `--pin-rollout-manager-to-head` help that the flag now also pins the session servers, since this PR makes them follow it. Verified with `git range-diff` showing both replayed commits identical and black clean on the help change; CI on the new head has not run yet.

PR #2852, opened 2026-09-01
perf(memory): toggle expandable segments around the training step

xiaohong42:
## Summary

- Enable PyTorch's `expandable_segments` allocator mode during actor training and disable it before offload/rollout, gated by `MILES_EXPANDABLE_SEGMENTS_DURING_TRAIN=1` (default off)
- Uses the runtime allocator API (`torch._C._accelerator_setAllocatorSettings`) rather than `PYTORCH_ALLOC_CONF` so the setting does not propagate to SGLang child processes, which would break IPC weight transfer and `torch_memory_saver` offloading

### Known limitation under colocate

In colocated mode, tensors still referenced when the window closes (e.g. `train_actor`'s return value) land in VMM segments that `torch_memory_saver`'s `pause()` cannot reach (it hooks `hipMalloc`/`hipFree`, not `hipMemCreate`/`hipMemMap`). This raises the device-level peak during rollout. The training-phase peak still drops, so the net effect depends on which phase is the bottleneck. This is a pre-existing TMS limitation, not introduced by this change; a follow-up could extend the window to include `offload_train` or release the training output before the window closes.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -23,7 +23,7 @@ from miles.utils.context_utils import with_defer
 from miles.utils.distributed_utils import get_gloo_group
 from miles.utils.ft_utils.indep_dp import IndepDPInfo
 from miles.utils.hf_config import load_hf_config
-from miles.utils.memory_utils import clear_memory, print_memory
+from miles.utils.memory_utils import clear_memory, expandable_segments_during_training, print_memory
 from miles.utils.multi_lora import is_multi_lora_enabled
 from miles.utils.processing_utils import load_tokenizer
 from miles.utils.ray_utils import Box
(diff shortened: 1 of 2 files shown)


xiaohong42:
Field report from a 4-node gfx942 (MI308X) bring-up.

**The `except AttributeError` fallback is load-bearing on this stack.** `torch._C._accelerator_setAllocatorSettings` does not exist on torch 2.9.0a0+git7bcbafe; `torch.cuda.memory._set_allocator_settings` does. Without the fallback this would have failed at the first training step rather than at import.

**One interaction worth putting in the description.** Expandable segments serve memory through the HIP VMM API, and VMM memory cannot be exported as a CUDA IPC handle -- which is exactly how Miles hands weights to SGLang in colocate mode. The failure is `CUDA error: invalid argument` during the weight update, far from the allocator setting that caused it.

The context manager calls `clear_memory()` before switching back, which should be enough in principle, but **nobody has run a colocate RL round with `MILES_EXPANDABLE_SEGMENTS_DURING_TRAIN=1`** as far as I can tell. Since the flag defaults off, merging is safe either way; it is using it that is unverified. A note in the docstring, or a guard that refuses to enable it under `--colocate`, would keep the next person from spending a run on it.

PR #2863, opened 2026-09-01
openenv tb2 recipe: pin mcp below 2 in the server layer

nblintao:
Freshly baked TB2 images fail every episode with `camel-ai (TerminalToolkit) is required for TB2`. The real error inside the image is `ImportError: cannot import name 'FastMCP' from 'mcp.server'`: camel-ai constrains only `mcp>=1.3`, and mcp 2.0 (2026-07-28) dropped that re-export. Anything baked before late July still works, which is why this only shows up on new bakes.

One constraint in the server layer: `mcp<2`. The template alias hashes the recipe commands, so the pin re-bakes every task image on first use and the already-broken templates are simply never referenced again. Verified in a baked sandbox on the dev sandbox-service cluster — the import fails as resolved today, works with 1.29.1 (what `<2` resolves to) — and `scan_golden.py` then scores 1.0 on `fix-git` and `adaptive-rejection-sampler`, both freshly baked.


Shi-Dong:
@claude review always

Shi-Dong (approved):
LGTM!

PR #2866, opened 2026-09-01
Generalize --dsv4-impl into --model-impl

yueming-yuan:
One lever for which side implements a model's non-standard layers: `--model-impl {miles, megatron}`, default `megatron`.

## What

- **`--dsv4-impl` is gone**; `--model-impl` is a general argument. For DeepSeek-V4 it means exactly what `--dsv4-impl` meant (`megatron` → `dsv4_hybrid` variant, `miles` → the ported `dsv4` variant; the TP/kernel-backend hard errors are unchanged). Every other family accepts only `megatron` — the standard stack is the identity choice — and `miles` raises.
- **Model identity comes from the checkpoint, not the launcher**: `set_default_megatron_args` resolves `args.model_family` from the HF config's `model_type`, so every megatron entry point (training, checkpoint conversion, the replay worker) gets it; `is_dsv4_model` no longer sniffs the `--spec` string. Families with a miles implementation register a normalize hook in one table; the dispatch replaces the `if is_dsv4_model(...)` chain.
- **Bridge rule**: `--megatron-to-hf-mode bridge` builds megatron-native modules by construction (Megatron-Bridge providers, e.g. `qwen35_bridge.py` sets `experimental_attention_variant`), so `--model-impl miles` under bridge is a hard error, family-independent.

## Migration

- `scripts/run_deepseek_v4.py` / `scripts/amd/run_deepseek_v4.py`: `dsv4_impl` field and flags renamed (env binding `MILES_SCRIPT_DSV4_IMPL` → `MILES_SCRIPT_MODEL_IMPL`); dsv4 e2e tests follow; launcher snapshots regenerated.
- The standalone megatron-parse tools registered the dsv4 arg themselves; `tools/convert_hf_to_torch_dist.py` and the replay worker now take the shared `add_model_impl_argument` — the converter keeps selecting the impl, which matters because the two impls' torch_dist checkpoints are not interchangeable.
- `test_dsv4_arguments.py` is deleted: every case restated a directly visible raise or one-line mapping.

## Verification

- `tests/fast/utils/test_arguments.py` — 156 passed.
- `tests/fast/launch_scripts` + `tests/manual/launch_scripts` — 494 passed; the 6 failures + 4 errors (amd glm5_2/inkling/kimi snapshots) reproduce identically on a clean main checkout.
- Not executed: a dsv4 training run; the change is argument plumbing with unchanged resolved fields, covered by the unit tests above.

Next (separate PR, stacked): register the `qwen3_5` family — its `miles` impl is the `miles_plugins.models.qwen3_5` spec (fla GDN), its `megatron` impl is the bridge/`gated_delta_net` path — and switch the Qwen3.5 LoRA recipe to thd.


miles/utils/arguments.py
```diff
@@ -100,6 +100,21 @@ def reset_arg(parser, name, **kwargs):
 _FT_CHOICES = ["rollout", "train"]
 
 
+def add_model_impl_argument(parser):
+    """Shared with the standalone megatron-parse tools (checkpoint conversion, replay worker)."""
+    parser.add_argument(
+        "--model-impl",
```
Zhichenzzz, line 106: could we explicitly tell users which models are now both supported in "miles" and "megatron", e.g., qwen35 and ds-v4

Zhichenzzz (approved):

PR #2869, opened 2026-09-01
fix(lora): pass config to the bridge value head instead of sequence_parallel

Shi-Dong:
## What this changes

`_make_value_model_hook` in `bridge_lora_helpers.py` swaps a bridge-built model's language-model head for a scalar value head. It constructed that head by passing `sequence_parallel=` to `LinearForLastLayer` — but that class takes a keyword-only `config` and has no `sequence_parallel` parameter at all. The call raised `TypeError` before it could build anything.

The fix passes the model chunk's own `config` instead. `LinearForLastLayer` reads exactly one field off it, `config.sequence_parallel`, and every chunk already carries the config the provider produced. The value that used to be threaded in separately from `provider.sequence_parallel` now comes from the chunk, so the redundant factory parameter is dropped.

## Why it is needed

The hook is reachable whenever LoRA or multi-LoRA builds a model through the bridge and the HF checkpoint declares a `ForSequenceClassification` or `ForTokenClassification` architecture — in other words, LoRA training on a reward or value model. Any such run died at model construction.

The mismatch has been there since the hook was first written: `LinearForLastLayer` has required a keyword-only `config` since it was introduced, so this call site never matched the constructor. It survived because every existing test that touches `_setup_lora_model_via_bridge` patches the function out, leaving the hook body with no coverage at all.

## Tests

New `tests/fast/backends/megatron_utils/test_bridge_value_model_hook.py` drives the hook directly against a stubbed megatron surface. No GPU, registered as a CPU CI test.

- the value head replaces the output layer on the last pipeline stage
- the head receives the chunk's config, carrying the correct `sequence_parallel`
- the factory accepts only `hidden_size`
- under virtual pipelining, non-last stages keep their original head
- a drift guard that parses the real `LinearForLastLayer` signature out of `model_provider.py` and binds the kwargs the hook passes, so a future change to that constructor fails here instead of shipping

I checked these tests genuinely fail against the unfixed code, in two separate ways: reverting the factory arity, and reverting only the constructor keyword while leaving arity intact. The second reproduces the original `TypeError` at the exact call site. All five pass with the fix, and pre-commit is clean.

Caveat on local verification: I ran the new test file with `--noconftest`, because the repo `conftest.py` needs a newer sglang than my local environment has. The rest of the suite was not run locally and is left to CI.

## Not addressed here

The value hook mutates the model chunks in place and returns `None`, while the sibling LoRA hook registered just above it returns the transformed chunks. If `register_pre_wrap_hook` chains hooks by passing return values forward, that difference would matter. I could not check megatron-bridge's semantics from my environment, so I left it alone rather than guess. Flagging it for anyone who can confirm quickly.

### DIFF
--- a/miles/backends/megatron_utils/bridge_lora_helpers.py
+++ b/miles/backends/megatron_utils/bridge_lora_helpers.py
@@ -33,7 +33,7 @@ def _ensure_model_list(model):
     return model if isinstance(model, list) else [model]
 
 
-def _make_value_model_hook(hidden_size: int, sequence_parallel: bool):
+def _make_value_model_hook(hidden_size: int):
     """Create a pre-wrap hook that replaces the output layer with a value head."""
     from megatron.core import parallel_state
 
@@ -59,7 +59,7 @@ def _make_value_model_hook(hidden_size: int, sequence_parallel: bool):
             model_chunk.output_layer = LinearForLastLayer(
                 input_size=hidden_size,
                 output_size=1,
-                sequence_parallel=sequence_parallel,
+                config=model_chunk.config,
             )
 
     return hook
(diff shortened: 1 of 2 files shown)


Shi-Dong:
@claude review always

Zhichenzzz (approved):

PR #2877, opened 2026-09-02
fix(ci): mint the command-App token with pull-requests: write for label commands

guapisolo:
## Summary

Fix HTTP 403 on label commands: the command-App token needs `pull-requests: write`.

## Symptom & Reproduction

- **Symptom:** `Comment command gateway` job `handle-command` failed at its App-token step with `##[error]GitHub API returned HTTP 403` right after `Mint the issues-scoped App token` succeeded; no label was added, no downstream CI started.

| Run | Commenter | Authorization path | Default-branch code |
|---|---|---|---|
| [33561328725](https://github.com/radixark/miles/actions/runs/33561328725) | guapisolo | repository role `admin` | 9d83a7f |
| [33567525760](https://github.com/radixark/miles/actions/runs/33567525760) | zyzshishui | explicit allowlist user ID | 2a7d68b |

- **Reproduction:** post a comment whose entire body is `/run-ci-sglang` on an open PR while the mint step requests `permission-pull-requests: read`; expected the label added, observed the 403 above. The zyzshishui run pins the failing call because its path performs only `GET /pulls/{n}` and the label `POST`.

## Root Cause

1. `comment-ci-command.yml` mints the App token with `permission-issues: write` and `permission-pull-requests: read`.
2. `GitHubAPI.add_label` sends `POST /repos/radixark/miles/issues/{n}/labels` for an issue that is a pull request.
3. GitHub authorizes issues-API writes on a pull request against `pull_requests`, so the call returns 403.
4. `GitHubAPI._request` reported only `HTTP 403`, hiding which call and which permission failed.

## Fix

The mint step requests `permission-pull-requests: write`. The App registration and its `radixark/miles` installation already hold `Pull requests: Read and write` (accepted 2026-09-01), and the current `read` request is a subset, so merging has no failure window. `GitHubAPI._request` now names the failing method and path and appends GitHub's `X-Accepted-GitHub-Permissions` header, never the token, so a permission failure identifies the call and the missing permission. `docs/ci/01-label.md` and `docs/ci/05-command-identity.md` state the new contract, the installation-approval requirement, and that `Pull requests: write` permits submitting reviews; the mitigations are the fixed handler code, which never calls a review endpoint, and the job-scoped token revoked when the job ends. The code-owner review rule on `main` covers only code-owned paths, so it is not counted as a mitigation. The explicit-ID and repository-role authorization paths are unchanged; `permission-issues: write` stays, and dropping it is a separate experiment.

## Verification

- `tests/ci/test/test_comment_ci_command.py`: 286 passed on c7b220e with `--noconftest`; the workflow test asserts the exact indented `permission-pull-requests: write` line (a commented-out key fails), asserts `read` is absent, plus two tests pin the error message with or without the accepted-permissions header.
- `black`, `isort`, `ruff check` pass on the changed Python files; the workflow YAML parses with the expected `with:` block.
- Live retest needs this change on the default branch because comment workflows run default-branch code only: (1) an explicit-allowlist user posts `/run-ci-<label>` on a PR where that label is absent; (2) a repository admin not on the explicit list posts `/clear-labels`, then `/run-ci-<label>`, which exercises `GET /collaborators/{login}/permission` on the App token; record each gateway run ID, its decision JSON (`ALLOW_ADDED` or `ALLOW_CLEARED`, not `ALLOW_ALREADY_PRESENT`), and the `labeled` consumer runs on the frozen head; label commands get no reaction, only `/rerun-test` does.

## Review Focus

- `comment-ci-command.yml` mint step: `pull-requests: write` is the minimum GitHub accepts for label writes on a pull request.
- `GitHubAPI._request`: the message carries method, path, plus the accepted-permissions header; every path segment is a constant, an integer, or a `urllib.parse.quote(..., safe="")` value.
- `GET /collaborators/{login}/permission` with the App token needs only `Metadata: read` per GitHub docs; on a 403 the new message names the call, and `accepted permissions: metadata=read` would mean adding `permission-metadata: read` to the mint step.

Generated with [Claude Code](https://claude.com/claude-code)


guapisolo:
bypass merge to unblock. need e2e test.

PR #2878, opened 2026-09-02
[PPO] Support the critic under --megatron-to-hf-mode bridge

yueming-yuan:
## Why

PPO rejected `--megatron-to-hf-mode bridge` with "Critic models are not supported". The reason was in the bridge model provider: it ignored `role`, so the critic came out as a plain LM (vocab-sized output layer, tied to the embedding per HF `tie_word_embeddings`) instead of a value head, and `get_values` would have gotten `[1, T, vocab]` where it needs `[1, T, 1]`.

The critic never touches the weight-update path (it returns from `init` before `weight_updater` is built and has no `rollout_manager`), so no changes there.

## What

- **Provider** (`model_provider.py`): the bridge branch now does what the raw and custom branches already do for `role == "critic"` — `share_embeddings_and_output_weights = False` on the provider before `finalize()`, and swap `output_layer` for `LinearForLastLayer([hidden, 1])` on post-process stages.
- **HF load** (`checkpoint.py`): the critic's weights come from the HF checkpoint via `bridge.load_hf_weights()`. That walk maps `output_layer.weight` onto `lm_head.weight`; with the `[1, hidden]` head it raises on shape for untied models at TP=1 and takes a negative-pad path in `ColumnParallelMapping.hf_to_megatron` at TP>1 (`allowed_mismatched_params` does not reach that path). `_hide_critic_value_head_from_hf_load` detaches every `LinearForLastLayer` child for the duration of the load — same trick miles already uses with `hide_adapters` for multi-LoRA base loads. The head keeps its fresh init, matching the raw path (tied model → key missing → ignored by `assume_ok_unexpected`).
- Drop the assert in `arguments.py`; flip `test_bridge_mode_rejects_critic` → `test_bridge_mode_accepts_critic`.
- Docs: `examples/ppo/README.md` + regenerated `docs/examples/ppo.md`; `docs/advanced/lora.md` no longer cites the bridge-critic rejection as the reason PPO+Bridge LoRA is unsupported (it is now merely untested — the critic never gets adapters).

## Validation

- `pytest tests/fast/utils/test_arguments.py` on an H200 devbox — 156 passed.
- E2E (run on the commit before the final `_hide_critic_value_head_from_hf_load` rewrite from `_modules.pop` to `delattr`/`setattr`; intended behavior-equivalent, **not re-run since**), 3 rollouts each, `tests/e2e/megatron/test_qwen3_4B_ppo.py` args plus `--megatron-to-hf-mode bridge` (HF dirs for `--hf-checkpoint`/`--ref-load`, no torch_dist), 4×H200 colocated:
  - **Qwen3-4B (tied), TP1/PP2/CP2** — exit 0. Critic loaded via the bridge with no `output_layer` mapping/shape warnings; rollout 0 critic-only; `critic-value_loss` 9.81 → 4.90 → 3.43 → 2.93 → 1.51 → 3.40.
  - **Qwen3-8B (untied, `tie_word_embeddings=false`), TP2/PP2** — exit 0. This is the configuration where the unhidden head would have hit the `lm_head.weight` shape path; no warnings, `critic-value_loss` 3.96 → 4.17 → 1.94 → 2.08 → 1.46 → 1.42.

Not executed: the full fast suite (CI covers it); PPO + Bridge LoRA; virtual pipeline (VPP) with a bridge critic.

## Follow-ups not in this PR

- `critic_load` defaults to the actor's `--load`, while `critic_save` is `--save + "_critic"`; on resume the critic therefore loads from the actor's checkpoint dir. Pre-existing, raw and bridge alike.
- `bridge_lora_helpers._make_value_model_hook` constructs `LinearForLastLayer(sequence_parallel=...)`, a kwarg the class does not take; it has no callers.

### DIFF
--- a/docs/advanced/lora.md
+++ b/docs/advanced/lora.md
@@ -177,9 +177,9 @@ alternative aligned-expert path.
 - **Losses and algorithms.** LoRA wraps the actor model independently of the
   policy loss. The maintained RL recipes use critic-free estimators such as
   GRPO. The SFT loss path can use the same adapter hooks, although the repository
-  does not yet have a dedicated LoRA-SFT E2E test. Shared actor/critic PPO is not
-  supported by the general Bridge LoRA path because the critic path rejects
-  Bridge mode.
+  does not yet have a dedicated LoRA-SFT E2E test. Shared actor/critic PPO with
+  Bridge LoRA is untested: the critic is always built as a full model without
+  adapters.
 - **Checkpoints.** miles saves native per-rank adapter shards and
   optimizer/scheduler state. Exact resume expects the same TP/PP topology. It
   also attempts a best-effort HF PEFT `adapter_model.bin` plus
@@ -374,7 +374,8 @@ dataset-driven driver.
   is model-specific to Inkling. General native coverage is pending PR #1792.
 - **Remote transport:** NCCL broadcast only, with PP1. P2P/RDMA and disk-delta
   reject LoRA.
-- **PPO:** shared actor/critic PPO is incompatible with general Bridge LoRA.
+- **PPO:** shared actor/critic PPO with Bridge LoRA is untested; the critic never
+  gets adapters.
 - **Resume:** miles adapter shards are resumable with the matching parallel
   topology. Direct HF PEFT import into the Bridge model is not yet implemented;
   native Inkling has a custom importer.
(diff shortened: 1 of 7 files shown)


Zhichenzzz (approved):

yushengsu-thu (approved):

yueming-yuan:
/rerun-test tests/e2e/megatron/model_scripts/test_inkling_small_4layer_lora_ci.py

PR #2882, opened 2026-09-02
feat(eval): support snapshot eval during pure SFT

Shi-Dong:
## Summary

- allow `--debug-train-only` jobs to reserve dedicated snapshot-eval GPUs
- start eval engines and dispatch exported snapshots without starting training rollout or session services
- initialize the shared HTTP client for external snapshot evaluators even when no rollout GPUs are reserved
- document that configured snapshot evaluation is the exception to train-only's usual no-engine behavior

## Why

Miles already has an asynchronous snapshot-eval dispatcher, but the pure-SFT path exits early under `--debug-train-only`: it drops eval GPU bundles, skips server and HTTP-client initialization, rejects snapshot-eval arguments, and suppresses evaluation generation. Consequently, SFT cannot use the existing dispatcher.

This PR connects pure SFT to that machinery. It does not change the async scheduler itself and contains no benchmark-specific evaluator code.

## Testing

- `pytest -q tests/fast/ray/rollout/test_rollout_server.py tests/fast/ray/test_placement_group_shared_ppo.py tests/fast/rollout/test_checkpoint_eval.py tests/fast/utils/test_arguments.py` (`206 passed`)
- `uvx ruff check` on all changed Python files (`All checks passed`)
- exercised end to end in a multi-node SFT smoke run with concurrent snapshot evaluation


Shi-Dong:
@claude review always

PR #3079, opened 2026-09-02
fix(offload): give each training role its own disk-offload directory

Michaelsqj:
### Problem

Under shared actor/critic PPO the critic reuses the actor's placement group, so rank *r* of both roles lands on the same GPU. `allocate_gpus_for_actor` derives the disk-offload directory from the cell and rank only:

```python
rank_dir = os.path.join(args.offload_train_disk_dir, f"cell{cell_index}_rank{rank}")
```

Both roles therefore receive the identical `TMS_DISK_BACKUP_DIR`, and two processes evict different models into one directory.

### Fix

Prefix the role for non-actor roles, leaving the actor's existing paths unchanged so nothing else has to move:

```
<dir>/cell0_rank16          # actor
<dir>/critic_cell0_rank16   # critic
```

### Scope

Two lines, `--offload-train-target disk` only. Independent of `--megatron-to-hf-mode`, LoRA, and #2878.

### Validation

GLM-5.2 744B LoRA PPO, four 8xH200 nodes, `--offload-train-target disk` for both roles. Before: one directory per rank shared by both roles. After: 8 actor and 8 critic directories coexist per node, each ~50 GB/rank, and the run completes rollout and the optimizer step.


Zhichenzzz:
@claude review

PR #3080, opened 2026-09-02
feat(critic): build the LoRA critic through the bridge helper

Michaelsqj:
Follow-up to #2878, targeting that branch. #2878 makes the critic **correct** under the bridge; this makes it **affordable**.

### Problem

`model.py` gates the bridge LoRA route on `role == "actor"`:

```python
is_lora_enabled(args) and role == "actor" and args.megatron_to_hf_mode == "bridge"
```

so a critic built through the bridge is full-parameter. At GLM-5.2's 744B that is roughly 279 GB of host optimizer state per rank, on top of an actor sharing the same GPUs. The model constructs correctly and then cannot be trained.

### Change

- Route `role in ("actor", "critic")` through `_setup_lora_model_via_bridge`.
- `is_value_model` also triggers on `role == "critic"`. `AutoBridge` only loads causal-LM architectures, so the existing `For*Classification` trigger is unreachable for any bridge-built model; the role is the only signal available.
- Register `_make_save_set_refresh_hook` after the value head exists. `apply_lora_hook` snapshots the PEFT save set before the pre-wrap hooks run, so without the refresh the freshly initialized head is silently dropped from the adapter checkpoint — training looks healthy and the saved adapter is unusable.

+20/-5 over #2878.

### Validation

GLM-5.2 744B LoRA PPO under `--megatron-to-hf-mode bridge`, four 8xH200 nodes, 32 GPUs, agentic terminal rollouts.

- Critic builds as a value model: **26,245,490,689** params/rank against the actor's **26,364,432,384**. The difference, 118,941,695, is exactly the vocab head (154880 x 6144 / 8) minus the scalar head (6145).
- Critic is LoRA-shaped, not full-parameter: its disk offload is **50 GB/rank, identical to the actor's**.
- Gradients flow: first optimizer step reports `critic-value_loss 3.8548`, `critic-grad_norm 46.709`, `critic-value_clipfrac 0.0`.

**Not yet validated:** that the value head reaches the adapter checkpoint on disk. That needs a checkpoint write, which lands at step 10 of the run currently in progress. I will confirm here before this is merged — the save-set refresh is precisely the part no training metric can prove.


yueming-yuan:
Reopened and retargeted to `main`: the base branch `yueming/critic-bridge` was deleted when #2878 squash-merged, which auto-closed this PR.

Because #2878 was squashed, this PR now lists its two original commits (`aed17ddadc`, `4e1fa78b97`) again and the diff replays them. @Michaelsqj could you rebase `feat/ppo-critic-lora-bridge` onto `main` and force-push? `edba3fd88d` cherry-picks onto current `main` with no conflicts (+20/-5, same two files), so the rebase should be a no-op resolution.

yueming-yuan:
Two important issues reported by claude, please check


**1. The value head does not reach a checkpoint miles can read back.**

`_make_save_set_refresh_hook` refreshes Megatron-Bridge's PEFT save set, which only feeds
`bridge.export_adapter_weights` and therefore the HF-format `adapter_model.bin`. miles' native
save in `save_lora_checkpoint` (`lora_utils.py`) selects tensors with `_is_adapter_param_name`,
which matches `lora_` / `.adapter.` names only, so `output_layer.weight` / `output_layer.bias`
never enter `adapter_megatron_rank*.pt`. On resume, `load_lora_adapter` reads the native file
only (the HF branch is marked "future work") and restores just the names present in it.

Net effect: a resumed critic gets its adapters and optimizer state back while the value head
is re-initialized at random, with no error and no metric that would flag it.

So the "not yet validated" item in the description cannot pass as written: even if the head
shows up in `adapter_model.bin`, the resume path never reads it. The native save/load should
select the trainable set (`requires_grad`, or the bridge save set) instead of a name pattern.

**2. Actor and critic share one `--lora-adapter-path`.**

`critic_args` is a `deepcopy(args)` and there is no `--critic-lora-adapter-path`, so
`checkpoint.py` (`if is_lora_enabled(args): adapter_path = args.lora_adapter_path`) applies the
same path to both roles. Now that the critic is LoRA-shaped its parameter names match the
actor's, so passing the actor's adapter silently loads it into the critic. Conversely there is
no way to point the critic at its own saved adapter for resume. At minimum
`critic_args.lora_adapter_path` should be cleared with an assert; the complete fix is a
`--critic-lora-adapter-path`.

yueming-yuan (changes requested):
(please see comment)

Zhichenzzz:
@claude review

PR #2711, opened 2026-08-22
test(e2e): verify multi-turn Anthropic sessions

guapisolo:
## Summary
Verify six-request Anthropic sessions and pin SGLang conversion behavior in Miles CI.

## Motivation
Six-request tool/result/text histories, reasoning replay, and supported intermediate-system insertion must preserve TITO prefixes. Incomplete Nemotron and GLM-4.7 responses must retry without advancing invalid history. SGLang PR [#35127](https://github.com/sgl-project/sglang/pull/35127) refactors the runtime-independent Anthropic conversion seams, but its external-PR policy does not run the owning CPU suites, so Miles needs a temporary paired-checkout gate for behavior preservation.

## Usage
```bash
python3 tests/e2e/sglang/test_session_server_multi_role/test_qwen36.py
SGLANG_SOURCE_ROOT=/path/to/sglang/python python3 -m pytest -q tests/fast/router/test_sglang_anthropic_conversion.py
```

## Design Notes
- **Mental model:** Each per-model test drives OpenAI v1/v2 and Anthropic v2 through the session tree.
- **Verification gates:** The verifier admits only a valid six-request leaf, then independently gates hard token drift and model/endpoint-specific assistant serialization drift.
- **CPU ownership:** Miles runs SGLang's conversion tests from `SGLANG_SOURCE_ROOT` pinned by `ci-sglang-pr: #35127` at `e95eae48`.
- **Retries:** Invalid responses retry from the same parent.
- **Intermediate system:** Insertion requires both live-route and template support.
- **MiniMax:** The test skips Anthropic and rejects direct agent use.
- **Port isolation:** Auto allocation selects each worker port independently; an explicit `--session-server-port` still anchors a consecutive range.
- **Formatting boundary:** Qwen3/Qwen3.5/Qwen3.6 keep Anthropic-only `1.0` assistant-text thresholds for conversion-normalized delimiters; OpenAI thresholds and all hard mismatch gates remain unchanged.

## Verification
- Exact-image focused tests on `miles-h200-dev-2`: `38 passed` for session verifier routing on `b9bea31d`, then `12 passed` for router allocation and `146 passed` for argument parsing on `713df081c`; changed-file pre-commit passed.
- Qwen3 full v1/v2/Anthropic E2E on two H200s: two consecutive runs passed.
[...]

PR #3082, opened 2026-09-02
test: declare the Anthropic intermediate-system expectation for qwen38small

yueming-yuan:
## Motivation

`tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py` fails before its Anthropic pass:

```text
ValueError: Anthropic per-model verification requires an intermediate-system expectation
```

#2711 made `anthropic_intermediate_system_expectation` mandatory whenever `verify_anthropic` is on (the default). #2760 added this config without it, and #2760 merged without any e2e label, so the test never ran. Every sibling config declares one (`required` for qwen3 / nemotron3 / glm47 / inkling, `forbidden` for qwen35 / qwen36 / dsv4, minimax turns `verify_anthropic` off).

## Modifications

Add `anthropic_intermediate_system_expectation="forbidden"` to the qwen38small `ModelConfig`. The agent derives the live capability as `route_supports and "system" in fixed_template_append_roles(tito_model)`; `Qwen38SmallTITOTokenizer` appends only `tool/user/assistant`, the same as Qwen3.5 / Qwen3.6, whose tests declare `forbidden`.

## Validation

Not executed here: the test needs `run-ci-sglang` (2× H200). First observed on radixark/miles#2756 (stage-c-2-gpu-h200 (0)), where the training job succeeded and only this harness check raised.

### DIFF
--- a/tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py
+++ b/tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py
@@ -17,6 +17,7 @@ CONFIG = ModelConfig(
     mamba_full_memory_ratio=4.59,
     cycles=2,
     tool_call_failure_mode="append_tool",
+    anthropic_intermediate_system_expectation="forbidden",
 )
 
 


guapisolo:
/rerun-test tests/e2e/sglang/test_session_server_multi_role/test_qwen38small.py

PR #3085, opened 2026-09-02
perf(p2p): make PD publication destination-aware and add DSV4 param mappings

ilyasher-harmonic:
#### Summary
- derive the P2P transfer plan from the actual update-enabled rollout-engine topology
- add the DeepSeek-V4 parameter mappings needed by the CPU staging replica
- propagate background P2P transfer failures instead of logging and discarding them

#### Commit structure

The changes are intentionally split into two commits:

1. Destination-aware P2P planning and DeepSeek-V4 staging support.
2. Transfer-error propagation. This is isolated because it overlaps conceptually with #1732 and can be dropped independently.

This change lowered our update_weights time from 30s to 7s for DSV4-Flash-FP8


guapisolo:
This PR is currently blocked by Yueming's weight update refactor Part 2. And the ETA is next week.

PR #3087, opened 2026-09-02
feat: expand GLM 53 TITO model support

guapisolo:
## Summary

Add `glm53` TITO support for GLM-5.3 and GLM-5.3-Flash text sessions.

## Motivation

The verified [GLM-5.3 template](https://huggingface.co/zai-org/GLM-5.3/blob/aca966e4e02791568aa6a4ced368624b3d897f42/chat_template.jinja) starts generation with `<think>` even when `enable_thinking=False`, so it needs a model-specific rendering contract. GLM-5.1 is intentionally unsupported: its native template is incompatible with the `glm47` template contract and would require separate repair and verification.

## Usage

Use `--tito-model glm53` for GLM-5.3 and GLM-5.3-Flash text sessions. GLM-4.7, GLM-5, and GLM-5.2 continue to use `glm47`; Flash multimodal inputs are out of scope.

## Design Notes

- **Mental model:** `glm53` uses each checkpoint's native template, pins fixed kwargs (`clear_thinking=False`, `enable_thinking=True`), then reuses the established GLM token-boundary and parser behavior.
- No bundled template, `glm5_next` selector, or unverified `glm51` selector is added.

## Verification

- GLM-5.3 FP8 `aca966e4`: Stage 2 passed `14/14` plus the actual-TITO assistant-append check; Stage 3 passed `4/4` on 8×H200 with every TITO mismatch metric at `0.0`.
- GLM-5.3-Flash FP8 text passed the same Stage 3 gate on feature revision `2b9b190`.
- `/run-ci-sglang` on head `63ffe2d6`: 22 checks passed, 9 expected skips, and no failures or pending checks.

## Review Focus

- `GLM53TITOTokenizer`: native-template kwargs versus inherited GLM token boundaries and parsers.
- GLM-5.3-Flash remains text-only.


guapisolo:
@claude always

PR #3091, opened 2026-09-03
feat(megatron): reset optimizer state by optimizer class

Zhichenzzz:
Corrected version of #2769, whose commit is kept as the first one here. Three things stop #2769 from working: its `dist_muon` branch can never run, because Megatron's `validate_args` renames the optimizer to `muon` before training starts; it rejects `HybridDeviceOptimizer` at any offload fraction below 1.0 (the AMD DSv4 recipe uses 0.75), which the loop it replaced handled; and its new top-level imports break `test_model_initialize.py` under that file's stub, taking stage-a-cpu and every GPU stage gated on it down.

The reset now dispatches on each leaf optimizer's class and walks `chained_optimizers` / `sub_optimizers` / `.optimizer` recursively, so nested `ChainedOptimizer`s and the GPU child of a partially offloaded optimizer are reached too. The key allowlists stay: `master_param` is the fp32 weights and must not be zeroed.

Verified on an H200 devbox: 13 fast tests pass with the real TE and Muon classes; an A/B against #2769's module raises `NotImplementedError` / `TypeError` where this branch resets; Qwen3-4B end to end with `--optimizer dist_muon` and with `adam --optimizer-cpu-offload --optimizer-offload-fraction 0.75`, both under `--reset-optimizer-states`, shows every moment tensor nonzero before the reset and zero after, and both jobs succeeded.

Follow-up, not here: share `_zero_step` with `multi_lora_utils.zero_optimizer_state_for_adapter`.

### DIFF
--- a/miles/backends/megatron_utils/model.py
+++ b/miles/backends/megatron_utils/model.py
@@ -28,6 +28,7 @@ from megatron.training.training import get_model
 from miles.backends.megatron_utils.ft.indep_dp import allreduce_grads_and_losses_across_replicas
 from miles.backends.megatron_utils.ft.types import TrainStepOutcome
 from miles.backends.megatron_utils.local_weight_checksum import dump_local_weight_checksums
+from miles.backends.megatron_utils.optimizer_state_reset import reset_optimizer_states
 from miles.utils.audit_utils.witness.allocator import WitnessInfo
 from miles.utils.audit_utils.witness.module import witness_dump_and_clear_stale
 from miles.utils.dumper_utils import DumperMegatronUtil, DumperPhase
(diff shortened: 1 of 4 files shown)


guapisolo:
@codex review

yueming-yuan (approved):
LGTM in human review

yueming-yuan:
codex reviews:
[P1] Avoid importing the Muon-only module on every Megatron backend — miles/backends/megatron_utils/model.py:31
The supported NPU setup pins Megatron-LM commit 3714d81d, which has no megatron.core.optimizer.emerging_optimizers module, and the NPU patches do not add it. This unconditional import therefore makes model.py raise ModuleNotFoundError even when --reset-optimizer-states is disabled, breaking every Megatron NPU run. Load the reset helper lazily or make its Muon import version-compatible.
[P2] Reset Adaptive Muon’s second-moment buffer — miles/backends/megatron_utils/optimizer_state_reset.py:35
TensorParallelAdaptiveMuon subclasses TensorParallelMuon, so this branch assigns it only the momentum_buffer schema. Adaptive Muon initializes an additional moment2_buffer; _reset consequently rejects that key. A valid --optimizer adaptive_muon --reset-optimizer-states run crashes once optimizer state exists instead of clearing it. Match the adaptive subclass first and include moment2_buffer.
Overall, the Adam and plain-Muon traversal appears sound, but the NPU compatibility regression blocks merging. CPU checks passed; all GPU checks were skipped. Tests omit Adaptive Muon and the pinned NPU Megatron import environment.

PR #3096, opened 2026-09-03
fix(qwen3.6): annotate packed QKV projection for Muon

Shi-Dong:
## Summary

- mark Qwen3.5/Qwen3.6 GatedDeltaNet's packed QKV projection for Muon
- attach parameter-specific Q, K, and V split sizes because this projection does not use Megatron's standard `linear_qkv` name
- add a CPU regression test for the optimizer metadata

## Why

Megatron automatically recognizes standard `linear_qkv.weight` parameters, but this model uses `linear_attn.in_proj_qkv.weight`. Without explicit metadata, Muon orthogonalizes the packed projection as one matrix instead of splitting Q, K, and V first.

The attributes are ignored by other optimizers, so this does not change Adam training.

## Testing

- `pre-commit run --files miles_plugins/models/qwen3_5.py tests/fast/optimizers/test_qwen3_5_muon_metadata.py`
- `pytest -q --confcutdir=tests/fast/optimizers tests/fast/optimizers/test_qwen3_5_muon_metadata.py`

### DIFF
--- a/miles_plugins/models/qwen3_5.py
+++ b/miles_plugins/models/qwen3_5.py
@@ -67,6 +67,11 @@ class Qwen3_5GatedDeltaNet(nn.Module):
         projection_size_qkv = self.key_dim * 2 + self.value_dim
         projection_size_z = self.value_dim
         self.in_proj_qkv = nn.Linear(self.hidden_size, projection_size_qkv, bias=False)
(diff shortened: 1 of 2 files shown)


Shi-Dong:
@claude review always

PR #3100, opened 2026-09-03
Support --colocate-memory-peak-device gpu under LoRA

yueming-yuan:
`--colocate-memory-peak-device gpu` (#2781) refused LoRA. gpu mode resumes the engine weights before the trainer sleeps, and the trainer has to give back what it can first: full-parameter runs only need `clear_memory()`, while LoRA also holds its grad buffers in the no-backup `grad_buffer` region that `patch_param_grad_buffer_for_colocate_mode_lora` creates. This adds that step and lifts the restriction.

- `sleep()` under LoRA now pauses `grad_buffer` alongside `default`, and `wake_up()` resumes both. Until now the LoRA sleep left the grad buffers on the GPU for the whole rollout.
- `offload_grad_buffer()` pauses the region ahead of `sleep()`; `train.py` calls it in gpu mode between `clear_memory()` and the engine weight resume. `sleep()` skips the region it already paused.

Split out of the Kimi K3 work (#1922, stacked on #1825, is superseded by #2781 plus this PR); K3's full-model runs on GB300 need gpu mode because the engine weight mirror and the trainer backup do not fit host RAM together.

## Validation

- `tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py`: the tag sequence of a LoRA sleep/wake, and that `offload_grad_buffer()` followed by `sleep()` pauses `grad_buffer` once.
- Qwen3.5-35B-A3B LoRA colocate smoke on 8xH200 (`radixark/miles:dev-202609031229`, TP2/EP8, rank 16), 3 rollouts, once per `--colocate-memory-peak-device` mode, node host memory (`MemTotal - MemAvailable`, the quantity `log_cpu_memory` reports) sampled every 0.5 s through the trainer-sleep / engine-resume handoff. Steady state during training is 230 GB in both modes (engine host mirror resident). cpu mode: host climbs to 324 GB (+95 GB, the trainer backup written while the engine mirror is still there), then the engine resume releases the mirror. gpu mode: the engine resume drops host to 161 GB first (-69 GB), then the trainer backup brings it to 257 GB (+28 GB net). Handoff peak 324 GB vs 257 GB, three cycles each within 1 GB; the 67 GB difference matches the 35B bf16 mirror (70 GB). `offload_grad_buffer()` itself frees 0.16 GB of GPU per rank (device used 29.01 -> 28.85 GB, `print_memory` around the pause); that is the LoRA grad buffer (24.8M adapter params) and is small by construction. `tests/fast/ray/train` + `tests/fast/ray/test_actor_factory.py`: 120 passed.
- Not executed: a host-RAM-bound run where gpu mode is the difference between fitting and OOM (that is the GB300 K3 recipe).


Zhichenzzz (approved):

xyuzh:
Would this impact multi-lora?
`grad_buffer` pause with no-backup could unmap DDP `main_grad` storage

yueming-yuan:
> Would this impact multi-lora? `grad_buffer` pause with no-backup could unmap DDP `main_grad` storage

It would not, this does not touch the multi-lora path

maocheng23 (approved):

Issue #543, opened 2026-02-02
Sync arguments help string and argument documentation


zhaochenyang20:
https://github.com/radixark/miles/pull/517#issuecomment-3833058685

PR #3103, opened 2026-09-04
test(docs): guard cli-reference.md against undocumented train flags

bo-lyu:
## What

Adds a CPU test that keeps `docs/user-guide/cli-reference.md` in step with the
CLI flags `arguments.py` actually defines, plus a baseline file listing the flags
undocumented today.

- `tests/fast/utils/test_cli_reference_sync.py` — AST-scans the
  `add_argument("--...")` calls in `miles/utils/arguments.py` and
  `miles/dashboard/args.py` and diffs them against the flags named in
  `cli-reference.md`.
- `tests/fast/utils/cli_reference_undocumented.txt` — the current backlog (248
  flags). Built to shrink, in the spirit of `_KNOWN_ORPHANS` in
  `tests/ci/test/test_ci_discovery_coverage.py`: document a flag, delete its
  line; a second assertion fails if a line goes stale (flag renamed, removed, or
  since documented).

## Why

`arguments.py` defines ~339 `train.py` / `train_async.py` flags; only ~91 appear
in `cli-reference.md`, and nothing keeps the two in step, so a new flag lands
undocumented and stays invisible until someone greps the source.

This is the "sync" half of #543 (the follow-up deferred from #517: *"I am
thinking how to do a high level sync/auto-generation with arguments.py and this
document"*), enforced rather than generated. It does not write the missing
entries, but it stops the gap growing and turns "finish the document" into an
incremental, CI-enforced task.

Relates to #543 

## Scope

Out of scope, and excluded: Megatron's inherited flags (the doc covers those by
reference) and the `--sglang-*` / `--eval-sglang-*` passthrough families (covered
as families). The check reads source and Markdown only — it never imports
`miles` — so it runs in `stage-a-cpu` with no training dependencies.

## Checklist

- [x] `pre-commit run --all-files` clean on the changed files (black, isort, ruff,
  autoflake, the Miles bans).
- [x] `pytest tests/fast/utils` passes in `radixark/miles:latest` — 3235 passed,
  11 skipped; the new file is 3 passed.
- [x] New behavior has a test, auto-registered in `stage-a-cpu` (verified with
  `run_suite.py --list-only`).
- [x] Adds no flags; no `doc-dev:` governed file touched.


behroozazarkhalili:
Ran this on top of current main (2ed189c0d); the branch merges cleanly but test_no_new_undocumented_flag and test_baseline_has_no_stale_entry fail because the baseline is behind main: --control-server-port was renamed in #1971 to --api-server-host / --api-server-port, and --ci-metric-checker-expect-num is new.

I opened https://github.com/bo-lyu/miles/pull/1 against your branch with the 4-line baseline refresh; with it all 3 tests pass and the rest of tests/fast/utils is 4840 passed. ruff/black/isort are clean and run_suite --list-only picks the file up in stage-a-cpu.

Since every flag rename on main re-reds the branch until merge, a regeneration one-liner in the file header (or a --regen mode on the test) would help whoever merges. One note on the matcher: _documented_flags regexes the whole markdown, so a flag mentioned in prose counts as documented without a table row; fine for a backlog guard, worth a line in the docstring.

cc @fzyzcjy @Ying1123

bo-lyu:
Thanks behroozazarkhalili@:
- merged your 4 line change to re-base on the head, 
- added a `--regen` mode (python tests/fast/utils/test_cli_reference_sync.py --regen) that re-snapshots the baseline in one pass
-  noted in the docstring that the matcher scans the whole markdown

behroozazarkhalili:
Re-ran the new head (25f2bbbaa merged onto current main 2f9bb0248, no conflicts): `tests/fast/utils/test_cli_reference_sync.py` plus `test_arguments.py`, 186 passed, and `--regen` rewrites `cli_reference_undocumented.txt` to an identical file, so the baseline is current again. Nothing left from my side.

PR #3104, opened 2026-09-04
[multi-lora] Push adapter versions without pausing the rollout engines

yushengsu-thu:
Implements the review "Optimization direction 1 (Async LoRA Rollout)" on #2846, rebuilt on the rewritten head (`06716481e`, on `main`, streamed adapter push): a `save_weights_for_sampler` push no longer stops the rollout fleet.

## What changes

- `WeightUpdater._sync` (`miles/backends/training_utils/weight_update/updater.py`): `pause_engines` / `resume_engines` run only for a base sync (`sync_base=True`). An adapter push (`push_adapter`, `sync_base=False`) keeps the session frame the streamed API needs — `register_lora_adapter → begin_weight_update(sync_base=False) → stream → end_weight_update` — but the engines keep serving while it runs.
- `TinkerService` save_weights_for_sampler (`miles/tinker/core/service.py`): `sampler_version` advances only after every engine applied the push.
- New fast test `TestAdapterOnlyPush` pins the adapter-only frame: register, begin(sync_base=False), end on every engine; no pause, flush, resume or version bump.

## Why it is safe

The adapter lands under a fresh `{model_id}@{version}`. The engine registers that name (zero weights, new `lora_id`), stashes the streamed tensors, and applies them at `end_weight_update`; SGLang keys its radix cache by `lora_id` (`schedule_batch.py`: `extra_key += lora_id`), so no KV entry of any other adapter, or of an older version of this one, can ever be read with the new weights. The pause/flush only ever protected in-place base writes, and `update_weights` keeps it.

Publishing the version *before* the push (the previous order) is what made the in-place apply dangerous: between `register_lora_adapter` and `end_weight_update` a client could already resolve `sampler_weights/<v>`, sample the zero-weight adapter, and leave KV under its `lora_id` that the stash then invalidates. With the version published after the push that window does not exist.

## What this does not yet give you

On the engine side an unpaused session RPC takes `model_update_lock.writer_lock`, and `generate_request` holds the reader lock for a request's whole lifetime (`tokenizer_manager.py:823-837`, through `_wait_one_response`); the lock is writer-preferring (`aio_rwlock.py`). So with this branch alone each `begin` / bucket / `end` RPC waits for the engine's in-flight requests to finish and holds back new admissions while it waits. Other tenants' requests are no longer retracted, flushed and re-prefilled, but the publish itself still serializes behind them: in the probe the `B@1` push took 0.55 s starting 1.0 s into a 1.5 s batch, and `A@2` pushed during a full `B@1` batch took 1.55 s, versus 0.17 s when paused. Real concurrency needs a sglang-miles change: a session path for fresh, unpublished adapter names that skips the writer lock (still serialized against other sessions, rejecting base tensors and overwrites of published names). The miles-side gate should then be "new unpublished name", not `sync_base=False` alone.

## Verification

- `tests/fast/backends/training_utils/weight_update` + `tests/fast/tinker`: 167 passed (pod, image sglang 0.5.19.dev56, i.e. the streamed LoRA API this head targets).
- GPU probe, Qwen3-0.6B, 2 trainer GPUs (TP1, DP2) + 2 SGLang engines (1 GPU each), rank 8, 4 slots. Tenant A trains and publishes `A@1`; 8 concurrent 512-token samples run on `A@1` while tenant B publishes `B@1`; then A publishes `A@2` while `B@1` samples are in flight. Same probe against this branch and against the unmodified head `06716481e`.

| | this branch | head `06716481e` |
|---|---|---|
| `B@1` push while 8×512-token samples in flight on `A@1` | 0.55 s | 0.17 s (engines paused) |
| `A@1` sample latency, max, concurrent push / idle | 1.49 s / 1.49 s (1.00×) | 1.76 s / 1.49 s (1.18×) |
| `pause_generation` → `flush_cache` … `continue_generation` on the engines | 0 | one per engine per push |
| `register_lora_adapter` → `begin_weight_update(sync_base=False)` → stream → `end_weight_update` | one per engine per push | one per engine per push |
| `B@1`, `A@1`, `A@2`, `A@1` after `A@2` greedy samples reproduce their tenant's marker | 4/4 | 4/4 |

Functionally the two runs are identical. The push itself takes longer on this branch because the engines keep serving while they receive, and in exchange the samples in flight on another adapter are no longer retracted, flushed and re-prefilled on every push (0.27 s per push on this tiny model; the cost scales with everything the fleet has in flight).

## Notes

- Not included on purpose: retiring old versions (SGLang's unload drains under `lora_update_lock`, so an unload racing a push would stall the collective fleet-wide; needs a fork change first) and try/finally around the push (a slot-command exception already kills the trainer cell, so there is nothing to resume).


yushengsu-thu:
CI at 2312b4988 (rebuilt on the rewritten #2846 head 06716481e): pre-commit green. The five CPU failures are the same ones the base head shows on its own run — stage-a-cpu (3) collection `ImportError: cannot import name 'AdapterRef' from 'miles.utils.types'` in `test_prefill_logprobs.py`, (0) `test_sample_utils::TestMergeSamples::test_basic_merge`, (1) `test_session_samples_op_v2::test_assembled_samples_golden_merged`, (2) `test_group_reconcile_adapters`, stage-b `test_multi_turn::test_two_turns_with_tool_call`. None of them touch the three files here; the weight_update and tinker suites pass (167 tests in the PR environment).

yueming-yuan:
the new multi lora has been rewritten to the refactored weight update, so we do not need the PR in the new version

PR #3114, opened 2026-09-04
Tolerate missing optional SGLang Anthropic helpers

Shi-Dong:
## Summary

- Keep the Miles session server importable when optional SGLang Anthropic helper modules are unavailable.
- Leave OpenAI-compatible and TITO session routes usable.
- Return HTTP 501 from the Anthropic Messages route when that adapter is unavailable.

## Why

SGLang installations can expose the core OpenAI-compatible serving APIs without including every optional Anthropic adapter helper. Miles currently imports those optional helpers when the session server module loads, so a missing helper prevents the entire server from starting—even when the caller only needs an unrelated API. Detecting the optional capability at runtime isolates that limitation to the Anthropic route and keeps the rest of the session interface usable.

## Verification

- Full `tests/fast/router/test_session_anthropic.py`: 46 passed, 1 skipped.
- Focused missing-module probes: 4 passed.
- Ruff check: passed.
- `git diff --check`: passed.

### DIFF
--- a/miles/rollout/session/anthropic_adapter.py
+++ b/miles/rollout/session/anthropic_adapter.py
@@ -3,9 +3,13 @@
 import json
 
 from pydantic import ValidationError
-from sglang.srt.entrypoints.anthropic import utils as anthropic_utils
 from sglang.srt.entrypoints.anthropic.protocol import AnthropicMessagesRequest, is_server_tool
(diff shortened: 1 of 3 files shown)


Shi-Dong:
@claude review always

guapisolo (approved):

PR #3115, opened 2026-09-05
Scan full responses for repetition metrics

Shi-Dong:
## Summary

- scan the full response for repetition using overlapping 10,000-character windows with a 5,000-character stride
- preserve coverage of the exact final 10,000-character suffix when it is not stride-aligned
- add focused tests for middle-response repetition, boundary lengths, suffix coverage, and the existing strict compression threshold

## Why

`rollout/repetition_frac` currently checks only the final 10,000 characters of each response. A long response can contain severe repetition earlier and end with non-repetitive text, causing the metric to miss the behavior. Scanning bounded overlapping windows makes the metric represent repetition anywhere in the response while retaining short-circuit evaluation once repetition is found.

## Testing

- `uv run --active --no-sync pytest -q --confcutdir=tests/fast/utils tests/fast/utils/test_metric_utils.py` — 5 passed
- `uv run --active --no-sync pre-commit run --files miles/utils/metric_utils.py tests/fast/utils/test_metric_utils.py` — passed

### DIFF
--- a/miles/utils/metric_utils.py
+++ b/miles/utils/metric_utils.py
@@ -1,7 +1,13 @@
 import math
+from collections.abc import Iterator
 from typing import Any, Literal
 
 import numpy as np
+
+
+REPETITION_WINDOW_SIZE_CHARS = 10_000
+REPETITION_WINDOW_STRIDE_CHARS = 5_000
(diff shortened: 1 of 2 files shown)


Shi-Dong:
@claude review always

Zhichenzzz (approved):

PR #3127, opened 2026-09-06
[LoRA] Publish immutable adapter versions alongside generation

yushengsu-thu:
ci-sglang-pr: #38165

> Scope after the split: this PR keeps the engine-facing weight-update mechanics, rebased onto #2843 (versioned adapters). The gateway half (disk commit point, request-carried backfill, tinker tests) moved into #2844/#2845/#2849.

## Motivation and Modification

- Scenario: `lora_a`, `lora_b`, and `lora_c` are separate LoRA adapters sharing one base model. A has finished rollout and training; its updated weights will become `lora_a@2`. `lora_b`, and `lora_c` are still generating.

[Miles PR](https://github.com/radixark/miles/pull/3127) — skip engine-wide interruption
Most important file: [updater.py](https://github.com/radixark/miles/blob/44ed544316f8a4e0b6ceb9ba0a95318c75de78b6/miles/backends/training_utils/weight_update/updater.py#L203-L225).
Previously, Miles paused target engines before publication. The default retract flow also interrupted `lora_b`, and `lora_c` and cleared KV cache.
The PR creates a session for fresh versions and skips the pause/resume sequence. Relevant excerpts from three functions:

```python
# _sync()
session_id = adapters[0][0] if new_version else None

# _prepare_sync()
if session_id is None:
    pause_engines(self.args, engines)

self._register_new_lora_adapters(
    engines, adapters, defer_publish=session_id is not None
)

# _finish_sync()
if session_id is None:
    resume_engines(engines)
```

Miles verifies deferred registration, then transfers weights through begin → send_bucket → end. The new sampler version becomes available only after all target engines report success.
Together, Miles removes the outer pause, while SGLang allows publication to coexist with generation and prevents incomplete weights from being sampled. `lora_b`, and `lora_c` can continue rollout, although transfer and loading may still affect performance.

- Need this sglang pr: https://github.com/sgl-project/sglang/pull/38165

ci-sglang-pr: 7fef95661fbaa911c19fa9533dcc8cea454eb232


yushengsu-thu:
@claude review

PR #3128, opened 2026-09-06
fix: use host weights when syncing an offloaded actor

Shi-Dong:
## Problem

Weight synchronization can run while a trainer is offloaded, including when trainers and rollout engines use separate GPUs. The active model tag can still be `actor` even though its GPU storage is no longer resident. Selecting weights using only that tag and colocation can therefore hand inaccessible GPU tensors to the weight updater.

There is a second part to the same issue: without a reference model, teacher, old actor, or colocation, the host weight backup can be disabled entirely. Offloading still needs a readable source for the next weight synchronization.

## When this can occur

- Trainer offloading is enabled (`--offload-train`), trainers and rollout engines are not colocated (`--colocate` is not set), and weight synchronization runs while the trainer is asleep.
- The active model tag is still `actor`, but its GPU storage has been offloaded. This can happen when actor and critic workers time-share training GPUs while rollout engines run on separate GPUs; it is a worker-lifecycle condition, not specific to an RL algorithm or to asynchronous training.
- If no reference model, teacher, or old-actor snapshot is configured, there may also be no retained host backup before this fix.

Resident trainers continue to read live GPU weights. The colocated path already selects host-backed weights and is not the missing-asleep-check case described above.

## Changes

- Read the actor's host backup whenever the trainer is asleep.
- Retain the host backup when trainer offloading is enabled, even without model switching.
- Preserve the existing memory-saver backup fast path and live-GPU path for resident trainers.
- Restore on every model switch, including when the requested tag is already active. With `--offload-train` miles sets `disable_param_buffers_cpu_backup`, so `sleep()` discards the param buffers' contents and `wake_up()` brings the storage back without values; the pinned backup is the only source, and the main-cast backuper already rebuilds the actor every cycle. Unknown-tag validation is preserved.
- Cost: every non-colocated `--offload-train` run now keeps a full pinned host copy of the actor per rank and pays one host-to-device restore before each training step. The leaner alternative (keep the memory-saver backup of the param buffers when no model switching is needed and read it through `translate_gpu_to_cpu=True` while asleep) is a follow-up option, not part of this change.
- Add `tests/e2e/megatron/test_qwen3_4B_offload_disaggregated.py`: four training GPUs beside four rollout GPUs without `--colocate`, the configuration no existing suite covers.
- Add regression coverage for asleep versus resident trainers and offloading with no reference model.

## Validation

Review revision: the 10 model-switch regression cases and the offload/asleep cases in `tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py` pass; the new offload-specific checks fail against unpatched `main` and pass with this fix.

H200 A/B (Qwen3-4B, 4 training GPUs + 4 rollout GPUs, `--offload-train --offload-train-target cpu`, no `--colocate`, `--ci-test --check-weight-update-equal`, 2 rollouts; each branch rebased onto current main):

| branch | result |
|---|---|
| `main` | first `update_weights` (actor asleep after init) dies in `connect_rollout_engines` → `dist.all_gather_object`: `CUDA error: invalid argument` (#3129's bug) |
| #3129 only | connection succeeds, then the transfer reads the paused param buffers: illegal memory access (this PR's bug) |
| this PR only | dies in `connect_rollout_engines` like `main` |
| this PR + #3129 | passes, with and without a reference model; the engines report every synced tensor equal |

Merge order: #3129 first, then this PR. The new e2e passes only once both have landed, so the GPU labels go on after #3129 is in.


Shi-Dong:
@claude review always

miles/backends/megatron_utils/actor.py
```diff
@@ -369,16 +369,26 @@ def _weight_sync_reads_tms_backup(self) -> bool:
...
 
     def _switch_model(self, target_tag: str) -> None:
         if not self._enable_weight_backup:
             return
         if target_tag not in self.weights_backuper.backup_tags:
             raise ValueError(f"Cannot switch to unknown model tag: {target_tag}")
+        if target_tag == self._active_model_tag and not self.weights_backuper.restore_required_when_active(target_tag):
+            # A value-copy restore of the already-active tag is a no-op; a
+            # rebuilding restore (main-cast) must still run every cycle.
+            return
```
guapisolo: codex cmt

[P1] Restore ordinary parameter buffers after offload even when the model tag is unchanged

With full-parameter training, `colocate=True`, `offload_train=True`, the normal backuper, and no ref/teacher/old_actor, this shortcut skips the necessary CPU-to-GPU restore after wake-up. The next forward/training step can then use unrestored parameter values.

`offload_train` sets `disable_param_buffers_cpu_backup=True` (`arguments.py:3421-3423`). Megatron consequently allocates `param_buffer` with TMS CPU backup disabled: `sleep()` discards its contents, and `wake_up()` reallocates that storage without restoring the values. Meanwhile, `_active_model_tag` remains `"actor"`. The normal backuper inherits `restore_required_when_active() == False`, so this branch returns instead of copying the retained host weights back. The MainCast override protects rematerialization mode, but an ordinary value-copy restore is also necessary when the destination's previous contents were discarded.

A focused H200 CUDA/TMS probe using the frozen base/head switch implementations and their real normal backupers confirmed this: after pause/resume, base restored all 262,144 parameter values; head left all 262,144 mismatched. This was an allocator/backuper mechanism check, not a full training run.

Please remove this shortcut and retain the existing restore behavior. The asleep-source selection and `offload_train` backup enablement can still fix the original synchronization issue independently. Regression coverage should check actual parameter values after pause/resume with the normal backuper, rather than only asserting whether `restore()` was called.

guapisolo: After human check, seems sensible 🤔

guapisolo (approved):

Zhichenzzz (approved):

PR #3129, opened 2026-09-06
fix: guard connection allocations during offloaded weight sync

Shi-Dong:
## Problem

Creating or reconnecting a weight-sync connection can allocate CUDA tensors, for example during NCCL object collectives. With trainer offloading enabled, those allocations can otherwise reuse memory-saver blocks whose GPU storage has already been unmapped, causing invalid memory accesses during connection setup.

The existing `torch_memory_saver.disable()` guard protects the subsequent weight transfer, but not the earlier connection setup. Protecting only the transfer is therefore insufficient.

## When this can occur

- Trainer offloading is enabled (`--offload-train`), and the trainer's memory-saver allocations have been paused/unmapped when `update_weights()` is called.
- The weight updater needs its first connection to rollout engines, or must reconnect after rollout-engine membership changes.
- Connection setup uses a path that allocates CUDA tensors, such as NCCL object collectives. Without the guard, those temporary allocations can reuse an offloaded allocator block.

An already-connected updater skips this setup path; its actual weight-transfer phase already has the allocation guard. Trainers without offloading do not have the unmapped-block condition and retain their existing behavior.

## Changes

- Run connection setup, connection-state bookkeeping, and its synchronization barrier inside the allocation guard when trainer offloading is enabled.
- Leave the non-offloaded path unchanged.
- Add coverage for successful and failing connections, both with and without offloading, including restoration of the allocation context after an exception.

## Validation

The focused suite passes on this branch; the new allocation-guard checks fail against unpatched `main` with offloading enabled and pass with this fix; the non-offloaded controls pass in both cases.

H200 A/B (Qwen3-4B, 4 training GPUs + 4 rollout GPUs, `--offload-train --offload-train-target cpu`, no `--colocate`, `--ci-test --check-weight-update-equal`, 2 rollouts; branches rebased onto current main): `main` dies at the first `update_weights` — the actor is asleep after init — inside `connect_rollout_engines` → `dist.all_gather_object` with `CUDA error: invalid argument`. With this PR the connection succeeds and the run proceeds to the transfer, which then needs #3128 (host weights while asleep); with both, the run passes and the engines report every synced tensor equal.

Merge order: this PR first, then #3128, which also adds a disaggregated `--offload-train` e2e (`tests/e2e/megatron/test_qwen3_4B_offload_disaggregated.py`) covering this configuration.


Shi-Dong:
@claude review always

guapisolo (approved):
Make sense. good fix. But the code placement of the disable() func need code owner double check.

Zhichenzzz (approved):

yueming-yuan:

yueming-yuan:

PR #3130, opened 2026-09-06
fix: isolate trainer disk-offload backups by role

Shi-Dong:
## Problem

Actor and critic trainer pools each start their cell and rank numbering at zero. When both use the same disk-offload root, a directory named only by cell and rank can therefore belong to two different workers.

Each worker reclaims stale backup files in its assigned directory at startup and exit. A newly started critic can consequently delete a paused actor's live backups, making the actor unable to restore its offloaded state. The reverse collision is also possible.

## When this can occur

- Both actor and critic trainer pools use disk offloading (`--offload-train --offload-train-target disk`).
- They share a disk-offload root on the same underlying filesystem and have overlapping cell/rank IDs, for example when both roles share a training node.
- One worker starts, restarts, or exits while the other still has live backup files in the formerly shared directory.

CPU-only offloading does not use these files. Identical path strings on separate, node-local filesystems do not by themselves cause a collision; both workers must see the same directory.

## Changes

- Include the trainer role in `TMS_DISK_BACKUP_DIR`: `<root>/<role>/cell<cell_index>_rank<rank>`.
- Pass the role through trainer environment construction.
- Add tests for separation between actor/critic roles, cells, and ranks.
- Update the disk-offload documentation and cleanup contract to match the directory layout.

This isolates workers within a job; independent jobs should still use separate disk-offload roots.

## Validation

The focused suite passes on this branch: **75 passed**.

The new actor/critic isolation check fails against unpatched `main` and passes with this fix.

```bash
python -m pytest -q -o addopts= \
  tests/fast/backends/megatron_utils/test_shared_ppo_lifecycle.py \
  tests/fast/ray/specs/test_train.py \
  tests/fast/ray/test_placement_group_shared_ppo.py
```

All pre-commit checks pass for the changed files. These tests validate worker environment generation and lifecycle behavior; they do not perform a full distributed offload/restore cycle.

### DIFF
--- a/docs/advanced/disk-offload.md
+++ b/docs/advanced/disk-offload.md
@@ -20,9 +20,10 @@ does not fit, and disk offload is the alternative.
 Instead of a pinned host copy, the paused actor is streamed to per-rank files through a
 fixed-size pinned staging buffer, so host memory stays bounded by
 `--offload-train-disk-chunk-mb` regardless of how much is offloaded. Each rank writes to
-its own directory under `--offload-train-disk-dir` (defaults to
-`$SCRATCH/miles_train_offload_<uid>`), the files are overwritten in place every step, and
-they are removed when the actor exits.
+its own `<role>/cell<cell_index>_rank<rank>` directory under `--offload-train-disk-dir` (defaults to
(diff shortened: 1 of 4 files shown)


Shi-Dong:
@claude review always

Issue #574, opened 2026-02-08
Training hangs indefinitely after 74 steps due to NCCL deadlock in weight conversion


euclidgame:
**Environment:**
- Miles image: latest
- Hardware: 8×H100 (80GB)
- Dataset: dapo-math-17k
- Model: unsloth/gpt-oss-20b-BF16

**Issue:**
Training completes 74 steps successfully, then hangs indefinitely during `update_weights`. No error messages appear—only sglang health check output. The hang persists for hours with no progress.

**Root Cause (confirmed via py-spy + Ray dashboard):**
All 8 MegatronTrainRayActor workers deadlock in `torch.distributed.all_gather` inside `gather_from_tp_ranks` during Megatron→HF weight conversion (mbridge mode). All actors remain ALIVE but blocked waiting on each other in the NCCL collective operation. SGLang engines remain healthy and are not involved.

**Configuration:**
- TP=8, EP=4, PP=1, CP=1
- Sequence parallel enabled
- `--megatron-to-hf-mode bridge`
- `--optimizer-cpu-offload` enabled
- `--qkv-format bshd` with `--attention-backend fused`

**Attempted mitigations:**
- Considering NCCL timeout variables (`NCCL_TIMEOUT=1800`, `TORCH_NCCL_ASYNC_ERROR_HANDLING=1`, `NCCL_DEBUG=WARN`) to surface diagnostics
- Considering `--use-fault-tolerance` flag (unsure if relevant to this deadlock)

**Full training script:**
```bash
#!/bin/bash

# for rerun the task
pkill -9 sglang
sleep 3
ray stop --force
pkill -9 ray
pkill -9 python
sleep 3
pkill -9 ray
pkill -9 python

set -ex

# will prevent ray from buffering stdout/stderr
export PYTHONBUFFERED=16
export HF_HOME=/workspace/hf_cache

# Load model architecture config
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
source "${SCRIPT_DIR}/models/gpt-oss-20b.sh"

CKPT_ARGS=(
   --hf-checkpoint /workspace/models/gpt-oss-20b-BF16
   --megatron-to-hf-mode bridge
   --save /workspace/checkpoints/gpt-oss-20b
   --save-interval 50
)

ROLLOUT_ARGS=(
   --prompt-data /workspace/datasets/dapo-math-17k/dapo-math-17k.jsonl
[...]

Issue #384, opened 2026-01-03
Add barrier to synchronize all ranks in reload_process_groups()?


jianguoz:
Hi, I face NCCL issues when fine-tuning qwen-30b-a3b on `two nodes after 1600 steps`.

```txt
(RolloutManager pid=2681781) [2026-01-01 10:29:20] rollout.py:601 - perf 1391

   Reloading 45 process groups in pid 2421204
   Reloading 45 process groups in pid 2696524
   [Rank 7] Memory-Usage after torch distributed error

Race condition in reloadable_process_group.py - not all 45 process groups across 16 GPUs were synchronized before the all_gather collective started.
Each rank reloads 45 process groups independently
new_group() is a collective operation, but there's no barrier after all groups are reloaded
Fast ranks can start all_gather while slower ranks are still reloading → NCCL fails
```

Potential fix: Add a barrier at the end of reload_process_groups().

```python
from miles.utils.distributed_utils import get_gloo_group

@staticmethod
def reload_process_groups():
    pid = os.getpid()
    reloadable_groups = ReloadableProcessGroup.GROUPS.get(pid, [])
    logger.info(f"Reloading {len(reloadable_groups)} process groups in pid {pid}")
    old_new_group = old_new_group_dict.get(pid)
    for reloadable_group in reloadable_groups:
        if reloadable_group.group is not None:
            continue
        group = old_new_group(ranks=reloadable_group.group_info["ranks"], backend="nccl")
        reloadable_group.group = group
    
    # Add barrier to sync all ranks after reload
    torch.cuda.synchronize()
    dist.barrier(group=get_gloo_group())
    logger.info(f"Process group reload complete and synchronized in pid {pid}")
```

> Note: Bug fixed in the new main branch 

 _Originally posted by @KennyYao2001 in [#354](https://github.com/radixark/miles/issues/354#issuecomment-3693713727)_

NJX-njx:
Created PR #705 which adds torch.cuda.synchronize() and dist.barrier(gloo) at the end of reload_process_groups() to ensure all ranks finish reloading before collective operations begin.

Issue #354, opened 2025-12-27
[Bug] KeyError in destroy_process_groups() when no process group was created (single-GPU)


KennyYao2001:
I'm running Qwen3-4B example on single H200 GPU instance, the script is adapted from [quick start guide](https://github.com/radixark/miles/blob/main/docs/en/get_started/quick_start.md)
I found there is KeyError when launch ray in a single GPU case. Although Miles is a distributed pipeline, some users may only have access to a single GPU and still want to try the example. In this single-GPU (world_size = 1) setup, I consistently encounter a KeyError during MegatronTrainRayActor initialization.

Config:
```
#!/bin/bash

# for rerun the task
pkill -9 sglang
sleep 3
ray stop --force
pkill -9 ray
pkill -9 python
sleep 3
pkill -9 ray
pkill -9 python

set -ex

# will prevent ray from buffering stdout/stderr
export PYTHONBUFFERED=16

NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
if [ "$NVLINK_COUNT" -gt 0 ]; then
    HAS_NVLINK=1
else
    HAS_NVLINK=0
fi
echo "HAS_NVLINK: $HAS_NVLINK (detected $NVLINK_COUNT NVLink references)"

SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
source "${SCRIPT_DIR}/models/qwen3-4B.sh"

CKPT_ARGS=(
   --hf-checkpoint /root/Qwen3-4B
   #--hf-checkpoint /root/Qwen3-4B-FP8
   --ref-load /root/Qwen3-4B_torch_dist
   --load /root/Qwen3-4B_miles/
   --save /root/Qwen3-4B_miles/
   --save-interval 20
)

ROLLOUT_ARGS=(
   --prompt-data /root/dapo-math-17k/dapo-math-17k.jsonl
   --input-key prompt
   --label-key label
   --apply-chat-template
   # Control if the model is thinking mode. we disabled think in example to avoid high truncation ratio and avoid reward hacking.
   # if you want to enable thinking, need to set --rollout-max-response-len to 16384
   --apply-chat-template-kwargs '{"enable_thinking":false}'
   --rollout-shuffle
   --rm-type deepscaler
   --num-rollout 3000
   --rollout-batch-size 32
   --n-samples-per-prompt 8
[...]

PR #3131, opened 2026-09-06
test(dist): cover the reloadable process-group destroy/reload cycle

David553:
## Why

`miles/utils/reloadable_process_group.py` is the machinery colocate runs on every rollout→train switch: `sleep()` destroys every wrapped process group, `wake_up()` rebuilds them. It has been behind several multi-node hangs (#384, #574; the same phase is reported in THUDM/slime#1487), and the fix in #1594 — replay the original `new_group` args on reload instead of a hardcoded `(ranks, backend="nccl")` — landed without a regression test. There is currently no `tests/fast/utils/test_reloadable_process_group.py`.

Before #1594 the rebuild silently dropped `timeout`, `pg_options` and `group_desc`. In practice that meant `--distributed-timeout-minutes 120` stopped applying after the first offload cycle and every rebuilt group fell back to the 10-minute default — which is exactly the "ignores the timeout flag, dies at 600 s" shape those reports describe.

## What

Adds `tests/fast/utils/test_reloadable_process_group.py` (CPU suite, auto-registered in `stage-a-cpu`, no GPU or `run-ci-*` label needed):

- `test_reload_replays_new_group_args` — a group created with `timeout=` and `group_desc=` is wrapped, `inner_kwargs` carries both, and a collective still works after `destroy_process_groups()` + `reload_process_groups()`.
- `test_gloo_group_is_not_wrapped` — explicit gloo (keyword or positional) is returned unwrapped.
- `test_single_rank_group_is_not_wrapped` — one-rank groups are returned unwrapped.
- `TestCleanupWithoutGroups` — `destroy_process_groups()` / `reload_process_groups()` are no-ops in a process that never registered a group (#354).

It runs on CPU because the monkey patch only skips groups whose *arguments* name gloo; a group created with the default backend on a gloo world is wrapped like any NCCL group. That also makes the pre-#1594 behaviour a hard failure rather than a silent one: the rebuilt group claims an NCCL backend and the next collective raises `No backend type associated with device type cpu`.

## Verification

Same file run against both versions of `reloadable_process_group.py`, torch 2.11, gloo, 2 processes:

| `reloadable_process_group.py` | result |
|---|---|
| `main` (with #1594) | `5 passed` |
| pre-#1594 | `1 failed, 4 passed` — `test_reload_replays_new_group_args`, at `assert group.inner_kwargs["timeout"] == _TIMEOUT` |

`pre-commit` hooks (ruff 0.14.7, autoflake, isort 5.13.2, black 24.3.0 @ 119) leave the file unchanged.

Not touched here: the reload path still has no synchronization point after the rebuild loop; that is #705's proposal and is left to that discussion.

### DIFF
--- a/tests/fast/utils/test_reloadable_process_group.py
+++ b/tests/fast/utils/test_reloadable_process_group.py
@@ -1,0 +1,93 @@
+"""Tests for reloadable_process_group: the destroy/reload cycle colocate runs on every step.
+
+The reload path has caused repeated multi-node hangs (#384, #574), and the fix that made
+reload replay the original ``new_group`` contract (#1594) landed without a regression test.
+These run on CPU: the monkey patch only skips groups whose *arguments* name gloo, so a
+group created with the default backend on a gloo world is wrapped like any NCCL group.
+"""
+
+from datetime import timedelta
+
+import torch
+import torch.distributed as dist
+from tests.fast.dist_utils import init_gloo, run_multiprocess
+
+from miles.utils.reloadable_process_group import (
+    ReloadableProcessGroup,
+    destroy_process_groups,
+    monkey_patch_torch_dist,
+    reload_process_groups,
+)
+
+_TIMEOUT = timedelta(seconds=1234)
+
+
+def _worker_reload_replays_new_group_args(rank: int, world_size: int, port: int) -> None:
+    """Reload must rebuild the group from the original new_group args, not a hardcoded pair.
+
+    Before #1594 the reload hardcoded ``(ranks, backend="nccl")``, silently dropping timeout,
+    pg_options and group_desc. On a CPU world that regression surfaces as a hard failure:
+    the rebuilt group claims an NCCL backend and the next collective raises.
+    """
+    init_gloo(rank, world_size, port=port)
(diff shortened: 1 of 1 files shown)


David553:
Gentle ping — this one has no reviewer auto-requested, since CODEOWNERS doesn't cover `tests/`.

It adds the regression test that #1594 landed without. Before that fix the reload path rebuilt every group as `(ranks, backend="nccl")`, dropping `timeout`, `pg_options` and `group_desc` — which is the shape behind the multi-node hangs in #384 and #574, where a run appears to ignore `--distributed-timeout-minutes` and dies at the 600 s default instead.

Cheap to take: CPU-only, auto-registered in `stage-a-cpu`, no GPU cost and no `run-ci-*` label needed for the test itself — though the fork's CI is currently held, so a `run-ci` label would release it.

Verified both ways against `miles/utils/reloadable_process_group.py`: 5/5 pass on `main`, and on the pre-#1594 file it fails exactly at the `inner_kwargs["timeout"]` assertion, so it does guard the thing it claims to.

@fzyzcjy (you wrote #1594) @yueming-yuan

PR #3135, opened 2026-09-07
feat(dsv4): direct-HF with MXFP8 training and rollout (P1)

bobboli:
## Summary

P1 only: initialize DeepSeek-V4 Flash Preview RL directly from the official HF checkpoint through Megatron-Bridge, with TE MXFP8 training and trainer-owned MXFP8 rollout. No offline BF16/torch_dist weight conversion or runtime source patches are required for this path.

This PR has been narrowed following review. Packed MXFP4 rollout and routed-expert QAT (P2/P3) are moved to the stacked Draft https://github.com/bobboli/miles/pull/18; they are not included here. Importing an official checkpoint that contains MXFP4 weights through Bridge is still supported; this does not imply MXFP4 rollout.

## P2/P3 follow-up

The follow-up Draft is temporarily hosted in `bobboli/miles`, with this PR's head branch (`github/dsv4-native-hybrid-consolidated`) as its base. This gives it a true incremental P2/P3 diff without duplicating P1. GitHub requires the base branch to exist in the PR's destination repository, and we do not have permission to create that staging branch in `radixark/miles`.

P1 remains ready for review independently. After P1 merges, rebase the remaining P2/P3 changes onto `radixark/miles main` and open the upstream follow-up PR. Keep the fork follow-up Draft while the quantization configuration and QAT review items are being addressed.

## Included

- Direct-HF trainer initialization, native hybrid attention/cuDNN setup, and selecting the HF trainer seed for Bridge export mappings.
- Metadata-only MXFP8 rollout schema and dummy SGLang initialization followed by initial trainer-to-rollout weight synchronization.
- Shared atomic weight-update groups and producer CUDA IPC lifetime synchronization.
- Canonicalize release-native Bridge export names before postprocessing and quantization, reusing SGLang's mapping and a single atomic-group table.
- FlashMLA installation/import checks for native DSA; associated build documentation. The unrelated wheels directory changes remain removed.

The new trainer-owned path requires `--rollout-mxfp8`. Existing converted-checkpoint recipes remain available.

```bash
python scripts/run_deepseek_v4.py full-train \
  --model-name DeepSeek-V4-Flash \
  --init-model-source hf --rollout-weight-source trainer \
  --no-train-fp8 --train-mxfp8 --no-rollout-fp8 --rollout-mxfp8
```

## Companion PR dependencies

This is a multi-repository integration, not a standalone Miles change. Use compatible companion source checkouts until these changes are integrated into the runtime image.

| Repository | Required PR | Role in P1 | Current companion head |
| --- | --- | --- | --- |
| Megatron-Bridge | https://github.com/radixark/Megatron-Bridge/pull/36 | Native-hybrid direct-HF import and hash-router mappings | `73905738b` |
| Megatron-LM | https://github.com/radixark/Megatron-LM/pull/96 | Native DSV4 stack with deterministic CSA no-grad/grad parity fixes | `3f0b8cceb` |
| SGLang | https://github.com/sgl-project/sglang/pull/39392 | R3 HashTopK capture and weight-update IPC synchronization | `0d4f00c7a` |

SGLang #39392 is the cleaned-up PR targeting `sglang-miles`, replacing #38282. Its MXFP4 reload/clamp changes are needed by P2/P3, not by the MXFP8 expert layout in P1. NVIDIA/Megatron-LM #7169 is the corresponding CSA fix on NVIDIA's dev base, not an interchangeable runtime pin.

The Dockerfile's repository branch defaults do not automatically select the above PR heads. Companion integration or explicit source pins remain necessary before claiming the stock image runs this configuration.

## Validation and review status

- Rebased onto `radixark/miles main` at `f6d0257d83` on 2026-09-16. All 11 original commits were patch-equivalent across the rebase (`git range-diff`).
- At the split, P1 and the P2/P3 child reproduced the rebased pre-split tree exactly. Subsequent terminology cleanup only replaces phase labels in docstrings, an assertion message, and test names/variables with functional descriptions. The original combined head is preserved at `bobboli:backup/dsv4-p1-p3-before-split-20260916` (`a6ebac3b48`).
- Syntax and whitespace checks only for this split. No unit tests, launcher snapshot runs, image builds, or RL runs were repeated; historical validation of the combined implementation is not validation of these new heads.
- Ready for review, not a claim that all integration/CI gates have passed. The inherited Bridge-wide `mtp_num_layers` override remains a review consideration. Native Transformers DSV4 config support is required.

## Export-name review follow-up (2026-09-17)

Commit `cbd26cf850` resolves names once from the selected Bridge checkpoint's local config and safetensors index, then remaps base-weight exports before postprocessing/quantization. Already-canonical checkpoints use identity mapping. Missing local config/index preserves export names with a warning; canonicalization is not claimed for that fallback. The duplicate checkpoint-layout atomic-group table and iterator override are removed; SGLang is unchanged.

Validation: 23 targeted CPU tests passed (iterator, atomic groups, factory, schema; run without unrelated root conftest fixtures), Ruff/isort checks passed, and Black passed on the modified production/iterator test files. An existing unrelated formatting discrepancy remains in `test_atomic_groups.py`. A static audit executed Bridge #36's mapping registry with recording-only constructors and SGLang's actual remapping function: all 74 unique name patterns, including 33 MTP patterns, passed normalization/idempotence checks. This is not a full distributed export, GPU quantization or RL validation. The Draft child was rebased with its incremental patch unchanged.


miles/backends/megatron_utils/megatron_to_hf/processors/__init__.py
```diff
@@ -18,6 +20,8 @@ def quantize_params(args, megatron_name, converted_named_params, quantization_co
     if quantization_config is None:
         return converted_named_params
     elif quantization_config["quant_method"] == "fp8":
+        if getattr(args, "rollout_fp4_experts", False) and is_routed_expert_param(megatron_name):
```
HJSang: do we want to also check if it is dsv4 arch here?

wheels/.gitkeep
```diff
@@ -0,0 +1 @@
+
```
HJSang: unrelated change

.gitignore
```diff
@@ -19,7 +19,8 @@ lib64/
 parts/
 sdist/
 var/
-wheels/
+wheels/*
+!wheels/.gitkeep
```
HJSang: unrelated change

HJSang:
The change LGTM but I need more time to test it carefully. 
for `mxfp4_quantize_dequantize`, can we reuse implementation from sglang or megatron?
I would recommend to split ` direct-HF init with the P1 schema ` out and get it merged first and then with a following up PR to enable rollout mxfp4? WDTY?

wheels/.gitkeep
bobboli: Removed in a6ebac3b4899879e15c3024ef0c7ce9c1f429f9d, together with the related .gitignore change. Restored the original COPY wheel[s]/ behavior; FlashMLA installation is unchanged.

.gitignore
bobboli: Reverted in a6ebac3b4899879e15c3024ef0c7ce9c1f429f9d. The wheels directory portability changes are no longer part of this PR.

bobboli:
Split as suggested: this PR now covers direct-HF initialization and MXFP8 training/rollout only. MXFP4 rollout and routed-expert QAT are in the stacked Draft https://github.com/bobboli/miles/pull/18, based on this PR’s head branch. The follow-up is temporarily hosted in my fork to preserve the stack and will be submitted to radixark after this PR lands. Both descriptions list the three companion-repository dependencies.

bobboli:
@HJSang Regarding `mxfp4_quantize_dequantize` (now in the stacked Draft https://github.com/bobboli/miles/pull/18): SGLang has `MXFP4QuantizeUtil`, used in weight-loading/conversion paths, while the fake-QAT implementation in our current Megatron branch is signed INT4 rather than MXFP4. Our Triton implementation fuses quantize/dequantize because it runs during training forwards. Would you prefer reusing SGLang’s implementation, which may need to be adapted into a fused kernel for the training hot path, or retaining our fused kernel with parity tests against the SGLang implementation? We have not benchmarked the two options yet.

HJSang:
> @HJSang Regarding `mxfp4_quantize_dequantize` (now in the stacked Draft [bobboli#18](https://github.com/bobboli/miles/pull/18)): SGLang has `MXFP4QuantizeUtil`, used in weight-loading/conversion paths, while the fake-QAT implementation in our current Megatron branch is signed INT4 rather than MXFP4. Our Triton implementation fuses quantize/dequantize because it runs during training forwards. Would you prefer reusing SGLang’s implementation, which may need to be adapted into a fused kernel for the training hot path, or retaining our fused kernel with parity tests against the SGLang implementation? We have not benchmarked the two options yet.

i see. I am ok with the fused kernel as long as parity tests look good.

HJSang:
**Discussion, not a merge blocker** — a thought on how the atomic-group change could be organized. Happy to defer to a follow-up if you'd rather land P1 as is.

The `dsv4_checkpoint_layout` flag, the second `_DEEPSEEK_V4_CHECKPOINT_GROUPS` table, and the `_hf_atomic_update_groups` override in the Bridge iterator all exist because Bridge exports release-native `.attn.*` names while everything else downstream of the exporter assumes HF `.self_attn.*` names (raw converter, `quantize_params`, the MXFP8 schema's `modules_to_not_convert`, and SGLang's own `remap_weight_name_to_dpsk_hf_format`).

One alternative is to canonicalize names at the export boundary instead of teaching the atomic-group table about both layouts:

- In `HfWeightIteratorBridge.__init__`, derive a name remap from the selected bridge checkpoint via `get_param_name_remap` — the helper this PR already adds in `miles/utils/hf_parameter_names.py` for the schema builder. It returns SGLang's native→HF remap for a release-native DSV4 checkpoint and identity otherwise.
- Apply it to `hf_name` in `_iter_hf_param_units`.
- Then `atomic_groups.py` keeps one table and the original signature (only the hyphen/underscore normalization stays), and the override in the Bridge iterator goes away.

What I like about it:

- One source of truth for the layout difference, and it's SGLang's own mapping rather than a hand-maintained second suffix list.
- It resolves the "hardcoded `dsv4_checkpoint_layout=True` is wrong for re-saved HF-format DSV4" item in the description automatically, since the remap is chosen from the checkpoint's `architectures` + `embed.weight` marker.
- The atomic-group table stays a description of what SGLang fuses, independent of which exporter produced the names.

Two things to confirm if you think it's worth trying: (1) the remap covers every name Bridge #36 exports — an assertion that no exported name still contains `.attn.` after remapping would make any gap obvious; (2) `_select_bridge_checkpoint` returns a directory with a weight index, since `get_param_name_remap` reads `weight_map` (true for P1's HF seed; a missing index could fall back to identity with a log line).

If normalizing names turns out to be undesirable on the SGLang side, a smaller step would still avoid two tables: keep the fused pairs relative to the attention module (`("wqkv_a", ("wq_a.weight", "wkv.weight"))`, …) and build suffixes from a single `attention_module` prefix derived from the first exported name, rather than a DSV4-specific boolean.

Separately and very minor: the hyphen/underscore normalization could live where `model_name` is derived in `actor.py`, so the raw converter dispatcher in `megatron_to_hf/__init__.py` picks it up too.

HJSang:
**Also discussion-only, organization not correctness** — on where `miles/utils/hf_rollout_schema.py` lives.

Its only consumers are prepare-time: `scripts/run_deepseek_v4.py` (`create_mxfp8_rollout_schema` in the new prepare stage), `tools/convert_hf_to_mxfp8.py` (`MXFP8_SKIP_WEIGHT_SUBSTRINGS`), and the tests. Nothing in the training runtime imports it. It has to stay importable from `miles` (scripts/ and tools/ aren't packages and the launcher reaches tools by subprocess), so I wouldn't move it to `scripts/` — but two smaller moves might read better:

- `MXFP8_SKIP_WEIGHT_SUBSTRINGS` + `build_mxfp8_quantization_config` describe the MXFP8 layout; they'd sit naturally beside `MXFP8_GROUP_SIZE` in `miles/utils/mxfp8.py` (or a sibling `mxfp8_layout.py`). The trainer-side `quantizer_mxfp8` is the other place that may eventually want the same skip list.
- `create_mxfp8_rollout_schema` (writes the directory, copies tokenizer files, drops the marker) is launcher-support behaviour, and `miles/utils/external_utils/` is the existing home for that (`command_utils` is the one `miles` module every launcher imports).

Optionally a thin `tools/create_rollout_schema.py` CLI over the library function, so it can be run by hand like the other prepare stages.

No need to hold P1 on this; happy for it to be a follow-up or left as is.

HJSang (approved):
LGTM. 
The change is clean and good to me. Just two comments about code structure. the functionality is fine.

bobboli:
@HJSang Implemented the [export-boundary canonicalization suggestion](https://github.com/radixark/miles/pull/3135#issuecomment-5702148887) in cbd26cf850; 492727f86c adds a concrete checkpoint-name → SGLang-name example in the comment.

The iterator resolves the existing SGLang remapper once from the selected checkpoint and applies it before postprocessing/quantization, preserving tensors and Megatron source names. Removed the duplicate atomic-group table, layout flag, and iterator override. Missing local config/index falls back to identity with a warning; no SGLang changes are needed.

Validation: 23 targeted CPU tests passed. A static audit of Bridge #36’s 74 export-name patterns (including 33 MTP patterns) against SGLang’s actual mapper found no remaining `.attn.`/`.ffn.` names and passed repeat-mapping/MTP compatibility checks. This was not a distributed export or RL smoke. The stacked Draft has been rebased with its incremental patch unchanged.

bobboli:
Regarding the [model-name normalization suggestion](https://github.com/radixark/miles/pull/3135#issuecomment-5702148887), I kept it local for now:

- [Atomic-group lookup strips hyphens/underscores into a local variable](https://github.com/bobboli/miles/blob/492727f86c6fcfb407f16f2ddd41b760771a20cd/miles/backends/training_utils/weight_update/hf_weight_iterator/atomic_groups.py#L28-L37), leaving the shared `model_name` unchanged.
- The raw converter [only lowercases its input](https://github.com/bobboli/miles/blob/492727f86c6fcfb407f16f2ddd41b760771a20cd/miles/backends/megatron_utils/megatron_to_hf/__init__.py#L33-L34), then [selects the specialized Qwen3.5/3.6 converter using `qwen3_5` / `qwen3_6`](https://github.com/bobboli/miles/blob/492727f86c6fcfb407f16f2ddd41b760771a20cd/miles/backends/megatron_utils/megatron_to_hf/__init__.py#L50-L51).
- The [next branch matches generic `qwen3`](https://github.com/bobboli/miles/blob/492727f86c6fcfb407f16f2ddd41b760771a20cd/miles/backends/megatron_utils/megatron_to_hf/__init__.py#L52-L53).

If actor.py normalizes the shared name, `qwen3_5 → qwen35`: the specialized check no longer matches, but the generic `qwen3` check does, selecting the wrong converter. Centralizing normalization therefore needs coordinated dispatcher updates, rather than just moving the line. I'd keep that separate from this DSV4 change.

PR #3138, opened 2026-09-07
fix(ray): use IP literal for SGLang router bind address

### LINKED ISSUE #3133: [Bug] SGLang router fails to start when Ray's node address is a hostname: launch_router --host <hostname> raises "Invalid address: invalid socket address syntax"
## Summary

When miles launches the SGLang router through the Ray worker manager, the router's bind address is the worker's *placed* address. On clusters where Ray reports a **hostname** rather than an IP literal (e.g. Slurm-managed nodes), `sglang_router.launch_router` fails immediately:

```
RuntimeError: Invalid address: invalid socket address syntax
```

The rollout therefore never comes up and the run dies with:

```
RuntimeError: Server at <node-hostname>:<port> not ready after 120s
```

## Affected revisions

- Last known good: `6b7ac372839606ddee25c36f00e274a1191924c1` — rollout starts, GRPO training completes.
- First observed bad: `d2fc97ce581577e255e494801d7568747d5a10d7` (current `main`) — router fails to start; still reproducing on 2026-09-06.

Both revisions were exercised with the same hardware, model, dataset, topology and default flags; the miles revision was the only changed axis (miles wheel `0.1.0+git6b7ac3728` → `0.1.0+gitd2fc97ce5`). The boundary has not been narrowed to a single commit inside that range; the defect is nevertheless present in `main` by direct source inspection (below).

## Where

`miles/ray/specs/inference.py::_compute_spec_router` builds the router launch command from the worker's placed address:

```python
primary = ctx.self_addrs["primary"]
...
router_args = compute_sglang_router_args(
    args,
    host=primary.host,
    port=primary.port,
    prometheus_port=ctx.self_addrs["prometheus"].port,
    has_pd_disaggregation=model_cfg.has_pd_disaggregation,
)
launch_argv = [*interpreter_prefix, "-m", "sglang_router.launch_router", *router_args_to_argv(router_args)]
```

`primary.host` comes from the Ray node address probed by the worker manager, which on a Slurm-managed cluster is the node's hostname. `sglang-router` 0.3.2 parses the `--host`/`--port` pair into a Rust `std::net::SocketAddr`, which accepts IP literals only — a hostname raises `invalid socket address syntax`.

Note that other consumers of the same address are unaffected: `dist_init_addr` and the engine HTTP base URLs are connect-side and resolve hostnames normally. Only the router's **bind** address must be an IP literal.

## Reproduction

Observed end-to-end through the normal miles path:

1. Run on a cluster where the Ray node address is a hostname (Slurm-managed nodes; the worker-manager node probe returns e.g. `node1234` instead of `10.x.y.z`).
2. Launch a GRPO run with the SGLang rollout and the SGLang router enabled, using default flags — no `--sglang-router-ip` and no `--miles-host-ip`.
3. The router `CommandActor` raises `RuntimeError: Invalid address: invalid socket address syntax` during startup; the health wait then fails with `Server at <node-hostname>:<port> not ready after 120s`.

Expected: the router binds and serves on the placed node.
Actual: the router process exits before binding, and the whole run fails.

A minimal check that exercises the same parse (derived from the source path above; we have not run it standalone) is:

```
python -m sglang_router.launch_router --host "$(hostname)" --port 20000
```

## Environment

- miles `main` @ `d2fc97ce5`; `sglang-router` 0.3.2; `sglang` 0.5.19.dev56; Ray 2.58.0; Python 3.12; torch 2.13.0+cu130; CUDA 13.0.3
- aarch64 (Neoverse-V2), 1 node × 4 GPUs, on both GB200 NVL and GB300 NVL nodes
- Qwen/Qwen3-30B-A3B, GRPO on gsm8k, Megatron train backend, SGLang rollout, bf16
- Reproduced on both hardware families on 2026-09-05 and again on 2026-09-06

## Suggested fix

Resolve the placed host to an IP literal on the bind side before it reaches `launch_router` — for example in `_compute_spec_router`, or by exposing an `ip` alongside `host` on the worker address type / node probe, falling back to `getaddrinfo`/`gethostbyname` when `host` is not already an IP literal — while keeping the hostname for connect-side and HTTP URLs. Alternatively, `sglang-router` could resolve hostnames instead of requiring a `SocketAddr`.

Setting `--miles-host-ip 0.0.0.0` is not a general answer: it changes the bind address of every service, not just the router, and the PR that restored that fallback under the worker manager (#2850) was closed unmerged.

---

_This issue was drafted with assistance from the `opus` AI model._

200lz:
## Motivation

The worker manager advertises each placed worker under Ray's node address (`NodeProbeMixin._get_node_ip()`, i.e. `MILES_HOST_IP` or `ray._private.services.get_node_ip_address()`). On Slurm-managed clusters that address is the node's hostname. Every connect-side consumer handles that fine (engine HTTP base URLs, `dist_init_addr`, the session server's backend URL, `wait_tcp_ready`), but `sglang_router` parses `--host:--port` as a Rust `SocketAddr`, which only accepts an IP literal. The router exits with `Invalid address: invalid socket address syntax` before binding and the whole run dies with `Server at <hostname>:<port> not ready after 120s`.

Reproduced CPU-only with `sglang-router` 0.3.2: `--host <hostname>` and `--host ::1` both fail that way, `--host 127.0.0.1` and `--host [::1]` bind.

There is no separate IP-typed value to carry from the placement layer: the probed node address *is* what Ray was started with, and the existing tests pin that `MILES_HOST_IP` may be a non-numeric name. So the fix stays on the router's bind side.

## Changes

- `compute_sglang_router_bind_host(host)` in `router_args_utils.py` turns the advertised host into what the router can bind: IP literals pass through unchanged (IPv6 stays bracketed, the only form the router parses), a hostname is resolved with `socket.getaddrinfo()` over both address families, an unresolvable name raises a `RuntimeError` naming the host and the `MILES_HOST_IP` / `--node-ip-address` remedies instead of letting the router die on bind, and a name that resolves to several addresses binds the first resolver candidate and logs the full candidate list.
- `_compute_spec_router()` passes that bind host to `compute_sglang_router_args(host=...)`. `primary.host` is still what the launch context advertises; the miles router, session server and engine launch paths are untouched.
- Regression tests in `tests/fast/ray/specs/test_inference.py` render the real launch command with a fake resolver (no DNS, no dependency on the machine's addresses): hostname advertised + resolved IP bound, IPv4 literal and bracketed IPv6 literal unchanged and never resolved, hostname resolving to IPv6 bound bracketed, multi-address warning, unresolvable-name error, and the miles router / session server keeping the hostname. Unit tests in `test_router_args_utils.py` cover bare-IPv6 bracketing, duplicate resolver rows and an empty resolver answer.

## Validation

- `tests/fast/ray/specs/test_inference.py::TestSglRouterBindHost` fails on current `main` for the intended reason (`--host worker-01.cluster.internal` instead of `10.20.30.40`) and passes with the fix.
- `pytest tests/fast/ray/specs tests/fast/backends/sglang_utils tests/fast/ray/rollout/test_engine_addressing.py tests/fast/ray/rollout/test_server_cell_addressing.py tests/fast/ray/rollout/test_router_manager.py tests/fast/utils/test_misc.py tests/fast/utils/workers`: 2014 passed, 5 xfailed.
- `pre-commit run --files <changed files>`: ruff, autoflake, isort, black all pass. `git diff --check` clean.
- Original-environment validation: the issue reporter validated this exact commit on the original Slurm-managed aarch64 GB200 NVL and GB300 NVL environments. Four end-to-end runs passed on the first attempt with Ray still advertising the node hostname and without `MILES_HOST_IP` or `--sglang-router-ip` workarounds. Engines registered, rollout became ready, and all runs completed five GRPO steps. The original `invalid socket address syntax` and readiness-timeout signatures did not recur. See the reporter's validation comment below.

## Scope

- Connect-side hostname behaviour is unchanged: `HostAndPort.host`, `addr`, `dist_init_addr` and the session server backend URL still carry the advertised hostname.
- Miles addresses are not globally normalized; the literal is derived only where the SGLang router's listening socket needs it.
- Resolution is performed when the router launch command is rendered, so the selected bind address reflects the manager's resolver view of the advertised hostname.
- No GPU is needed for the local regression; this is control-plane addressing logic. The original Slurm GB200/GB300 environment has now also passed end-to-end validation on this exact commit.

Fixes #3133

### DIFF
--- a/miles/backends/sglang_utils/router_args_utils.py
+++ b/miles/backends/sglang_utils/router_args_utils.py
@@ -1,9 +1,15 @@
 import argparse
+import ipaddress
+import logging
+import socket
 from collections.abc import Mapping
 
 from sglang_router.launch_router import RouterArgs
 
+from miles.utils.http_utils import MILES_HOST_IP_ENV, _wrap_ipv6
 from miles.utils.workers.argv_utils import render_cli_argv
+
+logger = logging.getLogger(__name__)
 
 _ROUTER_DEST_PREFIX = "router_"
 
@@ -32,6 +38,49 @@ def compute_sglang_router_args(
         router_args["pd_disaggregation"] = True
 
     return router_args
+
+
+def compute_sglang_router_bind_host(host: str) -> str:
+    """The ``--host`` sglang_router can bind for a worker advertised as ``host``.
+
+    The router parses ``host:port`` as a socket address, so it takes an IPv4 literal or a bracketed IPv6
+    literal and nothing else. The worker manager may advertise a placed node under a hostname (Ray on Slurm
+    reports nodes that way, and so may ``MILES_HOST_IP``); that name stays the connect-side address and is
+    resolved here only for the router's own listening socket.
+    """
+    if _is_ip_literal(host):
+        return _wrap_ipv6(host)
+
+    try:
+        infos = socket.getaddrinfo(host, None, type=socket.SOCK_STREAM)
+    except socket.gaierror as e:
+        raise RuntimeError(_unresolvable_bind_host_message(host, reason=str(e))) from e
+    candidates = list(dict.fromkeys(info[4][0] for info in infos))
+    if not candidates:
+        raise RuntimeError(_unresolvable_bind_host_message(host, reason="the resolver returned no address"))
+    if len(candidates) > 1:
+        logger.warning(
+            f"The sglang router's advertised host {host!r} resolves to {candidates}; binding the first resolver "
+            f"candidate {candidates[0]!r}. Configure {MILES_HOST_IP_ENV} or Ray's --node-ip-address as a specific "
+            f"IP literal if deterministic interface selection is required."
+        )
+    return _wrap_ipv6(candidates[0])
+
+
+def _is_ip_literal(host: str) -> bool:
+    try:
+        ipaddress.ip_address(host.strip("[]"))
+        return True
+    except ValueError:
+        return False
+
+
+def _unresolvable_bind_host_message(host: str, *, reason: str) -> str:
+    return (
+        f"Cannot bind the sglang router on {host!r}: {reason}. sglang_router needs an IP literal for --host, "
+        f"so configure {MILES_HOST_IP_ENV} on that node, or Ray's --node-ip-address, as a specific IP literal "
+        f"(another hostname would fail the same way)."
+    )
 
 
 def router_args_to_argv(router_args: Mapping[str, object]) -> list[str]:
--- a/miles/ray/specs/inference.py
+++ b/miles/ray/specs/inference.py
@@ -2,7 +2,11 @@ import logging
 import os
 import shlex
 
-from miles.backends.sglang_utils.router_args_utils import compute_sglang_router_args, router_args_to_argv
+from miles.backends.sglang_utils.router_args_utils import (
+    compute_sglang_router_args,
+    compute_sglang_router_bind_host,
+    router_args_to_argv,
+)
 from miles.backends.sglang_utils.sglang_config import ModelConfig, ServerGroupConfig, resolve_sglang_config
 from miles.backends.sglang_utils.sglang_engine import compute_engine_launch_cmd
 from miles.ray.utils import NOSET_VISIBLE_DEVICES_ENV_VARS_LIST
@@ -44,7 +48,8 @@ def _compute_spec_router(args, model_idx: int, model_cfg: ModelConfig) -> Comman
         else:
             router_args = compute_sglang_router_args(
                 args,
-                host=primary.host,
+                # peers keep reaching the router by primary.host; only its listening socket needs the literal
+                host=compute_sglang_router_bind_host(primary.host),
                 port=primary.port,
                 prometheus_port=ctx.self_addrs["prometheus"].port,
                 has_pd_disaggregation=model_cfg.has_pd_disaggregation,
--- a/tests/fast/backends/sglang_utils/test_router_args_utils.py
+++ b/tests/fast/backends/sglang_utils/test_router_args_utils.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import argparse
+import logging
+import socket
 from argparse import Namespace
 
 import pytest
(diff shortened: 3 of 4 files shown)


janbernloehr:
Environment validation is complete for `9578a292f6d9e4fb975dbbf53aeedee000322b11`, and the result is green.

I tested the PR on the original Slurm-managed aarch64 GB200 NVL and GB300 NVL environments where #3133 reproduced. The matrix covered both GB200 and GB300 workload configurations on both hardware environments, for four end-to-end runs total. All four passed on their first attempt.

The workload was kept representative of the original failure:

- one node with four GPUs;
- Qwen3-30B-A3B;
- GRPO on gsm8k;
- Megatron training with SGLang rollout;
- bf16, batch size 128, and five steps;
- `sglang-router` 0.3.2 and Ray 2.58.0;
- no `--sglang-router-ip` or `MILES_HOST_IP` workaround.

Most importantly, Ray was still started with its node address set to the Slurm hostname, so the condition that originally triggered the bug was preserved.

The logs confirm the intended bind/connect split from this PR:

- the SGLang router process was launched with `--host <resolved IP literal>`;
- the engine continued to use `--host <node hostname>`;
- `dist_init_addr` and the other connect-side addresses continued to use the hostname;
- the router was reported ready through the original advertised hostname.

After startup, the engines registered successfully, rollout became ready, and all four runs completed the full five GRPO steps. Each run also produced its expected performance metrics.

Across all four logs there were zero occurrences of either original failure signature:

- `Invalid address: invalid socket address syntax`
- `Server at <hostname>:<port> not ready after 120s`

I also repeated the regression bracket in a frozen environment using exact source revisions. The last-known-good revision, `6b7ac372839606ddee25c36f00e274a1191924c1`, completed all five steps. The first-bad revision, `d2fc97ce581577e255e494801d7568747d5a10d7`, failed before training with the original `invalid socket address syntax` error followed by the readiness timeout.

That comparison held the workload, system, libraries, Ray version, SGLang/router versions, and base environment fixed; the Miles revision was the only package difference.

From the original environment-validation perspective, #3138 resolves #3133: it fixes the router’s bind address without changing the hostname-based advertisement and connect-side behavior.

200lz:
Thank you for the thorough validation — especially for preserving the original hostname-based Ray setup and repeating the good/bad revision bracket.

The four GB200/GB300 end-to-end passes cover exactly the remaining system-level evidence I was hoping to get. I’ll leave the implementation unchanged and let the PR proceed through CI/review.

PR #3142, opened 2026-09-08
fix(optimizer): rematerialize streamed main weights

bobboli:
## Summary

- allow colocated CPU trainer offload to be combined with NVMe-streamed Adam state when parameter rematerialization is enabled
- restore model parameter shards from streamed FP32 main weights one bucket at a time
- release each main-weight bucket immediately, without loading Adam moments during rematerialization

This is stacked on #3135; its diff will reduce to this commit after the parent stack lands.

## Why

DeepSeek V4 training fits during the first backward, but the Adam state created by the first optimizer step remains resident and causes the next backward to OOM. Trainer disk offload avoids host pressure but currently crashes in TorchMemorySaver/CUDACachingAllocator during hot weight update. CPU trainer offload has a stable hot-update lifecycle; streaming optimizer state to NVMe removes the active-training Adam-state peak, but previously could not rematerialize evicted FP32 main weights.

## Tests

- 40 focused CPU tests passed
- 2 focused CUDA tests passed on B300, including clearing a parameter buffer, restoring it bit-exactly from NVMe, and verifying the FP32 main storage returns to zero bytes
- Ruff and `git diff --check` passed

### DIFF
--- a/miles/backends/megatron_utils/megatron_to_hf/processors/__init__.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/processors/__init__.py
@@ -1,6 +1,7 @@
 from .padding_remover import remove_padding
 from .quantizer_compressed_tensors import quantize_params_compressed_tensors
 from .quantizer_fp8 import quantize_params_fp8
+from .quantizer_mxfp4 import is_routed_expert_param, quantize_params_mxfp4
 from .quantizer_mxfp8 import quantize_params_mxfp8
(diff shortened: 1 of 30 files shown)


bobboli:
Update: I notice that https://github.com/radixark/miles/pull/2706 used `cudnn` while the OOM was encountered using pytorch kernels. Will try if `cudnn` could prevent OOM. Closing this now.

PR #3144, opened 2026-09-08
rollout: make sandbox credentials file-only

### LINKED ISSUE #3111: sandbox credentials: make the key file the only supply path (env supply masks broken file relays)
## Problem

Sandbox provider credentials currently have two supported supply paths, resolved env-first (`miles/rollout/agentic/credentials.py`):

- **A. env var** (`E2B_API_KEY`, ...) — the SDK's native form;
- **B. key file** (`E2B_API_KEY_FILE` → path, launcher forwards the path, worker resolves) — exists because worker env rides ray's `runtime_env`, which is echoed into driver logs and persisted in job metadata in plaintext.

Path A's presence lets it silently shadow path B, and that just cost us: the Harbor leg shipped without the worker-side file resolver, every validation still passed because the key was exported in the env, and the gap only surfaced on the first GPU e2e run (fixed in #2806). The hazard is structural — any future leg or provider can repeat it, and env-vs-file conflicts resolve by silent precedence.

## Proposal

Make the key file the only supported supply:

- `resolve_provider_api_key` stops honoring the env var (or first deprecation-warns loudly when it would have used one, then errors in a later release);
- preflights reject a set key env var with a message naming the file to create;
- docs/examples provision by file everywhere (the sandbox smoke already switched);
- K8s deployments use secret-as-file mounts (the standard practice anyway — env vars leak via /proc, child processes, and crash dumps);
- CI runners whose secrets are env-shaped write the file in a one-line step before launching.

## Costs

One-time file creation for developers; a mount-type switch for env-injected pods; friction for users following the provider's own "export the key" docs, mitigated by the error message. Migration needs one deprecation cycle for existing openenv users.

Out of scope for the current Harbor stack; touches the shared credentials layer, both sandbox legs, and docs together.

chi030303:
Fixes #3111.

Sandbox provider keys used to resolve **env-first**. That let `E2B_API_KEY` / `DAYTONA_API_KEY` / Modal token env vars silently satisfy preflight while the key *file* was missing or unread — the Harbor gap that only showed up on GPU e2e. The file is now the only supply path.

- `sandbox_key_supply` forwards the file path (never the secret) and **rejects a set key env var**, naming the file to create.
- `resolve_provider_api_key` reads the file only; a set key env var is a hard error that names the file.
- OpenEnv / AgentENV / glm52 docs and the glm52 Daytona launcher provision by file. Harbor and HUD keep their own copies (Harbor is out of scope on the issue).

## Tests

CPU-only, no GPU, no E2B/Harbor e2e:

```bash
pytest tests/fast/rollout/agentic/test_credentials.py tests/fast/examples/experimental/openenv -q
```

124 passed, 3 skipped.

Inverted the old env-wins tests: file-only succeeds and never puts the secret in the forwarded env; a set key env var fails and names the file path.

### DIFF
--- a/examples/experimental/agentenv/README.md
+++ b/examples/experimental/agentenv/README.md
@@ -53,7 +53,6 @@ documented in the [AgentENV docs](https://kvcache-ai.github.io/AgentENV/).
 
 ## 2. Point the E2B SDK at it
 
-<!-- 2026-09-09, tianqi, file-only sandbox credentials (#3111) -->
 ```bash
 pip install -e '<miles>[e2b]'   # e2b>=2.12: older releases send the template name in a field AgentENV does not read
 export E2B_API_URL=http://<server>:8000       # control plane
@@ -64,7 +63,6 @@ export OPENENV_E2B_URL_SCHEME=http
 # recent SDKs do validate the format client-side
 mkdir -p ~/.config/e2b && echo e2b_0000000000000000000000000000000000000000 > ~/.config/e2b/api_key
 ```
-<!-- end -->
 
 ## 3. Train
 
--- a/examples/experimental/agentenv/README.md
+++ b/examples/experimental/agentenv/README.md
@@ -53,6 +53,7 @@ documented in the [AgentENV docs](https://kvcache-ai.github.io/AgentENV/).
 
 ## 2. Point the E2B SDK at it
 
+<!-- 2026-09-09, tianqi, file-only sandbox credentials (#3111) -->
 ```bash
 pip install -e '<miles>[e2b]'   # e2b>=2.12: older releases send the template name in a field AgentENV does not read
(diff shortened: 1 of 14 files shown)


behroozazarkhalili:
Verified on top of current main (2ed189c0d); merges with no conflicts.

- tests/fast/rollout/agentic + tests/fast/examples/experimental/openenv on the merge: 125 passed, 3 skipped.
- Same suites with credentials.py reverted to main: the 7 new rejection tests fail (env-supplied key, unreadable path fallback, worker env, modal token, and the two sandbox resolve_api_key cases), the other 118 pass, so the tests pin the behavior change.
- ruff/black/isort at the pre-commit pins are clean on all 10 changed Python files.

cc @guapisolo @Shi-Dong

PR #3146, opened 2026-09-08
fix(megatron): align BSHD routing replay token order

guapisolo:
## Summary
Align Megatron BSHD routing replay with the router's sequence-first token order.

## Symptom & Reproduction
- Megatron runs using `--use-rollout-routing-replay --qkv-format bshd --micro-batch-size 2` can replay another token's expert IDs, even with TP=CP=1.
- For two three-token samples, replay rows are `[0,1,-1,2,3,-1]` instead of the required `[0,2,1,3,-1,-1]`: three of four valid tokens are misaligned. Matching tensor shapes do not detect this.
- The new multi-sample regression cases fail before the transpose and pass afterward.

## Root Cause
1. `fill_replay_data` stacks samples as `[B,S,L,K]` and flattens tokens in batch-first order.
2. `LanguageModelEmbedding.forward` converts hidden states to `[S,B,H]`.
3. `TopKRouter.routing` flattens `[S,B,E]`, pairing sequence-first token scores with batch-first replay rows.

The consumer ordering is visible in the pinned [Megatron router](https://github.com/radixark/Megatron-LM/blob/8c1e05747eb612b382df2632783df5c83a853646/megatron/core/transformer/moe/router.py#L816).

## Fix
Let the Megatron actor request sequence-first replay only for its MoE routing manager. Transpose after CP slicing and before flattening/SP slicing. Keep indexer replay batch-first; packed THD and FSDP retain their existing behavior.

## Verification
- `python -B -m pytest tests/fast/backends/training_utils/test_replay_data.py --confcutdir=tests/fast/backends/training_utils -o addopts= -p no:cacheprovider -q --tb=short`: **61 passed**, covering equal/unequal sample lengths, single samples, padding, multiple streams/top-k slots, TP/SP, both CP layouts through CPU rank simulation, batch-first ordering, and THD preservation.
- Black 24.3.0, isort 5.13.2, Ruff 0.14.7, and `git diff --check`: passed.
- Local global pytest fixtures cannot load because the installed SGLang lacks `encoding_dsv4`; the focused command excludes those unrelated rollout fixtures. No loader or CP slicing implementation is mocked.
- GPU distributed training has not been run.

## Review Focus
- `sequence_first`: only the Megatron MoE consumer opts in; the transpose must precede SP slicing.

### DIFF
--- a/miles/backends/megatron_utils/actor.py
+++ b/miles/backends/megatron_utils/actor.py
@@ -522,6 +522,7 @@ class MegatronTrainRayActor(TrainRayActor):
                     register_replay_list_func=m.register_replay_list_func,
                     if_sp_region=m.if_sp_region,
                     indices_are_token_positions=m.replay_indices_are_token_positions,
+                    sequence_first=m is routing_replay_manager,
                 )
 
         with inverse_timer("train_wait"), timer("train"):
--- a/miles/backends/training_utils/replay_data.py
+++ b/miles/backends/training_utils/replay_data.py
@@ -39,6 +39,7 @@ def fill_replay_data(
     register_replay_list_func: RegisterReplayListFunc,
     if_sp_region=True,
     indices_are_token_positions=False,
+    sequence_first=False,
 ):
     """Load rollout replay tensors into module replay queues.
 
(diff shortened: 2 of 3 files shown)


guapisolo:
@claude always

guapisolo:
@claude review always

guapisolo:
/rerun-test tests/e2e/megatron/test_qwen3_30B_A3B/test_r3_baseline.py

PR #3151, opened 2026-09-09
[DO NOT MERGE] e2e probe for the suite failure tail

yueming-yuan:
Throwaway. Carries #3149 plus a CPU test that prints 60k lines and then fails, so one real CI run exercises child pipe -> failure tail -> summary -> job log -> notifier evidence. Will be closed once the run is read.

### DIFF
--- a/tests/ci/ci_utils.py
+++ b/tests/ci/ci_utils.py
@@ -6,8 +6,10 @@ import os
 import re
 import signal
 import subprocess
+import sys
 import threading
 import time
+from collections import deque
 from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


yueming-yuan:
Closing: the probe cannot work. stage-a-cpu runs its files in one pytest invocation (run_suite.py:230, "CPU tests (fast/) use pytest; CUDA tests use python3 per-file"), so it never reaches run_unittest_files, which is the function #3149 changes. Exercising that path for real needs a GPU stage failure; not worth occupying a runner when nightlies fail daily on their own.

PR #3155, opened 2026-09-09
feat(dsv4): run the sparse-MLA and indexer kernels on a 64 KiB shared-memory GPU

xiaohong42:
## Problem

The DeepSeek-V4 tilelang kernels ship with tilings tuned for a 160 KiB shared-memory budget. On
gfx942 (MI300X / MI308X / MI325X, 64 KiB per workgroup) they do not compile, and the failure is a
hard error on the first forward pass:

```
RuntimeError: Initialization failed: Requested dynamic shared memory 141312 exceeds device limit 65536
```

Lowering each entry point at the shapes a real DeepSeek-V4-Flash checkpoint uses (`heads=64`,
`kv_lora_rank=512`, `index_topk=512`, indexer `heads=64 dim=128`) on an MI308X:

| kernel | layout | asks for | budget | over by |
|---|---|---:|---:|---:|
| `sparse_mqa_fwd` | CP=4 / TP=2, 32 local heads | 172032 B | 65536 B | 2.6x |
| `sparse_mqa_fwd` | CP=8 / TP=1, 64 local heads | 141312 B | 65536 B | 2.2x |
| `tl_indexer_fwd` | 64 heads, dim 128, `block_N=256` | 100352 B | 65536 B | 1.5x |

CP=4 overshoots by *more* than CP=8: `sparse_mqa_fwd` clamps `num_stages` to 1 only when
`H_per_block == 64`, so at 32 heads the KV pipeline still stages twice. `bwd` caps `max_block_H` at
32, which is not enough on its own; the indexer has no HIP branch at all. DeepSeek-V4 therefore
trains on gfx942 at no parallel layout, and the smaller one fails hardest.

### 1. Pick a tiling the target can host

Two numbers tilelang already has, neither advertised as API. The **budget** is in the target
description it builds for every compilation. The **requirement** is the `dyn_shared_memory_buf`
attribute on the lowered device function — the same figure it formats into the error above. Reading
it needs lowering but not code generation, about 4x cheaper than a build (2.0 s against 8.2 s on
gfx942), and it is exact: the three numbers above are read from the attribute and match the error
text byte for byte.

The requested tiling is compiled optimistically, so a device that fits it pays nothing — not a
wasted compilation, not even a lowering. Only a rejection starts the search, which then compiles
exactly one candidate: the largest that fits.

Three things it deliberately avoids:

- **No model of tilelang's allocator.** An earlier revision had one and got it wrong: `Q_shared` and
  `O_shared` share storage, so counting both over-estimated by 27% at `block_H=32` and rejected a
  tiling that builds.
- **No per-architecture tables.** Candidates are derived from the target's limits, so there is no
  `torch.version.hip`, no `get_device_properties` and no architecture-string parsing.
- **No branching on error text.** Gating the search on a recognised message would let a tilelang
  reword disable it *silently*. Any rejection starts the search; a caller error fails every
  candidate alike and ends at the same re-raise of the request's own exception.

### 2. Normalise `o` and `do` in the backward

`sparse_mqa_bwd_interface` asserts contiguity on four of its six tensors, but not on `do` — the one
tensor autograd may legitimately hand back non-contiguous, since `out.sum().backward()` produces a
stride-0 broadcast gradient.

The consequence is not a clean error. tilelang's stride check runs inside an `Exception ignored in:`
context, so it prints and returns; the kernel then walks the stride-0 pointer as though it were
dense and the process is killed by the GPU:

```
ValueError: Static stride mismatch for parameter dO: expected 16384 at index 1, got 0
Exception ignored in: 'tilelang_cython_wrapper.CythonKernelWrapper._check_static_strides'
...
Memory access fault by GPU node-2 (Agent handle: 0x...) on address 0x... Reason: Unknown.
```

Nothing in that output names the tensor that was wrong. A fifth assertion would look consistent but
be wrong: autograd is *allowed* to pass a non-contiguous gradient, so asserting only moves the crash
to a different line. `.contiguous()` is a no-op on an already-contiguous tensor.

## Numerics

Unchanged. Shrinking `block_N` only changes how much of the KV axis one workgroup sweeps per
iteration. `block_H` is plumbing rather than new math — the forward already blocked over heads
through `REPLICATE_H` and the `H0` offset, only the block size was hardwired to 64 — and shrinking
it is safe because V4 attention is MQA (`KV_shared` has no head axis) and the online-softmax state
is already per-head. On a target that fits the shipped tiling, the compiled kernel is the one
compiled before, because the request is always tried first.

## Verification

`tests/fast/test_dsv4_tiling_fallback.py`, 36 tests, no GPU required. They cover the search (a
fitting request costs one compilation and zero lowerings; a refused one measures until the first
candidate that fits and compiles only that; nothing fitting raises the *request's* rejection) and
pin the three tilelang couplings: `get_tir` on the kernel builders, `engine.lower` accepting the
no-device-compilation call, and the `dyn_shared_memory_buf` attribute name. An unreadable
requirement degrades to compiling candidates rather than concluding everything fits.

### On hardware

8x MI308X (gfx942, 65536 B per workgroup). The search resolves to:

```
[sparse_mqa_fwd]  num_stages=1 block_I=16 threads=64 block_H=None                       51184 B / 65536   (32 local heads)
[sparse_mqa_fwd]  num_stages=1 block_I=16 threads=64 block_H=32                         51184 B / 65536   (64 local heads)
[tl_indexer_fwd]  block_N=64 (requested 256)                                            51200 B / 65536
[sparse_mqa_bwd]  block_size=16 threads=64 split_store=16 stage_dq_through_shared=True
                  max_block_H=16                                                        52704 B / 65536
```

The real autograd entry point, `sparse_attn_tilelang` forward then `out.sum().backward()` — exactly
the stride-0 broadcast case — passes at `S=4096 H=32 topk=128` and `S=2048 H=64 topk=128`, output
shapes correct and `q`/`kv` gradients finite. Without the `do` normalisation the same call takes the
process down with the fault above.

A 4-layer DeepSeek-V4-Flash then trains end to end at 131072 tokens, both single-node layouts:

| | CP=4 / TP=2 | CP=8 / TP=1 |
|---|---|---|
| iterations | 3 / 3 | 3 / 3 |
| `context_parallel_size` / `allgather_cp` | 4 / True | 8 / True |
| Megatron→SGLang weight syncs | 32 | 32 |
| step time | 127.2 / 126.7 s | 120.4 / 120.2 s |
| OOM / NaN / shared-memory overflow | 0 / 0 / 0 | 0 / 0 / 0 |

Both runs are from one container whose Miles is this branch on top of `8d9826ea` with nothing else
applied, built from a Dockerfile based on an official `rocm/sgl-dev` image with no post-build
patching — so the tiling search is the only thing between the published kernels and a training run
on this hardware.

## Note for reviewers

The candidates *between* the shipped tiling and the one gfx942 settles on are only reachable on a
budget strictly between 65536 B and 160 KiB, which no device in reach has, so that part of the
order is reasoned rather than measured.

The right long-term home for the measurement is tilelang itself, which computes the number and then
throws it away into a formatted string. Until it exposes it, this reads the attribute.

Supersedes #2857, which solved the same problem with a hand-written per-architecture candidate list
and a pure-torch reference backward. Both are gone here: candidates are derived from the target, and
once the search makes the kernel always buildable there is nothing for a reference path to do.

### DIFF
--- a/miles_plugins/models/deepseek_v4/ops/kernel/tilelang_indexer_fwd.py
+++ b/miles_plugins/models/deepseek_v4/ops/kernel/tilelang_indexer_fwd.py
@@ -8,6 +8,14 @@ import tilelang
 import torch
 from tilelang import language as T
 
+from miles_plugins.models.deepseek_v4.ops.kernel.tiling import (
+    DeviceLimits,
+    build_with_largest_fitting_tiling,
+    current_target,
+    indexer_forward_block_ns,
+    shared_memory_required,
+)
+
 
 @tilelang.jit(
     pass_configs={
@@ -133,6 +141,40 @@ def _make_causal_cu_seqlens(seq_len_q, seq_len_kv, compress_ratio, device):
     return cu_seqlen_ks, cu_seqlen_ke
 
 
+_fitted_block_N: dict[tuple, int] = {}
+
+
+def _indexer_fwd_within_shared_mem(heads, index_dim, block_N=256):
+    """Build tl_indexer_fwd_impl with the largest block_N this target can host.
+
+    block_N only sets how much of the KV axis one workgroup sweeps per iteration, so shrinking it
+    costs performance and nothing else. Memoized per shape.
+    """
+
+    def build(n):
+        return tl_indexer_fwd_impl(heads=heads, index_dim=index_dim, block_N=n)
+
+    key = (heads, index_dim, block_N)
+    if key in _fitted_block_N:
+        return build(_fitted_block_N[key])
+
+    target = current_target()
+    limits = DeviceLimits.from_target(target)
+    kernel, fitted = build_with_largest_fitting_tiling(
+        requested=block_N,
+        derived=indexer_forward_block_ns(block_N=block_N, limits=limits),
+        compile_tiling=build,
+        required_bytes=lambda n: shared_memory_required(
+            tl_indexer_fwd_impl.get_tir(heads=heads, index_dim=index_dim, block_N=n), target
+        ),
+        budget=limits.shared_memory_per_block,
+        describe=lambda n: f"block_N={n} (requested {block_N})",
+        what="tl_indexer_fwd",
+    )
+    _fitted_block_N[key] = fitted
+    return kernel
+
+
 def indexer_fwd_interface(q, kv, weights, cu_seqlen_ks, cu_seqlen_ke, clean_logits=True):
     """Forward interface matching GLM-5's API but for a single batch element.
 
@@ -150,7 +192,7 @@ def indexer_fwd_interface(q, kv, weights, cu_seqlen_ks, cu_seqlen_ke, clean_logi
     seq_len_kv = kv.shape[0]
 
     clean_logits_kernel = clean_logits_()
-    tl_indexer_fwd_kernel = tl_indexer_fwd_impl(heads=heads, index_dim=index_dim)
+    tl_indexer_fwd_kernel = _indexer_fwd_within_shared_mem(heads=heads, index_dim=index_dim)
 
     logits = torch.empty([seq_len, seq_len_kv], device=q.device, dtype=torch.float32)
     tl_indexer_fwd_kernel(
--- a/miles_plugins/models/deepseek_v4/ops/kernel/tilelang_sparse_mla_bwd.py
+++ b/miles_plugins/models/deepseek_v4/ops/kernel/tilelang_sparse_mla_bwd.py
@@ -9,6 +9,15 @@ import tilelang
 import torch
 from tilelang import language as T
 
+from miles_plugins.models.deepseek_v4.ops.kernel.tiling import (
+    BackwardTiling,
+    DeviceLimits,
+    build_with_largest_fitting_tiling,
+    current_target,
+    shared_memory_required,
+    sparse_mla_backward_tilings,
+)
+
 
 @tilelang.jit(out_idx=[-1])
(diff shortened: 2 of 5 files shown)


xiaohong42:
Field report from a 4-node gfx942 (MI308X) bring-up, in case it is useful for review.

**The fallback is a no-op where it should be.** `build_with_largest_fitting_tiling` compiles the requested tiling first, so a device that fits it pays nothing -- confirmed on this stack: only gfx942 ever entered the search. `tests/fast/test_dsv4_tiling_fallback.py` is 36 passed here.

**One thing worth adding to the description.** On gfx942 the backward lands on `split_store=16 / stage_dq_through_shared=True`, and under that tiling dKV is not bitwise reproducible run to run: `tilelang_sparse_mla_bwd.py`'s `T.atomic_addx4` accumulates into a zero-initialised global fp32 buffer unconditionally. Measured 1.147e-3 (32 heads) and 1.404e-3 (64 heads) between two runs with identical inputs and tiling.

This is **not introduced by this PR** -- the original 160 KiB tiling has the same atomic add -- but anyone who picks up the fallback path to build a bitwise-reproducibility gate will hit it, and the number is useful to have up front. `--deterministic-mode` does not remove it either, since it constrains aten ops and this is inside the tilelang kernel.

Measured separately: switching `block_H` is bitwise identical (0.00e+00) as the description claims; switching `block_I` moves `out` by ~3e-3, which is bf16 reduction noise.

PR #3156, opened 2026-09-09
[AMD][LoRA] Add a Qwen3-8B LoRA RL launcher

Arist12:
# [AMD][LoRA] Add a Qwen3-8B LoRA RL launcher

## Problem

There is no maintained Qwen3-8B LoRA RL recipe for AMD.

## Change

Adds `scripts/amd/run_qwen3_8b_lora.py`: colocated GRPO over a rank-32 adapter on
MI350X / MI355X, through the Megatron-Bridge path, plus its three snapshot recordings, the
hardware pin the snapshot harness needs for an AMD-only launcher, and the doc row that
indexes LoRA recipes.

One setting is ROCm-specific: `--sglang-attention-backend triton`, because `fa3`, `fa4` and
`flashinfer` have no HIP kernels. Rollout length, batch shape, rank, alpha and learning rate
are taken from the existing recipes (`scripts/amd/run_qwen3_4b.py`,
`scripts/run_qwen3_dense.py`, `scripts/run_qwen3_5_35b_a3b_lora.py`) rather than tuned here.

## Corresponding modifications in the sibling repos

- **SGLang: none.** Exercised at `sglang-miles` @ `a8e5c632f`, which predates the `v0.5.19`
  bump on current `main`.
- **Megatron-Bridge: none.** `0.5.0+582783a0` works as-is.
- **Megatron-LM: none.**

One packaging note, not a code change: the published `rocm/sgl-dev:miles-rocm720-mi35x`
image ships an SGLang predating `/register_lora_adapter`, which miles `main` calls, so LoRA
RL on that image fails at the first weight update with a 404. `docker/Dockerfile.rocm`
already builds from `SGLANG_BRANCH=sglang-miles`, so the image tag only needs a rebuild.

## Validation

8× MI350X (gfx950), ROCm 7.2.0, torch 2.9.1+rocm7.2.0, SGLang `sglang-miles` @ `a8e5c632f`.

### Logprob diff against the published reference

SGLang's own regression test for this model,
`test/registered/lora/test_lora_qwen3_8b_logprob_diff.py`, consumes
[`yushengsu/lora-diff-Qwen3-8B`](https://huggingface.co/datasets/yushengsu/lora-diff-Qwen3-8B)
and is registered CUDA-only. It runs on MI350X with a three-line diff, changing only the
backends that have no HIP kernels:

```diff
- PREFILL_ATTENTION_BACKEND = "fa4"
- DECODE_ATTENTION_BACKEND = "fa4"
+ PREFILL_ATTENTION_BACKEND = "triton"
+ DECODE_ATTENTION_BACKEND = "triton"
-             attention_backend="flashinfer",
+             attention_backend="triton",
```

Base model, `LORA_BACKEND=triton`, `MAX_LORA_RANK=32`, `TP_SIZE=1` and the
`KL_THRESHOLD = 5e-3` assertion are untouched. **`Ran 2 tests ... OK`**:

```
[VERIFY] base vs lora: mean_diff=0.946310, max_diff=14.375022, identical=False
KL(orig_sampler, trainer) = 2.803054e-04
KL(sglang, trainer)       = 2.818798e-04
KL(sglang, orig_sampler)  = 2.795781e-04
```

`KL(sglang, trainer)` is the assertion, against a `5e-3` threshold.
`KL(orig_sampler, trainer)` is the reference dataset's own CUDA-sampler-vs-trainer gap.
ROCm lands within 1% of it (2.819e-04 against 2.803e-04), and the two samplers agree with
each other about as closely as either agrees with the trainer
(`KL(sglang, orig_sampler) = 2.796e-04`), so the ROCm and CUDA errors are strongly
correlated rather than compounding.

On the "< 1e-2" framing in the original request: the mean absolute per-token logprob
difference is **1.02e-2** on ROCm against **1.04e-2** for the reference dataset's own CUDA
sampler. ROCm is marginally the closer of the two, but both sit just above 1e-2 — that gap
is a property of this reference data, not of the backend, and 1e-2 is not reachable on it on
any backend. The `kl_v2 <= 5e-3` figure above is the assertion SGLang's own test makes.

### End-to-end LoRA RL

30 rollouts of GRPO on dapo-math-17k at the launcher's defaults, job `SUCCEEDED`, three
checkpoints. `train/train_rollout_logprob_abs_diff` stays flat at **0.0117–0.0138** across
60 optimizer steps, against miles' own `0.03` gate; `train/train_rollout_kl` ~7e-4.

wandb: https://wandb.ai/arist12/miles-run_qwen3_8b_lora/reports/Qwen3-8B-LoRA-RL-on-8xMI350X-ROCm--VmlldzoxNzg5NzgyNg

Re-scoring the checkpoints on a 128-problem held-out DAPO-math split at 16 samples per
problem (2048 generations per point, graded with miles' own `grade_answer_verl`):

| checkpoint | held-out accuracy | response len | truncated |
|---|---:|---:|---:|
| base | 0.767 | 9820 | 0.191 |
| `iter_0000029` | 0.774 | 9296 | 0.164 |

The accuracy delta is **not significant**: the independent unit is the problem, not the
generation, and at 128 problems the problem-level clustering puts the standard error near
±0.03 rather than the ±0.009 a naive binomial over 2048 generations would give. The run is
reported here as evidence that the recipe executes correctly and stays on-policy, not as
evidence of a learning effect.

## Tests

- `pytest tests/fast/launch_scripts tests/manual/launch_scripts` — 494 passed against 488
  on `main`, i.e. exactly the six new cases; the 12 pre-existing failures and 4 pre-existing
  errors are unchanged.
- `ruff` / `black` / `isort` clean.


Arist12:
Closing while I confirm scope with the maintainer — reopening or resubmitting once we agree on what's wanted.

PR #3159, opened 2026-09-09
Opt into frozen multimodal tower sync with --mm-tower-sync, for any tower name

dzorlu:
## Problem

The weight update can re-send the frozen vision/audio towers from the HF checkpoint, so a colocated engine can release its weights during training and get them all back at the next update. That re-send is gated on the provider path containing `inkling_mm_model_provider` (`_iter_mm_tower_units` in `hf_weight_iterator.py`). Every other VLM that trains only its language model cannot use it, and has to keep the engine weights resident with `--offload-rollout-level kv_cache`.

Measured on B300 (267 GB) nodes with the engine and trainer colocated:

| Model | Engine weights held through log-prob and train | Tower that blocks releasing them |
|---|---|---|
| GLM-5.3-Flash, 6 nodes | 87.35 GB per GPU | 347 tensors, 1.13 GB |
| Qwen3.8-27B, 4 nodes | 73 GB per GPU | 333 tensors, 0.92 GB |

The GLM run OOMs in the log-prob pass with the engine resident. Nothing in the re-send is Inkling-specific except the gate: it already selects any `.visual.` or `.audio.` key from the checkpoint index.

## Change

- `--mm-tower-sync TOWER [TOWER ...]` names the tower modules to re-send. A checkpoint key belongs to a tower when `.TOWER.` appears in `.key`, so `model.visual.blocks.0.attn.proj.weight` and `visual.merger.ln_q.weight` both match `visual`, and `model.visualizer.x` does not.
- `_iter_mm_tower_units` gates on `mm_tower_names(args)` instead of the provider string. The body is unchanged: read the index once, load the tower tensors to CPU, cache, yield after the language weights on every update.
- Inkling: the MM provider still implies `visual audio` when the flag is absent, with a one-time warning asking for the explicit flag. `scripts/run_inkling.py` passes `--mm-tower-sync visual audio`. Passing the flag with no values switches the implicit sync off.

## Compatibility

- Flag absent and provider not Inkling: the function returns at its first line, as before. Text runs are unaffected.
- Inkling launches with only the provider path: same towers as before, plus a warning.
- Engines load the same bytes they loaded at boot, in place. Sending them again is idempotent.

## Test

`tests/fast/backends/megatron_utils/test_mm_tower_sync.py`: default off, flag names the towers, Inkling fallback and its override, key matching, no checkpoint read when off, non-materializing ranks yield nothing, the named towers come back from a two-shard checkpoint, the checkpoint is read once. Ran with the rest of `tests/fast/backends/megatron_utils` and `tests/fast/backends/training_utils/weight_update` inside the miles image: 305 passed.

Downstream use: fleet-ai runs GLM-5.3-Flash and Qwen3.8-27B browser RL with `--mm-tower-sync visual` and the default offload level, replacing `--offload-rollout-level kv_cache` and `--check-weight-update-skip-list visual.`.

### DIFF
--- a/miles/backends/megatron_utils/update_weight/hf_weight_iterator.py
+++ b/miles/backends/megatron_utils/update_weight/hf_weight_iterator.py
@@ -127,17 +127,49 @@ def _gather_pp_full_adapter(
 
 
 _MM_TOWER_CACHE: list[tuple[str, "torch.Tensor"]] | None = None
+_INKLING_MM_PROVIDER = "inkling_mm_model_provider"
+_INKLING_MM_TOWERS = ("visual", "audio")
+_warned_implicit_mm_tower_sync = False
+
+
+def mm_tower_names(args) -> tuple[str, ...]:
+    """Module names of the frozen multimodal towers re-sent on every weight
+    update; empty when tower sync is off. `--mm-tower-sync` decides. The
+    Inkling MM provider implies its two towers so launches that predate the
+    flag keep syncing."""
+    global _warned_implicit_mm_tower_sync
+    names = getattr(args, "mm_tower_sync", None)
+    if names is not None:
+        return tuple(names)
+    if _INKLING_MM_PROVIDER in (getattr(args, "custom_model_provider_path", None) or ""):
+        if not _warned_implicit_mm_tower_sync:
+            logger.warning(
+                "mm tower sync: enabled implicitly by the %s provider; pass --mm-tower-sync %s",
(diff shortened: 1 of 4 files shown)


dzorlu:
Closing in favour of #3165, which carries the same change on the zhichen/glm53-flash branch (#2786), where the GLM-5.3 architecture lives and where downstream images are built from. Once #2786 lands on main, this change will be resubmitted against main in the form here (hf_weight_iterator.py).

Issue #3113, opened 2026-09-04
TITO strict gate trips on model-side behavior (alternate EOS, mid-turn control tokens); consumers must pass the family's parsers


nblintao:
## Summary

Running the TITO strict gate (`--ci-test`) on a real agent harness — terminus-2 × Qwen3-0.6B, the Harbor GPU e2e in #2876 — trips it for three independent reasons. One is a consumer misconfiguration the gate is right to catch; two are model-side behaviors the comparator currently files as hard TITO bugs. Evidence comes from a 2026-09-09 run with `--dump-details`; recipe at the end.

## Mechanism 1 — the consumer did not pass the family's parsers (fixed in #2876)

Without `--sglang-reasoning-parser qwen3` the engine returns no `reasoning_content`; harbor's `Chat(interleaved_thinking=True)` only carries reasoning back into history when that field is non-empty, so the qwen3 template re-rendered an empty `<think></think>` skeleton against a recorded stream with real think tokens → `special_token_count` on every sample. Passing the parsers the session-verify lane uses (`qwen3` / `qwen25`) removes this class. The NeMo-Gym smoke's rate=1.0 (2026-07) most likely had the same cause.

Hardening idea: `TITOTokenizer` subclasses already declare `reasoning_parser` / `tool_call_parser`; a consumer that resolves a tito family could default the sglang flags from it instead of every launcher restating them.

## Mechanism 2 — the model ends some turns with `<|endoftext|>` (fix in progress)

Qwen3-0.6B's `generation_config.eos_token_id` is `[<|im_end|>, <|endoftext|>]` (the second is also its pad token) and the small model samples the second one for some turns. The template can only render `<|im_end|>\n`, and `Qwen3TITOTokenizer` assumes the `<|im_end|>` boundary — `merge_tokens` inserts the missing newline only after it — so the recorded stream and the canonical render disagree structurally:

```
expected: ... [<|im_start|>] (279 tokens) [<|im_end|>] (1 tokens) [<|im_start|>] ...
actual:   ... [<|im_start|>] (279 tokens) [<|endoftext|>]         [<|im_start|>] ...
```

[...]

PR #3161, opened 2026-09-09
tito: Qwen3's second EOS is a turn end, and an alias for the comparator

nblintao:
## Purpose

Stop the TITO strict gate from filing a model choice as a TITO bug: Qwen3 models may end a turn with their second EOS, `<|endoftext|>`, which the chat template can never render (it always emits `<|im_end|>\n`). Mechanism 2 of #3113, with the dumped evidence there.

## What

- `Qwen3TITOTokenizer`: both EOS ids are turn ends for the `\n` boundary in `merge_tokens` (previously only `<|im_end|>`), and the pair is declared as an EOS alias group. The Qwen3.5 / 3.6 / 3.8 / Next subclasses inherit.
- `TokenSeqComparator`: new `eos_alias_groups` parameter; an aligned special segment whose two ids fall in one group is reported as the new non-severe type `eos_alias` instead of `special_token_type`.
- `metrics.py`: logs `eos_alias` per type. The strict tuple (`special_token_count` / `special_token_type` / `non_assistant_text`) is unchanged.
- Tests: comparator alias classification; Qwen3 `merge_tokens` after `<|endoftext|>`; an end-to-end case through the family's comparator where an `<|endoftext|>`-ended turn yields exactly one `eos_alias`; metrics logging/non-strictness.

## Behaviour change

Sessions where the model ended a turn with `<|endoftext|>` no longer fail the ci_test strict gate; they show up as `tito_session_mismatch_rate/<v>/eos_alias`. Training tokens are untouched.

## Verification

Stage 2: `verify_chat_template.py --model Qwen/Qwen3-0.6B --tito-model qwen3 --thinking both` 74/74 PASS. Stage 3 (2026-09-09, 2×H200): the Harbor e2e of #2876 with the strict gate armed passes on a healthy terminus-2 × Qwen3-0.6B session (`assistant_text` ×2, `eos_alias` ×1). A degenerate session still trips `non_assistant_text` for an unrelated reason — the model emitting `<|im_start|>` mid-turn — tracked as mechanism 3 in #3113, deliberately not in this PR.

## Follow-up

#2876 keeps `--ci-disable-tito-strict-checker` until mechanism 3 of #3113 is also resolved; then it drops the flag and its plumbing and reruns the e2e with the gate armed.

### DIFF
--- a/miles/ray/rollout/metrics.py
+++ b/miles/ray/rollout/metrics.py
@@ -119,7 +119,13 @@ def _compute_metrics_from_samples(args, samples):
         assert session_server_version in ("v1", "v2"), "TITO metrics require session server v1 or v2"
         metric_prefix = f"tito_session_mismatch_rate/{session_server_version}"
         log_dict[metric_prefix] = np.mean([len(v) > 0 for v in tito_vals]).item()
-        for mtype in ("special_token_count", "special_token_type", "non_assistant_text", "assistant_text"):
+        for mtype in (
+            "special_token_count",
+            "special_token_type",
+            "non_assistant_text",
+            "assistant_text",
+            "eos_alias",
+        ):
             log_dict[f"{metric_prefix}/{mtype}"] = np.mean(
                 [any(m.get("type") == mtype for m in v) for v in tito_vals]
             ).item()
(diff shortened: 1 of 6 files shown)


nblintao:
Closing. The consumer that needed this (#2876) now runs the recipe's model rollout-only with the strict gate armed and clean, and the TITO owner's read is that ending a turn with `<|endoftext|>` is a weak-model symptom rather than something TITO should absorb. The branch and the analysis in #3113 stay available if a production Qwen3 model shows this at a rate worth handling.

PR #3163, opened 2026-09-09
Fix missing Nemotron-H Mamba convolution mappings

Shi-Dong:
Recent Megatron versions expose Mamba convolution parameters as `mixer.conv1d_weight` and `mixer.conv1d_bias`, while the Nemotron-H bridge maps the older dotted names. The missing mappings leave those tensors unloaded and omit them from rollout weight updates.

Add the current names using `MambaConv1dMapping`, preserving its x/B/C-aware tensor-parallel conversion and the legacy mappings. Skip aliases already present upstream. This PR contains only that compatibility fix, a short model-doc note, and focused regression tests.

It also re-enables `tests/e2e/megatron/model_scripts/test_nemotron_3_ultra_4layer_ci.py`, which was disabled for exactly this missing mapping and is the end-to-end acceptance test for it (labels `run-ci-megatron`, `run-ci-model-scripts`).

Validation:
- Six regression cases pass in the full Miles pytest setup; the unpatched base fails four of them. Covers weight/bias aliases, both registry attribute variants, and existing/partially existing upstream mappings.
- During model bring-up, the same mapping fix changed a full HF → Megatron → HF audit from 6,197 matching tensors plus 46 missing tensors to all 6,243 matching tensors with no missing or duplicate names. Passed at TP1/PP1/EP1 and TP2/PP2/EP2/ETP1, comparing against the source checkpoint after dtype conversion to BF16.
- All applicable pre-commit hooks pass.

The checkpoint audit used NVIDIA-Nemotron-3-Nano-30B-A3B-BF16 with Megatron-LM `8c1e05747` and Megatron-Bridge `582783a0`. Full training smoke validation with the combined fixes remains pending.


Shi-Dong:
@claude review always

Zhichenzzz (approved):

PR #3164, opened 2026-09-09
Fix Nemotron-H forward compatibility when MTP is disabled

Shi-Dong:
The Nemotron-H forward shim always passes `mtp_labels`, including `None`. Megatron's `HybridModel.forward` does not accept that keyword, so an MTP-disabled run fails on its first training forward.

Only forward MTP labels when they are present. Preserve the existing next-token shift and loss-mask forwarding, and correct the shim's stale docstring.

Validation:
- Four regression cases pass in the full Miles pytest setup; the unpatched base fails the three MTP-disabled cases.
- Tests cover absent/empty/None MTP labels, a strict forward signature, loss-mask propagation, one-time label shifting, input preservation, and idempotent shim installation.
- All applicable pre-commit hooks pass.

The non-None label test uses a forward implementation that accepts MTP labels; it does not claim end-to-end MTP-enabled model support.


Shi-Dong:
@claude review always

Zhichenzzz (approved):

PR #3171, opened 2026-09-09
feat (trainer): trainer refactor with torchtitan backend support

Zhichenzzz:
Adds torchtitan as a third training backend, and unifies what the torch-native backends (FSDP, torchtitan) had each written on their own into one actor, one step-runner seam and one weight-update path.

## The problem

Before this PR the FSDP actor carried its own copy of the RL step (fetch rollout, reference and actor log-prob passes, advantages, optimizer loop), its own weight-sync classes (`update_weight_utils.py`, ~300 lines duplicating the engine handshake Megatron already had in `weight_update/`), its own host offload, and its own rank-serialized config/tokenizer loading. Adding torchtitan the same way would have meant a third copy of all of it. The refactor pulls each of those into the layer that actually owns it, and torchtitan is built on the result.

## Where things live now

```
miles/ray/train_actor.py            TrainRayActor: load_hf_assets()        <- all three backends
miles/backends/training_utils/
  torch_native/actor.py             TorchNativeTrainRayActor: the RL step  <- FSDP, torchtitan
  torch_native/step_runner.py       StepRunner protocol, LinearStepRunner
  torch_native/routing_replay.py    manager surface for R3 (stages, fill, ...)
  parallel.py                       ParallelState.is_metrics_rank
  weight_update/                    WeightUpdater, transports, HfWeightIteratorBase.build
miles/backends/fsdp_utils/          actor (model build only), hf_weight_iterator
miles/backends/torchtitan_utils/    actor, trainer, config, loss, components, compat,
                                    weight_bridge, routing_replay (install only), parallel
```

### The actor base and the step-runner seam

`TorchNativeTrainRayActor` owns the RL step for any backend whose model is a set of torch modules: `train` → `_train_core` runs the reference and actor `_log_probs` passes under the routing-replay stages, computes advantages, runs `_optimizer_steps`, and pushes weights through `update_weights`. Host offload (`sleep`/`wake_up`/`_move_to`) and the `WeightUpdater` construction live here too. A subclass builds its model in `init` and fills in the provider surface declared on the class (`model_parts`, `optimizers`, `weight_updater`, `prof`, `hf_config`, optionally `ref_runner` and `align_token_side_channel`).

The backend's own training step is reached through `StepRunner`: `forward_only_step(batches, compute)`, `forward_backward_step(batches, loss_closure)`, `zero_grad()`, `apply_step() -> StepMetrics`. Its unit is one optimizer step's worth of microbatches, not one forward, because under pipeline parallelism the schedule owns the microbatch ordering and a per-microbatch hook cannot exist there (pytorch/torchtitan#3856 settled on the same seam). `batches` is an iterator so a linear runner keeps fetch and compute interleaved while a schedule-based runner drains it up front. `LinearStepRunner` is the no-schedule implementation FSDP uses; `TrainerStepRunner` adapts torchtitan's trainer.

Metrics are logged from `ParallelState.is_metrics_rank` (last pipeline stage, tp 0, dp-cp 0), which is the rank experiment tracking was initialized on. Global rank 0 is a first-stage rank under PP and holds no loss.

### One weight-update path

The FSDP export becomes an `HfWeightIteratorBase` (`fsdp_utils/hf_weight_iterator.py`): gather each shard to a full tensor, apply the registered per-architecture transform, cast fp32 masters to the engine's dtype. That is the same protocol Megatron's iterators implement, so all three backends now go through `WeightUpdater` with the same bucketing, atomic groups, transports and engine session. `HfWeightIteratorBase.build` is the one factory: it resolves the protocol's required placement against the class's forced placement, so a backend hands `WeightUpdater` the class method instead of a function of its own. The two pieces every backend's `update_weights` repeated around the updater, reconnecting when the engine cell snapshot changes and verifying the engine's weight version under `--ci-test`, are `WeightUpdater.reconnect_if_needed` and `verify_engine_version`.

### Routing replay

The manager-facing surface of R3 (the four stage constants, `enable`, `fill`, `log_prob_stage`, `stage`, `rewind`, `reset`) was identical in the FSDP and torchtitan modules; it is one implementation in `training_utils/torch_native/routing_replay.py`, which the actor base uses directly. Each backend module keeps only its own `install`: how it finds the routers and rebinds their forward.

### torchtitan: the Trainer is the black box

`TitanTrainer` subclasses torchtitan's `Trainer` without touching construction. `Trainer.__init__` owns the whole build (parallel dims, model, optimizer, PP schedule, checkpointer) and `forward_backward_step` hides the PP/non-PP split; miles only translates the step-runner protocol onto that machinery. The pieces:

- `config.py` builds one `Trainer.Config` from miles arguments. The parallelism flags are torchtitan's own field names copied verbatim. A tied-embedding checkpoint is refused under PP: torchtitan would train a separate `lm_head` the HF export has no tensor to ship into.
- `loss.py` (`RLLossAdapter`) is miles' RL loss as torchtitan's `loss_fn`. The schedule hands it one `(pred, target)` per microbatch; the target carries the microbatch index so the adapter can find its batch, and under CP it gathers the logits back and undoes the balancer's permutation before the loss sees them.
- `components.py` stands in for the dataloader (the RL loop feeds microbatches directly) and the checkpointer (its HF load tolerates the `lm_head.weight` a tied checkpoint does not ship).
- `weight_bridge.py` streams the trainer's shards as HF-named full tensors through the model's own `state_dict_adapter`, reassembled across dp/tp/ep and, when the protocol's placement asks for it, completed across pipeline stages one tensor at a time. Multimodal towers the trainer never holds are streamed from the checkpoint files. Tensors go out in the dtype the checkpoint holds, not the fp32 masters.
- `routing_replay.py` rebinds `TokenChoiceTopKRouter.forward` with the manager's topk. A pipeline schedule runs a real forward and backward over microbatch 0 to infer stage shapes, on every eval-to-train switch; anything consumed per call would be shifted by one microbatch, so the initialization window is bypassed and every pass asserts the streams advanced by exactly the microbatch count.
- `parallel.py` maps torchtitan's meshes onto `ParallelState`. Context parallelism is deliberately reported as 1: the trainer shards the sequence and gathers the logits back itself.
- `compat.py` carries three feature-detected torch backports that each become a no-op once the pinned torch has the upstream form. Everything torchtitan re-exports is imported through torchtitan, never torch.

The package `__init__` is empty, so the arguments and the parallel-state mapping are importable on a machine without torchtitan; `config.py`, the first module that does import it, installs `compat` itself.

## What each backend still owns

- **Megatron**: unchanged in shape; it calls `load_hf_assets` and the two updater helpers.
- **FSDP**: building the HF model on rank 0 and loading it by broadcast, the precision policy, the reference model parked on the host, DCP checkpoints.
- **torchtitan**: the config tree, the loss adapter, the two component stand-ins, the weight bridge, the router rebinding.

## Scope

Supported on torchtitan: tp, pp, cp (headtail balancer), ep, dp-replicate, activation checkpointing, MoE with R3, colocated and disaggregated rollout, fully-async, every transport, a reference model. Not in this PR: LoRA, optimizers other than adam, on-policy distillation, checkpoint healing.

## Behaviour fixes folded in after review

- **LR schedule.** `Trainer.Config.lr_scheduler` was never set, so torchtitan's defaults ran (200 warmup steps, linear decay to 0) regardless of `--lr-decay-style` / `--lr-warmup-iters`. The config now maps them: `warmup_steps` from `--lr-warmup-iters`; `constant` pins `min_lr_factor = 1.0`; `linear` / `cosine` set `decay_type` and `min_lr_factor = min_lr / lr`. Styles torchtitan cannot express (WSD, inverse-square-root, warmup fraction, decay iters) are rejected at argument validation, as is `--fp16`.
- **Resume path.** The checkpoint folder derives from `--load`, then `--save`, then a fresh temp dir; it no longer defaults to `./outputs` in the actor's working directory, where a stale `step-N` would have been resumed silently.
- **Pipeline batch shape** is validated at config time instead of failing on GPU after model build.
- **`--start-rollout-id 0`** is honoured (it was treated as unset).
- **FSDP weight export** keeps the async all-gather overlap from `update_weight_utils` (a byte-bounded window of in-flight gathers) instead of gathering one parameter at a time.
- **p2p protocol** loads each CPU replica under its own `ParallelismContext`; previously every replica but the last loaded under whichever context `_create_cpu_replica` left behind. This is a Megatron-visible fix and is covered by the weight-update CI suite.
- Megatron's tower cache for Inkling now uses the shared `checkpoint_towers` module; `verify_engine_version` skips only when no sync has happened yet (weight version 0), which the Megatron `ci_test` path did not need before the delta protocol's baseline capture.
- **Gradient scale.** torchtitan disables FSDP's gradient division and every CP rank backpropagates the full gathered loss, so the shared loss's `× dp` convention (written for FSDP2's averaging) inflated torchtitan gradients by `dp × cp`. The loss adapter now scales the training loss by `1 / (dp_replicate × dp_shard × cp)`; logged losses are unchanged.
- **Checkpoints are resumable.** Saves write the full trainer state (`last_save_model_only=False`); `--load` resolves the newest `step-N` under its folder into `initial_load_path` while `--save` alone owns the dump folder, so the two never alias.
- **Reference trainer** applies the same loss-reduction setup as the actor, including the CP logits gather.
- `weight_bridge.py` is now `hf_weight_iterator.py`, matching the FSDP and Megatron modules.


docs/user-guide/training-backend.md
```diff
@@ -440,6 +449,118 @@ For profiling: `--use-pytorch-profiler` with `--profile-step-start` / `--profile
...
+kernel later.
+
+### 2. Choosing the parallelism
+
+```bash
+--titan-tensor-parallel-degree 2 \
+--titan-pipeline-parallel-degree 2 \
+--titan-context-parallel-degree 2 \
+--titan-expert-parallel-degree 2 \
+--titan-data-parallel-replicate-degree 1
```
HJSang, line 490: Can we uniform those params between titan and megatron and then parse them to trainer based on different backend? Then we don't have to change them if we just want to switch the trainer backend but keep the parallelism params. WDYT?

HJSang: Not a block

miles/backends/torchtitan_utils/hf_weight_iterator.py
yueming-yuan, line 1: very tiny nit: use "weight_iterator.py" as file name to keep consistent?

miles/backends/fsdp_utils/hf_weight_iterator.py
```diff
@@ -0,0 +1,59 @@
...
+            if pending_bytes >= self.args.update_weight_buffer_size:
+                yield from self._drain(pending, materialize=materialize)
+                pending_bytes = 0
+        yield from self._drain(pending, materialize=materialize)
+
+    def _drain(self, pending: deque, *, materialize: bool) -> Iterator[list[tuple[str, torch.Tensor]]]:
+        model_type = self.model.config.model_type
+        while pending:
+            name, param, full = pending.popleft()
+            if isinstance(param, DTensor):
```
yueming-yuan: codex cmt:

[P1] Preserve synchronous gather results — [fsdp_utils/hf_weight_iterator.py:35](https://github.com/radixark/miles/pull/3171/files)
Single-rank FSDP’s gather_full_param() returns a regular tensor through full_tensor(). Checking the input’s DTensor type then unconditionally calling full.wait() breaks the first weight update with AttributeError: 'Tensor' object has no attribute 'wait'. Reproduced with a real singleton DTensor; the previous drain handles this correctly.

miles/backends/torchtitan_utils/actor.py
```diff
@@ -0,0 +1,128 @@
...
+        finally:
+            for part in self._ref_parts:
+                part.cpu()
+            torch.cuda.empty_cache()
+
+    def save_model(self, rollout_id: int, force_sync: bool = False) -> None:
+        if self.args.debug_rollout_only or self.args.save is None:
+            return
+        assert not self.args.async_save, "TorchtitanTrainRayActor does not support async_save yet."
+        self.trainer.checkpointer.save(self.trainer.step, last_step=True)
```
yueming-yuan, line 129: codex cmt:

[P1] Save complete trainer state for resume — [torchtitan_utils/actor.py:128](https://github.com/radixark/miles/pull/3171/files)
Every save passes last_step=True, while the pinned torchtitan defaults last_save_model_only=True. Consequently, checkpoints omit optimizer, scheduler, and trainer state. Loading any positive checkpoint step requests all three, so these checkpoints cannot resume training. Executing the pinned save/load-selection methods confirms the missing states; configure full-state saves.

miles/backends/torchtitan_utils/actor.py
```diff
@@ -0,0 +1,128 @@
...
+        return args.start_rollout_id if args.start_rollout_id is not None else start_rollout_id
+
+    def step_runner(self):
+        return self.trainer.step_runner()
+
+    def _build_ref_runner(self, args: Namespace):
+        if not args.ref_load:
+            raise ValueError("--ref-load is required to build a torchtitan reference model")
+        ref_config = build_trainer_config(args, hf_assets_path=args.ref_load, lr_total_steps=1, dump_subdir="ref")
+        ref_trainer = TitanTrainer(ref_config)
```
yueming-yuan, line 102: codex cmt:

[P2] Enable CP gathering on the reference trainer — [torchtitan_utils/actor.py:102](https://github.com/radixark/miles/pull/3171/files)
_build_ref_runner() never calls enable_context_parallel_gather(), although its trainer still shards inputs over CP. The reference loss therefore receives local, permuted logits while Miles assumes the full sequence. For a 512-token sequence with CP=2 and a 128-token response, the existing response slicer produces zero logit rows for 128 target tokens. CP with a reference model fails or computes incorrect reference probabilities.

miles/backends/torchtitan_utils/config.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    config = Trainer.Config()
+    config.model_spec = resolve_model_spec(args)
+    if ties_embeddings:
+        config.model_spec.model.enable_weight_tying = True
+        logger.info("Checkpoint ties lm_head to the embedding; excluding lm_head.weight from the HF export")
+
+    config.hf_assets_path = hf_assets_path
+    dump_root = args.load or args.save or tempfile.mkdtemp(prefix="miles-torchtitan-")
```
yueming-yuan: codex cmt:

[P2] Separate checkpoint input and output directories — [torchtitan_utils/config.py:53](https://github.com/radixark/miles/pull/3171/files)
With --load /checkpoints/source --save /checkpoints/new, args.load wins and becomes the checkpointer’s shared load/save directory. Subsequent saves write into the source directory and ignore --save; torchtitan’s default retention can also purge source checkpoints. Configure loading independently from the output folder.

miles/backends/torchtitan_utils/loss.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if self._cp_mesh is not None:
+            pred = self._gather_context_parallel(pred)
+
+        index = int(target.flatten()[0])
+        batch = self._batches[index]
+        if self._is_training:
+            loss, log_dict = self._closure(pred, batch)
+            self._results[index] = log_dict
+            return loss, {}
```
HJSang: `_step_loss` multiplies the sample-mean loss by `DP`, assuming averaged gradients, but torchtitan [disables gradient averaging](https://github.com/pytorch/torchtitan/blob/2f10a2590c76e77d6b732dd0229ab59964535318/torchtitan/distributed/fsdp.py#L27-L41). CP also computes the full gathered loss on every rank and sums those copies during backward, so returning this loss unchanged inflates pre-clip gradients and grad norms by `DP × CP`.

docs/user-guide/training-backend.md
Zhichenzzz: Done in the stacked #3248: torchtitan now takes Megatron's names (--tensor-model-parallel-size, --pipeline-model-parallel-size, --context-parallel-size, --expert-model-parallel-size, --seq-length), and FSDP/torchtitan also accept --recompute-granularity full and --bf16 with Megatron's meaning. --titan-model-name/--titan-model-flavor stay since they name a torchtitan model package.

miles/backends/torchtitan_utils/hf_weight_iterator.py
Zhichenzzz: Renamed to hf_weight_iterator.py.

miles/backends/fsdp_utils/hf_weight_iterator.py
Zhichenzzz: Right, the single-rank path returns a plain tensor from full_tensor(). The drain now waits only when the gather handed back an AsyncCollectiveTensor.

miles/backends/torchtitan_utils/actor.py
Zhichenzzz: Fixed: last_save_model_only=False, so saves carry optimizer, scheduler and trainer state.

miles/backends/torchtitan_utils/actor.py
Zhichenzzz: Fixed: the reference trainer runs the same configure_loss_reduction() as the actor, which installs the CP logits gather (and the gradient scale below).

miles/backends/torchtitan_utils/config.py
Zhichenzzz: Fixed: --save (or a temp dir) owns the dump folder; --load only feeds initial_load_path with its newest step-N, with initial_load_model_only=False. Covered by test_resume_reads_from_load_and_writes_to_save.

miles/backends/torchtitan_utils/loss.py
Zhichenzzz: You are right, thanks. torchtitan sets gradient_divide_factor=1 and each CP rank backpropagates the full gathered loss, so the shared loss's x dp convention inflated gradients by dp x cp. The loss adapter now scales the training loss by 1/(dp_replicate x dp_shard x cp); logged losses are unchanged. Test: test_the_loss_adapter_undoes_the_summed_dp_and_cp_gradients.

miles/backends/training_utils/torch_native/actor.py
```diff
@@ -0,0 +1,277 @@
...
+    def sleep(self) -> None:
+        if self.args.offload_train:
+            self._move_to("cpu")
+
+    @timer
+    def wake_up(self) -> None:
+        if self.args.offload_train:
+            self._move_to("cuda")
+
+    def _move_to(self, device: str) -> None:
```
Rockdu: Do we want to pin weights to host memory when offloading as well?

PR #3172, opened 2026-09-09
[feat] snapshot eval during --debug-train-only by reusing the eval dispatcher

Zhichenzzz:
## What

Let a pure-SFT `--debug-train-only` run evaluate its checkpoints with the existing snapshot-eval machinery (`EvalDispatcher` / `EvalFleet`), instead of adding a train-only-specific eval path. Supersedes #2882.

## How

- `sglang_config.starts_inference_engines(args)` decides whether any engine starts: none for plain train-only, only the eval fleet when `--eval-num-gpus > 0`. The inference specs, `InferenceController.init` and the rollout GPU offset follow the config instead of carrying their own `debug_train_only` short-circuits.
- `train.py` drives eval through `EvalDispatcher` in every mode (as `train_async.py` already does), so train-only gets snapshot export, async dispatch and snapshot GC for free.
- `EvalDispatcher._export` onloads/offloads the actor around the export under `--offload-train`, mirroring `save_training_model`.
- The http client is created whenever snapshot eval is on, sized from the eval fleet when there are no rollout GPUs.
- `miles_validate_args`: train-only sets `rollout_num_gpus = 0`; snapshot eval rejects `--load-debug-rollout-data` (no rollout functions are loaded) and requires an explicit, fleet-aware `--eval-function-path` under `--debug-train-only` (the SFT rollout function cannot evaluate).

Net: +48 lines of production code, no new modules.

## Validation

8xH200, Qwen3-4B-Base, gsm8k SFT (`miles.rollout.sft_rollout.generate_rollout`, `--loss-type sft_loss`), 4 actor GPUs (TP2) + `--eval-num-gpus 4`, `--eval-interval 2`, `--eval-function-path miles.rollout.inference_rollout.inference_rollout_common.InferenceRolloutFn`, `--num-rollout 4`. Both entrypoints finish with `SUCCEEDED`, snapshots `step_1` / `step_3` exported and the pre-train eval runs against `--hf-checkpoint`:

| entrypoint | eval 0 (base) | eval 1 | eval 3 |
|---|---|---|---|
| `train_async.py` | 0.359 | 0.306 | 0.193 |
| `train.py` | 0.376 | 0.331 | 0.202 |

(The downward curve is the synthetic SFT data carrying gsm8k's `<<a/b=c>>` calculator annotations, which the math parser then trips on; train loss goes 0.58 → 0.33 as expected.)

New e2e `tests/e2e/short/test_qwen3_0.6B_sft_snapshot_eval.py` (stage-c-2-gpu-h200, labels short/eval/megatron): Qwen3-0.6B, one actor GPU + one-GPU eval fleet, metric checker demands exactly three eval points, so any train-only short-circuit on the engine or eval path fails it. Passes on H200 (gsm8k 0.590 / 0.570 / 0.582).

Fast tests: `tests/fast/{utils/test_arguments,utils/test_http_utils,rollout/test_checkpoint_eval,backends/sglang_utils/test_sglang_config,ray/rollout/test_rollout_server,ray/test_placement_group_shared_ppo,test_train,ray/rollout/test_eval_dispatch}.py` all pass.

## Notes

- A fleet-aware eval function is required: the legacy `miles.rollout.sglang_rollout.generate_rollout` ignores `RolloutFnEvalInput.generate_state` and sizes its own semaphore from `args.rollout_num_gpus`, so it cannot target the eval fleet (pre-existing).
- Under train-only, `--apply-chat-template` would hand `sft_rollout` a rendered string instead of the message list, so eval prompts need a pre-rendered column via `--eval-input-key`.


yueming-yuan (approved):
LGTHuman, with one codex comment

yueming-yuan:
[P2] Use the completed rollout’s version for resume eval — [train.py:108](https://github.com/radixark/miles/pull/3172/files)
On resume, start_rollout_id is the loaded checkpoint’s rollout plus one. Dispatching that next ID has two consequences: checkpoint-reuse mode requests a nonexistent checkpoint, producing ckpt_missing; staging mode exports under the same step_N directory that the first resumed training iteration may subsequently overwrite. With --eval-interval 1, I reproduced both pending evaluations reading the post-training weights, losing the intended pre-training measurement. Use the loaded checkpoint’s completed rollout ID.

Issue #3158, opened 2026-09-09
Where should TRITON_CACHE_DIR live for multi-node runs? Recipes disagree and the framework is silent


David553:
## The question

Where should `TRITON_CACHE_DIR` live for a multi-node run? Today each recipe decides on its own, the two that decide disagree with the default, and nothing in `miles/` sets or documents one — so the first training step of a multi-node run costs whatever the launcher happened to pick.

Current state in the tree:

- `examples/experimental/openenv/glm52_tbench2/run_glm5_2_744b_a40b_daytona.py:315-318` pins the caches node-local, with the rationale in the comment: *"Node-local caches: the NFS defaults (`~/.triton`, `~/.cache/tvm-ffi`) race across nodes under many-process cold compiles."*
- `docs/models/qwen/qwen3-8-flash-next.md:238` pins Triton and Inductor caches to `/tmp` the same way, plus `TORCHINDUCTOR_COMPILE_THREADS=1`.
- `grep -rn TRITON_CACHE miles/ miles_plugins/` → nothing. Every other recipe inherits `~/.triton`, wherever that lands in the image or on the cluster.

## Why it matters more than it looks

The cost of an unwarmed cache lands on the first training step after a rollout, where it is easy to mistake for a hang. On 2×8 H20, Qwen3.6-35B-A3B (TP2/PP2/EP8, colocate), with `~/.triton` container-local so every node started cold:

| pass | cold | warm |
|---|---|---|
| first `ref_log_probs` | 394 s | 13 s |
| first `actor_train` | 639 s | 60 s |

While the first pipeline stage compiles, the next one sits in `recv_forward`, and it is that stage's NCCL watchdog that fires. #3132 documents how to tell the two apart.

## What I measured about the race

The node-local pin exists to avoid cross-node races, so I tried to reproduce them. 16 ranks across two nodes, barrier-aligned (clock skew 0.17 s), all compiling the same `fla` GDN kernels concurrently into one NFS `TRITON_CACHE_DIR`:

| round | shared NFS cache | node A (8 ranks) | node B (8 ranks) | errors |
|---|---|---|---|---|
| 1 | cold, 16-way concurrent write | 144.2–144.4 s | 158.8–159.5 s | 0 |
| 2 | warm | 0.70 s | 0.72–0.74 s | 0 |
| ref | single process, node-local, cold | 109.1 s | — | 0 |

[...]

Issue #807, opened 2026-03-26
Bug fixes: rollout engine dist_init_addr wrong-node assignment, numpy 2.x assertion, deep_ep tms flag corruption


ashutoshuiuc:
1. miles/ray/rollout.py — SGLang engine dist_init_addr assigned to wrong node _allocate_rollout_engine_addr_and_ports_normal uses node_index = local_rank // num_engines_per_node to determine which node an engine is on. This assumes logical rank order matches physical node placement, which breaks
  when Ray schedules engines differently. All engines can get node 0's IP for dist_init_addr, causing TCP timeouts on engines physically on node 1. Fix: query each engine's actual IP via _get_current_node_ip_and_free_port.remote() and group by physical node.

2. miles/backends/megatron_utils/initialize.py — numpy 2.x assertion
  Hard assert np.__version__.startswith("1.") blocks numpy 2.x. The linked Megatron-LM issue (#1563) was about np.product() removal, but the current Megatron fork doesn't use it. Safe to remove.

3. miles/backends/megatron_utils/__init__.py — deep_ep monkey-patch corrupts tms_interesting_region flag
  The deep_ep Buffer.__init__ wrapper unconditionally sets tms_set_interesting_region(True) after init, but the flag may have been False before (e.g. during non-interesting allocations). Fix: save/restore the previous
  value instead of hardcoding True.

I'll raise a PR soon for this, just wanted to get feedback before i do.

behroozazarkhalili:
Two of the three are fixed on main:

- numpy 2.x assertion: removed in 00a698ea7 (#1587, sglang bump); the `assert np.__version__.startswith("1.")` line in miles/backends/megatron_utils/initialize.py is gone.
- dist_init_addr wrong-node assignment: a8ee4a59a (#1875) replaced allocate_rollout_engine_addr_and_ports_normal with a per-engine node-IP lookup in miles/ray/rollout/server_group.py, so the address comes from the engine's actual node.

Still open: miles/backends/megatron_utils/__init__.py L12-17 hardcodes tms_set_interesting_region(True) after deep_ep.Buffer.__init__ instead of restoring the prior value. This is the only call site in miles/, and I did not find a read-back API in torch_memory_saver, so a fix would have to track the flag on the Python side. If a maintainer confirms there is no getter, I can open that as a small change with a mocked test.

cc @fzyzcjy @yueming-yuan

PR #3177, opened 2026-09-10
fix(megatron_to_hf): name bridge mode when a raw convert hits VL wrapper params

### LINKED ISSUE #634: SFT Post Training for "Qwen3-VL-8B-Instruct" fails.

behroozazarkhalili:
Fixes #634.

### What

`_convert_to_hf_core` wraps the per-model dispatch. When a converter raises `ValueError` for a parameter whose name carries one of the megatron.bridge VL wrappers (`vision_model.`, `vision_projector.`, `language_model.`), the error keeps the converter's "Unknown parameter name" text and adds the flag to pass, `--megatron-to-hf-mode bridge`, as the geo3k_vlm launchers do. Converters that already map those wrappers (kimi_vl) are untouched, and decoder names still reach their converters unchanged.

An SFT run on Qwen3-VL-8B left on the default raw mode died in the converter with no pointer to the supported path. Qwen3-VL is bridge-only (miles_plugins/models/qwen3_vl.py), and the only place that said so was one flag inside the geo3k_vlm launchers.

### Test

`tests/fast/backends/megatron_utils/test_megatron_to_hf_bridge_only_hint.py`: three wrapper names raise with the flag in the message and the original text preserved; a Kimi-VL `vision_model` name still converts through its own converter; a plain decoder name still converts.

### Verification

- The six cases pass; with `megatron_to_hf/__init__.py` restored from main the three hint cases fail.
- `tests/fast/backends/megatron_utils`: 289 passed, 11 skipped.
- ruff, black 24.3.0 and isort at the pre-commit pins are clean on both files.

### Scope

- `tools/convert_torch_dist_to_hf.py` has no `--megatron-to-hf-mode` flag, so the checkpoint-conversion path in the second report on #634 still cannot pick bridge mode. Separate change.

cc @yueming-yuan @Zhichenzzz

### DIFF
--- a/miles/backends/megatron_utils/megatron_to_hf/__init__.py
+++ b/miles/backends/megatron_utils/megatron_to_hf/__init__.py
@@ -29,8 +29,30 @@ def convert_to_hf(args, model_name, name, param, quantization_config=None):
     return quantize_params(args, name, converted_named_tensors, quantization_config)
 
 
-# TODO optimize code details
+# Wrapper prefixes that megatron.bridge-built VL models produce. A converter that knows them (kimi_vl) maps
+# them; for every other model the raw converter raises a bare "Unknown parameter name" that hides the real
+# cause, so _convert_to_hf_core adds the fix to that error instead of guessing up front.
+_BRIDGE_ONLY_PREFIXES = ("vision_model.", "vision_projector.", "language_model.")
+
+
+def _bridge_mode_hint(error: ValueError) -> ValueError:
+    return ValueError(
+        f"{error}. Raw-mode conversion does not support the vision_model / vision_projector / language_model "
+        "wrappers that megatron.bridge VL models use; pass --megatron-to-hf-mode bridge (see examples/geo3k_vlm)."
+    )
+
+
 def _convert_to_hf_core(args, model_name, name, param):
+    try:
+        return _dispatch_to_model_converter(args, model_name, name, param)
+    except ValueError as error:
+        if any(prefix in name for prefix in _BRIDGE_ONLY_PREFIXES):
+            raise _bridge_mode_hint(error) from error
+        raise
+
+
+# TODO optimize code details
+def _dispatch_to_model_converter(args, model_name, name, param):
     model_name = model_name.lower()
     if (
         "glm4moelite" in model_name
--- a/tests/fast/backends/megatron_utils/test_megatron_to_hf_bridge_only_hint.py
+++ b/tests/fast/backends/megatron_utils/test_megatron_to_hf_bridge_only_hint.py
@@ -1,0 +1,38 @@
+"""Raw-mode conversion of a megatron.bridge VL checkpoint must say how to fix it (#634)."""
+
+from argparse import Namespace
+
+import pytest
+import torch
+
+from miles.backends.megatron_utils.megatron_to_hf import _convert_to_hf_core
+
+
+@pytest.mark.parametrize(
+    "name",
+    [
+        "module.module.vision_model.patch_embed.proj.weight",
+        "module.module.vision_projector.encoder.linear_fc1.weight",
+        "module.module.language_model.embedding.word_embeddings.weight",
+    ],
+)
+def test_bridge_only_param_names_point_at_bridge_mode(name):
+    args = Namespace(hidden_size=4, kv_channels=2, num_attention_heads=2, num_query_groups=1)
+    with pytest.raises(ValueError, match=r"--megatron-to-hf-mode bridge") as excinfo:
+        _convert_to_hf_core(args, "qwen3", name, torch.zeros(1))
+    assert f"Unknown parameter name: {name}" in str(excinfo.value)
+
+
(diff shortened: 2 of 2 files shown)


behroozazarkhalili:
@fzyzcjy @yueming-yuan one note instead of nineteen pings.

I have 19 PRs open, all rebased onto today's main and green on the CPU stages, but every one of them has the GPU stages skipped because none carries a `run-ci-*` label. The contributor guide says that label is what approves fork CI, so I think they are stuck one click short of the pipeline that would tell you whether they are worth your time. A label on any of them is all I am asking for here; review can follow once the GPU stages have spoken.

Grouped by area so one pass can take a batch. Each fixes a filed issue or ports a known technique, and each body carries the exact verification that was run.

**Bugs reported by users this month**

| PR | Fixes | What |
|---|---|---|
| #3279 | #3270 | AMD GLM-5.2 launcher called an exec helper that does not exist; adds a scripts-wide check |
| #3280 | #3265 | Rollout metrics crashed on dict rewards; also closes a silently wrong pass@k |
| #3281 | #3271 | FSDP checkpoint step counters were always zero |
| #3282 | #3277 | Retract-mode routing replay warned only with the session server on; retract is the default |
| #3303 | | Follow-up to #3280: nonzero-std filter fails clearly on dict rewards |

**Correctness fixes from the earlier bug sweep**

| PR | Fixes | What |
|---|---|---|
| #3177 | #634 | Raw megatron-to-hf convert names bridge mode when it hits VL wrapper params |
| #3178 | #956 | Session-not-found responses say why and which server answered |
| #3185 | #807 | Memory saver's region flag restored after deep_ep init |
| #3190 | #634 | convert_torch_dist_to_hf can export a bridge checkpoint |
| #3206 | #3158 | Node-local JIT caches for every trainer and inference actor |

**Ports of established RL techniques, all off by default**

| PR | What |
|---|---|
| #3200 | `--mask-truncated-completions` |
| #3201 | trackio tracking backend |
| #3202 | `--log-sample-completions` |
| #3205 | `--top-entropy-quantile` |
| #3220 | adaptive entropy coefficient (Skywork-OR1) |
| #3221 | `--rm-timeout` for reward calls |
| #3222 | adaptive KL controller |
| #3223 | ReMax advantage estimator |
| #3225 | `--reward-funcs` / `--reward-weights` composite rewards |

If any of these are not something you want to carry, say so and I will close them rather than keep rebasing; two needed conflict fixes this morning already. If a batch would be easier as one PR, I can combine. And if the team would rather I stop opening new ones until this queue drains, that is a reasonable answer too.

PR #3192, opened 2026-09-10
fix(docker): bump torch_memory_saver to pick up the free-while-paused fix

JessicaJiang-123:
## Problem

`test_deepseek_v4_flash_4layer_ci.py` has been failing in the [nightly](https://github.com/radixark/miles/actions/runs/34368332492) since 09-09 with:

```text
[torch_memory_saver.cpp] CUresult error: 1 (invalid argument) file=csrc/core.cpp func=free
```

The failure started after weight sync began calling `torch.cuda.empty_cache()` while the trainer is paused by `torch_memory_saver`.

With the current CUDA pin (`f05a8754d`), if `empty_cache()` frees a cached segment inside a paused region, `torch_memory_saver` tries to release the same allocation again and crashes. This is [torch_memory_saver#92](https://github.com/fzyzcjy/torch_memory_saver/issues/92), fixed by [#95](https://github.com/fzyzcjy/torch_memory_saver/pull/95).

The test passed in the 09-08 nightly and in PR CI on 09-09, then started failing after the `empty_cache()` change landed on main.

## Fix

Bump the CUDA `torch_memory_saver` pin to [b5588e83d](https://github.com/fzyzcjy/torch_memory_saver/commit/b5588e83de86412a48689a6583a4b567e75f7acc), the merge commit of [#95](https://github.com/fzyzcjy/torch_memory_saver/pull/95).

This is the smallest bump containing the fix. The package version remains `0.0.9.post1` and no build changes are needed.

ROCm already pins `06caa5348`, which includes the fix, so it is unchanged.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -163,7 +163,7 @@ RUN pip install "git+https://github.com/NVIDIA-NeMo/Emerging-Optimizers.git@v0.3
 
 # torch_memory_saver's source build needs TMS_CUDA_MAJOR; take it from torch.
 RUN TMS_CUDA_MAJOR=$(python3 -c "import torch; print(torch.version.cuda.split('.')[0])") \
-    pip install git+https://github.com/fzyzcjy/torch_memory_saver.git@f05a8754daf68238d54e4cf31cb3ba866684bbaf --no-cache-dir --force-reinstall
+    pip install git+https://github.com/fzyzcjy/torch_memory_saver.git@b5588e83de86412a48689a6583a4b567e75f7acc --no-cache-dir --force-reinstall
 RUN pip install "nvidia-modelopt[torch]>=0.37.0" --no-build-isolation
 RUN pip install git+https://github.com/radixark/Megatron-Bridge.git@bridge --no-deps --no-build-isolation
 RUN pip install megatron-energon --no-deps


yueming-yuan:
The three red NVIDIA shards here are the bug this PR fixes, not evidence against it. This PR comes from a fork, and `docker-decide` in `_build-pr-ci-image.yml` refuses to publish a `pr-N` image for forks (GitHub does not pass repository secrets to fork `pull_request` runs, so the Docker Hub login could not succeed anyway):

```
##[notice]Fork PRs cannot publish an image; suites use the released image.
```

All three PR Test runs on this branch therefore executed the GPU suites on `radixark/miles:dev`, which still pins `f05a8754`. Each red shard died on `CUresult error: 1 ... func=free line=104`:

| shard | test |
|---|---|
| `stage-c-4-gpu-h200 (1)` | `test_deepseek_v4_flash_4layer_ci.py` |
| `stage-c-4-gpu-h200 (2)` | `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` |
| `stage-c-8-gpu-h200 (0)` | `test_qwen3_4B_offload_disaggregated.py` |

The `rebuild-ci-image` label cannot help — the fork check runs before the label is read.

To get the bump actually built and tested, #3243 carries the same commit (`ef7ac9165`, cherry-picked with your authorship preserved) on an in-repo branch. There `docker-decide` took the other path — `No usable pr-3243 image; building one.` — and is building `pr-3243` on a `docker-build` node now; the GPU suites will run on it once published.

Independent confirmation of the mechanism: `/rerun-test` of `test_deepseek_v4_flash_4layer_ci.py` on #3242 — same test, same old `dev` image, the only difference being that `after_engines_resumed()` (and thus `torch.cuda.empty_cache()`) does not run — went green, with the weight-update path fully exercised and zero `CUresult` errors. That isolates the trigger to `empty_cache()` while the trainer's TMS regions are paused, which is exactly what `d7a3a51` in the new pin guards.

I'll post the #3243 result here when its GPU shards finish.

yueming-yuan:
#3243 (same commit, in-repo so the image actually builds) has run the GPU suites on `radixark/miles:pr-3243`, which carries `torch_memory_saver@b5588e83`. The three shards that were red here:

| test | on `dev` (old pin, this PR's runs) | on `pr-3243` (new pin) |
|---|---|---|
| `test_deepseek_v4_flash_4layer_ci.py` | `CUresult error: 1` ×5, FAIL — and on every nightly since 09-09 | **pass**, 10/10 in its shard, `CUresult error: 1` = 0 |
| `test_qwen3_4B_offload_disaggregated.py` | `CUresult error: 1` ×1, FAIL | still FAIL, but `CUresult error: 1` = 0 — dies on `Reloading 21 process groups` → `Segfault encountered`, line-for-line the same failure the 09-11 nightly showed; a separate bug |
| `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` | `CUresult error: 1` ×1, FAIL | still FAIL, but `CUresult error: 1` = 0 — `torch.OutOfMemoryError` in backward (6.93 GiB wanted, 8.00 GiB free), same shape as the nightly's 6.36 / 9.30; a separate bug |

Zero runtime `CUresult error: 1` across every h200 shard that has finished so far; the passing shards exercised 21 weight updates between them. The bump does what it says. The two remaining failures have their own causes and are unaffected in either direction.

This PR is correct as written; it just cannot demonstrate it from a fork. Either this or #3243 can merge — they carry the identical commit.

yueming-yuan (approved):

PR #3198, opened 2026-09-10
Share checkpoint publication across LoRA saves and HF exports

yueming-yuan:
Single-LoRA saves and HF exports currently write directly into their public directories. Reuse the shared checkpoint directory transaction so readers see a completed version only after all files have been written.

## Changes

- Save native single-LoRA shards, optional optimizer/scheduler state, and the HF adapter in one transaction. Reuse Tinker's HF iterator and adapter writer; adapter export errors now fail the save instead of producing a partial checkpoint.
- Apply the same publication flow to full-model and merged-LoRA HF exports, including checkpoint eval. LoRA HF exports contain a merged model and an HF-only `adapter/` subdirectory.
- Resolve the version before loading native LoRA state, and update eval snapshot GC to remove both published links and their backing directories.
- Give each fully-async eval E2E case its own temporary snapshot directory; CI containers share host IPC and must not reuse checkpoints from previous runs.

Native training filenames remain unchanged. HF adapters use `adapter_model.safetensors` and `adapter_config.json`; HF exports retain `.complete` for existing readers. Megatron's native full-model checkpoint and async-save protocol are outside this change.

## Remaining issues

- Rank-local publication failures can strand peers in collectives when periodic HF export catches and logs the error. Recoverable IO needs a consistent outcome across ranks.
- Failed exports can leave staging or unpublished version directories; cleanup must happen after writers stop and preserve the previous published checkpoint.
- Async eval must pin a version when submitting a checkpoint whose public name may be reused. Distinct per-step paths avoid this case.

## Validation

- `pre-commit run --all-files` and `git diff --check` passed.
- Local filesystem probes confirmed preservation of the old checkpoint on write failure and reproduced the staging leak and mutable-alias behavior. Distributed primitives were mocked.
- Existing single-LoRA tests were updated; no new tests were added during cleanup.
- CI on `bc815fdfc8`: CPU and pre-commit passed. Five NVIDIA jobs failed during Bridge import because Megatron lacks `mla_qk_norm_config`; the fully-async eval job separately failed on an existing ordinary snapshot directory.
- A local filesystem probe reproduced that directory collision before the E2E fix and passed afterward for fleet/external cases, preserving the previous checkpoint. Distributed primitives and the training launcher were mocked. GPU validation on `0efaee7953`: fully-async eval passed all three cases (`shared`, `fleet`, `external`), including snapshot exports for steps 0 and 1 in both dedicated eval modes. [CI run](https://github.com/radixark/miles/actions/runs/35478750409): CPU/pre-commit passed; NVIDIA jobs finished with 3 passed and 5 blocked by the same Bridge/Megatron import incompatibility. The B200 and stage-c 2-GPU H200 suites were skipped.


yushengsu-thu:
@claude review

yushengsu-thu (approved):
approve - the same way used in multi-lora.
But agent catch some issue. So I list below.

- Reviewed d94d93b47088; reproduced against base using CPU/Gloo with stubbed model conversion, not GPU E2E.

- Regression: Failed exports leave _tmp_step_*; [eval cleanup](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/ray/rollout/eval_dispatch.py#L48-L55) deletes only final paths. Two failed exports left four files versus zero on base.
- Regression: [LoRA saving](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/megatron_utils/lora/utils.py#L495) now uses [rank-0-only directory creation/publication](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/training_utils/checkpoint_io.py#L39-L47), breaking node-local checkpoints on other nodes. Base succeeds; head reports a missing directory.
- Existing issue remains: With empty HF exports, [nested save phases](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/megatron_utils/hf_export.py#L153-L161) mismatch across ranks; peers time out. Coordinate validation before entering adapter saving.

yushengsu-thu (commented):
Reviewed `d94d93b47088`; reproduced against base using CPU/Gloo with stubbed model conversion, not GPU E2E.

- **Regression:** Failed exports leave `_tmp_step_*`; [eval cleanup](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/ray/rollout/eval_dispatch.py#L48-L55) deletes only final paths. Two failed exports left four files versus zero on base.
- **Regression:** [LoRA saving](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/megatron_utils/lora/utils.py#L495) now uses [rank-0-only directory creation/publication](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/training_utils/checkpoint_io.py#L39-L47), breaking node-local checkpoints on other nodes. Base succeeds; head reports a missing directory.
- **Existing issue remains:** With empty HF exports, [nested save phases](https://github.com/radixark/miles/blob/d94d93b47088ad0101360c7451720b780862c487/miles/backends/megatron_utils/hf_export.py#L153-L161) mismatch across ranks; peers time out. Coordinate validation before entering adapter saving.

yushengsu-thu:
@claude review

miles/backends/megatron_utils/hf_export.py
```diff
@@ -84,21 +80,17 @@ def export_hf_model_direct(
...
-    finally:
-        # In a finally: rank 0 is the only rank that can fail above.
-        torch.distributed.barrier()
+    if is_writer:
+        assert weight_map, f"HF export to {path} produced no weights"
+        base_checkpoint = Path(args.hf_checkpoint)
+        if base_checkpoint.is_dir():
+            for meta_file in base_checkpoint.iterdir():
+                if _is_hf_metadata_file(meta_file):
+                    shutil.copy2(meta_file, path / meta_file.name)
```
Zhichenzzz: codex commented (zhichen reviewed)

[P1] Propagate writer failures consistently across ranks before returning

If this metadata copy or the index write raises on rank 0, that rank exits `write_shards` and the outer `save_hf_model` handler logs the exception and returns. Other ranks instead enter `write_checkpoint_dir`'s post-write Gloo barrier, which rank 0 never reaches. Removing the previous `finally: barrier()` therefore turns a recoverable metadata-write error into a collective timeout.

Reproduced with a real Megatron model, two H200 GPUs, TP=2, and actual NCCL/Gloo collectives: injecting an I/O error after the weight shards were written let both base ranks return without a barrier error, whereas this head returned on rank 0 and timed out on rank 1 after the explicitly configured 5-second Gloo timeout. Please coordinate failure status across ranks before any rank returns or enters the remaining transaction phases.

miles/backends/megatron_utils/hf_export.py
```diff
@@ -117,84 +109,55 @@ def save_hf_model(
...
+            write_lora_weights(model, args, tmp_dir / "adapter")
+        if torch.distributed.get_rank() == 0:
+            # eval readers also accept legacy directories and still require this marker
+            (tmp_dir / HF_EXPORT_COMPLETE_MARKER).touch()
+
+    if should_log:
+        logger.info(f"Saving model in HuggingFace format to {path}")
+    try:
+        write_checkpoint_dir(path, write_shards)
     except Exception as e:
```
Zhichenzzz, line 79: codex commented (zhichen reviewed)

[P2] Prevent failed exports from dispatching the previous checkpoint as the current step

With a fixed `--save-hf` path such as `/checkpoints/latest` and snapshot eval reusing that path (`eval_hf_dir=None`), a failed overwrite now preserves the previous version and its `.complete` marker. This handler suppresses the failure without returning an export status, so training still dispatches eval with the new rollout ID. The eval completeness check accepts the old marker and can attribute the previous weights' results to the new step.

Reproduced on H200 by saving step 0, modifying the actual model weights, and injecting a metadata-copy failure during step 1: the PR preserved the old weights byte-for-byte but dispatched step 1 with `is_complete_hf_export=True`; base had no marker and the check returned false. Please propagate the committed version/export outcome to the dispatcher and skip eval when the current step was not exported, while retaining the last good checkpoint.

miles/backends/megatron_utils/hf_export.py
```diff
@@ -117,84 +109,55 @@ def save_hf_model(
...
+        if is_lora_model(model):
+            write_lora_weights(model, args, tmp_dir / "adapter")
+        if torch.distributed.get_rank() == 0:
+            # eval readers also accept legacy directories and still require this marker
+            (tmp_dir / HF_EXPORT_COMPLETE_MARKER).touch()
+
+    if should_log:
+        logger.info(f"Saving model in HuggingFace format to {path}")
+    try:
+        write_checkpoint_dir(path, write_shards)
```
Zhichenzzz: codex commented (zhichen reviewed)

[P2] Clean up failed transaction staging directories on the eval export path

Switching to `write_checkpoint_dir` moves partial output to `_tmp_step_<id>`, but `EvalDispatcher.dispatch()` still handles export failures by deleting only the public `step_<id>` path. The transaction does not clean up its staging directory on failure, and failed exports never enter the dispatcher's retirement queue. Since subsequent eval points use different step IDs, their partial model files accumulate independently of `eval_keep_snapshots`.

Confirmed on H200 with real model exports and an injected metadata-copy error: three failed eval exports with `eval_keep_snapshots=0` left `_tmp_step_1`, `_tmp_step_2`, and `_tmp_step_3`, containing 2,376,048 bytes for the tiny test model. The same base run left no directories and zero bytes. Please reclaim failed staging/unpublished versions after all writers have stopped, preserving any previously published checkpoint; otherwise repeated failures can fill the recommended `/dev/shm` snapshot storage.

miles/backends/megatron_utils/hf_export.py
yushengsu-thu: Fixed in `97c3eecab2`; H200 metadata-failure probe returns on both ranks without timeout.

miles/backends/megatron_utils/hf_export.py
yushengsu-thu, line 79: Fixed in `97c3eecab2`; failed H200 exports no longer retain `.complete`.

miles/backends/megatron_utils/hf_export.py
yushengsu-thu: Fixed in `97c3eecab2`; three failed H200 exports leave no staging directories.

yushengsu-thu (commented):
(codex) Three issues in CPU/Gloo; - checked

miles/backends/megatron_utils/hf_export.py
yueming-yuan: Allowing saving on the same path in different steps will be risky for async eval, so i fixed it through add `step_xx` subdirectory when `rollout_id` is not in `save_hf`
```
if not any(field == "rollout_id" for _, field, _, _ in Formatter().parse(args.save_hf)):
                args.save_hf = os.path.join(args.save_hf, "step_{rollout_id}")
```

miles/backends/training_utils/checkpoint_io.py
```diff
@@ -1,70 +1,81 @@
...
+    checkpoint_dir = Path(path)
+    distributed = dist.is_initialized()
+    is_rank0 = not distributed or dist.get_rank() == 0
+    prepare_error = [None]
+    if is_rank0:
+        try:
+            if checkpoint_dir.exists():
+                if not overwrite:
+                    raise FileExistsError(f"checkpoint {checkpoint_dir} already exists")
+                shutil.rmtree(checkpoint_dir)
```
yushengsu-thu, line 39: `rmtree` fails on symlink checkpoints written by the current `write_checkpoint_dir`; unlink symlinks (and dangling ones) before `mkdir`.

miles/backends/megatron_utils/lora/utils.py
```diff
@@ -407,116 +407,37 @@ def save_lora_checkpoint(
...
                 "opt_param_scheduler": opt_param_scheduler.state_dict() if opt_param_scheduler else None,
-            },
-            save_path / f"training_state_rank{rank}.pt",
-        )
-        logger.info(f"Saved {'optimizer/scheduler' if save_optimizer else 'scheduler'} state to {save_path}")
+            }
 
-    if dist.is_initialized():
-        dist.barrier(group=get_gloo_group())
+        publisher.write_adapter(None, checkpoint_dir)
```
yushengsu-thu: A failed HF adapter export now fails the whole save and deletes the native shards; previously best-effort.

miles/backends/megatron_utils/lora/utils.py
```diff
@@ -407,116 +407,37 @@ def save_lora_checkpoint(
...
+        torch.save(adapter_state, checkpoint_dir / f"adapter_megatron_rank{global_rank}.pt")
+        if training_state is not None:
+            torch.save(training_state, checkpoint_dir / f"training_state_rank{global_rank}.pt")
 
-    return str(save_path)
+    write_checkpoint_dir(save_dir, write_shards)
```
yushengsu-thu, line 442: [P2] Node-local LoRA saves fail on nonzero nodes; create checkpoint directories on each node.

yushengsu-thu (commented):
Three H200 async E2Es pass; LoRA fails on head and base. Two checkpoint regressions remain.

Zhichenzzz (approved):

miles/backends/training_utils/checkpoint_io.py
yueming-yuan: because the symlink is just added in multi-lora, I think there will not be users now, so we can do this for simplicity

miles/backends/megatron_utils/lora/utils.py
yueming-yuan: given full checkpoint is written to only rank 0, I think this unification is ok

PR #3210, opened 2026-09-11
[multi-lora] Serve LoRA on the output layer

yueming-yuan:
The SDK defaults `train_unembed=True` and the cookbook recipes (sl_loop / rl_loop) expose no way to turn it off, so a gateway whose adapter layout lacks the output layer rejects every official client at `create_lora_training_client`. This adds the output layer to the servable layout:

- map `lm_head <-> output_layer` in the LoRA target-module tables; the sglang engine flag list picks up `lm_head` through the existing reverse conversion, and sglang's built-in lm_head LoRA path (vocab-parallel B shard) serves it unchanged
- put `output_layer` in the example gateway's default `--target-modules`, so `GatewayConfig.trains_unembed` is true and the SDK default is accepted

Three fixes surfaced while validating on a 6xH200 devbox, each its own commit; the first two belong to the stack and can be absorbed upstream on its next rebase:

1. `self.slot_optimizers` init was lost in the last stack rebuild — any `create_model` killed the trainer (CPU tests cannot see actor init)
2. the serve launcher hardcoded `megatron_model_type=qwen3-30B-A3B`, so any other `--hf-checkpoint` failed `hf_validate_args`; the cookbook e2e moves to Qwen3-4B-Instruct-2507 (the cookbook renderer registry has no bare Qwen3-4B entry)
3. idempotency keys are 0-based (the SDK counts `model_seq_id` from 0); the `>= 1` rule stays on stream seq_ids only

**Depends on** radixark/Megatron-Bridge branch `yueming/multi-lora-tied-output` (one commit): a tied output layer allocates no weight of its own, so `MultiLoRALinear.__init__`'s `next(to_wrap.parameters())` raised StopIteration. Untied models (the lora CI suite's Qwen3-30B-A3B) do not hit this path.

Validated on the devbox: gateway serves Qwen3-4B-Instruct-2507 with `output_layer` wrapped and `lm_head` in the engine's `--lora-target-modules`; cookbook sl/rl recipe run and a 30B `output_layer` training run are in progress and results will be posted here. Not executed: the disabled cookbook e2e in CI.

Launcher snapshot regenerated for the serve entrypoint. Pre-existing snapshot drift in kimi_k25/inkling/amd launchers (`--no-gradient-accumulation-fusion`) is not touched.

### DIFF
--- a/docs/examples/multi-lora.md
+++ b/docs/examples/multi-lora.md
@@ -15,7 +15,8 @@ One 8-GPU node, disaggregated (multi-LoRA forbids `--colocate`):
 - 4 training GPUs: TP2 for the dense layers, EP4 for the 128 routed experts.
 - 4 sampling GPUs: two SGLang engines of 2 GPUs each, serving adapter versions by name.
 - 4 adapter slots (`--multi-lora-n-adapters`), rank up to 32, covering attention
-  (`linear_qkv`, `linear_proj`) and the per-expert MoE projections (`linear_fc1`, `linear_fc2`).
+  (`linear_qkv`, `linear_proj`), the per-expert MoE projections (`linear_fc1`, `linear_fc2`),
+  and the output layer (`output_layer`) so the cookbook's default `train_unembed=True` is servable.
 
 ## Run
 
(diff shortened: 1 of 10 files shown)


yushengsu-thu (approved):
tbh, it's not much model set lm_head for lora training 🤣 but for SDK default

PR #3211, opened 2026-09-11
Fix MTP gradient detachment in AutoBridge mode

Shi-Dong:
With `--megatron-to-hf-mode bridge --enable-mtp-training`, the bridge provider retained Megatron's default `mtp_detach_heads=False`, allowing the MTP auxiliary loss to update shared policy weights. Apply the same detachment rule as the regular model provider before finalization, so the auxiliary loss trains only MTP-specific parameters.

The production change is four added lines. The CPU regression test covers enabled, disabled, and missing MTP flags, including preservation of an existing provider setting.

Validation:

- All 6 regression cases pass; unpatched `main` fails the enabled-MTP case.
- All applicable pre-commit checks pass.
- Actual Nemotron Lightning AutoBridge provider construction and finalization on CPU now produces `mtp_detach_heads=True` with MTP training enabled.
- Native Megatron MTP block, embedding, loss, and autoscaler CPU probe: shared backbone/embedding/output auxiliary gradients become zero; draft gradients remain nonzero, and a main-model loss still updates the backbone. This probe uses small CPU layers and mocked pipeline metadata; it is not a full-model GPU training test.

This PR addresses detachment when MTP training is enabled. Environment-variable parsing, auxiliary-loss scaling, and label-mask alignment are separate changes.

### DIFF
--- a/miles/backends/megatron_utils/model_provider.py
+++ b/miles/backends/megatron_utils/model_provider.py
@@ -49,6 +49,10 @@ def _apply_bridge_runtime_config(provider, args: argparse.Namespace) -> None:
     provider.calculate_per_token_loss = args.calculate_per_token_loss  # CP>1 VL models assert this
     provider.variable_seq_lengths = args.variable_seq_lengths
 
+    # Match the non-bridge path: MTP must only train its own draft parameters.
+    if getattr(args, "enable_mtp_training", False):
+        provider.mtp_detach_heads = True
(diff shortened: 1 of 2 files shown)


Shi-Dong:
@claude review always

guapisolo (approved):

Zhichenzzz (approved):

PR #3219, opened 2026-09-12
fix: align MTP loss masks with input tokens before context-parallel slicing

Shi-Dong:
## Summary

Fix auxiliary multi-token-prediction (MTP) supervision using Miles changes only. Megatron is unchanged, and the main SFT/policy loss keeps its existing mask.

## Bug

Miles builds `full_loss_masks` in next-token/prediction alignment, but calls Megatron with `labels=None`. Megatron's MTP path derives next-token labels from `input_ids` and shifts the supplied mask along with them before applying the additional MTP-depth shift. Passing an already prediction-aligned mask therefore advances the mask one token too far relative to the MTP targets.

For example, an intended supervised target span at positions 4–8 can instead select 3–7 for the auxiliary loss. The targets themselves are not relabeled; the wrong boundary positions receive loss. This does not demonstrate a downstream accuracy impact or a problem with the main SFT loss.

## Changes

- Add opt-in `get_input_loss_masks` to the shared batch builder. It returns a separate `input_loss_masks` tensor without changing `full_loss_masks`.
- Construct alignment from the original response mask before padding, packing, or context-parallel slicing. A local roll after sharding would not correctly handle shard/packed-sequence boundaries.
- Pass the input-aligned mask to Megatron in training and forward-only calls when MTP training is enabled. Non-MTP consumers and the combined schedule-plan path retain their previous behavior.
- Add an independent token-position oracle for both supported layouts, zigzag/contiguous CP, CP sizes 1/2/4, padding, multiple supervised spans, empty/all-zero responses, and prompt-free inputs.
- Exercise the real optional Megatron `process_mtp_loss`, sequence rolling, normalization, and `MTPLossAutoScaler.backward` on CPU. Only the expensive output projection/CE is replaced with differentiable per-token losses; the tests inspect which target positions receive gradients. Covers supplied/inferred labels, one/two MTP layers, both layouts and normalization modes, and all-masked inputs.

## Validation

- **80 focused cases passed** (0.72 s): 48 batch/layout/CP mask cases and 32 cases through the real Megatron MTP loss/backward consumer.
- **324 training-utility cases passed** (177.85 s), including the 48 new batch cases and existing distributed CP2 loss/advantage checks. There are **356 distinct passing cases** across these two commands.
- All changed files passed pre-commit (including Ruff, isort, Black and secret checks) and `git diff --check`.
- Negative control: replacing the new input-aligned mask with the old prediction-aligned mask reproduces a failing target-gradient assertion in the real Megatron consumer; no source file was changed for this control.
- Megatron consumer validation used unchanged `radixark/Megatron-LM` commit `235952df607b3820716e5e67728a5ab470ca33ae`. The optional consumer tests skip if Megatron is not installed; they actually ran in this validation.
- Scope: CPU mask/gradient validation, simulated CP1/2/4 batch layouts, and neighboring distributed CPU tests. No full-model GPU retraining or downstream benchmark comparison was performed.

```sh
python -m pytest tests/fast/backends/training_utils -q --durations=5
python -m pytest tests/fast/backends/training_utils/test_input_loss_masks.py \
  tests/fast/backends/megatron_utils/test_mtp_loss_mask_alignment.py -q --durations=5
```

No new dependency or user-facing launch flag is required. The batch-builder docstring documents the internal alignment contract; no README change is needed.

### DIFF
--- a/miles/backends/megatron_utils/model.py
+++ b/miles/backends/megatron_utils/model.py
@@ -329,6 +329,7 @@ def forward_only(
             args.data_pad_size_multiplier,
             args.qkv_format,
             allgather_cp=args.allgather_cp,
+            get_input_loss_masks=args.enable_mtp_training,
         )
         unconcat_tokens = batch["unconcat_tokens"]
         tokens = batch["tokens"]
@@ -347,7 +348,7 @@ def forward_only(
             attention_mask=None,
             labels=None,
             packed_seq_params=packed_seq_params,
-            loss_mask=batch["full_loss_masks"],
+            loss_mask=batch["input_loss_masks"] if args.enable_mtp_training else batch["full_loss_masks"],
             **(filter_keys(batch, ["witness_ids"]) if args.enable_witness else {}),
             **(batch["multimodal_train_inputs"] if batch["multimodal_train_inputs"] is not None else {}),
             fp32_output=fp32_output,
@@ -512,6 +513,7 @@ def train_one_step(
             args.data_pad_size_multiplier,
             args.qkv_format,
             allgather_cp=args.allgather_cp,
+            get_input_loss_masks=args.enable_mtp_training,
         )
 
         if "adapter_token_counts" in batch:
@@ -543,7 +545,8 @@ def train_one_step(
                 "attention_mask": None,
                 "labels": None,
                 "packed_seq_params": get_packed_seq_params(batch, args),
-                "loss_mask": batch["full_loss_masks"],
+                # With labels=None, Megatron derives MTP labels and shifts this mask.
+                "loss_mask": batch["input_loss_masks"] if args.enable_mtp_training else batch["full_loss_masks"],
                 **(filter_keys(batch, ["witness_ids"]) if args.enable_witness else {}),
             }
 
--- a/miles/backends/training_utils/data.py
+++ b/miles/backends/training_utils/data.py
@@ -121,6 +121,52 @@ def get_rollout_data(
     return rollout_data, store_get_result
 
 
+def _prepare_loss_masks(
+    loss_masks: Sequence[torch.Tensor],
+    total_lengths: Sequence[int],
+    response_lengths: Sequence[int],
+    *,
+    qkv_format: str,
+    max_seqlen: int,
+    pad: int,
+    allgather_cp: bool,
+    input_aligned: bool,
+) -> torch.Tensor:
+    """Align response masks before packing and CP slicing.
+
+    The main loss uses next-token/prediction alignment. Megatron MTP with
+    ``labels=None`` instead requires input-token alignment: it derives labels
+    and shifts the supplied mask itself. Shifting an already CP-sharded mask
+    cannot recover the values across shard or packed-sequence boundaries.
+    """
+    parallel_state = get_parallel_state()
+    cp_size = parallel_state.cp.size
+    cp_rank = parallel_state.cp.rank
+    shift = 0 if input_aligned else 1
+    aligned_masks = []
+    for loss_mask, total_length, response_length in zip(loss_masks, total_lengths, response_lengths, strict=True):
+        prompt_length = total_length - response_length
+        loss_mask = F.pad(loss_mask, (prompt_length - shift, shift), value=0)
+        if not allgather_cp:
+            loss_mask = slice_with_cp(loss_mask, 0, qkv_format, max_seqlen)
+        aligned_masks.append(loss_mask)
+
+    if qkv_format == "bshd":
+        if allgather_cp:
+            local_len = max_seqlen // cp_size
+            start = cp_rank * local_len
+            aligned_masks = [
+                F.pad(mask, (0, max_seqlen - mask.size(0)), value=0)[start : start + local_len]
+                for mask in aligned_masks
+            ]
+        return torch.stack(aligned_masks)
+
+    loss_mask = F.pad(torch.cat(aligned_masks), (0, pad), value=0)
+    if allgather_cp:
+        loss_mask = loss_mask.chunk(cp_size, dim=0)[cp_rank]
+    return loss_mask.unsqueeze(0)
+
+
 def get_batch(
     data_iterator: "DataIterator",
     keys: Sequence[str],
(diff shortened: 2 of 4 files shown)


Shi-Dong:
@claude review always

guapisolo:
codex cmt (human verifying)

I suggest using one stable mask contract at the Miles → Megatron boundary:

> `loss_mask[i]` indicates whether `input_ids[i]` should receive supervision when it is a prediction target.

The motivation is to separate **which tokens receive supervision** from **where predictions for those tokens are produced**. The MTP flag should control auxiliary prediction, while the mask’s meaning stays fixed.

Both forward entry points pass `labels=None`. With the Megatron version examined here, MTP expects an input-aligned mask; the built-in non-MTP path returns logits without using this mask to compute the main loss. Miles computes its main loss from the original response `loss_masks`. I therefore don’t see a built-in consumer that requires a second, prediction-aligned full-sequence mask.

Proposed plan:

1. Keep the original response `loss_masks` as the supervision source.
2. Generate one input-aligned `full_loss_masks` tensor by padding the prompt positions with zeros **before packing and CP slicing**. Document this fixed contract.
3. Pass it consistently through training, forward-only, and combined-schedule entry points. Remove `input_loss_masks`, `get_input_loss_masks`, and the MTP-dependent mask selection.
4. Leave Megatron’s label/mask shifting and Miles’ main-loss calculation unchanged.

Before migrating the existing field’s semantics, check custom model/loss consumers for any dependency on prediction alignment. Validation should confirm unchanged non-MTP logits/main loss, intended MTP target gradients, and correct packed-sequence/CP boundaries, including all-masked samples.

This keeps one supervision source, one full-sequence mask representation, and prediction shifting owned by the consumer.

guapisolo:
cc cmt

I suggest one stable mask contract at the Miles → Megatron boundary instead of two alignments selected by a flag: `loss_mask[i]` says whether `input_ids[i]` should receive supervision when it is a prediction target, and whoever derives labels owns the prediction shift. #3233 implements this on top of this branch; the notes below are the reasoning.

Why one input-aligned mask is enough (checked against radixark/Megatron-LM 235952df, current `miles-main` for these files):

- All three Miles call sites pass `labels=None`: `forward_only`, the normal train forward, and the `build_schedule_plan` branch in `miles/backends/megatron_utils/model.py`.
- With `labels=None`, Megatron's non-MTP path returns logits and never reads `loss_mask`. The only readers in `GPTModel._postprocess` are `prefetch_halos` and `process_mtp_loss`, both under `config.mtp_num_layers`, plus `output_processor`, which Miles never passes.
- `process_mtp_loss` derives labels from `input_ids` and rolls `loss_mask` with them, so it needs the input-aligned mask.
- Miles' main loss uses the original response `loss_masks` (`loss.py` via `get_local_response_loss_masks`). Nothing in `miles/`, `miles_plugins/` or `examples/` reads `full_loss_masks` except the three `loss_mask=` arguments above; FSDP calls `get_batch` but reads neither mask.

So the prediction-aligned `full_loss_masks` has no remaining consumer, and the second tensor plus the MTP-dependent selection is what #3233 drops.

[P1] The selection key does not match the consumer's gate. Megatron runs `process_mtp_loss` whenever `config.mtp_num_layers` is truthy, and `model_provider.py` builds the MTP block from `args.mtp_num_layers` alone. Several presets carry `--mtp-num-layers 1` without `--enable-mtp-training`: `scripts/models/qwen3.5-35B-A3B.py`, `qwen3.6-35B-A3B.py`, `glm4.7-flash.py`, `mimo-7B-rl.py`, `qwen3-next-80B-A3B.py`, `examples/on_policy_distillation/qwen3_5_35b_selfdistill/*.sh`, and the e2e `test_mtp0_spec_v2.py` case. In those runs this PR still passes the prediction-aligned `full_loss_masks`, Megatron rolls it again, and the MTP mask stays one token early. By code reading the auxiliary loss is live there: `schedules.py` sets the MTP loss scale whenever `mtp_num_layers` is set, and `mtp_detach_heads` is only set under `--enable-mtp-training`, so the misaligned MTP gradient also reaches the shared trunk and output layer. I have not confirmed this on a GPU run; a one-step gradient probe on one of these presets would settle it. A flag-free input-aligned mask removes this class of gap. If you prefer to keep a selection, key it on `bool(args.mtp_num_layers)`.

[P2] The `build_schedule_plan` branch still passes `full_loss_masks` and only asserts `not args.enable_mtp_training`. Megatron's `PostProcessNode` forwards `chunk_state.loss_mask` into the same MTP gate, so an MTP-bearing preset plus combined 1F1B gets the misaligned mask. No shipped preset combines the two today, hence medium.

[P3] Scope note, pre-existing and unchanged by #3233: with `--allgather-cp` and CP>1 Miles never sets `PackedSeqParams.cp_partition_mode`, so Megatron rolls contiguous shards as zigzag for thd and has no contiguous unpacked roll at all for bshd. A single-mask design cannot fix that. I would assert that `allgather_cp` with `mtp_num_layers` is unsupported, or set `cp_partition_mode="contiguous"` for thd, and not claim contiguous-CP correctness in the docstring. No live preset combines them, hence low.

What #3233 does:

1. Keeps the original response `loss_masks` as the only supervision source.
2. Builds one input-aligned full-sequence mask by left-padding the prompt positions with zeros before packing and CP slicing (what `_prepare_loss_masks(input_aligned=True)` already does), keeps it under the existing `input_loss_masks` key and drops `full_loss_masks`, so any reader that assumed prediction alignment fails loudly instead of silently mis-masking. The contract is documented on the helper.
3. Passes it from all three entry points and removes `get_input_loss_masks`, the `input_aligned` switch and both `enable_mtp_training` ternaries. Megatron's label/mask shifting and Miles' main-loss calculation are untouched.
4. Tests: drops the `get_input_loss_masks` and `labels_provided=True` axes (Miles never passes labels), derives the prompt boundary in the oracle instead of `ids[3:]`, and uses production `torch.int` masks. The remaining gap is CP>1 through the real Megatron roll, which neither test exercises (the consumer test runs `cp_group=None`); the zigzag layout matches `slice_with_cp` by code reading, so this is a coverage note, not a defect claim.

Happy to fold #3233 into this PR instead if you prefer a single change.

guapisolo (approved):

Zhichenzzz (approved):
LGTM after offline discussion

PR #3233, opened 2026-09-14
fix: use one input-aligned loss mask for every Megatron entry point

guapisolo:
> **Depends on:** #3219

## Summary
Pass one input-aligned loss mask to Megatron from every entry point, regardless of MTP flags.

## Symptom & Reproduction
- **Symptom:** With `--mtp-num-layers 1` but no `--enable-mtp-training`, #3219 still hands Megatron the prediction-aligned `full_loss_masks`; Megatron rolls it once more, so the MTP mask stays one token early (last prompt token supervised, last response token dropped).
- **Affected presets:** `scripts/models/qwen3.5-35B-A3B.py`, `qwen3.6-35B-A3B.py`, `glm4.7-flash.py`, `mimo-7B-rl.py`, `qwen3-next-80B-A3B.py`, `examples/on_policy_distillation/qwen3_5_35b_selfdistill/*.sh`; the combined-1f1b `build_schedule_plan` branch has the same gap for any MTP-bearing preset.
- **Reproduction:** Code reading against radixark/Megatron-LM `235952df` (`gpt_model.py` `_postprocess`, `multi_token_prediction.py` `process_mtp_loss`); no GPU run.
- **Mechanism witness:** the parent PR's negative control, `test_mtp_loss_mask_alignment.py` with `full_loss_masks` passed for `labels=None`, expects the supervised response tokens as MTP targets and observes them shifted by one; not re-run here (no `triton` on macOS).

## Root Cause
1. `GPTModel._postprocess` runs `process_mtp_loss` whenever `config.mtp_num_layers` is truthy.
2. `model_provider.py` builds the MTP block from `args.mtp_num_layers` alone.
3. `process_mtp_loss` with `labels=None` rolls `input_ids` and `loss_mask` together, so it needs an input-aligned mask.
4. `model.py` picked `input_loss_masks` only under `args.enable_mtp_training`; other MTP-bearing runs kept `full_loss_masks`.
5. `full_loss_masks` has no other consumer: the non-MTP path never reads `loss_mask`, and `loss.py` uses the response `loss_masks`.

## Fix
`get_batch` emits one `input_loss_masks` (`mask[i]` is 1 iff `tokens[i]` is a supervised response token) and no `full_loss_masks`. All three Megatron call sites pass it; `get_input_loss_masks`, `input_aligned` and both `enable_mtp_training` ternaries are removed. The contract is documented on `_prepare_loss_masks`: whoever derives labels owns the prediction shift; Megatron does so when `labels=None` and ignores the mask when MTP is off. Megatron's shifting and Miles' main loss are unchanged. FSDP calls `get_batch` but reads neither mask.

## Verification
- `tests/fast/backends/training_utils/test_input_loss_masks.py`: 12 cases (thd/bshd, zigzag/allgather, CP 1/2/4) against an independent token-position oracle with production `torch.int` masks; passed locally.
- `tests/fast/backends/training_utils/test_get_batch_multi_lora_cp.py`, `test_opd_cp_data.py`: 16 neighbouring `get_batch` cases passed locally.
- `tests/fast/backends/megatron_utils/test_mtp_loss_mask_alignment.py`: 16 cases through the real `process_mtp_loss` with `labels=None`; the oracle now derives the prompt boundary. Skipped locally (no `triton` on macOS); relies on the `stage-a-cpu` CI job.
- black 24.3.0, isort 5.13.2, ruff 0.14.7, autoflake 2.0.2, `git diff --check`: clean.

## Review Focus
- `_prepare_loss_masks` / `data.py`: left pad is `prompt_length` with no right pad; `pad` still only matters on the thd paths.
- `model.py` `build_schedule_plan` branch: now receives the input-aligned mask; `assert not args.enable_mtp_training` stays because combined 1F1B with MTP is unsupported.
- Out of scope, pre-existing: `--allgather-cp` with CP>1 never sets `PackedSeqParams.cp_partition_mode`, so Megatron's MTP roll treats contiguous shards as zigzag. No preset combines allgather CP with trainer-side MTP.


Shi-Dong:
@claude review always

Shi-Dong (approved):

PR #3236, opened 2026-09-14
feat(rollout): support an opaque external endpoint

nanjiangwill:
## Summary

- route rollout traffic to one externally managed endpoint
- launch no Miles router or inference engines for that topology
- support both direct `/generate` rollouts and optional session-server/TITO rollouts
- retain Miles' existing workload and per-engine concurrency primitives
- make the session-affinity header configurable when session serving is enabled

## Design

Miles owns rollout semantics and request admission. The external service owns GPU placement, per-engine admission, routing, scaling, and engine lifecycle. `--use-session-server` remains an orthogonal agentic/TITO choice rather than a requirement of disaggregated serving.

Miles cannot derive aggregate service capacity from a hidden replica count. Fully async runs use `--async-max-concurrent-samples` as their total producer bound; the external deployment must apply `--sglang-server-concurrency` consistently to each engine.

This PR does not add provider-specific request mutation or retry policy.

## Validation

- 68 focused argument, URL, concurrency, and session tests passed in the Miles image
- 209 broader fast tests passed before an unrelated multiprocessing test hung in the local container
- ruff, compileall, and diff checks passed
- Ray spec collection in the local CPU-only container requires `libcuda.so.1`; CI covers that environment


nanjiangwill:
Current head is green except for one unrelated CPU-shard failure. The complete job log shows `tests/fast/backends/training_utils/test_ppo_cp_advantages.py` ending in a native child `SIGABRT` (`terminate called without an active exception`), not a Python assertion or endpoint-path failure. The same shard is passing on the sibling PRs.

I attempted to rerun the failed job, but GitHub requires repository admin permission. A maintainer rerun should be sufficient; if the same native teardown failure repeats, it should be investigated independently of this PR.

PR #3238, opened 2026-09-14
perf(weight-sync): overlap initial delta baseline with rollout

nanjiangwill:
## Summary

- start the first rollout before capturing an opaque disk-delta service's initial baseline
- keep baseline completion as a hard barrier before training consumes rollout data
- preserve the original ordering for weight equality checks and initial evaluation

## Why

An opaque rollout service already serves the immutable base checkpoint. Disk-delta's first sync only captures that same checkpoint as its source baseline, so serializing the two operations adds latency without protecting mutable state.

Steady-state generation needs no special branch: the refactored fully-async rollout worker is already persistent while training and artifact publication proceed.

## Dependencies

This draft depends on #3236 and #3237. It is kept as an independent diff so the ordering change can be reviewed separately.

## Tests

- 10 async driver ordering/lifecycle tests
- ruff, isort, and diff checks


nanjiangwill:
Closing this startup-only optimization rather than upstreaming an implicit model-identity assumption. `external endpoint + disk-delta` does not prove that the service is serving the same base checkpoint the trainer uses for its delta baseline, especially across resume. A correct future optimization should use an explicit checkpoint/model-version identity contract; the one-time startup latency does not justify a trust flag or endpoint-specific branch in `train_async.py`.

PR #3242, opened 2026-09-14
fix(weight-update): call after_engines_resumed only for a LoRA weight sync

yueming-yuan:
`tests/e2e/megatron/model_scripts/test_deepseek_v4_flash_4layer_ci.py` passed on the 2026-09-08 nightly and has failed on every nightly since:

```
Timer finalize_and_resume_engines end (elapsed: 0.0s)
[torch_memory_saver.cpp] CUresult error: 1 (invalid argument)  file=csrc/core.cpp func=free line=104
-> ray.exceptions.ActorUnavailableError
-> WorkerUnreachableError: Worker died or is unreachable when calling 'update_weights'
```

## Cause

#3147 is mine, and its title says what it was for — bounding the trainer's GPU memory **during the LoRA weight sync**. The call it added is not restricted to that:

```python
        with timer("finalize_and_resume_engines"):
            ...
        protocol.after_engines_resumed()      # every update, not just LoRA
```

The `cuda_ipc` implementation is `torch.cuda.ipc_collect()` + `torch.cuda.empty_cache()`, and neither it nor the call site tests `is_lora`. So every colocated weight update now empties the cache the instant the engines resume, while torch_memory_saver still holds the regions it paused. TMS then frees a region in a state it does not expect and the process dies where it stands, which Ray only sees as an actor that stopped answering.

The log's ordering is exact: the TMS error lands between the `finalize_and_resume_engines` timer closing and the next statement, which is this call.

It reaches non-LoRA tests because the hook is on the protocol, not on the LoRA path — `test_deepseek_v4_flash_4layer_ci.py` is colocated (`[miles-colocate] Update weights` appears 117 times in its log) and carries no adapters.

## Scope of the claim

- The nightly that passed used `3de96596f`, the first that failed used `8d9826eac`. Eight commits merged in that window and #3147 is the only one touching the weight-update path.
- `test_deepseek_v4_flash_4layer_megatron_impl_ci.py`, which fails on the same stage, dies of a genuine `torch.OutOfMemoryError` in backward and is unrelated to this.

## Verification

Not reproducible locally — it needs colocated engines on a GPU. I will run the test on this branch with `/rerun-test` and report the result here; the change is one line and the prediction is falsifiable, so a pass on this branch against a fail on main is the evidence.

`pre-commit` clean on the changed file. `tests/fast/backends/training_utils/weight_update/test_dist_weight_update_lifecycle.py`, which #3147 also touched, passes 9/9.


yueming-yuan:
/rerun-test tests/e2e/megatron/model_scripts/test_deepseek_v4_flash_4layer_ci.py

yueming-yuan:
Superseded by #3243 (in-repo copy of #3192, bumping torch_memory_saver past the free-while-paused fix). Gating the hook on `is_lora` in the updater was the wrong layer and the wrong predicate: the crash is `empty_cache()` while the trainer's TMS regions are paused, which colocated LoRA runs hit too, and `empty_cache()` is load-bearing for #3147's wake-up fix. The `/rerun-test` on this branch stays useful as the reverse validation of the diagnosis.

PR #3243, opened 2026-09-14
fix(docker): bump torch_memory_saver past the free-while-paused fix (in-repo copy of #3192)

yueming-yuan:
In-repo copy of #3192 by @JessicaJiang-123 (`ef7ac9165`, cherry-picked with the original author preserved). Same one-line change; this PR exists only so CI can actually build and test it.

## Why a second PR

#3192 comes from a fork. `docker-decide` in `_build-pr-ci-image.yml` refuses to publish a `pr-N` image for forks — GitHub does not hand repository secrets to fork `pull_request` runs, so the Docker Hub login would fail anyway — and falls back to the released `dev` image:

```
##[notice]Fork PRs cannot publish an image; suites use the released image.
```

All three of #3192's PR Test runs therefore executed the GPU suites on `radixark/miles:dev`, which still pins the old `torch_memory_saver` (`f05a8754`). Its three red NVIDIA shards each died on the very bug the bump fixes:

| shard | test | `CUresult error: 1` |
|---|---|---|
| `stage-c-4-gpu-h200 (1)` | `test_deepseek_v4_flash_4layer_ci.py` | 5 |
| `stage-c-4-gpu-h200 (2)` | `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` | 1 |
| `stage-c-8-gpu-h200 (0)` | `test_qwen3_4B_offload_disaggregated.py` | 1 |

The `rebuild-ci-image` label cannot help there: the fork check runs before the label is read. From an in-repo branch `SAME_REPO` is true, so `docker-decide` builds `pr-N` and the suites run on the bumped TMS. This PR carries `run-ci-image` for the same GPU scope #3192 had.

## The bug, in one paragraph

`#3147` added `after_engines_resumed()` to the weight-update protocol; the `cuda_ipc` implementation calls `torch.cuda.empty_cache()` right after the engines resume. Under colocation the trainer is asleep during `update_weights` (it sends from its CPU backup), so every TMS region is `PAUSED`, and `empty_cache()` makes torch `cudaFree` fully-free cached segments whose regions are paused. TMS `f05a8754`'s `free()` then calls `cuMemUnmap` + `cuMemRelease` unconditionally; `pause()` had already released the handle, so the second `cuMemRelease` fails with `CUDA_ERROR_INVALID_VALUE` and `CURESULT_CHECK` exits the process:

```
Timer finalize_and_resume_engines end (elapsed: 0.0s)
[torch_memory_saver.cpp] CUresult error: 1 (invalid argument)  file=csrc/core.cpp func=free line=104
-> ActorUnavailableError -> WorkerUnreachableError on 'update_weights'
```

Upstream fixed exactly this in `d7a3a51` ("Do not release the VMM handle twice when freeing paused memory"), which guards unmap/release with `if (ACTIVE == metadata.state)` and still frees the VA and backups. `b5588e83` is the merge of that PR (torch_memory_saver #95); `f05a8754..b5588e83` is four commits, the other three unrelated `change_env` fixes.

`test_deepseek_v4_flash_4layer_ci.py` passed on the 2026-09-08 nightly and has failed this way on every nightly since; #3147 merged in that window.

## Why the bump and not a miles-side guard

`empty_cache()` is load-bearing for #3147: TMS tracks whole `cudaMalloc` segments, so torch freeing blocks into its own cache is invisible to TMS and the segments are resumed in full at the next wake-up. Only `empty_cache()` makes torch `cudaFree` them so TMS drops them. Skipping it would bring back the wake-up OOM #3147 fixed. With the TMS fix, freeing paused segments is correct and desirable. #3242 (gating the hook on `is_lora`) was the wrong layer and the wrong predicate and will be closed.

## Reverse validation of the diagnosis

`/rerun-test` of `test_deepseek_v4_flash_4layer_ci.py` on #3242 — same test, same old `dev` image, only difference is the hook not running — went green: 23 `Update weights` progress lines, 3 `finalize_and_resume_engines end`, training to step 1, zero `CUresult` errors, zero actor loss. That isolates the trigger to `empty_cache()` in the hook.

## What this PR's CI should show

`docker-decide` → "No usable pr-N image; building one." → GPU shards run on the new pin → the three tests above no longer die on `CUresult error: 1`. If `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` still fails, it should be on its `torch.OutOfMemoryError` in backward (seen on the 2026-09-11 nightly), which is unrelated.


yueming-yuan:

PR #3246, opened 2026-09-14
fix(workers): avoid forwarding attached Python entrypoints

bojiang-li:
When a Miles process is started with an attached Python entrypoint such as `python -cCODE` or `python -mmodule`, `python_argv_prefix()` currently forwards that entrypoint into worker launch commands. The appended worker module then becomes an argument to the parent's entrypoint instead of the program Python executes. Clustered forms such as `-ucCODE` have the same problem, and `-uW ignore` / `-uX dev` lose their separate option values.

Scan short-option clusters up to the command/module selector, preserving preceding interpreter flags and treating the remainder of `-W` / `-X` options as values. Add real-subprocess regression coverage for attached command and module selectors, clustered flags, and valued options. The command cases also execute a child using the returned prefix to verify its own entrypoint runs.

Validation on macOS, Python 3.13.2:

- Before the production fix: 7 new regression cases failed; 2 controls passed.
- `python -m pytest --noconftest -o addopts='' tests/fast/utils/workers/test_argv_utils.py -q`: 67 passed.
- Ruff 0.14.7, Black 24.3.0, isort 5.13.2, and `git diff --check`: passed for both changed files.
- The targeted run disables repository-wide conftest loading to avoid unrelated rollout/Ray dependencies. It imports the real implementation and runs actual Python subprocesses. Pytest reports an unknown `asyncio_mode` setting because pytest-asyncio is not installed in this minimal environment; these tests are synchronous.
- Full CPU CI and GPU/Ray end-to-end training have not been run locally. No training performance claim is made.

AI assistance: Codex inspected the source, implemented the change, and ran the checks. This is a draft for review; no human review is claimed.


guapisolo:
@claude always

guapisolo (approved):

guapisolo:
@claude review

PR #3249, opened 2026-09-14
[6/6] feat(session): save resolved request args for turn continuation

guapisolo:
## Summary

Save each turn's resolved request so continuations can use that turn's argument history.

## Motivation

Session callers currently pass `turn_args=None`, so the model resolver cannot use historical arguments. A continuation can lose omitted tools or model-constrained settings from the turn whose tokens it reuses. Recording each generation's resolved request lets continuation, rollback and branching select the matching history.

## Usage

Omitted tools inherit on continuation; sampling remains per-request. For Qwen3.8, `chat_template_kwargs={"reasoning_effort": "low"}` on a new root keeps subsequent turns at `low`, even if they request `medium`. New roots may choose again. Recorded absence is preserved and uses the template default `xhigh`.

## Design Notes

- **Mental model:** After a successful generation, save a deep copy of the full resolved backend request as `turn_args`. For the next request, select the continued checkpoint or parent node, then pass its `turn_args` into argument resolution before rendering.
- **Ownership:** The session records history; the [model resolver introduced in #3264](https://github.com/radixark/miles/blob/b65f107b59b8a5846772d67c50019f92efedcad2/miles/utils/chat_template_utils/tito_tokenizer.py#L142) keeps responsibility for inheritance, overrides and compatibility checks.
- **Storage:** v1 keeps `turn_args_history` aligned with token checkpoints during rollback; v2 stores `turn_args` on each tree node so branches use their own parent history.
- **Exports:** Session/sample metadata omits `input_ids` and `messages` before copying; stored snapshots remain complete. The default v2 postprocessor prevents agent metadata from replacing exported `turn_args`.

## Verification

- **441 CPU tests passed** at `87d7924ca`, covering tokenizer rules, fixed templates, CLI arguments and request preparation.
- Qwen3.8: ten regression cases changed from failing to passing.
- `pre-commit` and `git diff --check`: passed.

The original [H200 parity failure](https://github.com/radixark/miles/actions/runs/35193900074/job/105118343891) compared random tool-call IDs in exported messages. Exports now omit messages; randomized-ID integration and record relationship checks remain.

New-head CI is pending; GPU tests were not run locally.

## Review Focus

- Save the full resolved request, not only template kwargs; failed generations add no snapshot.
- Continuations use the selected turn's history; requests, exports and sibling branches cannot mutate it.


guapisolo:
@claude always

guapisolo:
@claude review always

guapisolo:
bypass merge to unblock.

PR #3254, opened 2026-09-14
[debug] Python stacks for the offload_disaggregated SIGSEGV

yueming-yuan:
Diagnostic only — do not merge.

`test_qwen3_4B_offload_disaggregated.py` (added by #3128) segfaults at the first `new_group` of `reload_process_groups()` on every CI execution, including the torch_memory_saver bump (#3192). The log shows one frame (`runtime.sigfwd`): tvm-ffi installs its handler from a shared-library constructor and Mooncake's Go runtime forwards through it, so the real stack is lost. This re-registers faulthandler at `update_weights` entry so the crash prints all Python threads.

Run: `/rerun-test tests/e2e/megatron/test_qwen3_4B_offload_disaggregated.py`


yueming-yuan:

yueming-yuan:
Closing: CI is not the place for this kind of debugging; the segfault will be reproduced on a devbox instead.

PR #3257, opened 2026-09-14
[LoRA] Mirror moe_router_dtype into the Megatron-Bridge provider

Arist12:
## Problem

`_setup_lora_model_via_bridge` copies a whitelist of Megatron args onto the Megatron-Bridge provider. `moe_router_dtype` is not in that whitelist, and `gpt_oss_bridge.py` does not set it either, so on the bridge LoRA path gpt-oss-20b runs with `moe_router_dtype=None` even though the recipe passes `--moe-router-dtype fp32`.

The provider *is* the model config (`provide()` passes it to `MCoreGPTModel` as `config`), so `TopKRouter` reads the unset field at `router.py:125` and keeps the router GEMM in the token dtype.

A real 8×MI355X run shows both sides of the contradiction. The resolved args table reports the request:

```
moe_router_dtype ................................ fp32
```

and Megatron core, from `provider.finalize()` on every trainer rank, reports the effective config:

```
transformer_config.py:3597: UserWarning: Using a large number of experts (e.g. >=32)
without fp32 routing. Consider enabling moe_router_dtype for better numerical stability.
[repeated 7x across cluster]
```

That warning fires only on `num_moe_experts >= 32 and not self.moe_router_dtype`. gpt-oss-20b has 32 experts. Building the provider exactly as miles does confirms `moe_router_dtype is None` both before and after `finalize()`.

Nothing else on this path sets the field: `finalize()` only coerces `activation_func`/`params_dtype`/`pipeline_dtype` and never sees `args`, and the sibling whitelist `_apply_bridge_runtime_config` is called only from the non-LoRA branch.

## Change

One assignment, next to the other MoE provider settings, guarded so it only overrides when the arg was explicitly set — thirteen bridges in megatron.bridge deliberately default this to `fp32`, and an unconditional copy would downgrade them to `None`.

Same shape and same insertion point as #3232, which mirrors `moe_shared_expert_overlap` for this exact class of gap.

## Affected recipes

The four launchers that reach this path all pass `--moe-router-dtype fp32`:

| launcher | effect |
| --- | --- |
| `scripts/amd/run_gpt_oss_20b_lora.py` | `None` → `fp32` |
| `scripts/run_qwen3_5_35b_a3b_lora.py` | `None` → `fp32` (`qwen35_bridge.py` does not set it either) |
| `scripts/run_glm5_1_744b_a40b_lora.py` | no-op, `glm5_bridge.py` already sets `fp32` |
| `scripts/run_glm5_2_744b_a40b_lora.py` | no-op, same |

`args.moe_router_dtype` is generated from the `TransformerConfig` dataclass and is always present, so bare attribute access is safe.

## Scope

This is a config-fidelity fix: the model now matches its own args table, and a warning Megatron raises on every rank goes away.

**It is not claimed to fix any train/rollout logprob discrepancy.** I found this while investigating an elevated `train_rollout_logprob_abs_diff` on gpt-oss-20b LoRA RL, but a layer-by-layer dumper comparison on that recipe shows the router *logits* already agree between trainer and rollout (relative diff 4.1e-4, inside the comparator's 1e-3 threshold), so this is unlikely to be the explanation there. Landing it on correctness grounds only.

## Follow-ups, deliberately not in this PR

- `_apply_bridge_runtime_config` has the identical gap, so full-FT bridge mode is still affected.
- `moe_router_use_torch_mm` is dropped the same way and feeds the same `router_gating_linear` call, despite miles force-setting it globally with the comment that the TE gemm "increase[d] log prob diff".

## Testing

- `provider.moe_router_dtype` becomes `fp32` for the gpt-oss-20b recipe and the `transformer_config.py:3597` warning stops firing.
- Recipes that do not pass the flag are unchanged, since the guard skips the assignment.


Arist12:
@yushengsu-thu Could you help review this? Thx

Arist12:
Superseded by #3263, which is a strict superset: the same LoRA-path `moe_router_dtype` line, plus the non-LoRA `_apply_bridge_runtime_config` path, plus `moe_router_use_torch_mm` on both paths, plus test coverage.

The two fields belong together: `moe_router_dtype` alone still leaves `RouterGatingLinearFunction` on the TE branch multiplying bf16 operands, so the gpt-oss-20b recipe does not actually reach the fp32 router GEMM its args table reports until `moe_router_use_torch_mm` is mirrored as well. #3263 is validated as that pair.

Suggest closing this in favour of #3263 rather than sequencing two 4-line PRs.

Arist12:
Closing in favour of #3263, which covers both bridge paths and both dropped fields (`moe_router_dtype` + `moe_router_use_torch_mm`) with the same guard.

PR #3258, opened 2026-09-15
[1/6] refactor(session): drop the v2 active_leaf cursor; the tree is the state

guapisolo:
## Summary

Remove `active_leaf`; derive the served v2 session path from the last committed node.

## Motivation

Requests moved a shared cursor before generation succeeded, so rejected or concurrent requests could change what `GET /sessions/{id}` showed. Commit parents and training samples already used the tree directly.

## Before / After

- **Before / After:** `GET` followed a mutable cursor; it now follows `state.latest()`.
- **What moved where:** `attach_point_for_request` returns the parent explicitly; rendering and commit use that parent.

## Behavior Preservation

Token reuse, branching and sample assembly remain unchanged. `GET` now shows the last committed path, including while a request is pending or rejected.

## Verification

At `25bd8a9c3`:

- Router tests: **402 passed, 1 skipped**; one Anthropic conversion test deselected.
- Session/tokenizer tests: **1947 passed**.

## Review Focus

`latest()` follows commit order; attachment lookup remains pure and rejects truncated nodes with HTTP 409.


guapisolo:
@claude always

guapisolo:
@claude review always

guapisolo:
bypass merge to unblock.

PR #3264, opened 2026-09-15
[3/6] feat(tito): resolve full request arguments before rendering

guapisolo:
## Summary

Resolve model-specific request settings once so local rendering and the backend use the same values.

## Motivation

Model-specific argument handling previously centered on `chat_template_kwargs` and tokenizer clones, but model rules may affect the entire request. This change centralizes those rules in an interface that resolves the full request, then passes the resolved arguments explicitly to rendering.

## Usage

Qwen3.5 overrides a requested `preserve_thinking=false` with its required `true` for both rendering and forwarding. Empty resolved fields remain `chat_template_kwargs={}` and `tools=None`.

## Design Notes

- **Mental model:** `prepare_chat_request` copies client input, applies server constraints, then runs model rules on the full request. The backend receives that result; `extract_template_args` selects its renderer fields.
- Rules merge kwargs (launch → history → request), resolve tools, restore history-constrained fields, then apply required settings. `FIXED_TEMPLATE.consistant_kwargs` declares historical constraints; `extra_kwargs` declares required values.
- DeepSeek's merge selects thinking aliases as one group (history > request > launch). Its final rule reads only the merged kwargs and normalizes `thinking` for [SGLang's parser](https://github.com/sgl-project/sglang/blob/32839114c4ada2ae237581405a8ff39dc0db9e25/python/sglang/srt/entrypoints/openai/serving_chat.py#L2483).
- Explicit `template_args` replaces tokenizer clones. A supplied dict is complete; `None` uses launch defaults.

Session callers still pass `turn_args=None`: #3302 unifies preparation; #3249 connects recorded history. Qwen3.8 remains fixed to `xhigh`; top-level reasoning mapping is follow-up work.

## Verification

- At `463111992`: **209 tokenizer/request tests passed**, including both DeepSeek resolved-input regressions. **68,608 comparisons** against `5a843bd06` preserved full-resolver results, field order and mutation isolation.
- At stack tip `7517ebcab`: **444 CPU tests passed** across 11 tokenizer/session modules; `pre-commit` passed.

Updated-head CI is pending; GPU tests were not run locally.

## Review Focus

- Preserve alias precedence and input isolation.
- Render and forward the same resolved settings.


miles/rollout/session/linear_trajectory.py
```diff
@@ -107,6 +107,7 @@ def prepare_pretokenized(
         Must be called under ``self.lock``.
         """
         matcher = message_matcher if message_matcher is not None else strict_message_matches
+        template_args = tito_tokenizer.default_template_args() if template_args is None else template_args
```
guapisolo, line 110: Will be replaced by later PRs.

guapisolo:
@claude always

guapisolo:
@claude review always

guapisolo:
bypass merge to unblock.

PR #3274, opened 2026-09-15
feat(amd/dsv4): let the 4-node profile run with context parallelism

xiaohong42:
## What

Give the AMD DeepSeek-V4 4-node profile a `context_parallel_size` knob and derive the parallel layout from it.

## Why

The 32-GPU tier pinned `--context-parallel-size 1` and carried `--allgather-cp` commented out, which caps the training actor at whatever sequence fits with no context parallelism. CP is the axis that matters for long context: the DSV4 indexer materialises a `[seqlen, seqlen_kv]` fp32 score matrix **per layer**, and only CP divides it. Pipeline parallelism shrinks how many layers a rank owns, but every one of them still allocates the matrix in full.

## How

CP is spent out of the tensor-parallel budget, so the pipeline split stays 11+11+11+10 and DP collapses to 1:

| context_parallel_size | layout |
|---|---|
| 1 (default) | TP4 / PP4 / EP8, DP2 — **byte-identical to before** |
| 2 | TP4 / PP4 / EP8, DP1 |
| 4 | TP2 / PP4 / EP8, DP1 |
| 8 | TP1 / PP4 / EP8, DP1 |

EP stays at 8 throughout. `--allgather-cp` is emitted automatically above CP=1, and `--sequence-parallel` is dropped at TP=1.

A CP value that does not divide `total_gpus / pipeline_size` raises `NotImplementedError` with the arithmetic in the message.

## Test

`cp=1` produces a byte-identical argument string to the previous literal.

End to end on 4 nodes x 8 MI308X with the full 43-layer DeepSeek-V4-Flash at `cp=8`, actual trained sequence 130594:

| weight sync | 32/32 ranks ok=true |
|---|---|
| rounds completed | 2 |
| ess_ratio | 1.0 |
| train_rollout_kl | 1.04e-2 |
| train_rollout_logprob_abs_diff | 4.45e-2 |

Note that long-context runs also want a larger `--distributed-timeout-minutes` than the current default; that is a separate PR.


xiaohong42:
Superseded by a combined PR that merges this with the 2-node profile and train-memory-margin changes — they share a dependency chain and are validated by the same run.

PR #3276, opened 2026-09-15
fix(amd/dsv4): pick the colocate memory split from the card, not the node count

xiaohong42:
## What

Choose the colocate memory split from the card's capacity instead of the node count. New `--colocate-memory-profile` with `auto` (default), `192gb`, `288gb`.

## Why

The 4-node branch keeps the training weights resident and gives SGLang half the card (`--sglang-mem-fraction-static 0.5`, `--no-offload-train`, `--optimizer-offload-fraction 0.75`). The pair is sized for a 288 GB card. On 192 GB the same split leaves SGLang about 102 GiB, which does not hold the weights and a KV cache, and the run dies during load:

```
Loaded weights leave no GPU memory for the KV cache under
--mem-fraction-static=0.5. Raise above 0.627
```

**The condition is the wrong one.** Card capacity decides this; the node count only correlated with it because the profile was written on MI355X.

## Behaviour

| profile | optimizer / train weights | --sglang-mem-fraction-static |
|---|---|---|
| 288gb | --optimizer-offload-fraction 0.75 --no-offload-train | 0.5 — **unchanged** |
| 192gb | --optimizer-offload-fraction 1.0 --offload-train | 0.85 |

`auto` reads `torch.cuda.get_device_properties(0).total_memory` and takes `288gb` at 256 GiB and above. Any failure to read a device also returns `288gb`, so a GPU-less launcher keeps exactly today's behaviour.

## Test

Verified on 4 nodes x 8 MI308X (192.0 GiB) with the 43-layer DeepSeek-V4-Flash:

* `auto` resolves to `192gb` on that card, and to `288gb` when CUDA is unavailable.
* `Load weight end ... avail mem=40.12 GB`, KV cache allocated, no refusal.
* Weight sync 32/32 `ok=true`, two complete rollout/train rounds.

The `288gb` path is unchanged, so MI355X users see no difference.


xiaohong42:
Superseded by a combined PR that merges this with the 2-node profile and train-memory-margin changes — they share a dependency chain and are validated by the same run.

PR #3296, opened 2026-09-17
docker: bump nvidia-cudnn-frontend 1.26.0 -> 1.28.0 for the compact DSA indexer

HJSang:
## Why

Upstream Megatron #5992 (ported to `miles-main` in radixark/Megatron-LM#99) adds a config-time check on SM100: with a compress-ratio-4 indexer and the fused cuDNN DSA backend, `cudnn.DSA.indexer_forward_top_k_wrapper` must exist and accept `deterministic` (and the MXFP8 scale parameters when `--dsa-indexer-precision mxfp8`). It is a hard check, not a fallback:

```
ValueError: Fused DSA indexer requires a compatible cuDNN Frontend compact wrapper; missing parameters: deterministic
```

cuDNN Frontend 1.26.0, the current pin, has no compact wrapper at all, so once #99 lands on `miles-main` every `--dsv4-impl megatron --dsa-kernel-backend cudnn` run fails at model init, including `test_deepseek_v4_flash_4layer_megatron_impl_ci.py`.

## What changes

`docker/Dockerfile`: `nvidia-cudnn-frontend==1.26.0` -> `==1.28.0` in the `ENABLE_CUDA_13` block, plus a comment explaining the requirement. Nothing else. The image build workflow should pick this up on the PR (Dockerfile change).

## Verified on a 4x B200 devbox running `radixark/miles:dev`

- `pip install --no-deps nvidia-cudnn-frontend==1.28.0` on the cu13 stack; `cudnn.dsa` namespace imports; `indexer_forward_top_k_wrapper` present with all parameters #5992 checks (`deterministic`, `precision`, `q_scale`, `k_scale`, `cu_seqlens_q_scale_padded`, `cu_seqlens_k_scale_padded`, `sf_vec_size`).
- With 1.26.0: radixark/Megatron-LM#99 fails at model init as above. With 1.28.0: DeepSeek-V4-Flash-FP8-4layer trains end to end on #99 and #101 (GSM8K, 4096-token responses, 3 rollouts, deterministic mode), and the CSA unit suite passes (`test_csa_fused_sparse_attention.py` 98 passed, `test_attention_variant_csa.py` 113 passed, `test_dsv4_hybrid_native_parity.py` 25 passed).
- Current `miles-main` (pre-#99) also runs fine with 1.28.0 (unpatched baseline arm), so this bump is safe to land ahead of the Megatron PRs.

1.27.0 was not tested. Upstream Megatron `dev` pins cudnn-frontend `develop` (>= 1.28).

## CI on this PR (run 35173591245, triggered by `run-ci-megatron`)

`docker-build` built the PR image with cuDNN Frontend 1.28.0 and every stage passed against it: stage-a/b CPU, stage-b 2x H200, stage-c 2x H200 (2 partitions), stage-c 4x H200 (3 partitions), stage-c 8x H100 (2), stage-c 8x H200 (2), stage-c 8x B200. DSv4 tests in the 4x H200 stage, all against the 1.28.0 image: `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` PASS (903 s, partition 1, CI gate trusted on 10 metrics) — this is the cuDNN DSA backend path that fails at model init on 1.26.0 once radixark/Megatron-LM#99 lands; `test_deepseek_v4_flash_4layer_ci.py` PASS (1078 s, partition 0); `test_dsv4_thd_cp_correctness.py` PASS.

## Sequencing

Land this and rebuild the image **before** radixark/Megatron-LM#99 merges, otherwise `miles-main` breaks Miles DSv4 CI (CI checks out `miles-main` at job start).


HJSang:
/rerun-test tests/e2e/megatron/model_scripts/test_deepseek_v4_flash_4layer_megatron_impl_ci.py

HJSang:
Cancelled the `/rerun-test` above: `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` already ran in run 35173591245, stage-c-4-gpu-h200 partition 1, and passed (903 s, CI gate trusted on 10 metrics). My earlier note that it was not scheduled was a misread of the partition logs.

HJSang:
Note on the red `stage-c-4-gpu-mi350 (0)` check: it is `tests/e2e/megatron/model_scripts/test_inkling_small_4layer_ci.py` failing with `ValueError: Cannot find model module. 'InklingForConditionalGeneration' is not a registered model in the Transformers library` inside the ROCm image. That stage runs from `docker/Dockerfile.rocm` with its own image resolution, and `docker/image_inputs.py` excludes `Dockerfile.rocm` on purpose, so the CUDA-13 `nvidia-cudnn-frontend` pin changed here cannot reach it. The `run-ci-megatron` label pulled that stage in because the same label selects the ROCm Inkling test. All CUDA stages, including both DeepSeek-V4 4-layer E2E tests on the new image, are green.

HJSang:
Kicked the GPU tests that the current label set was missing.

**Added `run-ci-precision`** (run [35290911859](https://github.com/radixark/miles/actions/runs/35290911859)). `run-ci-megatron` alone selects the two DeepSeek-V4 4-layer E2E tests but not the DSA-adjacent precision ones. The label adds:

- `tests/e2e/precision/test_dsv4_bshd_thd_rollout_parity.py` — BSHD vs THD rollout parity on `DeepSeek-V4-Flash-FP8-4layer`, Hopper-only, 4 GPU
- `tests/fast-gpu/test_dsv4_thd_cp_correctness.py` — DSv4 THD context-parallel compressor/boundary correctness
- `tests/fast-gpu/test_quantizer_ci.py` and `tests/fast-gpu/test_nvfp4_quantizer.py` — cheap, and they share the CUDA-13 wheel stack this pin sits in

**Scope of what this pin can affect today.** `nvidia-cudnn-frontend` only reaches the DSA indexer through `megatron/core/transformer/.../csa_utils/fused_sparse_attention.py`, and `miles-main` currently has no compact-indexer branch there at all: `indexer_forward_top_k_wrapper` appears only on the CSA-core port branch (radixark/Megatron-LM#99, and #101 on top of it). So on today's `miles-main` this bump is a no-op at runtime and lands as a **prerequisite** for those two ports, not as a behaviour change. The value of these runs is that the rebuilt CUDA-13 image is otherwise healthy.

**Blackwell pass to follow.** The compact wrapper is SM100-only, so once this run is green I will add `run-on-blackwell` for a second pass that moves the Blackwell-capable DSv4 tests onto B200 on the new image. Note that even that pass does not execute the compact kernel until #99/#101 merge, since the branch does not exist on `miles-main`; exercising the kernel itself needs a `ci-megatron-pr:` pin, which belongs on those PRs rather than here.

The red `stage-c-4-gpu-mi350 (0)` is still the unrelated ROCm Inkling failure explained above; `docker/image_inputs.py` excludes `Dockerfile.rocm`, so this pin cannot reach that stage.

HJSang:
`run-ci-precision` run [35290911859](https://github.com/radixark/miles/actions/runs/35290911859) is green on every stage against the 1.28.0 image. The DeepSeek-V4 tests all executed, confirmed from the partition logs rather than the stage colour:

| test | partition | result |
|---|---|---|
| `test_deepseek_v4_flash_4layer_ci.py` | 0 | PASS |
| `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` | 1 | PASS |
| `test_dsv4_bshd_thd_rollout_parity.py` | 2 | PASS, 1278 s |

The rollout parity test is the one the earlier label set was missing. It is the closest thing in the suite to a direct BSHD/THD indexer check on `DeepSeek-V4-Flash-FP8-4layer`, and it is Hopper-only by registration.

Now added `run-on-blackwell`, run [35307365136](https://github.com/radixark/miles/actions/runs/35307365136), to move the Blackwell-capable DSv4 tests onto B200 against the new image. The parity test will not appear there, since it registers `hardware=["hopper"]`.

The two red `stage-c-4-gpu-mi350 (0)` entries are the same unrelated ROCm Inkling failure counted once per run, not a new regression. The DSv4 ROCm variant is already `disabled=` in its own registration.

HJSang:
Blackwell pass done, run [35307365136](https://github.com/radixark/miles/actions/runs/35307365136). **Both DeepSeek-V4 tests pass on B200 against the 1.28.0 image**, which was the point of the exercise:

| test | B200 result |
|---|---|
| `test_deepseek_v4_flash_4layer_ci.py` | PASS, 1285 s |
| `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` | PASS, 615 s, CI gate trusted on 10 metrics |

The stage is red because the third test, `test_glm5_2_744b_a40b_5layer_ci.py`, failed on B200 and the suite stops on first failure, so the remaining 38 never ran.

**That failure is not caused by this pin.** The same test passes on 4x H200 with the identical image, same commit, in run [35290911859](https://github.com/radixark/miles/actions/runs/35290911859) at 666 s. Only the GPU generation differs. The crash is a SGLang fused-MoE weight-loader shape mismatch during the first weight update, nothing to do with cuDNN:

```
File "sglang/srt/layers/moe/fused_moe_triton/layer.py", line 768, in _load_w13
    expert_data.copy_(loaded_weight)
RuntimeError: The size of tensor a (64) must match the size of tensor b (6144) at non-singleton dimension 2
```

GLM-5.2 registers `hardware=["hopper", "blackwell"]` but its home suite is `stage-c-4-gpu-h200`, so `run-on-blackwell` moved it to the 8-GPU B200 host, where the MoE runner backend resolves differently. It looks like a pre-existing Blackwell gap that no scheduled suite exercises, worth its own issue rather than anything to fix here.

I have removed `run-on-blackwell` so the label set is back to `run-ci-megatron` + `run-ci-precision`. Removing a label does not re-trigger CI, so the rollup still shows that B200 stage red; the full green matrix for this exact commit is run 35290911859. `/rerun-failed-ci` will refresh it if you want the checks green before merging.

HJSang:
**Correction to my earlier comment.** I wrote that on today's `miles-main` this pin is "a no-op at runtime". That is wrong and understates the scope, so I want it on the record before review.

`miles-main` already depends on this wheel. `megatron/core/transformer/experimental_attention_variant/csa_utils/fused_sparse_attention.py` imports `from cudnn import DSA` and calls into it in 16 places today, including `indexer_forward_wrapper`, `indexer_top_k_wrapper`, `indexer_backward_wrapper`, `dense_indexer_backward_wrapper` and `compactify_wrapper`. Every `--dsv4-impl megatron --dsa-kernel-backend cudnn` run goes through those.

What is genuinely absent on `miles-main` is only the **compact** entry point, `indexer_forward_top_k_wrapper`, which arrives with radixark/Megatron-LM#99. So the accurate framing is:

- 1.26.0 -> 1.28.0 **does** change the kernels behind calls that already run today. It is a library swap under the existing dense indexer path, not an inert pin.
- It is **additionally** a hard prerequisite for #99, which fails at model init without the compact wrapper.

The evidence that the swap is safe is empirical rather than structural: on this exact image, both DeepSeek-V4 4-layer tests pass on 4x H200 and again on B200, and `test_dsv4_bshd_thd_rollout_parity.py` passes on H200. Those exercise the dense indexer path that the upgrade touches. I have not attempted a numerical diff of 1.26.0 against 1.28.0 kernel outputs, and the CI gates are the metric checks in those tests rather than a direct comparison.

HJSang:
**Second correction, and it supersedes the coverage claim in both of my comments above.** I said the passing DeepSeek-V4 tests "exercise the dense indexer path that the upgrade touches". That is true of one test, not three, and even that one does not check numerics. Reviewers should not read the green checks as validating the kernel swap.

**Only one CI test reaches cuDNN.** `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` sets `dsv4_impl="megatron"` and `dsa_kernel_backend="cudnn"`. The other two I cited both set `dsv4_impl="miles"`:

| test | impl | reaches `cudnn.DSA`? |
|---|---|---|
| `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` | megatron | yes |
| `test_deepseek_v4_flash_4layer_ci.py` | miles | no, TileLang |
| `test_dsv4_bshd_thd_rollout_parity.py` | miles | no, TileLang |

`miles_plugins/models/deepseek_v4/arguments.py` actively rejects `--dsa-kernel-backend cudnn` under `--dsv4-impl miles`, so those two cannot regress on a cuDNN change by construction.

**What the one cuDNN test asserts.** Its `extra_args` include `--ci-disable-logprobs-checker`, which switches off the block in `miles/backends/training_utils/log_utils.py` that compares `rollout/log_probs` against `ref_log_probs` at `abs_tol=1e-8` and against `rollout_log_probs` at `abs_tol=0.03`, plus the entropy band. With that flag the test asserts that two GSM8K rollouts complete without crashing, plus the weight-update check run with `--check-weight-update-allow-quant-error`.

**The metric gates are informational.** `tests/ci/ci_utils.py` wraps the hook so "the gate verdict can never change the test's pass/fail this round"; a PR run only logs `[CI Gate][shadow]`. `TRUSTED (10 metric(s))` means the run qualifies as a baseline to persist, not that ten metrics were inside a band. I quoted that line earlier as if it were evidence of numerical validation. It is not.

**Net.** CI covers the cuDNN path at smoke level: import, dispatch, no crash, no illegal memory access. It does not cover a change in selected Top-K indices or indexer score values. The suites that do cover that, `test_csa_fused_sparse_attention.py` and `test_attention_variant_csa.py`, live in the Megatron fork, and Miles CI never invokes Megatron's unit tests, so their results exist only as devbox runs quoted in radixark/Megatron-LM#99 and #101.

If you want real coverage before this lands, the cheapest option is dropping `--ci-disable-logprobs-checker` from that one test, which turns the existing 1e-8 trainer/rollout comparison into a genuine gate on the cuDNN path. I have not made that change here, since it is a CI policy question beyond a dependency pin, and it may be disabled for a known reason I have not traced. Happy to open it as a separate PR if you want it.

HJSang:
Measured the pin on hardware instead of reasoning about it. **The upgrade does change a code path that runs on `miles-main` today, and it is not the compact indexer.**

Setup: 4x B200 devbox, `radixark/miles:dev`, Miles `ef3807c0e`, Megatron `73b54618f` (`miles-main` tip, so no #99), `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` (`dsv4_impl=megatron`, `dsa_kernel_backend=cudnn`). Same box, same checkout; only the wheel changed between the last two arms.

| arm | frontend | `cudnn.csa` | eager-compressor fallbacks | trainer logprob | rollout logprob | gap |
|---|---|---|---|---|---|---|
| A | 1.26.0 | absent | yes | -1.683108 | -1.670478 | 0.012631 |
| A repeat | 1.26.0 | absent | 2 | -1.715303 | -1.703219 | 0.012084 |
| B | 1.28.0 | present | 0 | -1.688201 | -1.676762 | 0.011438 |

All three exit 0.

**What changes.** On 1.26.0 every run logs, from `fused_compressor.py:154`:

```
CSA fused compressor is enabled and this device is supported, but cudnn-frontend's
CSA compressor API is unavailable (ModuleNotFoundError: No module named 'cudnn.csa');
keeping the eager compressor.
```

I confirmed this against the wheels directly: 1.26.0 ships no `cudnn/csa` package, 1.28.0 does. So the pin silently moves the CSA compressor from the eager fallback onto the fused cuDNN path. That happens on today's `miles-main`, independent of #99. The fallback log line disappears in arm B, which is the observable confirmation.

**Effect on numerics.** The trainer/rollout gap is 0.011438 under 1.28.0 against 0.012084 and 0.012631 under 1.26.0. That is slightly outside the two-run 1.26.0 spread, but with one sample on the new version I would not call it a real shift. Absolute log-probabilities move much more than this between runs (-1.683 to -1.715 at a fixed version), so the gap, not the absolute, is the meaningful quantity. Either way both sit far inside the 0.03 tolerance.

**Net for review:** the pin is well-behaved here, but it should be described as enabling the fused CSA compressor plus unblocking #99, not as an inert version bump. Reviewers may want a deliberate look at the fused compressor path, since nothing in CI compares it against the eager one.

Separately, on the coverage question above: I re-ran this test with `--ci-disable-logprobs-checker` **removed** and it passes. The gain is smaller than it looks, though. The 1e-8 trainer-vs-reference assertion never fires, because `--kl-loss-coef 0.00` means no reference forward and `rollout/ref_log_probs` is absent; the entropy assertion is skipped for the same reason. Only the 0.03 trainer-vs-rollout comparison runs. That is still real coverage on the cuDNN path and is free, so dropping the flag looks worthwhile, but it is a cross-engine check rather than a determinism gate.

HJSang:
Rebased onto `main` (parent `12754e950`, the sglang v0.5.20 bump), so the evidence links in my comments above now point at the *old* head. Superseding them.

**The red ROCm checks are inherited from `main`, not caused by this branch.** The scheduled ROCm run on `main` at exactly our rebase parent, run [35617143599](https://github.com/radixark/miles/actions/runs/35617143599), fails both mi350 partitions with a superset of what we see here:

| test | fails on `main` @ `12754e950` | fails on this PR |
|---|---|---|
| `test_inkling_small_4layer_ci.py` | yes | yes |
| `test_qwen3_5_35B_A3B_mtp/test_amd_mtp1_spec_v2_r3.py` | yes | yes |
| `test_qwen3_4B_offload_disk_stream.py` | yes | not selected |
| `test_qwen3_4B_fsdp_hybrid_shard_r2s2.py` | yes | not selected |

This is not a one-off. Every scheduled ROCm run on `main` since 09-17 has failed: 35238581399, 35360810108, 35450765419, 35518679737, 35617143599. Five consecutive days. Combined with `docker/image_inputs.py` excluding `Dockerfile.rocm` from image resolution, a CUDA-13 pip pin cannot be the cause. ROCm on `main` looks like it needs its own owner and issue.

The CUDA `PR Test` run for this head is still queued on GPU runners; I will post the matrix when it lands. Previous CUDA evidence, now for the pre-rebase head, was run 35290911859 (full matrix green) and run 35307365136 (both DSv4 tests green on B200).

HJSang:
**Post-rebase CUDA CI is green: run [35626663381](https://github.com/radixark/miles/actions/runs/35626663381), head `9ecf17275` on top of `12754e950` (sglang v0.5.20).** Every CUDA stage passed, and the image was rebuilt with 1.28.0 by `docker-build` in this run.

Stages: docker-build, resolve-ci-image/policy, stage-a CPU (4 partitions), stage-b CPU, stage-b 2x H200, stage-c 2x H200 (2), stage-c 4x H200 (3), stage-c 8x H100 (2), stage-c 8x H200 (2), stage-c 8x B200. All success.

Per-test, the DSA-relevant ones, read from the partition logs rather than the stage colour:

| test | partition | result |
|---|---|---|
| `test_deepseek_v4_flash_4layer_megatron_impl_ci.py` (cuDNN DSA backend) | 4x H200 (1) | PASS |
| `test_deepseek_v4_flash_4layer_ci.py` | 4x H200 (0) | PASS, 1094 s |
| `test_dsv4_bshd_thd_rollout_parity.py` | 4x H200 (2) | PASS, 1303 s |
| `test_dsv4_thd_cp_correctness.py` | 4x H200 (2) | PASS, 11 s |
| `test_glm5_2_744b_a40b_5layer_ci.py` | 4x H200 (1) | PASS, 696 s |
| `test_glm5_2_744b_a40b_5layer_nvfp4.py` | 8x B200 | PASS, 1199 s |
| `test_glm5_2_744b_a40b_5layer_nvfp4_w4a16.py` | 8x B200 | PASS, 628 s |
| `test_nvfp4_quantizer.py` | 8x B200 | PASS, 30 s |

Note the GLM-5.2 5-layer test passes here on 4x H200. Its B200 failure reported earlier came from forcing it onto Blackwell with `run-on-blackwell`, which is a pre-existing gap on that hardware and unrelated to this pin; that label is no longer applied.

The only red left on this PR is the ROCm stage, which fails identically on `main` at our own rebase parent. Details in the comment above.

This is ready to merge from my side. Recommended order: land this and rebuild the image **before** radixark/Megatron-LM#99, since `miles-main` DSv4 CI breaks at model init on 1.26.0 once the compact indexer lands.

yueming-yuan (approved):

PR #3301, opened 2026-09-17
[2/6] fix(session): extract server argument resolution and validate client overrides

guapisolo:
> **Depends on:** #3258

## Summary

Apply session-server constraints to client request fields in one place.

## Symptom & Reproduction

- **Symptom:** Client token offsets, replay flags or adapter selectors could conflict with server settings.
- **Reproduction:** Send `input_ids=[1]` or `logprob_start_len=0`; the request now returns HTTP 400 before reaching the backend.

## Root Cause

1. `prepare_chat_request` handled only some server-owned fields.
2. `is_lora_enabled` also selected adapters intended only for training.

## Fix

`resolve_request_args_by_config` assigns metadata/replay flags directly and checks token/adapter controls inline. It validates template kwargs and removes backend streaming fields while retaining client stream intent. Adapter selection uses `lora_rollout_enabled`; parsing only decodes JSON.

## Verification

CPU request/config/v1-v2 HTTP tests at `cd1743c4d`: **135 passed**. All 1,776 resolver comparisons preserve outputs, mutations and errors; override warnings are removed. GPU not run.

## Review Focus

Replacement versus rejection, including null values, disabled replay and training-only LoRA.


guapisolo:
@claude always

guapisolo:
@claude review always

guapisolo:
bypass merge to unblock.

PR #3311, opened 2026-09-17
[5/6] fix(session): validate retry requests before rolling back

guapisolo:
> **Depends on:** #3302

## Summary

Preserve v1 session history when retry preparation fails.

## Symptom & Reproduction

- **Symptom:** A retry with invalid arguments discards the last committed turn before returning HTTP 400, as reported in [#3302](https://github.com/radixark/miles/pull/3302#discussion_r4040816247).
- **Reproduction:** Complete two turns, retry the second with forbidden `input_ids`, and inspect the session. At `858a6a88b`, expected two unchanged records, observed one after rejection.

## Root Cause

1. `_try_detect_and_rollback_to_assistant_checkpoint` truncates history before request preparation.
2. `prepare_chat_request` or `_render_token_ids` can then fail, leaving the truncated history behind.

## Fix

Select the checkpoint without mutation, resolve arguments and render using its messages and tokens, then apply rollback. Validation and rendering failures preserve history. The one-step rollback limit and v2 behavior remain unchanged; backend failures after preparation retain existing semantics.

## Verification

- Integration at stack tip `a085997b4`: **444 CPU tests passed** across 11 tokenizer/session modules, including aliases, request preparation, rollback, v1/v2 HTTP, snapshots and parity.
- `pre-commit` at the tested stack tip: passed.
- Concurrent preparation/rollback comment edits were preserved; both affected session modules have unchanged executable ASTs after docstrings are excluded.

CI for the updated head is pending; GPU tests were not run locally.

## Review Focus

- `_render_token_ids` must use the selected checkpoint without changing session state.
- `_rollback_to_checkpoint` runs only after argument resolution and rendering succeed.


guapisolo:
@claude review always

guapisolo:
bypass merge to unblock.

PR #3319, opened 2026-09-18
fix(rollout): isolate cancelled groups in fully async generation

Shi-Dong:
## Problem

A rollout sample can end in `asyncio.CancelledError` without the async rollout producer being cancelled—for example, when an agent's sandbox client leaks cancellation from a timeout.

`generate_and_rm_group` already cancels and settles the sibling samples, then re-raises. The producer calls `task.result()` on that cancelled group task, so the cancellation escapes into the long-lived worker. The next training-batch fetch then fails, even though other rollout groups are healthy.

## Fix

- Retain the original prompt group alongside each in-flight task.
- Keep scheduling in `_worker_loop`; `_collect_group_result` converts cancelled groups into aborted buffer entries at the completed-task boundary.
- Let the existing default buffer exclude that entire group from training, increment `rollout/fully_async/aborted_groups_filtered`, and apply `--async-unused-samples-handler=drop|retry` to the original prompts.
- Preserve real producer cancellation and fail-fast behavior for ordinary exceptions; do not broadly catch or suppress `CancelledError`.

This is Miles-side containment, not a fix for a sandbox SDK's timeout implementation. Custom buffers continue to own their aborted-group filtering policy.

## Validation

Original cancellation implementation (`d9d993e3`):

- **99 targeted CPU tests passed**, covering fully async rollout, sample backfill, lifecycle, async utilities, and async training orchestration.
- All four new sample-cancellation cases fail on unmodified `main` at `task.result()`; the producer-shutdown control still passes.
- The five new cases also passed 20 repeated runs with uvloop selected.
- Regression coverage exercises the real group fan-out and sibling cleanup, both submission granularities, drop/retry handling, and successful subsequent batches.
- All applicable pre-commit hooks passed.
- No training job was launched or restarted.

Cleanup (`9b9e66be0`):

- Extracted result collection without changing cancellation or buffer behavior.
- `pre-commit run --all-files` passed locally.
- CPU regression tests run in [PR CI](https://github.com/radixark/miles/actions/runs/35407569295); currently pending.

No new dependencies.


Shi-Dong:
@claude review always

miles/rollout/fully_async_rollout.py
```diff
@@ -127,15 +128,31 @@ async def _generate_group(self, prompt_group: list[Sample]) -> DataBufferInput:
...
+                task, prompt_group = self._submit_one_group()
+                active[task] = prompt_group
+            done, _ = await self._scheduler.wait_for_progress(set(active))
             for task in done:
-                await self._output.put(task.result())
+                prompt_group = active.pop(task)
+                if task.cancelled():
+                    # Reading a cancelled child's result would cancel the producer too.
+                    # Only handle finished children here; cancellation of this worker
+                    # at an await remains a shutdown signal and propagates normally.
```
Shi-Dong: Make these comments clearer.

Shi-Dong: Clarified in d9d993e3c44bcd2267519194fb293e145e6a1113. The comments now distinguish a cancelled child group from cancellation of the producer, explain why task.result() would stop all generation, and state why real producer shutdown still propagates. This is a comments-only change; 99 targeted tests and the applicable pre-commit hooks pass.

yueming-yuan (approved):

PR #3323, opened 2026-09-19
[DO NOT MERGE] fix(session): disable training replay outputs during evaluation

guapisolo:
## Summary

Stop evaluation sessions from inheriting training replay outputs and R3-only constraints.

## Symptom & Reproduction

- **Symptom:** Agentic evaluation on a replay-enabled fleet requests training-only routing/indexer data and rejects caller-supplied R3 offsets.
- **Reproduction:** Enable routing/indexer replay, create a session with `{"evaluation": true}`, and send a chat request. Previously the creation body was ignored, leaving replay outputs enabled instead of disabled; `tests/fast/router/test_session_evaluation.py` covers this HTTP behavior for both session versions.

## Root Cause

1. `OpenAIEndpointTracer.create` and the session registries did not carry evaluation purpose into request preparation.
2. `resolve_request_args_by_config` applied training replay flags and R3 offset validation to every session.

## Fix

Validate `POST /sessions` with its own `CreateSessionRequest`, then pass only `evaluation: bool` into `SessionCore.create_session`. Empty bodies default to training; invalid creation inputs return HTTP 400 before allocation. Chat completion parsing remains separate.

Propagate `GenerateFnInput.evaluation` through the tracer into v1/v2 session state. The purpose survives rollback and branching.

After model rules resolve the request, evaluation forces `return_sampling_mask`, `return_routed_experts`, and `return_indexer_topk` to `false` and removes `routed_experts_start_len`. Training behavior, sampling default/override precedence, TITO rendering and compatibility checks stay unchanged. Evaluation can vary temperature between turns; existing sample collection remains supported without replay data.

## Verification

- Focused session, agentic, tracer, router/tree-assembly CPU suites: **481 passed**, covering creation validation, continuation/retry, concurrent train/eval sessions, sampling resolution and collection without replay outputs.
- An initial parity run received empty HTTP 404 responses before reaching the mock backend; the complete unchanged rerun passed. The transient cause remains unconfirmed.
- `uvx pre-commit run --all-files`: all hooks passed; `git diff --check` passed.
- Backend checks at `6891d396cc`, before the creation-boundary refactor: H200/Qwen3-30B-A3B with SGLang `c9660a7e112c07fc2003e90d5fa3330323acf2f9` plus `sglang-router==0.3.2`: direct positive/negative capture, mixed requests, direct/gateway v1/v2 evaluation sessions passed. Evaluation responses omitted replay fields while retaining token logprobs.
- The gateway positive sampling-mask control did not return a mask; evaluation suppression checks passed separately. Positive DSA-indexer capture, CUDA graphs and full RL training were not exercised.

## Review Focus

- `CreateSessionRequest` / `SessionCore.create_session`: validate at the HTTP boundary; core receives the resolved boolean. `prepare_chat_request` retains final replay-policy enforcement.
- `SessionStateV2` / `LinearTrajectory`: purpose belongs to the session across continuation, retries, concurrent training/evaluation.
- Disabling response fields does not guarantee zero internal capture on an engine with capture enabled globally. Gateway positive sampling-mask compatibility remains outside this change.


guapisolo:
@claude review always

guapisolo:
@claude review always

miles/rollout/generate_hub/agentic_tool_call.py
```diff
@@ -50,7 +50,7 @@ async def generate(input: GenerateFnInput) -> GenerateFnOutput:
     )
     use_v2 = getattr(input.args, "use_session_server", None) == "v2"
     collect_spec_metrics = use_v2 and input.args.sglang_speculative_algorithm is not None
-    tracer = await OpenAIEndpointTracer.create(input.args)
+    tracer = await OpenAIEndpointTracer.create(input.args, evaluation=input.evaluation)
```
HJSang, line 53: instead of passing `evaluation` flag, how about explicitly passing "is_replay"? In case, we need to turn it off for non-evaluation requests too. What do you think?

guapisolo: I also thought of this... My idea is the `evaluation` semantic is easy to be understood by user, and also we might skip more TITO components for evaluation to ensure perf. So I think keep this param name is ok.

yueming-yuan (approved):

PR #3324, opened 2026-09-19
fix(rollout): bypass sessions for agentic evaluation

guapisolo:
## Summary
Let agentic evaluation run without a training session or TITO trajectory collection.

## Symptom & Reproduction
- **Symptom:** Agentic evaluation requires a training session.
- **Reproduction:** On `e2a5a3e592`, calling `agentic_tool_call.generate` with `evaluation=True` without session addresses raises `requires session_server_addrs`.

## Root Cause
1. `agentic_tool_call.generate` ignores `input.evaluation` before creating the training session.
2. `collect_samples` assembles training trajectories even though evaluation only consumes trial rewards.

## Fix
Branch before session creation and call the same agent against the selected inference router, including the eval fleet. Preserve sampling, chat-template and LoRA kwargs. Return one sample containing the agent's metadata and any supplied reward; otherwise the existing reward-function path computes the score. Agent exceptions propagate through the generate call. Skip unavailable token metrics. Training keeps its existing session path.

## Verification
- Targeted CPU regression: 75 passed across agentic eval, existing reward handling, training sessions, metrics and single/multi-turn rollout integration.
- Changed-file pre-commit checks passed.
- New tests are discovered automatically by `stage-a-cpu`.
- No GPU model or real-agent benchmark run.

## Review Focus
- Evaluation uses ordinary chat rendering without TITO continuation. Reward functions receive prompt and metadata, without a collected model transcript.


guapisolo:
@claude review always

PR #3326, opened 2026-09-19
[AMD] replace the ROCm 7.2.0 image with ROCm 7.2.4

XinyuJiangCMU:
## Summary

The Miles ROCm 7.2.0 image can no longer rebuild: its Python 3.10 base is incompatible with the current Python 3.12-only Megatron-Bridge.

- Replace `rocm720-mi35x` with the dated Python 3.12 `rocm724-mi35x` base and matching wheel shelf.
- Keep the base-provided Apex, install NVRx, and update the VMM overlay from ROCr 70200 to 70204. The overlay covers both system and torch-bundled HSA runtimes.
- Update the ROCm CI image default and related documentation to 7.2.4.

## Validation

- Built a precursor image on MI355X and verified Python 3.12, torch/ROCm, Transformer Engine, flash-attn, Apex, router, gateway, and Miles/SGLang imports.
- Verified VMM pause/resume with the 70204 overlay; `rocminfo` and torch load the patched HSA copies.
- `git diff --check`, Python compilation, workflow YAML parsing, and both ROCm variant dry runs pass.

The final image will be rebuilt and smoke-tested after review, then published once as `rocm/sgl-dev:miles-rocm724-mi35x` before this PR is opened. A follow-up SGLang PR will move the Miles nightly workflow to the new variant.


yushengsu-thu (approved):
LGTM

PR #3328, opened 2026-09-20
[Sampling Replay 4/8] Capture native rollout sampling support

nanjiangwill:
ci-sglang-pr: sglang-miles-v0.5.20

## Stack position

Part 4/8 of the bounded sampling-support replay series. Merge after #3327. External prerequisite: #3321.

Layer-only diff: https://github.com/nanjiangwill/miles/compare/feat/sampling-replay-actor-scoring...feat/sampling-replay-rollout-capture

## Summary

- Request SGLang v0.5.20 native realized sampling support for bounded training requests.
- Preserve SGLang support-normalized selected-token log probabilities.
- Validate complete per-token metadata and sampled-token membership.
- Cover standard, legacy, single-turn, and multi-turn rollout paths.
- Record environment and tool tokens with singleton support.
- Disable training-only capture during evaluation.

## Correctness boundary

Miles does not reconstruct top-p or top-k support. SGLang is the source of truth for the realized support and owns physical support capacity and custom-logit-processor compatibility. Miles validates only transformations its trainer must reproduce, including request temperature and unsupported relative-logit changes.

This layer excludes session-server and agentic transport, which is handled by the next PR.

## Validation

Focused tests cover request construction, native metadata parsing, abort behavior, ragged-support validation, forced tokens, multi-turn merging, and evaluation.

PR #3321, opened 2026-09-19
Bump sglang to v0.5.20

Zhichenzzz:
ci-sglang-pr: sglang-miles-v0.5.20
https://github.com/sgl-project/sglang/tree/sglang-miles-v0.5.20
Rebases the `sglang-miles` stack onto `v0.5.20` and repoints the miles image at it.

## sglang side

`sglang-miles-v0.5.20` = 54 commits on `v0.5.20`, replayed from `sglang-miles` (the v0.5.19 stack: 55 numbered + 6 unnumbered commits, frozen as `sglang-miles-v0.5.19-final` @c9660a7e11). 727 upstream commits of drift; 21 of the 57 picks conflicted.

**Dropped (already in v0.5.20 or superseded):**
- `Keep NVFP4 blockscale swizzle padding on device (#39142)` — byte-identical to upstream #39141.
- `state_capturer: pin the exact host-cache size via mmap + cudaHostRegister (#36562)` — a strict subset of upstream #37285, which also adds the `destroy()` hooks.
- `dsv4: gate the padded-head path on attn_tp_size, not tp_size` — v0.5.20's `MQALayer.forward` no longer has that block (`_kernel_num_heads()` / `_local_attn_sink(kernel_num_heads)` replaced it).
- `[38] Restore MqaAttentionBase.maybe_use_decode_attn_tp` — only re-added two imports that the Nemotron commit had removed; that removal is gone (below), so the commit became empty.
- `Remove obsolete PD retract guard (#39119)` squashed into `[8/54] Improve PD pause handling`, as its subject asked.

**Notable conflict decisions.** v0.5.20 split `ServerArgs` into `arg_groups/fields/*.py` (choices in `arg_groups/choices.py`, no `NS()` markers; `field_order.py` is a compat record, new fields sort last), removed the DSA prefill in-sequence-split CP path, moved the DSA prefill split-ops into `dsa_prefill_cuda_graph.py`, and made the tokenizer dispatch path async with its own `dispatched` / `abort_sent` bookkeeping. Our behaviour was re-expressed on those shapes:

- *True on-policy* (`[1]`): `true_on_policy_contract`, `enable_prefill_only_deterministic_inference` live in `fields/exec_.py`, `RL_ON_POLICY_TARGET_CHOICES` in `choices.py`; the decode CUDA-graph capture keeps upstream's `get_exec().graph` flag inside our prefill-only deterministic patch; `qwen3_moe` keeps upstream's empty-batch guard around our true-on-policy routing branch.
[...]

PR #3329, opened 2026-09-20
[Sampling Replay 5/8] Preserve sampling support through session rollouts

nanjiangwill:

## Stack position

Part 5/8 of the bounded sampling-support replay series. Merge after #3328. External prerequisite: #3321.

Layer-only diff: https://github.com/nanjiangwill/miles/compare/feat/sampling-replay-rollout-capture...feat/sampling-replay-sessions

## Summary

- Carry bounded-sampling request defaults through session server v1 and v2.
- Retain native support metadata in server-owned records while stripping it from client responses.
- Reconstruct supports and native rollout log probabilities during sample collection.
- Extend the typed session codec field selection without changing the disabled path.
- Support agentic and tool-call rollouts, including explicit evaluation opt-out.

## Why this layer is separate

Session rollouts have a distinct lifecycle: requests are proxied, metadata is stored server-side, turns may branch or be truncated, and samples are encoded later. Keeping this layer separate lets reviewers verify that lifecycle without re-reviewing trainer math or direct SGLang generation.

## Validation

Focused tests cover v1 and v2 configuration, request resolution, record conversion, truncation, aborts, client metadata stripping, codec selection, agentic collection, and router integration.


nanjiangwill:
Replaced by #3350, rebased onto current main. GitHub cannot reopen or retarget this PR because its deleted base branch (#3321) causes the API to reject both operations.

PR #3330, opened 2026-09-20
[Sampling Replay 6/8] Validate bounded sampling replay contracts

nanjiangwill:

## Stack position

Part 6/8 of the bounded sampling-support replay series. Merge after #3329. External prerequisite: #3321.

Layer-only diff: https://github.com/nanjiangwill/miles/compare/feat/sampling-replay-sessions...feat/sampling-replay-validation

## Summary

- Treat top-p below 1 or positive top-k as sampling-support replay.
- Require positive bounded top-k for finite top-p so captured support is bounded.
- Require the Miles router while the SGLang v0.5.20 gateway drops the request extension.
- Reject prefill log-probability recomputation and relative-logit transformations the trainer cannot replay.
- Temporarily reject KL and on-policy distillation because the current loss interface carries only one actor score per token.

## Ownership

SGLang owns sampling-mask capacity, overflow, custom logit processors, and speculative-decoding compatibility. Miles does not duplicate those policies. Request-specific positive top-k may differ from the launch default; SGLang enforces its physical limit.

The KL restriction is an interface limitation, not a mathematical incompatibility. The actor still has full-vocabulary logits, but the loss path must eventually carry separate support-normalized and full-policy actor scores.

## Validation

Argument tests cover invalid bounds, missing router support, prefill incompatibility, and mixed sampling-support objectives.


nanjiangwill:
Replaced by #3351, rebased onto current main. GitHub cannot reopen or retarget this PR because its deleted base branch (#3321) causes the API to reject both operations.

PR #3338, opened 2026-09-20
docker(rocm): round FP32->BF16 to nearest in the AITER attention kernels on MI300X

Arist12:
## Brief

On `"transport": "daemon"` the scheduler stops prefixing `. <launch_env>` onto the job command (`gpu-scheduler/gs/runner.py::LAUNCH_ENV_INHERITED`) and trusts the daemon's own process environment. That change silently turned every *way of starting a daemon* into a launch-configuration path. Two of them were told — `fleet-daemon/deploy/host-run.sh` and `deploy/install-node-service.sh` both source `job-daemon/deploy/launch-env.sh`, and the installer refuses to install without it — and the three payloads `fleet-daemon/deploy/` sends to a node over `/v1/exec` were not. They are strings inside `.py` files, so `tests/test_boot_paths_agree.py`, which walks `*.sh`, could never have seen them.

Measured on `mia1-p02-g45` on 2026-09-10: the only daemon in the fleet that cron's `@reboot` line had actually started (pid 26930, daemon uptime 190.6 h against a host uptime of 190.6 h), reporting `launch_env_sourced: false` and short exactly seven `JOBD_*` names against a node started the systemd way — `JOBD_GH_BIN`, `JOBD_GH_TOKEN_FILE`, `JOBD_GITCONFIG_FILE`, `JOBD_GIT_CREDENTIAL_BIN`, `JOBD_CACHE_ROOT`, `JOBD_DEFAULT_VISIBLE_DEVICES`, `JOBD_LAUNCH_PRECONDITIONS_MISSING`. Four are the result-delivery quartet: a job placed there gets `gh` and a credential helper with no token behind them, `git push` fails on authentication, no pull request is opened, and the job exits 0. Nothing refused it — `gpu-scheduler/gs/placer.py::_passport_reject` fires on a term that is literally `"no"` and an unmeasured launch term is `"unknown"`; g45 was out of the placer's way only because a reservation held all eight of its cards.

## Detail

**Which fix, and why not the other one.** The alternative was to make `deploy/install-node-service.sh` able to write a root system unit for an mia node and retire the cron path. That is not this change, for a reason that is about the machines and not about taste: `Linger=no` on all five mia nodes and `loginctl enable-linger` needs a privilege nobody in this repository has, so the installer's `--user` unit would not survive a reboot there (the script itself only warns). The unit that works on `mia1-p02-g46` is a **root system** unit installed by hand on 2026-09-08, and nothing in this repository can install a root unit on a machine it has no root on. Meanwhile the only channel into these nodes is `/v1/exec`, which is `autostart.py`'s channel, and four of the five still boot from cron. Fixing the path that is running beats writing a path that cannot be applied. Retiring cron stays a decision for whoever installed that unit; `fleet-daemon/README.md` now says so, with the shape of the unit that works.

**The fix, in three payloads and not one.** `autostart.py`'s `BOOT`, `beat.py`'s `RESTARTER` and `repoint.py`'s `RESTARTER` each resolve `${FD_LAUNCH_ENV:-${FD_SRC%/}/../job-daemon/deploy/launch-env.sh}` and source it. `beat.py` and `repoint.py` arrive as children of the daemon and do inherit its environment — which is exactly the problem: a daemon cron started has nothing to inherit, and a restart that only inherits carries the unconfigured launch forward one more time. `deploy/sidecar-run.sh` is the one daemon start left alone on purpose: it mounts no docker socket and shares only the network namespace, so it cannot launch work at all, and giving it a launch configuration would say it could.

**Order, and what is deliberately not carried.** The launch owner is sourced *after* the snapshot and `uplink.env`. Every value in that file is `${VAR:-default}`, so a measured per-node answer — the card fence, `JOBD_PASSWD`, the GPU floor — is never replaced by a repository default; the sourcing fills gaps. `JOBD_LAUNCH_PRECONDITIONS_MISSING` is the one value that file exports unconditionally, which is why it stays **out** of `KEEP`: a precondition list is a measurement of the boot in progress, never a copy of the last one. `FD_LAUNCH_ENV` does join `KEEP`, because an override that does not survive a reboot is an answer that works until the machine restarts. A missing `launch-env.sh` warns and does not refuse: a daemon that will not start is a node with no census, and the census is the only thing that would have said the launch was unconfigured.

**One unit name.** `install-node-service.sh` wrote a flat `amdpilotv2-fleet-daemon.service` while the g46 unit is `amdpilotv2-fleet-daemon-g46.service`. The per-node name wins — it is the naming every other per-node artefact already uses (`env-<node>`, `boot-<node>.sh`, `fleet-daemon-<node>.env`), and these machines are reached through shared jump hosts where `systemctl --user status amdpilotv2-fleet-daemon` answers about whichever host the shell landed on. The full hardware name, not g46's short suffix, because `banff-cyxtera-cx57-4` and `-5` have none. The flat name is still deployed on every node installed before today, so it is carried rather than deleted: `--off` disables both, `--status` warns when the legacy unit is also active, and an install **refuses** while it is enabled or active, because two supervisors for one daemon race for `FD_PORT` at boot.

**The guard**, `fleet-daemon/tests/test_autostart_sources_launch_env.py`, 19 tests. It discovers every daemon start in `deploy/` with a start-line pattern anchored so that `autostart.py`'s own prose about "a bare `setsid python -m fd`" is not counted, fails on any it cannot classify, and requires the launch owner to be *resolved* (assigned or sourced) rather than merely mentioned — the whole defect was a file that named `launch-env.sh` at length and did not source it. The behavioural half cuts each shipped resolution block out of the file between `# >>> launch-env` markers and runs it under `bash` three times: with `FD_LAUNCH_ENV` set, with nothing set, and with no launch owner anywhere. Both branches, per shape 10's rule that a sweep goes across the override *and* the thing being overridden — the default is the branch every node is actually on. It also asserts the closure that makes `fleet-daemon/run.sh` safe: it sources nothing, and it is only ever reached through a payload that did. Verified non-vacuous by deleting the fix and re-running: 12 of 19 fail.

**Documents in the same commit.** `FAILURE-SHAPES.md` gains **shape 15**, "Responsibility moves to the process, and only the supervised way of starting it is told" — a **new shape**, not a footnote on §10. §10 is about a literal (`pgrep`'s stale needle) appearing in N files; here the literal was never the problem, the *responsibility* moved, and the diff that moved it mentions none of the paths that now have to carry it. §10 gains one cross-reference sentence saying they are neighbours and not the same. The title changes from "Fourteen" to "Fifteen". Also corrected: `SEAMS.md`'s `launch-env` entry (the second class has three payloads that are not scripts in the tree); `fleet-daemon/README.md` (the new sourcing, the resolution and its order, the unit-name convention, and the false implication that a `--user` unit is installable on an mia node); and three stale `autostart.py` comments, including "the mia five have no systemd and read this file instead", which stopped being true for g46 on 2026-09-08.

**Verification.**

- `make test-fleet-daemon` → `672 passed` (19 of them new).
- `make test-overlay-native` → `5219 passed, 27 skipped`.
- `make test-job-daemon` → `830 passed, 1 skipped`. It caught a real omission first: `test_component_catalogs_cover_their_owned_literal_names` failed with `fd config schema omits ['FD_LAUNCH_ENV']`, so the name is now declared in `image-config/components/fd.json` under `host_tools` — nothing inside a container reads it.
- `make test-solbench-tasks` → `571 passed, 36 skipped`.
- The default resolution was checked against the real nodes, read-only, by sourcing each node's own `env-<node>` and evaluating the same expression: `READABLE /var/tmp/amdpilotv2/repo/fleet-daemon/../job-daemon/deploy/launch-env.sh` on **all five** of g05, g10, g23, g45, g46. The mia checkout layout is exactly what the derivation predicts, and the resolved path is the one the working g46 wrapper names literally.

## Not done

**The on-node half — a repo commit changes nothing on any node.** This is `FAILURE-SHAPES.md` shape 5 and it is the larger half of the work. The generated boot scripts on the mia five are dated 2026-08-14 to 2026-08-17 (`boot-mia1-p02-g23.sh` Aug 14 06:07, the other four Aug 17), all 2863 bytes, none of them containing the string `launch-env`. Each is replaced only by re-running `python fleet-daemon/deploy/autostart.py <node>` against that node, which reaches it at `http://ts-<node>:9110/v1/exec` — a mutating remote call that needs its own approval, so it was not attempted here. When it is done: keep a copy of the previous script, diff the regenerated one against it, and confirm with `--check` before trusting it. Note also that regenerating rewrites `env-<node>` from the running daemon's environ, so on g45 it would snapshot a daemon that is *still* unconfigured — run it there after the daemon has been restarted with the launch owner sourced, not before.

**`mia1-p02-g45` itself is not repaired by this PR.** Its running daemon (pid 26930) still reports `launch_env_sourced: false`. Restarting it is what fixes it, and it holds reservation `h-5393c6` (jinpan12, GLM-5.2-FP8, `glm-serve-ht` up 7 days, "release only through serving undeploy"), so that is the reservation holder's call. Do **not** repair it with `~/.fleet-daemon/boot-mia1-p02-g45.sh` as it stands on the node today: that is the old generated script and it reproduces the same state.

**Retiring the cron path is not decided here.** The repo cannot install a root system unit on a machine it has no root on — `sudo -n true` is refused on all five mia nodes, re-confirmed 2026-09-10 — so `install-node-service.sh` still writes a `--user` unit and `autostart.py` still writes a crontab line. What this PR does is make the cron line correct while it exists, and write down in `fleet-daemon/README.md` what the working shape on g46 is, so that whoever installed it can extend it to the other four. The `@reboot` line must be removed *before* a unit is armed on a node, never both at once: the boot script's `pgrep` guard races the unit and the loser can leave the wrong daemon holding `:9109`.

**The g46 unit's own gap is recorded and not fixed here.** It carries no `RequiresMountsFor=/home /var/tmp` and no `After=remote-fs.target`, while its wrapper exits 1 when `$HOME/.fleet-daemon/env-<node>` — on WekaFS — is unreadable. It has also not survived a boot yet (host uptime 1w 23h, unit active since 2026-09-08 19:30), so the pattern is not yet proven durable. That belongs to the change that installs the unit on a second node.

**`launch-env.sh` was deliberately not sourced on a live node to measure the seven variables directly.** It creates `JOBD_PASSWD` and the cache-root directories, and this account is not the account the mia daemons run as, so doing it would have left directories owned by the wrong uid on a node with a running job. The seven names were confirmed statically instead — all seven are `export`ed by `job-daemon/deploy/launch-env.sh` — and the resolution was checked with a read-only `test -r`.


Arist12:
Closing this. An 8-GPU GRPO run on MI300X (Qwen3-4B, gsm8k, SGLang rollout + FSDP trainer, 10 steps per arm) shows the change does not do what this PR claims.

**The Dockerfile change is inert for the images this tree builds.** `docker/build.py` defines only `rocm720-mi35x` and `rocm10-mi35x`, both `GPU_ARCH=gfx950`. On gfx950 ck_tile's `float_to_bf16` uses the native convert instruction and ignores `CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT` entirely, so setting it there changes nothing while the eviction `RUN` costs a pointless rebuild. The MI300 image (`rocm/sgl-dev:miles-rocm700-mi30x`) is not built from this tree.

**And the premise does not hold up.** `train/train_rollout_logprob_abs_diff` was 0.0576 ± 0.008 with the stock truncating build and 0.0556 ± 0.020 with the macro at 0 — Welch p = 0.77. The metric scores generated tokens, and rollout decode runs on the paged-attention kernel (`pa_ragged_*`), which is not a ck_tile kernel and does not read this macro. Only rollout prefill and RMSNorm are affected, so this macro cannot move most of what the metric measures.

The underlying gfx942/gfx950 rounding divergence is real and I have a one-line AITER patch for it (attention error −33%, RMSNorm −50% against an FP64 reference), but it belongs upstream in AITER and is an accuracy fix, not a train-inference-mismatch fix. Re-opening the MI300X train-inference investigation against the decode path instead.

